Issue #1013, opened 2023-04-13
Autosubmit describe only prints the autosubmit version


kinow:
In GitLab by @abaer-kit on Apr 13, 2023, 17:45

Hello @dbeltrankyl and @kinow,

this bug might be related to #1011 .

### Autosubmit Version

v4.0.76 (local installation)

### Expid affected(If applicable)

All on my local system.

### Summary

The output of autosubmit describe only prints the autosubmit version:
```bash
$ autosubmit describe a005
Autosubmit is running with 4.0.76
$ 
```
In addition, no AS log file is created.

### Steps to reproduce

On my system, I can run `autosubmit describe <expid>` for any expid - either newly create or existing.

### What is the current bug behavior?

See summary.

### What is the expected correct behavior?

According to the [documentation](https://autosubmit.readthedocs.io/en/master/userguide/monitor_and_check/index.html#how-to-get-details-about-the-experiment) it should display:  
owner,description_date,model,branch and hpc

kinow:
@abaer-kit thanks for reporting it.

>According to the [documentation](https://autosubmit.readthedocs.io/en/master/userguide/monitor_and_check/index.html#how-to-get-details-about-the-experiment) it should display:  
>owner,description_date,model,branch and hpc

I just finished debugging the code, and I can confirm that we have an exception happening due to an `AttributeError`, as an attribute was removed in an `autosubmit4-config-parser` object. This command prints the error:

```bash
(autosubmit4) kinow@ranma:~/Development/python/workspace/autosubmit$ autosubmit -lc WARNING describe a006
Autosubmit is running with 4.0.0b
Reloading configuration files, due a change in the following files: []
[WARNING] Some defined platforms have the following issues: []
[WARNING] Could not describe the following experiments:
['a006'][eCode=3000]
```

@dbeltrankyl, IMHO it would be better to show warnings by default in our logging. Is there a reason why only INFO is enabled by default? That would save some time troubleshooting issues like this one…

>In addition, no AS log file is created.

I can confirm that too. I deleted both log folders in my expid folder, executed `autosubmit describe a006`, and no files nor directories were created. I found a place of the code with a list of expid-less commands, that included `expid`, `readme`, `changelog`, etc., and `describe`.

The problem is that that `expid_less` control two features. It control whether an `expid` is expected and verified, and whether the command will produce a log file for stdout & stderr.

[...]

PR #1825, opened 2024-12-12
Draft: Fix autosubmit4-config-parser object attribute, use the folder instead of a config file

kinow:
In GitLab by @kinow on Apr 19, 2023, 14:05

Related to #1013

We were using `experiment_file` attribute, but with minimal/custom config that attribute was removed as that file is not mandatory any longer.

This MR is fixing the attribute error, but there are other issues in the linked issue #1013, related to the creation of log files, and to the default log level in AS4 (thus a draft for now).

- [x] fix attribute error producing a hidden warning
- [ ] display warnings by default (?)
- [ ] fix log files not created on ASLOGS folder
- [ ] add unit tests so this attribute error is captured in our CICD pipeline if it happens again in this part of the code

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -3371,8 +3371,8 @@ class Autosubmit:
         """
         Show details for specified experiment
 
-        :param experiments_id: experiments identifier:
-        :type experiments_id: str
+        :param input_experiment_list: experiments identifier:
+        :type input_experiment_list: list[str]
         :param get_from_user: user to get the experiments from
         :type get_from_user: str
         :return: str,str,str,str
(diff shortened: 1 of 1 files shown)


kinow:
In GitLab by @dbeltrankyl on Apr 27, 2023, 15:46

I needed this command to check some experiments and found the bug,


as_conf.experiment_file refers to expid/conf/expid.yml which doesn't exist anymore, now this check is done in the $expid/conf instead to check the user and created date

`                user = os.stat(as_conf.conf_folder_yaml).st_uid`

```
created = datetime.datetime.fromtimestamp(
                    os.path.getmtime(as_conf.conf_folder_yaml))
```

kinow:
In GitLab by @dbeltrankyl on Apr 28, 2023, 09:57

mentioned in commit a325afdd25c5e394c2c0de3670bd0ce43e2ef91f

kinow:
In GitLab by @kinow on May 9, 2023, 14:02

Thank you @dbeltrankyl !

This command now works:

```bash
autosubmit describe a005
```

So closing this MR.

Issue #864, opened 2022-09-06
conda installation


kinow:
In GitLab by @pablohe on Sep 6, 2022, 10:15

Hi! I have been trying to install it using the conda recipe:

https://autosubmit.readthedocs.io/en/latest/installation.html#sequence-of-instructions-to-install-autosubmit-and-its-dependencies-with-conda

- the link to miniconda download is not updated
I installed as
```bash
conda env update -f environment.yml -n autosubmit python=2
```

and tried to use as follow:

```bash 
conda activate autosubmit
(autosubmit) root@gisserv:~# autosubmit
autosubmit: command not found

(autosubmit) root@gisserv:~# autosubmit/bin/autosubmit -v
Traceback (most recent call last):
  File "/root/autosubmit/bin/autosubmit", line 25, in <module>
    from log.log import Log, AutosubmitCritical , AutosubmitError
ModuleNotFoundError: No module named 'log'

```

```bash
(autosubmit) root@gisserv:~# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.4 LTS"

```

```

kinow:
In GitLab by @dbeltrankyl on Sep 6, 2022, 10:59

Hello, thanks for the feedback

The conda link is working but was concatenated to the following line, I've updated the docs with the fix. 

Also, there was a missing chmod + x before the execute step. 

Due to some circumstances with the operational, 3.14.0 is only meant to be used with it and works under bsc machines. It will be updated in the near future with the 3.14.0b fixes. 

This 3.14.0b includes some fixes for outsiders, the log issue should work with 3.14.0b

Therefore, I recommend you use the 3.14.0b tag instead of 3.14.0

kinow:
In GitLab by @mcastril on Sep 7, 2022, 13:07

@dbeltrankyl can you also update the doc so that the `git clone` downloads 3\.14.0b until it is released as 3.14.0 ?

This is will be faster as we don't have a final release and there can be additional fixes.

kinow:
In GitLab by @pablohe on Sep 22, 2022, 11:09

Hi!

Just to let you know that will be useful to have conda installation for 3.14.0b

I couldn't get it "manually", and using `SPLIT` option will be very useful in my current setting (parameters exploration) unfortunately it is present just in 3.14.0b

many thanks!

[...]

Issue #1380, opened 2024-07-31
[discussion] Plan for the transition to GitHub


kinow:
This is an issue to document the steps to be taken, any possible risks and mitigations, responsibilities, etc., for the transition from GitLab to GitHub. Feel free to update the list with other items.

Action list:

- [x] Discuss the idea
  - [x] done in a few meetings
- [x] Discuss the location of the new project
  - [x] initially autosubmit/autosubmit, but it's taken, so we will use [bsc-es/autosubmit](https://github.com/bsc-es/)
  - [x] Francesco B. added Bruno to the team, and we are discussing about an Autosubmit team, managed by us so that it's easier to update the list of people with write access to the repo
  - [x] Create the Autosubmit team, and add others
- [ ] Move from GitLab issue & merge request templates, to GitHub issue & pull request templates
- [ ] Move from GitLab workers CICD to GitHub Actions
  - [ ] The coverage that's handled by GitLab will have to be moved somewhere like Codecov, for instance
  - [ ] Revise "skipped" tests ( per ex, https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/508#note_318531 )
- [ ] Choose a uniform main branch name (AS is `master`, others are `main`)
- [ ] Migrate code, issues (? or leave issues where they are) and merge requests
  - [x] autosubmit-gui
  - [x] autosubmit-api
  - [x] autosubmitconfigparser
  - [ ] autosubmit
  - This SO answer for “Transfer git repositories from GitLab to GitHub - can we, how to and pitfalls (if any)?” https://stackoverflow.com/a/22266000
  - This tool (from SO answer above) that migrates merge requests - https://github.com/piceaTech/node-gitlab-2-github?tab=readme-ov-file#transfermergerequests
- [ ] Update documentation
  - [ ] README
  - [ ] Sphinx
  - [ ] ES Wiki
  - [ ] [BSC page](https://www.bsc.es/research-and-development/software-and-apps/software-list/autosubmit) (it needs an update too)
  - [ ] References in Eduuni
  - [ ] GitLab Wiki (besides the README, there are many links there)
- [ ] GitLab Wiki (can we move part or most of the information to GitHub / README / RTD)?
- [ ] Update ReadTheDocs
- [x] Add the CITATION cff file (a comment from @manuel-g-castro in another issue reminded about this feature in GH) -- done in https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/472
[...]

PR #1849, opened 2024-12-12
Draft: Reduce duplication in the environment.yml

kinow:
In GitLab by @kinow on Jul 31, 2023, 10:54

Part of #864 

This needs more testing, both locally and with the CICD worker (can't recall if that uses `environment.yml`).

I also think that `nosecov` is only installed in `environment.yml`.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,44 +4,9 @@ channels:
 
 dependencies:
 - python=3.7
-- nose
-- nose-cov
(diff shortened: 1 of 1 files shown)


kinow:
In GitLab by @kinow on Aug 19, 2024, 10:45

Superseded by !435 and by the move to GH #1380

PR #1884, opened 2024-12-12
Draft: Create experiment, workflow, and tasks documentation from configuration, as well as metadata

### LINKED ISSUE #939: [enhancement] Create experiment, workflow, and tasks documentation from Autosubmit configuration files
This idea came from my induction training at the BSC, in a CES trello card. While trying Auto-Monarch and Auto-Earth, I noticed these tools had a similar documentation format, but parts of the documentation appeared to be out of sync with the current workflow code.

I reported the issues to both repositories. After the [response from Auto-Monarch by @gmontanepinto](https://earth.bsc.es/gitlab/es/auto-monarch/-/issues/513#note_196937) (thanks again!), I shared with him an idea to auto-generate the documentation in certain formats, like Markdown, directly from Autosubmit experiment configuration.

Other workflow managers support this already, like Cylc, ecFlow, and cwltool. The idea is to have new attributes in our YAML configuration for documentation. For example:

*jobs.yaml*

```yaml
JOBS:
  LOCAL_SETUP:
    FILE: template/local_setup.sh
    RUNNING: once
    TITLE: A task to prepare the local archives used by the model
    DESCRIPTION: |
      This verifies if the expected files exist in the local directory,
      `tar`'s these files, and creates a log line about its execution.
      Executed once, this step is optional if you have a pre-existing
      model binary, and can be disabled with the environment variable
      `ABCDE`...
  SIM:
    FILE: template/ini.sh
    TITLE: The model simulation task
    DESCRIPTION: |
      This is the most important task of this workflow. It will ...
```

*expdef.yaml*

```yaml
DEFAULT:
  EXPID: a000
  HPCARCH: local
  TITLE: Auto-Application
  DESCRIPTION: |
    Auto-Application is the Autosubmit workflow for the Application model.
    It is developed and maintained by the team ABC.

    It is a model of this type, that does this, and that...
    
    documentation:  <https://intranet...>
    issue tracking: <https://gitlab...>
    source code:    <https://gitlab...>
    contact point:  [ABC DEF](mailto:....)
```

In this example, the `TITLE` attribute would be a short title for the workflow node, by default the experiment ID or the task ID. The `DESCRIPTION` attribute would then be a longer text with links, documentation about the task or workflow, who to contact, what's the expected input, etc.

These values would then be used by, maybe, a new subcommand like `autosubmit autodoc`, to produce text, HTML, or Markdown output.

@gmontanepinto also suggested to use the [GitLab API](https://docs.gitlab.com/ee/api/wikis.html) to make it easier to update the Wiki when there's a change. I think that would definitely keep the sources and docs in sync!

Another suggestion from @gmontanepinto: “and we should define a common format between projects.”

By getting some feedback from users for this common format, I think we could start simple with a Markdown template using [Python stdlib's string](https://docs.python.org/3/library/string.html#template-strings) module to render it. This way we wouldn't have to add a new dependency to Autosubmit, and we could improve the solution later if needed (but that's more about the implementation, that could be discussed later).

kinow:
In GitLab by @kinow on Jan 19, 2024, 12:16

Closes #939 

Creating a merge request to demonstrate the idea of the automated documentation of experiments. This is not super new, as you can achieve almost the same with Cylc and ecFlow. In Cylc you have DOC and TITLE too. In ecFlow they use MANUAL.

I am extending the scope of the issue (created 11 months ago) to add something that was discussed more recently, about the documentation for models, metadata, ...

The idea is documented in the `docgen/__init__.py` docstrings. Pasting it here:

```
Goes over experiment configuration, gathering each job's
``TITLE`` and ``DOC`` to create the documentation for the
jobs.

Goes over the ``METADATA`` configuration entry, looking
for sections with ``key=value`` pairs, or links to
configuration keys, e.g.:

.. code-block:: yaml

  METADATA:
    MODEL:
      # This is a static value that will produce name = IFS
      - name: name
      # This is a dynamic value that will produce resolution = abc...
      - resolution: %RUN.IFS.RESOLUTION%
      # This resolves the key and value automatically and will produce RUN.IFS.INIPATH=/path/path/...
      - %RUN.IFS.INIPATH%
      # This is the extended format
      - name: source
        value: https://git@...
        documentation: |
          The source code is hosted at the private repository...
    DATA:
      - PROVENANCE: ...


Each property of the ``METADATA`` YAML object is rendered as
a separate section. These sections, in YAML, are arrays that
contain metadata references in each line. Extra documentation
about each key can be entered using the extended format.
```

The gist is that the maintainers of the experiment would be able to select configuration values from the workflow that must be exported as workflow metadata (in the `METADATA` YAML object/entry above).

The new command in Autosubmit will take care to go over the experiment configuration, resolve variables like `%EXPID%`, `%RUN.IFS.RESOLUTION.ETC%`, and produce the metadata list, with the **current values** used in the experiment. 

I think this is key here, to have the documentation generated with the values used in the workflow -- i.e. the workflow run configuration is the [single source of truth](https://en.wikipedia.org/wiki/Single_source_of_truth), to avoid having a PDF or Wiki page manually written that might be out of sync with what's actually happening in the workflow.


kinow:
In GitLab by @kinow on Jan 19, 2024, 12:19

The current code is a draft and only produces:

```
$ autosubmit docgen a001
Autosubmit is running with 4.0.99

Checking configuration files...
Expdef config file is correct
Platforms sections: OK
Jobs sections OK
Configuration files OK

# Workflow Jobs

These are the jobs in the workflow “a001
”.
## INI

**Initialize the experiment**
This task runs locally, the expid is %EXPID%

## SIM

## ASIM

## POST

# Metadata

```

This is valid Markdown, that could be used to generate a web page, PDF, docx, etc.. The contents of the `ini` job configuration:

```yaml
  ini:
      FILE: 1.sh
      RUNNING: once
      TITLE: Initialize the experiment
      doc: This task runs locally, the expid is %EXPID%
```

Pasting the content below so GitLab can render the markdown (expanding the `%EXPID%` is easy but I will add that later):

# Workflow Jobs

These are the jobs in the workflow “a001
”.
## INI

**Initialize the experiment**
This task runs locally, the expid is %EXPID%

## SIM

## ASIM

## POST

# Metadata

kinow:
In GitLab by @kinow on Jan 19, 2024, 12:20

/cc @mcastril , @GinkaVanThielen , @ainagaya , @Lerriola , @franra9 , @sebastianbeyer , @fdoblasreyes wrote a quick merge request just now to demonstrate the idea of automated documentation generation :point_up:

kinow:
In GitLab by @kinow on Jan 19, 2024, 12:20

Commented on [autosubmit/docgen/__init__.py line 1](https://github.com/BSC-ES/autosubmit/compare/1e37bc035176cd710b42f10e739844546bcf2450..52294be774e0e73ab5f04a4c0f55e54a5c812ace#diff-165c5b03bc8806cd1e6e876b118722ceR1)



Note to self, update to 2024

kinow:
In GitLab by @kinow on Jan 19, 2024, 12:22

And this doesn't need to replace any other documentation planned for models; this can be used to complement any documentation/wiki/deliverable/etc., but automating the parts that contain variables from the workflow.

kinow:
In GitLab by @kinow on Jan 19, 2024, 12:36

mentioned in issue digital-twins/de_340/project_management#579

kinow:
In GitLab by @mcastril on Jan 19, 2024, 19:09

This is wonderful Bruno. Can you clarify one doubt?

```
Goes over the ``METADATA`` configuration entry, looking
for sections with ``key=value`` pairs, or links to
configuration keys, e.g.:

```

Wouldn't it be `key: value`?

kinow:
In GitLab by @ainagaya on Jan 22, 2024, 09:55

That's great @kinow !! 

So, if I understood correctly, what should be done in the workflow is to create a configuration file with the METADATA entries and fill them with placeholders? Sounds good to me.

kinow:
In GitLab by @kinow on Jan 22, 2024, 10:01

Yup! That's the initial idea. But we can work on it and decide what would be the best way for that. I think with a few more hours (2...3) I can get it to work and produce the final markdown.

With a working prototype we can test in the climatedt workflow or another example workflow, and then discuss what needs changing.

kinow:
In GitLab by @kinow on Jan 22, 2024, 10:33

>Wouldn't it be `key: value`?

Ah, yes, so that it's valid YAML, good catch. Will fix that in the docs later. Thanks @mcastril !

kinow:
In GitLab by @ainagaya on Jan 24, 2024, 13:07

mentioned in issue digital-twins/de_340/workflow#174

dbeltrankyl:
Status update?

VindeeR:
@kinow I saw this branch and I think this is mostly done, maybe there's something we can migrate to a newer issue like the one @rajeshsuraj is working on.

we updated the documentation to have valid and working experiment structures as well as running experiments on the run checking if AS is working as intended.

This way we won`t lose track of it, wdyt?

PR #1891, opened 2024-12-12
Wrapper fix

### LINKED ISSUE #1226: Wrapper deadlock detection and splits
In GitLab by @dbeltrankyl on Feb 9, 2024, 10:21

Hello @kinow , 

I found an issue with infinite deadlock detection ( the normal deadlock works fine ) when the jobs wrapped have the split value set.

It is a design issue more than a bug 

To know if a job could be wrapped or not in the future, AS fills a list of "next_wrappeable_jobs" and checks if the child is a direct one or not ( for vertical and 2-D wrappers, for horizontal ones is not necessary but it also does it ) to know which jobs need to be in READY status to add more jobs in the wrapper. 

The problem is that the calculation only takes into account the chunks. So, for splits, this list could or could not be wrong. (For example, with the filter "previous" , this is not enough )

The idea is to change the algorithm to take into account other factors.

I think that max_jobs_to_submit, total_jobs, max_jobs, checking if other jobs can be submitted, and checking the list of active jobs could be enough to calculate the infinite deadlock without relying upon dependencies that, with the special filters, can be not so straightforward as in the past.

kinow:
In GitLab by @dbeltrankyl on Feb 20, 2024, 16:15

This fixes #1226 , 

I've changed the logic following the following points:

        # It is a deadlock when:
        # 1. There are no more non-wrapped jobs in ready status
        # 2. And there are no more jobs in the queue ( submitted, queuing, running, held )
        # 3. And all current wrappers being processed at this time are not wrappable.

The unit test seems to work fine with this version as well


kinow:
In GitLab by @dbeltrankyl on Feb 20, 2024, 16:15

requested review from @kinow

kinow:
In GitLab by @dbeltrankyl on Mar 4, 2024, 13:06

deleted the `squashed-Destine-1025` branch. This merge request now targets the `master` branch

kinow:
In GitLab by @dbeltrankyl on Mar 5, 2024, 12:57

con in !408

Issue #945, opened 2023-02-07
Change lock mechanism


kinow:
In GitLab by @dbeltrankyl on Feb 7, 2023, 14:22

Our current lock mechanism relies on a library that doesn't work properly if three or more instances of the experiment are running. ( mainly due to the shared filesystem ) 

* The lock is done properly on the first run, 
* The lock is released if another run or create on the same experiment is triggered, although the attempt is blocked if the first instance didn't finish. 
* The lock is done again, but the instance is never blocked. 

The idea is to change this library to a custom file-based lock.

kinow:
In GitLab by @mcastril on Sep 7, 2023, 16:47

mentioned in issue auto-ecearth3#1952

kinow:
In GitLab by @dbeltrankyl on Jun 27, 2024, 14:20

Update: 

I wanted to try another library I read about, called fcntl, but the same is happening. I think it is related to being in a shared filesystem, so unfortunately, we still need to implement our custom file lock.

kinow:
Maybe we can try the latest versions of the library to see if they fixed it :fingers_crossed:  https://pypi.org/project/portalocker/#history

kinow:
In GitLab by @dbeltrankyl on Aug 30, 2024, 10:54

Maybe we can move this to 4.2.0 and implement the postgres lock and file lock all together @LuiggiTenorioK , @kinow 

https://earth.bsc.es/gitlab/es/autosubmit/-/issues/1392? 

Or merge this one ( after the review) and address this comment https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/468#note_300848 in #1392

kinow:
>Maybe we can move this to 4.2.0 and implement the postgres lock and file lock all together @LuiggiTenorioK , @kinow

Works for me!

kinow:
In GitLab by @dbeltrankyl on Oct 3, 2024, 08:20

Also, related with #1392

kinow:
mentioned in commit 44b869f9a9aefe51da84afeed636d1420185ac58

kinow:
mentioned in commit 1d628b4778cb5ea23100e6d60a2308a4ca5a71e7

kinow:
mentioned in commit 1fc9cc344aa8f88a1d48c79778ee6f8e56d44797

kinow:
mentioned in commit a360eb748573467d99cf56758e5977b3796d15ae

kinow:
[...]

Issue #1342, opened 2024-07-08
[Regression test] Add the workflow dependencies comparison (4.0 vs 4.1 ) test to the pipeline


kinow:
In GitLab by @dbeltrankyl on Jul 8, 2024, 09:40

Hello @kinow 

Before I forget, I created this issue to upload the current regression test I have locally. This will be added when we enable the regression test.

Files:

1) [4.0_docs_test.txt](https://earth.bsc.es/gitlab/es/autosubmit/uploads/e45daa99704419129f1f4214dc11f117/4.0_docs_test.txt) # Do not touch this, unless more tests are added if so, modify 4) to run under 4.0

2) [4.1_docs_test.txt](https://earth.bsc.es/gitlab/es/autosubmit/uploads/e41db963e2a5626229a8ff41574ebcc6/4.1_docs_test.txt) # Generated by 4)

3) [autosubmit_docs.tar.gz](https://earth.bsc.es/gitlab/es/autosubmit/uploads/e637ad7dfb42896acdf26dea6dd614b0/autosubmit_docs.tar.gz) # db

4) test/regression/local_check_details.py # Execute script

5) CLI pipeline comparison # TODO, Right now, I use the pycharm ide

kinow:
In GitLab by @dbeltrankyl on Jul 8, 2024, 09:43

Added a new test case (a00l) derived from #1341

PR #1956, opened 2024-12-12
ASLock implemented

kinow:
In GitLab by @dbeltrankyl on Aug 7, 2024, 11:14

Implements #945.

This branch removes the portlocker library, and instead, it uses a new implementation without relying on libraries. 

This fixes the issue where the experiments could run multiple times, see the test to know how to use it, but it is similar to the portalocker ( use it with the with statement )

Ready to review @kinow @edgano, I don't know how much work you have right now as I was on holiday if nobody assigns itself I'll assign later one of you, thanks!.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -74,8 +74,8 @@ import re
 import random
 import signal
 import datetime
-# import log.fd_show as fd_show
-import portalocker
+import log.fd_show as fd_show
+from autosubmit.helpers.utils import ASLock
 from pkg_resources import require, resource_listdir, resource_string, resource_filename
 #import importlib
 from collections import defaultdict
@@ -2133,8 +2133,8 @@ class Autosubmit:
 
         # checking if there is a lock file to avoid multiple running on the same expid
         try:
-            # Portalocker is used to avoid multiple autosubmit running on the same experiment, we have to change this system in #806
-            with portalocker.Lock(os.path.join(tmp_path, 'autosubmit.lock'), timeout=1):
+            # ASLock is used to avoid multiple autosubmit running on the same experiment. #806
+            with ASLock(expid):
                 try:
                     Log.debug("Preparing run")
                     # This function is called only once, when the experiment is started. It is used to initialize the experiment and to check the correctness of the configuration files.
@@ -2325,9 +2325,6 @@ class Autosubmit:
                                 7051, e.message)
                     except AutosubmitCritical as e:  # Critical errors can't be recovered. Failed configuration or autosubmit error
                         raise AutosubmitCritical(e.message, e.code, e.trace)
-                    except (portalocker.AlreadyLocked, portalocker.LockException) as e:
-                        message = "We have detected that there is another Autosubmit instance using the experiment\n. Stop other Autosubmit instances that are using the experiment or delete autosubmit.lock file located on tmp folder"
-                        raise AutosubmitCritical(message, 7000)
                     except BaseException as e:
                         raise # If this happens, there is a bug in the code or an exception not-well caught
                 Log.result("No more jobs to run.")
@@ -2381,10 +2378,6 @@ class Autosubmit:
                         exp_history.finish_current_experiment_run()
                     except:
                         Log.warning("Database is locked")
-        except (portalocker.AlreadyLocked, portalocker.LockException) as e:
-            message = "We have detected that there is another Autosubmit instance using the experiment\n. Stop other Autosubmit instances that are using the experiment or delete autosubmit.lock file located on tmp folder"
-            terminate_child_process(expid)
-            raise AutosubmitCritical(message, 7000)
         except AutosubmitCritical as e:
             terminate_child_process(expid)
             raise
@@ -4020,7 +4013,7 @@ class Autosubmit:
         backup_pkl_path = os.path.join(
             pkl_folder_path, "job_list_{}_backup.pkl".format(expid))
         try:
-            with portalocker.Lock(os.path.join(tmp_path, 'autosubmit.lock'), timeout=1):
+            with ASLock(expid):
                 # Not locked
                 Log.info("Looking for backup file {}".format(backup_pkl_path))
                 if os.path.exists(backup_pkl_path):
@@ -4076,9 +4069,6 @@ class Autosubmit:
                 else:
                     Log.info(
                         "Backup file not found. Pkl restore operation stopped. No changes have been made.")
-        except (portalocker.AlreadyLocked, portalocker.LockException) as e:
-            message = "Another Autosubmit instance using the experiment\n. Stop other Autosubmit instances that are using the experiment or delete autosubmit.lock file located on the /tmp folder."
-            raise AutosubmitCritical(message, 7000)
         except AutosubmitCritical as e:
             raise AutosubmitCritical(e.message, e.code, e.trace)
         except BaseException as e:
@@ -4451,7 +4441,7 @@ class Autosubmit:
             exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, expid)
             tmp_path = os.path.join(exp_path, BasicConfig.LOCAL_TMP_DIR)
             # Encapsulating the lock
-            with portalocker.Lock(os.path.join(tmp_path, 'autosubmit.lock'), timeout=1) as fh:
+            with ASLock(expid):
                 try:
                     Log.info(
                         "Preparing .lock file to avoid multiple instances with same expid.")
@@ -4607,8 +4597,6 @@ class Autosubmit:
                     Log.result("\nJob list created successfully")
                     Log.warning(
                         "Remember to MODIFY the MODEL config files!")
-                    fh.flush()
-                    os.fsync(fh.fileno())
                     if detail:
                         Autosubmit.detail(job_list)
                     return True
@@ -4616,14 +4604,9 @@ class Autosubmit:
                 except KeyboardInterrupt as e:
                     # Setting signal handler to handle subsequent CTRL-C
                     signal.signal(signal.SIGINT, signal_handler_create)
-                    fh.flush()
-                    os.fsync(fh.fileno())
                     raise AutosubmitCritical("Stopped by user input", 7010)
                 except BaseException as e:
                     raise
-        except (portalocker.AlreadyLocked, portalocker.LockException) as e:
-            message = "We have detected that there is another Autosubmit instance using the experiment\n. Stop other Autosubmit instances that are using the experiment or delete autosubmit.lock file located on tmp folder"
-            raise AutosubmitCritical(message, 7000)
         except AutosubmitError as e:
             raise
         except AutosubmitCritical as e:
@@ -5160,7 +5143,7 @@ class Autosubmit:
         job_validation_message = " "
         # checking if there is a lock file to avoid multiple running on the same expid
         try:
-            with portalocker.Lock(os.path.join(tmp_path, 'autosubmit.lock'), timeout=1):
+            with ASLock(expid):
                 Log.info(
                     "Preparing .lock file to avoid multiple instances with same expid.")
 
@@ -5399,9 +5382,6 @@ class Autosubmit:
                                                 groups=groups_dict,
                                                 job_list_object=job_list)
                 return True
-        except (portalocker.AlreadyLocked, portalocker.LockException) as e:
-            message = "We have detected that there is another Autosubmit instance using the experiment\n. Stop other Autosubmit instances that are using the experiment or delete autosubmit.lock file located on tmp folder"
-            raise AutosubmitCritical(message, 7000)
         except (AutosubmitError, AutosubmitCritical):
             raise
         except BaseException as e:
--- a/autosubmit/database/repositories/locks.py
+++ b/autosubmit/database/repositories/locks.py
@@ -1,0 +1,204 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from datetime import datetime, timezone
+import os
+import json
+import sqlite3
+from typing import Optional
+from autosubmitconfigparser.config.basicconfig import BasicConfig
+
+
+@dataclass
+class LockModel:
+    """
+    Data model of the Autosubmit Lock
+    """
+
+    lock_id: str
+    hostname: str
+    pid: int
+    created: datetime = datetime.now(timezone.utc)
+
+
+class LocksRepository(ABC):
+    """
+    Interface for Autosubmit Locks Repository
+    """
+
+    @abstractmethod
+    def get_lock(self, lock_id: str) -> Optional[LockModel]:
+        """
+        Get the lock by lock_id. Return None if the lock does not exist.
+        """
+
+    @abstractmethod
+    def insert_lock_from_model(self, lock: LockModel):
+        """
+        Insert the lock
+        """
+
+    @abstractmethod
+    def insert_lock(self, lock_id: str, hostname: str, pid: int):
+        """
+        Insert the lock. Create a LockModel object first.
+        """
+
+    @abstractmethod
+    def delete_lock(self, lock_id: str):
+        """
+        Delete the lock
+        """
+
+
+class LocksFileRepository(LocksRepository):
+    """
+    Repository of Autosubmit Locks using files.
+    It creates a lock file for each lock which is stored in the directory.
+    """
+
+    def __init__(self, dir_path: str):
+        self.dir_path = dir_path
+        self._initialize()
+
+    def _initialize(self):
+        """
+        Create the locks directory if it does not exist
+        """
+        os.makedirs(self.dir_path, exist_ok=True)
+
+    def get_lock(self, lock_id: str) -> Optional[LockModel]:
+        """
+        Get the lock by lock_id. Return None if the lock does not exist.
+        """
+        lock_file = os.path.join(self.dir_path, lock_id)
+        if not os.path.exists(lock_file):
+            return None
+
+        # Load from JSON format
+        with open(lock_file, "r") as f:
+            lock_data = json.load(f)
+            return LockModel(
+                lock_data["lock_id"],
+                lock_data["hostname"],
+                lock_data["pid"],
+                datetime.fromisoformat(lock_data["created"]),
+            )
+
+    def insert_lock_from_model(self, lock: LockModel):
+        """
+        Insert the lock into the directory. Save content in JSON format.
+        """
+        lock_file = os.path.join(self.dir_path, lock.lock_id)
+        # Save in JSON format
+        with open(lock_file, "w") as f:
+            json.dump(
+                {
+                    "lock_id": lock.lock_id,
+                    "hostname": lock.hostname,
+                    "pid": lock.pid,
+                    "created": lock.created.isoformat(),
+                },
+                f,
+            )
+
+    def insert_lock(self, lock_id: str, hostname: str, pid: int):
+        """
+        Insert the lock into the directory. Create a LockModel object first.
+        """
+        new_lock = LockModel(lock_id, hostname, pid, datetime.now(timezone.utc))
+        self.insert_lock_from_model(new_lock)
+
+    def delete_lock(self, lock_id: str):
+        """
+        Delete the lock from the directory
+        """
+        lock_file = os.path.join(self.dir_path, lock_id)
+        os.remove(lock_file)
+
+
+class LocksDBRepository(LocksRepository):
+    """
+    Repository of Autosubmit Locks using SQLite
+    """
+
+    # TODO: Generalize with SQLAlchemy
+
+    def __init__(self, db_path: str):
+        self.db_path = db_path
+        self._initialize()
+
+    def _initialize(self):
+        """
+        Create the locks table if it does not exist
+        """
+        with sqlite3.connect(self.db_path) as conn:
+            conn.execute("""
+                CREATE TABLE IF NOT EXISTS locks (
+                    lock_id TEXT PRIMARY KEY,
+                    hostname TEXT,
+                    pid INTEGER,
+                    created TEXT
+                )
+            """)
+            conn.commit()
+
+    def get_lock(self, lock_id: str) -> Optional[LockModel]:
+        """
+        Get the lock by lock_id
+        """
+        with sqlite3.connect(self.db_path) as conn:
+            cursor = conn.cursor()
+            cursor.execute(
+                "SELECT lock_id, hostname, pid, created FROM locks WHERE lock_id = ?",
+                (lock_id,),
+            )
+            row = cursor.fetchone()
+            if row is None:
+                return None
+            return LockModel(
+                row[0],  # lock_id
+                row[1],  # hostname
+                row[2],  # pid
+                datetime.fromisoformat(row[3]),  # created
+            )
+
+    def insert_lock_from_model(self, lock: LockModel):
+        """
+        Insert the lock into the database
+        """
+        with sqlite3.connect(self.db_path) as conn:
+            conn.execute(
+                "INSERT INTO locks VALUES (?, ?, ?, ?)",
+                (lock.lock_id, lock.hostname, lock.pid, lock.created.isoformat()),
+            )
+            conn.commit()
+
+    def insert_lock(self, lock_id: str, hostname: str, pid: int):
+        """
+        Insert the lock into the database. Create a LockModel object first.
+        """
+        new_lock = LockModel(lock_id, hostname, pid, datetime.now(timezone.utc))
+        self.insert_lock_from_model(new_lock)
+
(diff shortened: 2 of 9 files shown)


kinow:
In GitLab by @kinow on Aug 7, 2024, 16:09

Commented on [test/regression/local_check_details.py line 47](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..3ba256fdcc4dbf99f648b4258244c3be38308312#diff-94d06f524eec539adab862de0fbda2b5R47)



Prints are ignored in pytest output. Maybe this can be commented out, so others can uncomment if they need to run it locally/debug? There are flags to enable the printing in pytest tests, but it's not really considered a good practice (e.g. most times that I would run the unit tests I would not be interested/know what these print statements mean, unless I am actually touching this code).

kinow:
In GitLab by @kinow on Aug 7, 2024, 16:09

Commented on [test/unit/test_utils.py line 21](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..a7ebbd3879a4c74919ae98a9fe72d6e826bc643b#diff-73b8caee56b662fa0be91fe41da42694R21)



`/tmp` may not exist/be readable depending on platform or if it's manipulated (`mktemp`, `$TEMP`, `tmp.dir`, etc...). Probably safer to use that pytest fixture for temporary paths (`tmp_path`, I think?).

kinow:
In GitLab by @kinow on Aug 7, 2024, 16:09

Commented on [setup.py line 48](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..7cb77c570e3644ba1b491fff7ce7bcb85d4148e6#diff-2eeaed663bd0d25b7e608891384b7298R48)



Less dependencies are good! 

Note that `pytest` is not included in the `setup.py`. Our dependencies are a bit of a mess right now... the requirements.txt probably doesn't match either.

Test dependencies are installed on our workers, and include the trio `pytest pytest-cov pytest-mock`. `pytest-mock` provides a fixture (`mocker`) that's compatible with the `mock`...

But I am not sure if [`pytest-mock`](https://github.com/pytest-dev/pytest-mock/blob/d7a69e35164b5edc2978b6565a670e0a9b8ecfa1/pyproject.toml) brings `mock`... we do have that in other tests... so interesting case. Will dig a bit more into this later to understand what can be done to simplify it a bit.

This should be fixed later by !435

kinow:
In GitLab by @kinow on Aug 7, 2024, 16:09

+1 to removing the `portalocker` dependency if possible, @dbeltrankyl ! :clap: 

Less dependencies is normally == to less security/upgrades/python version/API issues, especially for something small like this.

Left a few comments, but nothing major. What do you think of this final test for this review?

1. Create a small script with `portalocker` creating a lock that should work locally/MN5, but fail on LUMI
2. Create the equivalent script with this new code, that should work both on local/MN5, and also on LUMI

That way we should be able to be more confident that this will fix the issue on LUMI (where I think is the only platform where portalocker is not working?), while also keep backward compatibility.

I can create these scripts and do some tests later.

Thanks!!!!

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:17

Commented on [test/regression/local_check_details.py line 47](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..3ba256fdcc4dbf99f648b4258244c3be38308312#diff-94d06f524eec539adab862de0fbda2b5R47)



Ah, this didn't mean to be in the commit. Also, this test is not used by anybody ( only me ) (#1342).

I was using this to test the Leo issues with the performance of the testing suite

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:18

requested review from @kinow

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:29

Hello @kinow ,

Thanks for looking at this. 

The portalocker was not working fine on all the platforms, issue #945 has been open since the start of Autosubmit 4.0, so I am unsure if it is necessary to do that test, (This was also happening in autosubmit 3.13,3.14 and 3.15)

Also, while working on this, I found that the autosubmit always deleted the lock file on termination. I'm unsure when that was added, but that also caused issues with the new implementation. So, the port locker may be working in the Lumi VM in case that was removed, although I think the new solution is better as it will work under our Bsc machines.

> while also keep backward compatibility.

About this, I didn't think about what happens if an experiment is running with <4.1.9 and is launched with this branch. In that case, multiple-running experiments can happen, or autosubmit will complain in the new implementation that the file is locked and crashes.

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:31

Commented on [test/regression/local_check_details.py line 47](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..3ba256fdcc4dbf99f648b4258244c3be38308312#diff-94d06f524eec539adab862de0fbda2b5R47)



changed this line in [version 2 of the diff](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=29692&start_sha=3ba256fdcc4dbf99f648b4258244c3be38308312#f480549edf9c74e64e41b75655e8bb12118a3896_47_45)

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:31

added 1 commit

<ul><li>a7ebbd38 - Removed time elapsed</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=29692&start_sha=3ba256fdcc4dbf99f648b4258244c3be38308312)

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:56

Commented on [test/unit/test_utils.py line 21](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..a7ebbd3879a4c74919ae98a9fe72d6e826bc643b#diff-73b8caee56b662fa0be91fe41da42694R21)



changed this line in [version 3 of the diff](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=29695&start_sha=a7ebbd3879a4c74919ae98a9fe72d6e826bc643b#d0271c3b8d64cc1fe43e666ce63802cfdf216112_21_30)

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:56

added 1 commit

<ul><li>2b1e8b71 - Changed tmp_path</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=29695&start_sha=a7ebbd3879a4c74919ae98a9fe72d6e826bc643b)

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:56

Commented on [test/unit/test_utils.py line 21](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..a7ebbd3879a4c74919ae98a9fe72d6e826bc643b#diff-73b8caee56b662fa0be91fe41da42694R21)



I think that pytests also uses the /tmp folder by default but can be changed with pytest --basetemp=mydir.

The test is now using tmp_path_factory

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 08:57

added 1 commit

<ul><li>470417e2 - removed unused import</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=29697&start_sha=2b1e8b7177201b1d538180ad7c405ec647609acc)

kinow:
In GitLab by @dbeltrankyl on Aug 8, 2024, 09:15

added 1 commit

<ul><li>4bdb2c85 - Changed the other test aswell</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=29699&start_sha=470417e2d90b642015a9f1cceed001e9ab7481e3)

kinow:
In GitLab by @kinow on Aug 8, 2024, 10:18

Got it, so I can test on the BSC Hub, and review this part.

kinow:
In GitLab by @LuiggiTenorioK on Aug 19, 2024, 19:22

Commented on [autosubmit/helpers/utils.py line 22](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..4bdb2c85b70288ec0ea84882dd92a5379cd3f8b5#diff-cc3c5b640446e4718e7eca934be71647R22)



I think, instead of using always `autosubmit.lock` for the filename, it should be a variable. For example, in the API, we have another file called `calculation_in_progress.lock` to avoid interfering with this lock.

kinow:
In GitLab by @LuiggiTenorioK on Sep 30, 2024, 18:40

Commented on [autosubmit/helpers/utils.py line 22](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..4bdb2c85b70288ec0ea84882dd92a5379cd3f8b5#diff-cc3c5b640446e4718e7eca934be71647R22)



changed this line in [version 6 of the diff](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=32305&start_sha=4bdb2c85b70288ec0ea84882dd92a5379cd3f8b5#ff161687c56ba0a2dd5ce8f3f9c979f7cb2b6dbe_22_39)

kinow:
In GitLab by @LuiggiTenorioK on Sep 30, 2024, 18:40

added 1 commit

<ul><li>7cb77c57 - update ASLock to use SQLite</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=32305&start_sha=4bdb2c85b70288ec0ea84882dd92a5379cd3f8b5)

kinow:
In GitLab by @LuiggiTenorioK on Sep 30, 2024, 18:52

Commented on [setup.py line 48](https://github.com/BSC-ES/autosubmit/compare/30a1bd85e2eecbfbf5ffea92b68ba90196c92397..7cb77c570e3644ba1b491fff7ce7bcb85d4148e6#diff-2eeaed663bd0d25b7e608891384b7298R48)



changed this line in [version 7 of the diff](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=32307&start_sha=7cb77c570e3644ba1b491fff7ce7bcb85d4148e6#8e2edce0d507e1297474f25c00cae94258db38d8_48_98)

kinow:
In GitLab by @LuiggiTenorioK on Sep 30, 2024, 18:52

added 69 commits

<ul><li>7cb77c57...e91258be - 63 commits from branch <code>master</code></li><li>d06ca109 - ASLock implemented</li><li>0413737a - Removed time elapsed</li><li>2f140ce8 - Changed tmp_path</li><li>5c62a457 - removed unused import</li><li>137a05bd - Changed the other test aswell</li><li>5090ee01 - update ASLock to use SQLite</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=32307&start_sha=7cb77c570e3644ba1b491fff7ce7bcb85d4148e6)

kinow:
In GitLab by @LuiggiTenorioK on Sep 30, 2024, 18:58

@dbeltrankyl @kinow I changed the ASLock to acquire the lock by creating a row in an SQLite table instead of creating a file. This is because in the future we want to integrate it with Postgres easily. 

The only catchup here is that we will then need to implement a command to force-release the lock since the user can no longer do it by deleting the file.

If you feel this will not work, feel free to revert my commit.

kinow:
In GitLab by @kinow on Oct 1, 2024, 10:03

I think the reason for not using sqlite before was that under certain conditions sqlite may not effectively "lock". Even though it supports multi-threaded/user, there are cases where you may run into inconsistencies due to NFS (we also had that in NZ/NIWA, either due to how the filesystem was organized, or sometimes due to IBM bugs).

So as long as that's taken into account, and it works with older versions (backward comp) it should be OK, I think.

kinow:
In GitLab by @LuiggiTenorioK on Oct 1, 2024, 11:20

Sure, using SQLite for locking might not be the best (as well as using the NFS). However, the idea is to make space to extend it and use Postgres in the future. Then, we have to make it clear that using Postgres will be the best and using SQLite will be the choice for not highly concurrent environments where autosubmit is running.

kinow:
In GitLab by @kinow on Oct 1, 2024, 11:25

Yup! Postgres for environments where you need a distributed DB, and sqlite for when you want to use the old architecture of one DB mapped over NFS.

We could also write to disk by default, but have an option to write to a row on sqlite for when users are 100% sure that there is no risks of concurrency issues... but we might not need it for now...

kinow:
In GitLab by @dbeltrankyl on Oct 1, 2024, 11:25

Hello, 

What about using a file lock or/and sqlite lock and another option for Postgres?

> here are cases where you may run into inconsistencies due to NFS

I'm mainly thinking about our BSC environment, where the databases occasionally become locked/corrupted for no special reason other than being slow or being under the NFS system.

BSC ( hubs/auto submit machines) is a special case but also the most important place, and as Bruno commented, adding this SQLite lock by default could cause issues.

We could just activate one or the other through the autosubmitrc file

kinow:
In GitLab by @LuiggiTenorioK on Oct 1, 2024, 11:49

> BSC ( hubs/auto submit machines) is a special case but also the most important place, and as Bruno commented, adding this SQLite lock by default could cause issues.

Taking that into account, it is true that we should continue using the file system as a default to prevent other additional issues. Still, I'll keep this implementation and just change the persistence (from SQLite to FS again) and then extend it using SQLAlchemy.

The main idea of my latest change was to generalize the lock mechanism (each lock has its `lock_id` and is independent of the `expid`) rather than using SQLite.

kinow:
In GitLab by @LuiggiTenorioK on Oct 1, 2024, 18:02

added 1 commit

<ul><li>9e9bec29 - update ASLock to generalize persistance layer</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=32367&start_sha=5090ee01efa06ec0e7fa0f1b5c8fabea6b4b82ba)

kinow:
In GitLab by @LuiggiTenorioK on Oct 1, 2024, 18:13

added 1 commit

<ul><li>86bb67d8 - update ASLock to generalize persistance layer</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/468/diffs?diff_id=32371&start_sha=9e9bec29b9a0046451f4729d4ec39960683e7a45)

kinow:
In GitLab by @LuiggiTenorioK on Oct 1, 2024, 18:31

Changed back to saving the lock as a file. The difference now is that the locks will be stored inside the `{LOCAL_ROOT_DIR}/locks` directory.

Note: for extending it with another type of persistence (e.g. SQLite, Postgres) we will only need to add the child class in the `autosubmit.database.repositories.locks.create_locks_repository` function without affecting the actual `ASLock` class.

I left some TODOs for the future but, for now, I think it can be merged.

kinow:
In GitLab by @LuiggiTenorioK on Oct 3, 2024, 10:49

@kinow @dbeltrankyl As I mentioned in today's meeting, this is ready to be merged for Autosubmit 4.1.11, if desired.

If not, it can be moved once we close 4.1.* versions.

kinow:
In GitLab by @kinow on Oct 25, 2024, 12:59

Commented on [bin/autosubmit line 43](https://github.com/BSC-ES/autosubmit/compare/e91258be4e4424a3bd00ea12e8fc680d1c2ffc8a..86bb67d8322311d5f8f7b72ce3e9075ea8197108#diff-2fe7b98febd0e3dbdfc7ac517eb3edceL43)



Shouldn't the code here call the repository function to `delete_lock`?

kinow:
In GitLab by @kinow on Oct 25, 2024, 12:59

Commented on [bin/autosubmit line 61](https://github.com/BSC-ES/autosubmit/compare/e91258be4e4424a3bd00ea12e8fc680d1c2ffc8a..86bb67d8322311d5f8f7b72ce3e9075ea8197108#diff-2fe7b98febd0e3dbdfc7ac517eb3edceL61)



Ditto here.

kinow:
In GitLab by @kinow on Oct 25, 2024, 12:59

Finished reading the code, left a questoin, and `autosubmit/autosubmit.py` has Git conflicts which need to be fixed. For next week I will try to reproduce the bug with porta locker, and then confirm that with this branch I cannot reproduce this any longer.

kinow:
In GitLab by @LuiggiTenorioK on Oct 25, 2024, 15:06

Commented on [bin/autosubmit line 43](https://github.com/BSC-ES/autosubmit/compare/e91258be4e4424a3bd00ea12e8fc680d1c2ffc8a..86bb67d8322311d5f8f7b72ce3e9075ea8197108#diff-2fe7b98febd0e3dbdfc7ac517eb3edceL43)



I'm not sure this worked before since `Log.file_path` was never set. Then, we don't have a way to know which lock to release from here. (Remember that we can create multiple locks, usually per expid)

Either way, we discussed about having an additional command to force-release the existing locks.

kinow:
In GitLab by @kinow on Oct 28, 2024, 12:25

Trying to reproduce this today, as I think we should try to carefully test it on our laptops and on the VM if it's something simple, and maybe involve CSC (@utigerstedt and @Nortamo) early before we merge anything and release, to avoid regressions and issues later on.

Created a workflow that runs a single task for 5 minutes. I'm using the latest version deployed in the ClimateDT VM, and the workflow is running there.

Confirmed already the tmp file is created fine.

```bash
[brunokino@climatedt-wf tmp]$ ls -lah  | grep .lock
-rw-r-----. 1 brunokino autosubmit_users    0 Oct 28 12:37 autosubmit.lock
[brunokino@climatedt-wf tmp]$ lsof autosubmit.lock
      Output information may be incomplete.
COMMAND       PID      USER   FD   TYPE DEVICE SIZE/OFF      NODE NAME
autosubmi 1495282 brunokino    7wW  REG  253,2        0 940984867 autosubmit.lock
autosubmi 1495300 brunokino    7w   REG  253,2        0 940984867 autosubmit.lock
[brunokino@climatedt-wf tmp]$ ps -p 1495282,1495300
    PID TTY          TIME CMD
1492556 pts/3    00:00:03 autosubmit
1492567 pts/3    00:00:00 autosubmit log 
```

So the lock file was created, and it's used by two processes, `autosubmit run` and by `autosubmit log`. But the lock is owned/created by `autosubmit run`.

```bash
[brunokino@climatedt-wf tmp]$ lslocks -p 1495282
COMMAND             PID  TYPE SIZE MODE  M START END PATH
autosubmit      1495282 FLOCK   0B WRITE 0     0   0 /appl/AS/AUTOSUBMIT_DATA/a1ts/tmp/autosubmit.lock
[brunokino@climatedt-wf tmp]$ lslocks -p 1495300
```
While the workflow is still running, I ran an `autosubmit create`.

```bash
[brunokino@climatedt-wf tmp]$ autosubmit create a1ts
/appl/AS/4.1.10/lib64/python3.9/site-packages/paramiko-3.4.0-py3.9.egg/paramiko/pkey.py:100: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "cipher": algorithms.TripleDES,
/appl/AS/4.1.10/lib64/python3.9/site-packages/paramiko-3.4.0-py3.9.egg/paramiko/transport.py:259: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "class": algorithms.TripleDES,
Autosubmit is running with 4.1.10
 [CRITICAL] We have detected that there is another Autosubmit instance using the experiment
. Stop other Autosubmit instances that are using the experiment or delete autosubmit.lock file located on tmp folder [eCode=7000]
[brunokino@climatedt-wf tmp]$ lslocks -p 1495282
COMMAND             PID  TYPE SIZE MODE  M START END PATH
autosubmit      1495282 FLOCK   0B WRITE 0     0   0 /appl/AS/AUTOSUBMIT_DATA/a1ts/tmp/autosubmit.lock (deleted)
```

This shows that the lock has been deleted. Issuing `autosubmit create` again now succeeds, since there is no more lock (it was deleted). After running that experiment a few times, I decided the tests on the VM were good enough (I have nothing else running there).

I used a debugger locally, before trying `pdb` on the ClimateDT VM, and looks like the lock is removed not by an issue on PortaLocker, but as an exception is caught in `autosubmit create`, which then triggers the code `exit_from_error` from `autosubmit.py`.

That code will do this:

image

So the PortaLocker exceptions for when a file is locked raise an AutosubmitCritical, which is captured at the final callback for errors, which on its turn deletes the `auosubmit.lock` file.

To confirm this theory, I issued another `lslocks` just before letting the debugger move on.

image

Then as soon as the debugger moves forward, and the file has been deleted,

image

we can see in the `lslocks` that it confirms that the process is running, and it has a lock onto a file, but it has been deleted. Thus the portalocker bug on the ClimateDT VM, which is not locking and preventing multiple instances or running create multiple times.

image

I'm +1 to organizing the code so that Postgres changes can be accommodated. But I'm -1 to removing the existing library & code that has been working for a while and to use a new solution, especially with e/o-suites so close.

IMHO we should be able to re-work the logic in the exception handling to avoid deleting the lock for PortaLocker exceptions. That should fix the error, and then in another change or on this MR we remove the new lock mechanism and just add whatever is needed for Postgres, but we merge that in 4.2.0, after the o-suite is live.

kinow:
In GitLab by @LuiggiTenorioK on Oct 28, 2024, 14:39

> So the PortaLocker exceptions for when a file is locked raise an AutosubmitCritical, which is captured at the final callback for errors, which on its turn deletes the `auosubmit.lock` file.

Good catch! 

> IMHO we should be able to re-work the logic in the exception handling to avoid deleting the lock for PortaLocker exceptions. That should fix the error, and then in another change or on this MR we remove the new lock mechanism and just add whatever is needed for Postgres, but we merge that in 4.2.0, after the o-suite is live.

You are right! We have to be more careful about this feature since it could be more complex than it seems. Maybe just listing the requirements we must fulfill could help us arrive at a better logic and write tests to fulfill acceptance criteria. Let me start:

* Autosubmit should create locks that are independent between them (e.g. for each experiment)
* In lock creation, if there is already a lock with the same lock_id, it should fail
* **Autosubmit should delete only the locks that are related to the pid when exiting**
* [EDITO] Locks could be stored either in the file system or Postgres DB

Maybe I'm missing something but this should be easily translated to tests.

kinow:
In GitLab by @kinow on Oct 30, 2024, 08:45

>You are right! We have to be more careful about this feature since it could be more complex than it seems. Maybe just listing the requirements we must fulfill could help us arrive at a better logic and write tests to fulfill acceptance criteria. Let me start:

+1

>Autosubmit should create locks that are independent between them (e.g. for each experiment)
>In lock creation, if there is already a lock with the same lock_id, it should fail

:thumbsup:, I think that's how it worked before, with one lock file in the experiment's `tmp` folder.

>**Autosubmit should delete only the locks that are related to the pid when exiting**

Yup, and I think this worked, until some other change may have broken it. We can fix it and include a unit test to prevent future regressions like this one.

> \[EDITO\] Locks could be stored either in the file system or Postgres DB

+1! And this can be done in the Postgres branch. For 4.1.11 I think we must fix the lock first. Then for Postgres we basically use the approach here, or similar approach, to have both a file-based lock, and the Postgres lock.

kinow:
@dbeltrankyl I'm going through my list of ~16 open PRs to triage what I can update now. This PR appears as mine, but it's actually yours (GitHub migration). 

I think we solved the AS lock another way and this can be closed?

kinow (commented):
@dbeltrankyl I'm going through my list of ~16 open PRs to triage what I can update now. This PR appears as mine, but it's actually yours (GitHub migration). 

I think we solved the AS lock another way and this can be closed?

dbeltrankyl:
> I think we solved the AS lock another way and this can be closed?

We did? I mean, there were plans to do it for the postgres to do a DB-based lock, but are they applied? @LuiggiTenorioK

kinow:
> > I think we solved the AS lock another way and this can be closed?
> 
> We did? I mean, there were plans to do it for the postgres to do a DB-based lock, but are they applied? @LuiggiTenorioK

At least part of this. This PR replaces portalocker and also addresses the custom locker for DB & disk. Portalocker was working fine but there was some logic issue, fixed in this PR https://github.com/BSC-ES/autosubmit/issues/2000

I think doing one thing at a time may be easier? i.e. one PR to make the class that handles DB & portalocker, and then another PR to replace portalocker if we really need to.

dbeltrankyl:
Don't delete the branch

Issue #1389, opened 2024-08-12
Autosubmit 4.1.9 recover_job_logs function performance downgrade


kinow:
Hello @dbeltrankyl,

### Autosubmit Version

4.1.9

### Expid affected(If applicable)

Any

### Which task has issues? Where is the log(If applicable)

Any with remote logs to be retrieved

### Summary

The CSC team identified that there were several `ps` processes being launched by Autosubmit, basically DOS'ing the machine.

They identified the [part of the code](https://earth.bsc.es/gitlab/es/autosubmit/-/blob/46697dbc6501543267def17bc2643dfd22571095/autosubmit/platforms/platform.py#L835) where this is coming from.

It appears that this is from our solution to the orphaned processes in 4.1.8 (which had a solution for another issue with logs in 4.1.7, and ditto the same for 4.1.6...).

One possible solution for that is to add a sleep period, so the `ps` processes are launched in a more spaced manner.

Another possible solution is the use of `psutil` or another Python lib/code to verify that the main process is still alive. Or, yet, rethink the design (yet again, I am afraid) of these thread/processes and make it so that the spawned thread/process "works", searching for logs, until the parent process tells it to stop. We use `_exit` I think, which I suspect would kill the main process and leave orphan processes... maybe that would be required for a change like this.

Anyway, it would be nice to have 4.1.10 out with a fix for this, and avoid a new issue caused by a change in our releases.

I haven't worked on this part of the code yet, but either @edgano or I can help you if you need, @dbeltrankyl.

### Steps to reproduce

Launch several experiments, that will launch several log retrieval processes, that will launch even more `ps` commands.

### What is the current bug behavior?

The performance indicators of computers running Autosubmit show a clear increase in CPU usage after 4.1.9.

### What is the expected correct behavior?

No orphan/zombies, and performance is the same or better than previous versions.

### Relevant logs and/or screenshots(if applicable)

This is from the message on Mattermost, to show how severe this change was:

```bash
We are seeing something like 20-30 ps calls per second on the machine which is hammering the machine.
```

### Any other relevant information(if applicable)

IMHO, for the merge request to fix this, we **must** use code review, and do thorough testing. In past merge requests we skipped this due to time/deadlines, but it's risky as the testing suite only spots punctual issues in the workflow. Issues like this one are caught only running AS in an env like ClimateDT, or by scrutinizing the code in a code review.

()

kinow:
[...]

PR #1957, opened 2024-12-12
time sleep added

kinow:
In GitLab by @dbeltrankyl on Aug 12, 2024, 15:04

#1389: Added a sleep between log petitions


kinow:
In GitLab by @dbeltrankyl on Aug 12, 2024, 15:04

requested review from @kinow

kinow:
In GitLab by @dbeltrankyl on Aug 12, 2024, 15:12

#1389

kinow:
In GitLab by @kinow on Aug 12, 2024, 15:40

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



@Nortamo, @utigerstedt, this is a hotfix based on 4.1.9 to add a sleep, currently of 10 seconds between each loop. As we don't have this configurable at the moment, I thought it'd be worth checking with you before we prepare this release.

Do you think this interval is good for the VM?

Dani created this branch off the v4.1.9 tag, as we had other changes already in `master`, and we will tag/release 4.1.0 from this branch, and merge this commit into `master` once it's done. We are doing it this way to minimize the number of changes, and to avoid other issues as the other changes on `master` have not yet been tested with the testing suite.

Cheers

kinow:
In GitLab by @kinow on Aug 12, 2024, 15:47

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



I think maybe this value could be a bit higher... 60, 120, maybe 180 seconds...? In a simulation that takes normally hours to finish that should be acceptable?

While this loop sleep, the process might appear as a zombie as the parent could have exit'ed without waiting (or `.join()` in Python, which calls the wait...).

But I think it's OK that a process appears as a zombie briefly (although we can modify the code in the future, I think).

kinow:
In GitLab by @Nortamo on Aug 12, 2024, 15:55

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



A quick test showed that 8 experiments were producing about ~100 ps calls per second. 
-> 13 / process / second. So 10 sec sleep should already be a major improvement, but 60sec would leave a lot of margin to increase the number of concurrent experiments, so if you can live with that it would be great to set it to e.g 60. How long something appears as a zombie is not an issue as long as we don't have any cleaning -> if they anyway clean themselves up after a minute max then that's fine and no cleaning is required on the system side.

kinow:
In GitLab by @kinow on Aug 12, 2024, 15:58

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



All good with 60 seconds for me too. Thanks @Nortamo !

kinow:
In GitLab by @kinow on Aug 12, 2024, 15:59

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



@dbeltrankyl so I think we only need to change the timeout to 60 here, add a one-entry/line changelog for 4.1.10, and update the version too. Then it should be ready to be approved & released, and later merged.

kinow:
In GitLab by @dbeltrankyl on Aug 12, 2024, 16:25

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



changed this line in [version 2 of the diff](/gitlab/es/autosubmit/-/merge_requests/469/diffs?diff_id=29854&start_sha=8c0d14acea11311a4868dd9f4ed9188728c03480#e00ded6b6df8370954fc259c5e67a43145a32028_838_839)

kinow:
In GitLab by @dbeltrankyl on Aug 12, 2024, 16:25

added 1 commit

<ul><li>ff74beca - added a while for download all the last logs</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/469/diffs?diff_id=29854&start_sha=8c0d14acea11311a4868dd9f4ed9188728c03480)

kinow:
In GitLab by @dbeltrankyl on Aug 12, 2024, 16:29

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



Thanks to both, 

I had to modify the code slightly so it recovers all the remaining logs after the main process exit 

```python
        max_logs_to_process = 60
        while not event.is_set() and os.system(f"ps ax | grep {str(self.main_process_id)} | grep -v grep > /dev/null 2>&1") == 0:
            time.sleep(60)
            logs_processed = 0 # avoid deadlocks just in case
            try:
                while not self.recovery_queue.empty() and logs_processed < max_logs_to_process:
                    logs_processed += 1
```

if you're okay with this, tomorrow at 8:00am I can make a release with these changes

kinow:
In GitLab by @kinow on Aug 12, 2024, 16:40

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



That sounds like a separate issue, no @dbeltrankyl ? Wouldn't it be best to pass that through the testing suite/tests/etc? And have in this change only what's impacting the performance in the VM?

>if you're okay with this, tomorrow at 8:00am I can make a release with these changes

All good for me. I could prepare it today, but we will have to work with users to restart their experiments on this version anyway, so it won't be immediate even after the release. But if we can fix this before the holiday (Thursday in Spain) that'd be better, I think.

kinow:
In GitLab by @dbeltrankyl on Aug 12, 2024, 16:47

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



> That sounds like a separate issue, no @dbeltrankyl ? Wouldn't it be best to pass that through the testing suite/tests/etc? And have in this change only what's impacting the performance in the VM?

The problem is that this change, the time.sleep, will cause this issue. If there is no delay, the issue won't arise due to being a race condition, but with the timer set to 60s it will be shown for sure

kinow:
In GitLab by @utigerstedt on Aug 13, 2024, 07:41

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



ps ax | grep {str(self.process_id)} | grep -v grep 

will match 1%process_id, 2%process_id, %process_id4 and a whole lot more (including numbers in the process names or command line options)
Whoever is thinking up this, please just stop. 

If you are checking if the parent is alive to kill the child from within the child, you are doing it the wrong way. The parent should supervise the children.

kinow:
In GitLab by @kinow on Aug 13, 2024, 08:19

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



It's a bit unconventional for sure (I don't remember seeing this in Java/Python yet), but I think this was the fix for the orphaned processes (I believe there was a reason for that, something how AS launches the processes -- which used to be threads before?).

Said all that, for me the main concern is how long that'd take to improve.

kinow:
In GitLab by @kinow on Aug 13, 2024, 09:21

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



Hi @dbeltrankyl 

>The problem is that this change, the time.sleep, will cause this issue. If there is no delay, the issue won't arise due to being a race condition, but with the timer set to 60s it will be shown for sure

I'm trying to understand what else had to be changed.

Looks like you are adding a check if the queue is empty, `while not self.recovery_queue.empty()`, before retrieving the next item from the queue, which sounds fine.

Why did you have to limit it to process 60 logs as well?

Can you point where the race condition happened, if it's not too complicated?

kinow:
In GitLab by @dbeltrankyl on Aug 13, 2024, 09:38

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



Hi @kinow , 

> Why did you have to limit it to process 60 logs as well?

I wanted to be sure that it doesn't end in a deadlock in case the queue is never empty for some reason or is inaccessible in cases where the main process got kill -9.

> Can you point where the race condition happened, if it's not too complicated?

We will have stacked logs with the timeout set to 60seconds.

When the timeout was as fast as the CPU could process, this would unlikely happen as the queue would be emptied whenever there was a new log, and maybe we only had this issue when the autosubmit had force killed.

kinow:
In GitLab by @kinow on Aug 13, 2024, 09:44

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



Got it. We can review this later and see how to prevent that from happening, and making it configurable so site admins can configure it.

I think we can release this version, and since Leo is running the testing suite, I can ask him if he'd launch one of the experiments with this new version (once deployed to the VM), or I can do that myself if he's too busy.

kinow:
In GitLab by @kinow on Aug 13, 2024, 09:44

approved this merge request

kinow:
In GitLab by @dbeltrankyl on Aug 13, 2024, 10:00

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



@kinow uploaded

kinow:
In GitLab by @kinow on Aug 13, 2024, 14:38

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



I confirmed in the morning that it was successfully deployed, thanks @dbeltrankyl .

Leo had two test cases working with 4.1.9, so I picked one of these two, confirmed I had a copy (`t01v`), and prepared a new run.

It is now running (should take a few more hours, queues look bad), and CSC is aware of the new process/pid to analyze if this version is looking better. The testing suite test case is also a quick way to at least try to have some validation, even though the fix we applied was only a few lines of code.

kinow:
In GitLab by @kinow on Aug 14, 2024, 14:11

Commented on [autosubmit/platforms/platform.py line 838](https://github.com/BSC-ES/autosubmit/compare/ebb0aa027f081b27faf18d0909e83cf67476a702..8c0d14acea11311a4868dd9f4ed9188728c03480#diff-5d19b70fc06bfdd8c8b4686d4aa7fc24R838)



Should be good to be merged once it's uploaded to PyPI, I think :thumbsup:

kinow:
In GitLab by @dbeltrankyl on Aug 20, 2024, 09:11

marked this merge request as **ready**

kinow:
In GitLab by @dbeltrankyl on Aug 20, 2024, 09:12

mentioned in commit 007ecd318120a46b9fb6a212b742574e71fac716

PR #1963, opened 2024-12-13
4.1.11 Log recovery rework Zombies, orphan, performance...

kinow:
In GitLab by @dbeltrankyl on Aug 23, 2024, 14:22

Hello @kinow , @mcastril 

Removed original description as it is deprecated and misleading. 

Check this comment or the whole thread: https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/475#note_303398

* [ ] Doc
* [ ] Test in climate-dt
* [X] Unit test if possible, but it is hard as multiple processors need to connect to a remote platform...

Edit: 
* [ ] Check the Autosubmit stop command
* [X] Check https://earth.bsc.es/gitlab/es/autosubmit-gui/-/issues/113#note_303210. This was more complicated than expected. Merged on this branch, changes from https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/490


kinow:
In GitLab by @kinow on Aug 26, 2024, 08:57

Hi @dbeltrankyl ! Thanks for the diagram and explanation.

I think since we have other options, besides the intermediary process, our first step before reviewing this must be trying to reproduce the issue from the climate dt vm in a separated environment, preferably one that others can test too (e.g. simple instructions for any linux OS, or a container shared as in the edito issue where we had zombies #1372 ).

Once we reproduce the problem, then I think it'd be easy to assess and to review a fix like this one. We can ask for help from @Nortamo and @utigerstedt if we are not able to reproduce it easily on our laptops (but first we must try that ourselves for a while, I suggest).

I believe that this will also help us to fix this issue for once and all, since we had a few hiccups in 4.1.6, 4.1.7, 4.1.8, 4.1.9, and now 4.1.10, as we only tested this with the testing suite on the VM (which does not detect all issues), without doing some local repro & debugging.

kinow:
In GitLab by @utigerstedt on Aug 26, 2024, 10:05

The whole "I need to monitor my parent/grandparent to see if they are still alive" is ass-backwards and not the way you are supposed to do it.

The AS VM has 
[tigerste@climatedt-wf ~]$ cat /proc/sys/kernel/pid_max
4194304

so the PIDs wrap around. Having the child check for existens of something older is prone to random behaviour, 
the whole idea with children and wait() is that the dead childs PID is not reused until the parent checks it.

Have the main process do a loop with the python equivalent of waitpid(-1,&wstatus, WNOHANG) and send it a signal when it should die.

kinow:
In GitLab by @dbeltrankyl on Aug 27, 2024, 16:32

added 2 commits

<ul><li>c62f00c9 - changes to signals and waitpid</li><li>94281bd5 - Added cleanup signal</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=30455&start_sha=d323764e833b83af43d95f1c41dfa8fd6c228d27)

kinow:
In GitLab by @dbeltrankyl on Aug 27, 2024, 17:00

added 1 commit

<ul><li>8112fa74 - cleanup code</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=30460&start_sha=94281bd55587ca3902fd950294128ab0a4266b42)

kinow:
In GitLab by @dbeltrankyl on Aug 27, 2024, 17:01

added 1 commit

<ul><li>4d64c54b - cleanup code</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=30462&start_sha=8112fa74437d9491933f50260522f4695896e319)

kinow:
In GitLab by @dbeltrankyl on Aug 27, 2024, 17:12

Hello @utigerstedt , 

> The whole "I need to monitor my parent/grandparent to see if they are still alive" is ass-backwards and not the way you are supposed to do it.

It was unnecessary when Autosubmit was finished successfully or with a control+c or a Sigterm. 

However, it was required ( in my system ) only when Autosubmit received the signal kill -9 for the autosubmit main PID process, and only that process. 

Without the pid checker, the child would keep waiting for the event to be set to exit the loop, but since the main process was dead, the child never ended. ( This is the issue )

But you're right that looking at the PID is not a good solution, so I implement another approach.

> Have the main process do a loop with the python equivalent of waitpid(-1,&wstatus, WNOHANG) 

Changed, and now it is doing this instead of the other approach of the child/grandchild. Thanks this is less complex

> and send it a signal when it should die.

Autosubmit already did this, and it worked fine without the PID checker.

However, at least in my system, this was not enough when only the PID associated with the autosubmit main process was forced killed, so the PID checker was added.  (I tested that and got zombies, but I didn't test it in the climate-dt)

I see now why checking the parent PID is not a good solution. Thanks for that. 

------------------------------------------------------

So after the last commit, what Autosubmit does now is: 

- Removed the PID looking
- Removed the child/grandchild for waitpid(-1,&wstatus,WNOHANG)

Changed the loop condition of the log recovery process: 

- Do work only while the work event is set to 1 by the main process. The child process clears each iteration of the loop.
- Added cleanup flag, set by the main process, indicating that the run is finished and waiting for the child to finish. 

In the main process, set the work signal(a keepalive signal) for each iteration of the main loop ( and in other parts of the code, that can be heavy to process depending on the configuration, so the timeout is not reached) while the child cleans this signal each time is set to 1

_Another option would be to set a lifespan to the log recovery process, and when the timeout is reached, spawn another process when needed. Not for each "work" as it would mean to spawning a process/thread for each finished job_

If the Autosubmit main process finishes correctly, the child will receive a signal to gather the remaining job logs and finish. This is similar to what we had.

And in case the Autosubmit main process is sigkill killed -9 while there are still active jobs, the signal of work won't be set, and thus, the child process will eventually (timeout defaults to 60 seconds) finish.

So now, the child depends on two signals from the main process.

kinow:
In GitLab by @kinow on Aug 28, 2024, 13:05

Sounds good, @dbeltrankyl . Were you able to reproduce the issue before this change somewhere, and confirm that with this change the issue is fixed?

kinow:
In GitLab by @dbeltrankyl on Aug 28, 2024, 13:25

Thanks @kinow , 

Yes, I have to do more tests, but I was waiting for confirmation that this solution was good enough before continuing. 

Both the PID checker and the current commit worked, but PID can lead to inconsistencies, as commented. 

This commit what did is to replace the PID checker.

The grandchild/child change for waitpid, seems to work similarly in my system 

Grandchild/child

```
pstree -s -p 24771 # orphan process
systemd(1)───systemd(2064)───autosubmit log (24771)─┬─{autosubmit log }(24772)
```

waitpid

```
pstree -p -s 27292 # orphan process
systemd(1)───systemd(2060)───autosubmit log (27292)───{autosubmit log }(27293)
```

So, in summary, I see orphan processors when: 

- kill -9 $main_process_pid

- Old commit: Comment PID checker
- Current commit: no timeout or changing the while condition to rely only on the cleanup signal

Current commit with inf timeout and sigkill autosubmit: 

```bash
(python312_env) (base) dbeltran@ bsces107894:~/autosubmit/a002/tmp/LOG_a002$ ps ax | grep autosubmit

  27272 pts/0    Rl+    0:03 /home/dbeltran/python312_env/bin/python3 /home/dbeltran/Autosubmit_source/bin/autosubmit -lc DEBUG run a002 -v
  27292 pts/0    Sl+    0:00 autosubmit log a002 recovery marenostrum5copy
  27300 pts/0    Sl+    0:00 autosubmit log a002 recovery marenostrum5

(python312_env) (base) dbeltran@ bsces107894:~/autosubmit/a002/tmp/LOG_a002$ kill -9 27272

dbeltran@ bsces107894:~/autosubmit/a002/tmp/LOG_a002$ ps ax | grep autosubmit
  27292 pts/0    Sl+    0:00 autosubmit log a002 recovery marenostrum5copy
  27300 pts/0    Sl+    0:00 autosubmit log a002 recovery marenostrum5

dbeltran@ bsces107894:~/autosubmit/a002/tmp/LOG_a002$ autosubmit run a002

  27292 pts/0    Sl+    0:00 autosubmit log a002 recovery marenostrum5copy
  27300 pts/0    Sl+    0:00 autosubmit log a002 recovery marenostrum5
  27603 pts/0    Sl+    0:03 /home/dbeltran/python312_env/bin/python3 /home/dbeltran/Autosubmit_source/bin/autosubmit -lc DEBUG run a002 -v
  27625 pts/0    Sl+    0:00 autosubmit log a002 recovery marenostrum5
  27632 pts/0    Sl+    0:00 autosubmit log a002 recovery marenostrum5copy
  27641 pts/2    S+     0:00 grep --color=auto autosubmit


```

kinow:
In GitLab by @kinow on Aug 29, 2024, 10:18

@manuel-g-castro was testing `stop` for help messages (`master` branch) and the `stop` command was not showing... maybe helpful to the point about testing `stop` here...

kinow:
In GitLab by @manuel-g-castro on Aug 29, 2024, 11:48

Yeah, stop command does not show. It seems like it is an issue with the `argparser`

```
mgimenez@ bsces107930 ~/Documents/local/dev/autosubmit
 % python3 bin/autosubmit stop
error: argument command: invalid choice: 'stop' (choose from 'run', 'expid', 'delete', 'monitor', 'stats', 'clean', 'recovery', 'migrate', 'inspect', 'check', 'describe', 'report', 'create', 'configure', 'install', 'setstatus', 'testcase', 'dbfix', 'pklfix', 'updatedescrip', 'test', 'refresh', 'updateversion', 'archive', 'unarchive', 'upgrade', 'readme', 'changelog', 'cat-log')

```

kinow:
In GitLab by @manuel-g-castro on Aug 29, 2024, 11:49

I am on master on commit `69962` "Merge branch 'typo-in-stop' into 'master'".

kinow:
In GitLab by @dbeltrankyl on Aug 30, 2024, 12:33

Hello @manuel-g-castro , what python version are you using? 

For me, it is working on both branches using python3.12 and python3.9. 

I introduced autosubmit stop, and the help message was displayed. 

Tested with pycharm & the autosubmit that comes from pip.

However, maybe the issue is that you have the argsparser==1.4.0 installed, which is no longer part of the autosubmit dependencies. Can you check it? And if you have it, uninstall it.

kinow:
In GitLab by @manuel-g-castro on Aug 30, 2024, 14:28

Ahh! Good catch! I have python 3.8.10 on my not-so-used development virtual environment. Works with python 3.9.

Thank you!

kinow:
In GitLab by @mcastril on Aug 30, 2024, 17:42

Thank you Dani. So, as I see it, the only mission of the middle process is to spawn the children. Then, as it dies, the responsible for handling their ripping is PID 1. 

I have some questions:

In the explanations in the description:

> Finish force:

> * User or Sysadmin force kills the main process.
> * The log recovery process checks through psutils if the main process is still alive.
> * Log recovery process finishes.

But you said later than you remove the child/grandchild waits... How it works in that case?

> The main process sends the job through the queue to the log recovery process.

Does the log recovery also send jobs?

> * Do work only while the work event is set to 1 by the main process. The child process clears each iteration of the loop.
> * Added cleanup flag, set by the main process, indicating that the run is finished and waiting for the child to finish.

You mean that the child to an iteration and dies? What does it do in the iteration?

> In the main process, set the work signal(a keepalive signal) for each iteration of the main loop ( and in other parts of the code, that can be heavy to process depending on the configuration, so the timeout is not reached) while the child cleans this signal each time is set to 1

I didn't understand this part. The keep alive is in the main but the one cleaning it is the child. Could you explain this again?

> So now, the child depends on two signals from the main process.

What are the two signals?


I agree with @kinow that it would be helpful that someone else can reproduce both the error and the fix

kinow:
In GitLab by @dbeltrankyl on Sep 2, 2024, 08:48

> But you said later than you remove the child/grandchild waits... How it works in that case?

It works similar once the main process ends. Since there is now a waitpid() call, I can see, with the `pstree` command, that the log command is waiting for the init process. 

> I didn't understand this part. The keep alive is in the main, but the one cleaning it is the child. Could you explain this again?

* The main process sets a shared Event() to 1 in each iteration.

* The child receives this Event() signal and starts to do some work. 
* At the end of the work, the child clears the Event() signal, which means that the shared event is set to 0. 
* At this point, the child stops until this signal is set again by the main process ( and the timeout/sleep timer is completed to avoid hammering the CPU ) 

> What are the two signals?

There are two events ( signals ) instead of one now. 

* The Keepalive signal ( replaces the pid checker ) , which was explained in the previous point. This signal is sent to the child while the main process is doing stuff. If it is not set during 60 seconds ( default timeout ) the child will end. This is to avoid orphan children when the main process gets a kill -9. 

* The Cleanup is set once the main process ends successfully ( via control +c, killterm, or just finished correctly ) and waits for the child to end. Once this is set by the main process, the child will try to end its work as fast as possible, and the main process will wait for it.

kinow:
In GitLab by @dbeltrankyl on Sep 3, 2024, 14:57

added 1 commit

<ul><li>c2cbf87e - Added tests for signals</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=30882&start_sha=4d64c54b5dcecaa0daaa330034a807605d27f87e)

kinow:
In GitLab by @dbeltrankyl on Sep 3, 2024, 14:58

added 1 commit

<ul><li>5cb81bd5 - removed debug message</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=30883&start_sha=c2cbf87edd6fc2fca7c10c097e1e35ab93005e01)

kinow:
In GitLab by @dbeltrankyl on Sep 3, 2024, 15:00

added 1 commit

<ul><li>53bd6299 - Changed the timeout</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=30888&start_sha=5cb81bd5cc135867ffc3da60215ac23e8dd9d9e0)

kinow:
In GitLab by @dbeltrankyl on Sep 3, 2024, 15:29

added 37 commits

<ul><li>53bd6299...bee0114f - 26 commits from branch <code>master</code></li><li>ee530494 - 1 earlier commit</li><li>f76e71fd - WIP: Zombies/orphan processors logs fix</li><li>5c5cb49c - WIP: Looking good, some more fixes neccesaries and ready to be pushed</li><li>8c9cc5bf - A new orphan/zombies code approach</li><li>4981f20e - changes to signals and waitpid</li><li>d2afbcd4 - Added cleanup signal</li><li>692f7576 - cleanup code</li><li>05edb175 - cleanup code</li><li>3bad45b5 - Added tests for signals</li><li>05cadcae - removed debug message</li><li>7725905d - Changed the timeout</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=30892&start_sha=53bd62992d51d2607ca6eaf568afff4055f1dd90)

kinow:
In GitLab by @dbeltrankyl on Sep 3, 2024, 18:42

added 1 commit

<ul><li>4b49f227 - added more tests</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=30912&start_sha=7725905dfbfd73bbd06bd493743f4ceb108d021c)

kinow:
In GitLab by @dbeltrankyl on Sep 4, 2024, 16:20

added 2 commits

<ul><li>fba7193f - clean import</li><li>2907535e - Fix events queue being reset in a platform failure</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=31013&start_sha=4b49f227a72f4a41c5bdbceb52fff18de79769e8)

kinow:
In GitLab by @dbeltrankyl on Sep 5, 2024, 09:32

added 1 commit

<ul><li>d25c1d07 - Mock exists() so it works in the pipeline</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=31059&start_sha=2907535e913e42346991e2c13cb90a897ced69dc)

kinow:
In GitLab by @dbeltrankyl on Sep 5, 2024, 15:10

added 24 commits

<ul><li>d25c1d07...3b220e38 - 8 commits from branch <code>master</code></li><li>3b220e38...bec43320 - 6 earlier commits</li><li>d6b36478 - cleanup code</li><li>56e4b683 - cleanup code</li><li>d5ff3eab - Added tests for signals</li><li>5010ea19 - removed debug message</li><li>2715518b - Changed the timeout</li><li>5aa50cfb - added more tests</li><li>f8150bed - clean import</li><li>14bda46b - Fix events queue being reset in a platform failure</li><li>e3bd797b - Mock exists() so it works in the pipeline</li><li>292d6278 - reduced time.sleeps</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=31126&start_sha=d25c1d0714a761da4e42e03b7327012b10fb152e)

kinow:
In GitLab by @dbeltrankyl on Sep 10, 2024, 16:07

added 1 commit

<ul><li>2cce4b5e - Added database test, fixed some issues</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=31342&start_sha=292d6278e7726f7acd8bd8a6b8e87fad76f280f2)

kinow:
In GitLab by @dbeltrankyl on Sep 12, 2024, 09:00

mentioned in merge request autosubmit4-config-parser!12

kinow:
In GitLab by @dbeltrankyl on Sep 16, 2024, 12:12

added 2 commits

<ul><li>8bf3103a - Changes to stat, db and retrials (WIP)</li><li>58b7f957 - [Single job] Test working, retrials working</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=31542&start_sha=2cce4b5e515f6a45ea462022fb2bf94387ee7884)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 09:28

mentioned in commit d065912a94ad7a990e5b51679eb064355001d532

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 09:28

mentioned in commit 8a162ef89ca2cf49b6910df6d299d8fdd8278179

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 09:28

added 2 commits

<ul><li>8a162ef8 - !475 vertical regression test</li><li>d065912a - Merge branch &#39;!475-vertical-regression-test&#39; into &#39;4.1.11-zombies&#39;</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32026&start_sha=58b7f957a3cc45d0a229cb99c6182734fc9867e0)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 09:31

mentioned in commit 30a48b03496963f954fca9110bcb930170388032

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 09:31

added 26 commits

<ul><li>d065912a...2df20535 - 6 commits from branch <code>master</code></li><li>2df20535...b5e5a75a - 10 earlier commits</li><li>0f3f45fd - Changed the timeout</li><li>933db353 - added more tests</li><li>7a3146e4 - clean import</li><li>96d363f5 - Fix events queue being reset in a platform failure</li><li>6781d297 - Mock exists() so it works in the pipeline</li><li>100aa1a4 - reduced time.sleeps</li><li>b061fe5d - Added database test, fixed some issues</li><li>ba3dd2c1 - Changes to stat, db and retrials (WIP)</li><li>130887c4 - [Single job] Test working, retrials working</li><li>30a48b03 - !475 vertical regression test</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32028&start_sha=d065912a94ad7a990e5b51679eb064355001d532)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 09:47

marked the checklist item **Unit test if possible, but it is hard as multiple processors need to connect to a remote platform...** as completed

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 09:58

mentioned in issue autosubmit-gui#113

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 12:24

added 1 commit

<ul><li>d985a331 - Fix tests, added pytest-xdist</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32070&start_sha=30a48b03496963f954fca9110bcb930170388032)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 12:31

added 1 commit

<ul><li>7ff04769 - Update as version</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32071&start_sha=d985a3313f97c240ac114964ba21a92b62d5ff40)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 14:36

added 1 commit

<ul><li>86fa1ad0 - Test pipeline</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32103&start_sha=7ff047691294fb4105c3881f4eec6b9142b74e5d)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 14:51

added 1 commit

<ul><li>983bd348 - Test pipeline ( now with output)</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32111&start_sha=86fa1ad0a29faf195e0f9770e59793bb8d165318)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 14:55

added 1 commit

<ul><li>5886ced4 - Test pipeline ( now with output)</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32114&start_sha=983bd348e2f85992d7dd5ee4fbb5c04f7bfbf787)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 15:19

mentioned in commit aa14b011a92ce4cb8cb8a2828381068cab5b79d3

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 15:19

added 29 commits

<ul><li>5886ced4...91a59208 - 4 commits from branch <code>master</code></li><li>91a59208...b81d570c - 15 earlier commits</li><li>1051c8f5 - reduced time.sleeps</li><li>95453aa6 - Added database test, fixed some issues</li><li>bf382aa4 - Changes to stat, db and retrials (WIP)</li><li>5f666b1c - [Single job] Test working, retrials working</li><li>aa14b011 - !475 vertical regression test</li><li>5e8c9e62 - Fix tests, added pytest-xdist</li><li>8df256ce - Update as version</li><li>89aef5e9 - Test pipeline</li><li>6b918df1 - Test pipeline ( now with output)</li><li>8999df2a - Test pipeline ( now with output)</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32126&start_sha=5886ced4016d9b0f867eff6790253b14cfb96898)

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 15:52

* Code rebased.
* Unit/regression tests working.
* Code working. 
* database code working.
* Pipeline tested 

TODO: 

* [x] Check the quality of the code now that everything is written, and check the lint. Revise if something can be rewritten more easily.
* [x] Check/improve the comments may be outdated or missing.
* [x] Check/improve/reduce the verbosity of the logs.
* [x] Fine-tune the keepalive signal. It is set to 5 minutes, while the mandatory sleep is set to 60 seconds. For larger workflows, 5 min (for example, 1000 jobs generated in the same loop iteration ) could not be enough; 5 min is too much for shorter workflows. ( Remind that the cleanup signal exists, so the 5 min would be just if the run is killed -9 ) 
* [x] Explain the changes so it is easier to be reviewed.

kinow:
In GitLab by @kinow on Sep 26, 2024, 15:53

Looks like you made good progress!!! :tada: :star:

kinow:
In GitLab by @dbeltrankyl on Sep 26, 2024, 15:54

Thanks Bruno! Yes, I hope that tomorrow I'll have a version ready to be reviewed

kinow:
In GitLab by @dbeltrankyl on Sep 27, 2024, 10:28

mentioned in commit 7d8c7316ccda3ee5183179e07adac2ace2c7a5ad

kinow:
In GitLab by @dbeltrankyl on Sep 27, 2024, 10:28

added 29 commits

<ul><li>8999df2a...e91258be - 6 commits from branch <code>master</code></li><li>e91258be...d2484bfe - 13 earlier commits</li><li>f814a0a4 - Mock exists() so it works in the pipeline</li><li>c1cbaf5d - reduced time.sleeps</li><li>3fc2b835 - Added database test, fixed some issues</li><li>b77c4807 - Changes to stat, db and retrials (WIP)</li><li>b27605cb - [Single job] Test working, retrials working</li><li>7d8c7316 - !475 vertical regression test</li><li>f344f167 - Fix tests, added pytest-xdist</li><li>09034393 - Update as version</li><li>a152ba76 - Test pipeline</li><li>27c0a9fa - Test pipeline ( now with output)</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32161&start_sha=8999df2a952e8869e17c2b05aa133d440871c7d9)

kinow:
In GitLab by @dbeltrankyl on Sep 27, 2024, 18:00

added 2 commits

<ul><li>d4703385 - Added comments and typing</li><li>a706df8b - Added comments and typing (II )</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32226&start_sha=27c0a9fae647785aea58fd454ea74d322416d328)

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 09:54

added 1 commit

<ul><li>96821ed7 - Fix pipeline</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32242&start_sha=a706df8b0c3cf8e4784bff2e8f30e1a931aa9782)

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 10:16

added 1 commit

<ul><li>23e80a81 - Fix pipeline</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32248&start_sha=96821ed72ce8d96a06e300180bce71518c088876)

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 12:32

added 1 commit

<ul><li>e46ac917 - Fix pipeline</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32264&start_sha=23e80a814acab5a3f626e2208a0b4c13768ea0d1)

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 13:01

added 1 commit

<ul><li>ab581049 - Fix doc test</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32267&start_sha=e46ac9177514f5cd578800a27cd89b8577f554bb)

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 13:25

added 1 commit

<ul><li>c8c16dcd - PIPELINE</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32280&start_sha=ab581049acce443ed7f9b7ed2a58fa8abf0fa830)

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 13:45

added 1 commit

<ul><li>23b93b9c - PIPELINE</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32281&start_sha=c8c16dcdee7e1d551782b2ed420d6238d22b1c2d)

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 14:07

added 1 commit

<ul><li>850d9eab - PIPELINE</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=32282&start_sha=23b93b9c75b6914d1cdeae503d57ee7578f538cb)

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 14:50

Hello @kinow 

I think this is ready to be reviewed. I had some issues with the pipeline, but now they seem solved. 

All code is commented on, but I'll summarize there:

Main processor logs code:

* Logs are spawned with an async waitpid, as suggested in previous comments.

* Keep alive signal: set to 5 min. 
 * (main process ) Added a cls function to send this signal in the main running loop, check_job, check_wrappers, and submit jobs so when a platform is doing some stuff, it can communicate to the rest of the platforms that the main process didn't stop.

* Cleanup signal: This is called at the end of the run ( control + c and normal run ). The cleanup is sent to all platforms, followed by a .join() so the last logs can be downloaded. 

* The function to check if the log has been downloaded in the main process now correctly takes into account the failed jobs the same goes for the stats

Log process: 

* Has a mandatory sleep of 60 seconds ( or whatever is introduced in the autosubmitrc file) unless the cleanup signal is set. 
* Has an active wait of 5 minutes; it will exit once there is something in the queue, or there are any of the two signals set. The process will exit if there is nothing ( due to a kill -9 on the main process ). This may lead to up to 5 min orphan jobs that are active waiting in kill -9 cases

Log/Stats detection:

* Logs will now have the submission date as the name of the log. Retrials will have {submission_name}_retrial_{fail_count} as a name. The same goes for STAT files.
* Autosubmit has some variables to determinate the {ready,submitted,finished}
 timestamps used for the detection of the logs but also for timestamps in case the _STAT file doesn't have some of the dates ( that should only happen in case of single failed jobs ).

For further details in stats check https://earth.bsc.es/gitlab/es/autosubmit-gui/-/issues/113#note_308427

Wrappers:

* The vertical_wrapper.cmd has changed because the stat files weren't working as expected there. 

Local wrappers: 

* Changes that are not in platforms.py but they're in local_platforms are to enable them. 
* Additional changes to check_job and submission where needed as the detection mechanism of FAILED/COMPLETED job in the local platform is just running, not running. 

Others:

* Some functions were added to reduce complexity
* Comments added and docstrings to changed functions 
* Typing added, but I think I may have overused the "ANY" typing. We can revise it in other branches or this one. 

About the tests: 

The log one doesn't have a sleep mocked, as we discussed a few times in Slack. They're working as they are now and don't delay the pipeline much ( just a few seconds ). 

However, I will try mocking the sleep in a separate branch to improve them. The problem is not the sleep mock; instead, I somehow need a sync between the child and the main, and what I test in these tests is precisely that synchronization. 
 
* [x] What is left is to deploy this version (as some stuff has changed regarding the previous test) in the climate-dt machine and see the behavior but you can start the review of the code itself @kinow

kinow:
In GitLab by @dbeltrankyl on Sep 30, 2024, 14:50

requested review from @kinow

kinow:
In GitLab by @dbeltrankyl on Oct 1, 2024, 12:59

Tested sucesfully on climate-dt

Used 4.1.11-zombies module: 

```
autosubmit expid -H "local" -d "test-zombies"
autosubmit create a1qw -np
autosubmit run a1qw
```
* run from 0 to 100 -> No zombies, no orphan
* force killed (-9) run -> no zombies, orphan during 2-3 min ( expected 5 min, but ended earlier )
* force killed (-9) run + children -> no zombies, no orphan
* Control + c -> no zombies no orphan

* autosubmit stop $expid -> no zombies, no orphan
* autosubmit stop $expid -f -> no zombies, no orphan

kinow:
In GitLab by @LuiggiTenorioK on Oct 3, 2024, 09:58

Commented on [autosubmit/job/job.py line 887](https://github.com/BSC-ES/autosubmit/compare/e19e7606a95c640558fb907c2d893765f566e623..4751863a248337fa9c81123beab0cd4116c88fcc#diff-f8751fbf5d2712919c09b25f56122c3fL887)



I've seen this datetime format in many places in Autosubmit but: Does it need to be like that? Now that we have concurrent users in many timezones (e.g. Finland and Spain) I think that is time we follow [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) to avoid inconsistencies in the future.

kinow:
In GitLab by @dbeltrankyl on Oct 3, 2024, 10:59

Commented on [autosubmit/job/job.py line 887](https://github.com/BSC-ES/autosubmit/compare/e19e7606a95c640558fb907c2d893765f566e623..4751863a248337fa9c81123beab0cd4116c88fcc#diff-f8751fbf5d2712919c09b25f56122c3fL887)



Hello @LuiggiTenorioK 

I've used our usual format across the rest of the code.

As commented in the meeting, we should follow that standard as it is more appropriate, but I would do it in another branch ( as this is already big enough ) and in a major release as it could lead to retro compatible issues.

kinow:
In GitLab by @dbeltrankyl on Nov 7, 2024, 11:48

added 4 commits

<ul><li>b5e9a032 - Zombies orphan and db fixes</li><li>d66dc233 - Merge of !490 ( vertical db fixes )</li><li>9708f549 - Tests and fixes after vertical db fixes</li><li>3487c355 - Merge remote-tracking branch &#39;origin/4.1.11-zombies&#39; into 4.1.11-zombies</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=34879&start_sha=850d9eabdc4994b8abf35aa13925a1f3a72f2fab)

kinow:
In GitLab by @dbeltrankyl on Nov 7, 2024, 15:40

mentioned in commit f663984e519789bab10ff83e8c7080ac5c4ba7e9

kinow:
In GitLab by @dbeltrankyl on Nov 7, 2024, 15:40

added 1 commit

<ul><li>f663984e - Squashed zombies/orphan and database fixes.</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=34905&start_sha=3487c355ffefc0d93cb65225584928fbfdaf9400)

kinow:
In GitLab by @dbeltrankyl on Nov 7, 2024, 16:01

mentioned in commit f0266c900e19eec57bd2cb37e45c18adf7426d73

kinow:
In GitLab by @dbeltrankyl on Nov 7, 2024, 16:01

added 34 commits

<ul><li>f663984e...e49a8b6b - 33 commits from branch <code>master</code></li><li>f0266c90 - Squashed zombies/orphan and database fixes.</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=34911&start_sha=f663984e519789bab10ff83e8c7080ac5c4ba7e9)

kinow:
In GitLab by @dbeltrankyl on Nov 8, 2024, 10:13

added 1 commit

<ul><li>ed9148d5 - Fixed pipeline after rebase</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=34956&start_sha=f0266c900e19eec57bd2cb37e45c18adf7426d73)

kinow:
In GitLab by @kinow on Nov 19, 2024, 13:24

@dbeltrankyl I started rebasing this branch, but there were conflicts not with the the vertical wrapper changes, but this MR appears to be a few commits behind master (things merged some days ago are missing).

I started going through the changes trying to keep what I thought would be correct, but I am not 100% that I will always keep the correct code. Would you be able to have a go and see if you are able to rebase this onto `master`, please?

If not, I can try and push-forge and ask you to review the diff here on GitLab, whichever works best for you :thumbsup: 

Thanks!

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 13:35

I'll try to do it asap

kinow:
In GitLab by @utigerstedt on Nov 19, 2024, 13:49

You have probably met with a hard nemesis: Rebasing while some else is squashing merges.

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 14:13

mentioned in commit 9d2e673e08a5dba49fd9306e8b57b6a9aba6f107

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 14:13

added 13 commits

<ul><li>ed9148d5...28f25521 - 11 commits from branch <code>master</code></li><li>9d2e673e - Squashed zombies/orphan and database fixes.</li><li>88b775ba - Rebase fix</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=35633&start_sha=ed9148d55240dd0d0aa9f5006462e7a1c982b89a)

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 14:20

added 1 commit

<ul><li>cbd05cbc - Changed err_code function to return the err_code</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=35636&start_sha=88b775ba6cec919ae46fc6adbdff9afb7927da57)

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 14:22

added 1 commit

<ul><li>8e6c8771 - added a missing return</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=35638&start_sha=cbd05cbcf393b9f63fe700c173ce0c34bff75b0b)

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 14:22

added 1 commit

<ul><li>c5c7114b - updated typing</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=35640&start_sha=8e6c87714a25e2abcebe3d8dd7ad6f862a162e75)

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 14:29

added 1 commit

<ul><li>9c9b8197 - Reverting a conflict in the rebase to the other option</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=35646&start_sha=c5c7114bb927129f8512f65b2e04754410d260da)

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 14:33

added 1 commit

<ul><li>0fcf1bef - Reverting a conflict in the rebase to the other option II ( tests are only failing in the CI/CD )</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=35648&start_sha=9c9b8197c6b05132e51bed62dc052b29f7d0372b)

kinow:
In GitLab by @dbeltrankyl on Nov 19, 2024, 14:39

Hello @kinow 

Rebased, the tests are working, the rebase was a bit of a mess, so I hope everything is order now. 

> You have probably met with a hard nemesis: Rebasing while some else is squashing merges.

Could be most of the conflicts were typing/comments, I think they're derived from !515

Others come from other commits of the master branch, but they seemed more usual conflicts than the rest I encountered

@kinow  On a side note, the db tests with sleep 1 don't work in the pipeline. I'm unsure why I had to put sleep 2 again in the jobs. I'll revise it once this is merged

kinow:
In GitLab by @kinow on Nov 20, 2024, 09:39

Thanks for rebasing it Dani!

>@kinow On a side note, the db tests with sleep 1 don't work in the pipeline. I'm unsure why I had to put sleep 2 again in the jobs. I'll revise it once this is merged

I wonder if we could get rid of using `sleep` in tests, as eventually we might have issues either with the total time to run the tests, or complicated issues that are hard to debug. 

In Java I did something similar [to this article](https://medium.com/@Tom1212121/mock-the-clock-not-my-tests-how-java-util-clock-saved-my-sanity-b0b9730a2244) where a dev controls the clock. Manuel/BSC uses a Slurm emulator that also controls the clock (making time pass faster, or simply changing the time of clock), and in JavaScript it's [common practice](https://jestjs.io/docs/timer-mocks) to avoid time-dependent tests by controlling the time in the tests.

But we can work on this after we fix the children processes waiting/killing here. Thanks Dani!

kinow:
In GitLab by @kinow on Nov 20, 2024, 13:19

I think this discussion can be marked as resolved as this approach of an intermediate process was not necessary (can you confirm and hit the resolve thread button, please @dbeltrankyl ?).

kinow:
In GitLab by @kinow on Nov 20, 2024, 13:33

Summary of today's work on this review. Sorry the long text, it's more
a brain dump so I do not forget what I did, what worked or not, and to
later summarize my thoughts.

The log process is created here: https://earth.bsc.es/gitlab/es/autosubmit/-/blob/28f255218dc66d83b5a4054ae971d93b2c9154c2/autosubmit/platforms/platform.py#L843-844

To run this function: https://earth.bsc.es/gitlab/es/autosubmit/-/blob/28f255218dc66d83b5a4054ae971d93b2c9154c2/autosubmit/platforms/platform.py#L916-936

## Orphaned process

Using `master` (4.1.11), I first reproduced an orphaned process by
launching a workflow with a single task, that runs for 1 hour. Once
it started, I killed the main `autosubmit run` process.

In another terminal, I have `watch -n 1 pgrep --list-full autosubmit`,
to see which processes are running (handy as it also displays the
zombie status).

Once I killed the parent, the child kept running and its new parent
became systemd (confirmed looking at `ps -eo pid,ppid,cmd | grep ...`).

Looking at `strace`, it confirms the process is waiting on a resource
with `pselect6`. That looks correct, as the log recovery process waits
on an Event while checking for a timeout. And indeed, after the timeout
expired, the process exited (in the console I had strace'd the process,
it appeared the message ”local(log_recovery): Exiting.”), and the
`pgrep` command showed no more Autosubmit processes.

So not too bad. I think we could pass the parent pid to each child log
recovery process and, if the child detects the parent is different, I think
it could immediately exit as the `Event` won't be set by the scheduler.

But I think it's OK as if we get an orphaned process, it should exit
eventually after the timeout. So I think we don't have a big problem
with that right now on the VM (nor on the BSC hubs). Have you seen/heard
of any issues with orphaned processes in the Climate DT VM or hubs recently,
@dbeltrankyl?

## Zombies

Also using `master`, I didn't get any zombies when testing. And I haven't
seen zombie processes in the VM in a while, except for Autosubmit API
processes that will be eventually reaped by systemd. Do you know if that
has happened with 4.1.10, @dbeltrankyl?

```bash
[brunokino@climatedt-wf ~]$ ps aux | grep 'Z'
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
ltenorio  522088  0.0  0.0      0     0 ?        Z    Nov15   0:00 [id] <defunct>
asapi    2599613  0.0  0.0      0     0 ?        Z    12:26   0:00 [sh] <defunct>
asapi    2600203  0.0  0.0      0     0 ?        Z    12:28   0:00 [sh] <defunct>
asapi    2603337  0.0  0.0      0     0 ?        Z    12:39   0:00 [sh] <defunct>
```

So I think this is OK. (In fact I checked again after writing and one got
reaped already.)

We have [a call to `os.waitpid()`](https://earth.bsc.es/gitlab/es/autosubmit/-/blob/28f255218dc66d83b5a4054ae971d93b2c9154c2/autosubmit/platforms/platform.py#L847)
issued after the log recovery process is launched.

This call includes `os.WNOHANG`, which seems to make it more resilient (in case there
are no children), although [it looks like its doc is lacking](https://github.com/python/cpython/issues/78459).

However, looks like zombie processes are still possible even with `waitpid` + `WNOHANG`:
https://python-list.python.narkive.com/CkwlS91Z/i-used-os-waitpid-but-i-still-can-t-zombie-process

And systemd appears to be reaping the processes (at the of the ASAPI) correctly,
and we are not accummulating tens of processes on the VM right now.

## Logs not retrieved

I think this is the real issue that we have right now. I thought it was
happening due to orphaned or zombie processes, but reading the code, I
think the logic is well contained in a few places so it shouldn't be too
hard to isolate it in a test script and verify in which occasions could
the log retrieval process be left in a state where it stops retrieving
the logs.

If it's confirmed that the logs are not retrieved due to maybe the `os._exit`
call killing the main parent and leaving the events and/or queue in some
state the causes the log retrieval process confused, maybe we can reduce the
merge request changes, and improve the logic in the child process.

kinow:
In GitLab by @dbeltrankyl on Nov 20, 2024, 14:10

Hello @kinow 

> So not too bad. I think we could pass the parent pid to each child log recovery process and, if the child detects the parent is different, I think it could immediately exit as the `Event` won't be set by the scheduler. 

If I'm not wrong, that solution is what ( @utigerstedt ) wants to avoid. 

> But I think it's OK as if we get an orphaned process, it should exit eventually after the timeout.

The child will run as long as there are logs to recover and the parent didn't send the "clean up" signal. 

The clean-up signal is sent whenever there are no more logs to recover or the timeout is completed. 

The clean-up signal cannot be passed if the parent is killed -9. Otherwise, it will always be sent. In this case the child will end by itself after the timeout ( because it checks a signal sent by the parent if it is 0 or 1 ) and reap by the system


> Logs not retrieved

Are, right now, the logs not being recovered? in this branch I mean

kinow:
In GitLab by @dbeltrankyl on Nov 20, 2024, 14:17

 > Also using `master`, I didn't get any zombies when testing. And I haven't seen zombie processes in the VM in a while, except for Autosubmit API processes that will be eventually reaped by systemd. Do you know if that has happened with 4.1.10, @dbeltrankyl?

No, we had them in 4.1.8. 

- 4.1.9 fixed the zombies but added orphan without delay time.( with the method of checking the parent.pid)
- 4.1.10 band fixed the orphans by adding a delay but it broke the log recovery. ( with the method of checking the parent.pid)
- This branch aims to solve everything only relying on signals sent by the main process.

kinow:
In GitLab by @kinow on Nov 20, 2024, 14:25

>> So not too bad. I think we could pass the parent pid to each child log recovery process and, if the child detects the parent is different, I think it could immediately exit as the `Event` won't be set by the scheduler.
>
>If I'm not wrong, this solution is what ( @utigerstedt ) wants to avoid.

Ah, then I missed that, thanks. I thought that would be useful as the process will keep while/sleep/nothing to do... while/sleep/nothing to do... while in reality if the parent `autosubmit run` died, and nothing else will update the `Queue` nor the `Event`'s that the process is checking during the while-loop, then it'd be better if it just stopped and reported that there was nothing left to do and the parent process $pid-bla was not found. But not a problem if we don't do that now.

>> But I think it's OK as if we get an orphaned process, it should exit eventually after the timeout.
>
>The child will run as long as there are logs to recover and the parent didn't send the "clean up" signal.
>
>The clean-up signal is sent whenever there are no more logs to recover or the timeout is completed.
>
>The clean-up signal cannot be passed if the parent is killed -9. Otherwise, it will always be sent. In this case the child will end by itself after the timeout ( because it checks a signal sent by the parent if it is 0 or 1 ) and reap by the system

Great, that is what I understood reading the code (we can document this later). And in case the process is killed nicely, not only we have the `waitpid`, the process is also `p.join()`'ed, which calls `.wait` on the process, preventing zombies. :thumbsup: 

>> Logs not retrieved
>
>Are, right now, the logs not being recovered? in this branch I mean

They are, but what I recall (when I was running the testing suite) is that in some cases we could end up with logs not retrieved from the platform. And what my memory tells me is that it was the logs of the last task/job/maybe chunk? (but my memory is not very good, so I will confirm it searching the issues).

>> Also using `master`, I didn't get any zombies when testing. And I haven't seen zombie processes in the VM in a while, except for Autosubmit API processes that will be eventually reaped by systemd. Do you know if that has happened with 4.1.10, @dbeltrankyl?
>
>No, we had them in 4.1.8.
>
>* 4.1.9 fixed the zombies but added orphan without delay time.( with the method of checking the parent.pid)
>* 4.1.10 band fixed the orphans by adding a delay but it broke the log recovery. ( with the method of checking the parent.pid)
>* This branch aims to solve everything only relying on signals sent by the main process.

In 4.1.11, orphans and zombies appear to be under control. Is there a specific issue with log recovery in the latest version that we are trying to solve with this branch? (If so I can have a look at that one to reproduce/understand it).

kinow:
In GitLab by @dbeltrankyl on Nov 20, 2024, 15:09

Thanks! Great we're in the same page now then


> They are, but what I recall (when I was running the testing suite) is that in some cases we could end up with logs not retrieved from the platform.

Ah, great, I thought that some rebase broke it again but doesn't seem the case then

>  And what my memory tells me is that it was the logs of the last task/job/maybe chunk? (but my memory is not very good, so I will confirm it searching the issues).

Yes, if I recall correctly the issue was in the last jobs and also the local platform was unaffected

> In 4.1.11, orphans and zombies appear to be under control. Is there a specific issue with log recovery in the latest version that we are trying to solve with this branch? (If so I can have a look at that one to reproduce/understand it).

Nope. If I recall correctly(again), the origin of the problem was a race condition. Autosubmit v4.9's non-delay was hiding that. 

These functions in job_list contain fixes for the race condition and for the db:

` def get_uncompleted,    def check_if_log_is_recovered(self, job: Job) -> Path:`

Also, the https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/475/diffs#6bde465286156a9da245a31e64df703a55c76323_2388_2370 code related to the end of the execution. 

Also, something was wrong in the child's log_recovery code, but I'm afraid I don't recall what it was. I think it was related to the Queue... I think I'm overlapping some db fixes and log fixes because some changes to the DB fixes required changes to the logs and viceversa.

In the end, the problem was a combination of diverse problems in the functions that were all hidden by the non-existent delay, so I think that comparing the previous code with this one won't help you much. Sorry for that I should've commented on that more

kinow:
In GitLab by @kinow on Nov 26, 2024, 12:25

>Also, something was wrong in the child's log_recovery code, but I'm afraid I don't recall what it was. I think it was related to the Queue... I think I'm overlapping some db fixes and log fixes because some changes to the DB fixes required changes to the logs and viceversa.

Got it. If I am able to reproduce the issue on `master`, and then compare with this branch, that could help us understand what is/was broken, where it got fixed and how.

>In the end, the problem was a combination of diverse problems in the functions that were all hidden by the non-existent delay, so I think that comparing the previous code with this one won't help you much. Sorry for that I should've commented on that more

That's alright, that happens, and we lost momentum as I took a long time to review it (sorry too :bow:).

Linking some issues here that I'm using while testing this;

In this testing suite thread, the process exited and an orphaned process was left running but it never retrieved the logs: https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/issues/485#note_295677

This other issue reports a similar situation, but doesn't say whether there was an orphaned process for a short time (timeout), but could be related: https://earth.bsc.es/gitlab/es/autosubmit/-/issues/1414 (similar symptoms at least, even the part about the _COMPLETED missing).

This issue was when we consistently got a lot of orphaned processes (I think it may have been before the one-log-retrieval-per-platform change?): https://earth.bsc.es/gitlab/es/autosubmit/-/issues/1390 (we may be able to close this now even without merging this one).

kinow:
In GitLab by @kinow on Nov 26, 2024, 16:07

Interesting, messing up with the timeouts (to seed up my debugging sessions) resulted in the child log recovery process exiting earlier without collecting logs on the remote platform. So obviously I have to lower other timeouts to make everything clock faster. Are you aware of what else I need to reduce, @dbeltrankyl ?

My diff for now:

```diff
diff --git a/autosubmit/platforms/platform.py b/autosubmit/platforms/platform.py
index 9af2ca48..c53cc8ad 100644
--- a/autosubmit/platforms/platform.py
+++ b/autosubmit/platforms/platform.py
@@ -112,7 +112,7 @@ class Platform(object):
         self.work_event = Event()
         self.cleanup_event = Event()
         self.log_recovery_process = None
-        self.keep_alive_timeout = 60 * 5 # Useful in case of kill -9
+        self.keep_alive_timeout = 1 * 5 # Useful in case of kill -9
         self.processed_wrapper_logs = set()
         self.max_waiting_jobs = 20
 
@@ -927,8 +927,8 @@ class Platform(object):
         Log.get_logger("Autosubmit")  # Log needs to be initialised in the new process
         Log.result(f"{identifier} Sucessfully connected.")
         log_recovery_timeout = self.config.get("LOG_RECOVERY_TIMEOUT", 60)
-        self.keep_alive_timeout = max(log_recovery_timeout*5, 60*5)
-        while self.wait_for_work(sleep_time=max(log_recovery_timeout, 60)):
+        self.keep_alive_timeout = max(log_recovery_timeout*1, 1*5)
+        while self.wait_for_work(sleep_time=max(log_recovery_timeout, 5)):
             jobs_pending_to_process = self.recover_job_log(identifier, jobs_pending_to_process)
             if self.cleanup_event.is_set():  # Check if main process is waiting for this child to end.
                 self.recover_job_log(identifier, jobs_pending_to_process)
```

And as a note to self, in case I need to replicate it on my other laptop, my testing setup:

```yaml
# A single job, no wrappers, as in a1bh and t01s (DE) wrappers were not used, and there were
# only simple dependencies between tasks at chunk level (I don't think I need dependencies).
JOBS:
  A:
    FILE: wait_1_hour.sh
    RUNNING: chunk
    PLATFORM: fake-ssh
PLATFORMS:
  fake-ssh:
    TYPE: ps
    HOST: fake-ssh
    SCRATCH_DIR: /tmp
    USER: autosubmit
```

My platform entry in `~/.ssh/config`:

```ssh
Host fake-ssh
 HostName localhost
 Port 2222
 User autosubmit
 IdentityFile /home/kinow/Development/python/workspace/id_rsa
 ForwardX11 no
```

To generate the key (default settings, then copy it): `docker run --rm -it --entrypoint /keygen.sh linuxserver/openssh-server`

To start the fake-ssh container (same we gave to EDITO MOi team to use):

```bash
$ docker run --rm \
  --name=openssh-server \
  --hostname=fake-ssh \
  -e PUID=1000 \
  -e PGID=1000 \
  -e TZ=Etc/UTC \
  -e PUBLIC_KEY="<PASTE_PUB_KEY>" \
  -e SUDO_ACCESS=false \
  -e PASSWORD_ACCESS=false \
  -e USER_PASSWORD=autosubmit \
  -e USER_NAME=autosubmit \
  -e LOG_STDOUT=  \
  -p 2222:2222 \
  lscr.io/linuxserver/openssh-server:latest
```

Then `autosubmit run` with the debugger, `pgrep`, `kill`, etc.

kinow:
In GitLab by @dbeltrankyl on Nov 27, 2024, 08:54

Hello Bruno,

The problem with reducing that timer only (and being that short) there is that the main process has to:

* Check ready Jobs
* Create, submit Jobs
* Sleeps for safetytime(10 seconds by default you can change it)
* Send active signal
* Check job/ wrapper status (send active Signal during wrappers check)
* Mark job as completed
* Send the job to the queue

So, the templates has to dont have sleep time or very short and also change the safety time to 0 seconds.

What is happening to you is that the log recovery proccess is thinking that the main proccess died and ends itself

kinow:
In GitLab by @dbeltrankyl on Dec 2, 2024, 09:29

mentioned in commit 358d8439aa90141dc1bf545a1e12e380ca889bff

kinow:
In GitLab by @dbeltrankyl on Dec 2, 2024, 09:29

added 20 commits

<ul><li>0fcf1bef...e19e7606 - 13 commits from branch <code>master</code></li><li>358d8439 - Squashed zombies/orphan and database fixes.</li><li>3ed4a75f - Rebase fix</li><li>cfa69b56 - Changed err_code function to return the err_code</li><li>7ec463e1 - added a missing return</li><li>c28a47fa - updated typing</li><li>ecd9d248 - Reverting a conflict in the rebase to the other option</li><li>e62f2a2d - Reverting a conflict in the rebase to the other option II ( tests are only failing in the CI/CD )</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=36503&start_sha=0fcf1bef0a7891e327b5605b56929aefa41695ab)

kinow:
In GitLab by @dbeltrankyl on Dec 2, 2024, 09:47

added 1 commit

<ul><li>29bf6bfb - Autosubmit config version compatibility fix ( rebasing )</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=36515&start_sha=e62f2a2db449e51767c6ef60dcce4b1be3678051)

kinow:
In GitLab by @dbeltrankyl on Dec 2, 2024, 10:37

mentioned in commit e1f4e3d6638afb74c2023a6a1031f0acb113ad64

kinow:
In GitLab by @dbeltrankyl on Dec 2, 2024, 10:37

added 1 commit

<ul><li>e1f4e3d6 - Autosubmit config version compatibility fix ( rebasing )</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=36523&start_sha=29bf6bfb0297ca60219b9dc6f1c8ca41b6061ec1)

kinow:
In GitLab by @dbeltrankyl on Dec 2, 2024, 10:51

added 1 commit

<ul><li>6f3d7974 - Pipeline failing, debug (I)</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=36532&start_sha=e1f4e3d6638afb74c2023a6a1031f0acb113ad64)

kinow:
In GitLab by @dbeltrankyl on Dec 2, 2024, 11:41

added 1 commit

<ul><li>4751863a - Fix Stat_0 file being deleted when it shouldn&#39;t</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/475/diffs?diff_id=36534&start_sha=6f3d79741e3852733003d2851568034b49c68349)

kinow:
In GitLab by @dbeltrankyl on Dec 2, 2024, 11:52

Hello @kinow 

Rebased + squashed

After the rebase(I'm unsure which one, this or the previous one), the STAT_0 file was always deleted in the following retrials. The last commit contains the test and fixes added.

kinow:
In GitLab by @kinow on Dec 4, 2024, 15:48

mentioned in issue digital-twins/de_340-2/workflow#747

PR #1984, opened 2024-12-13
Removed unneeded sleep

kinow:
In GitLab by @dbeltrankyl on Oct 10, 2024, 15:56

This removes an unnecessary sleep at the end of the inspect command.

This sleep is only meant to use to wait between loops in an autosubmit run

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1605,7 +1605,6 @@ class Autosubmit:
                     as_conf, job_list, jobs_cw, packages_persistence, False)
 
             Log.info("no more scripts to generate, now proceed to check them manually")
-            time.sleep(safetysleeptime)
         except AutosubmitCritical as e:
             raise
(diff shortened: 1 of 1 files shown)


kinow:
In GitLab by @dbeltrankyl on Oct 10, 2024, 15:56

requested review from @kinow

kinow:
In GitLab by @kinow on Oct 11, 2024, 08:55

Ah, good catch! Let's hope inspect is faster for users now. We should remember to add it to the changelog too (you can do it now if you'd like, there are some for 4.1.11 already)

kinow:
In GitLab by @kinow on Oct 11, 2024, 08:55

approved this merge request

kinow:
In GitLab by @dbeltrankyl on Oct 21, 2024, 11:19

mentioned in commit 99a511d780bd7aab7cd3bd4bdc0a0ba7ded4b43a

Issue #1435, opened 2024-10-16
Possible bug in STATUS: RUNNING dependencies


kinow:
In GitLab by @franra9 on Oct 16, 2024, 15:52

Hello @dbeltrankyl and @kinow,

### Autosubmit Version

4.1.10

### Expid affected(If applicable)

a1q3

### Which task has issues? Where is the log(If applicable)

* **Full_name**: does not apply

* **Log_Path:** does not apply

### Summary

DN is submitted on SIM submission and not on running despite the jobs.yml does not say so.

### Steps to reproduce

Try copying a1q3 but there is some issue before.

### What is the current bug behavior?

DN submitted on SIM on submission instead of Running (SIM is on the queue but DN already running)

### What is the expected correct behavior?

DN submitted on SIM on running instead.

### Relevant logs and/or screenshots(if applicable)

```
Job a1q3_20200101_fc0_1_SIM is QUEUING
Checking 1 jobs for platform LUMI-LOGIN
Job a1q3_20200101_fc0_1_1_DN is RUNNING
```

```
(...)
  DN:
    CHECK: on_submission
    DEPENDENCIES:
      APP_ENERGY_ONSHORE-1:
        SPLITS_FROM:
          '0':
            SPLITS_TO: '%JOBS.DN.SPLITS%'
      DN:
        SPLITS_FROM:
          all:
            SPLITS_TO: previous
      REMOTE_SETUP:
        STATUS: COMPLETED
      SIM:
        STATUS: RUNNING
(...)
```
### Any other relevant information(if applicable)

()

kinow:
In GitLab by @dbeltrankyl on Oct 16, 2024, 15:54

But it didn't happen until now, no? 

Can you try to remove the 

```
REMOTE_SETUP:
        STATUS: COMPLETED # <- this
```

kinow:
In GitLab by @franra9 on Oct 18, 2024, 11:49

Hi @dbeltrankyl , you are right this is a misconfiguration in the jobs.yml

I don't know if this did not happen or we just didn't see it.


in any case,

Does it make sense though to have an OR instead of an AND here? I mean, tell me if what I say is not correct, but it seems that in this case AS is proceeding with the downstream execution when not all the conditions described in JOBS are fulfilled. Maybe you can explain to me how this works exactly via slack if you wish.

kinow:
In GitLab by @franra9 on Oct 18, 2024, 16:29

[...]

PR #1997, opened 2024-12-13
Draft: Add all expid_dir_path functions  to autosubmit.py and fix tests

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 12:27

All previous definitions all `expid_path`, `tmp_path`, `log_path` and `aslogs_path` (and similar) have been removed and their calls have been switched for `BasicConfig.foo()` calls. Impacted tests are mainly affected in terms of adding Mock objects since `BasicConfig.foo()` functions are not available in autosubmitconfigparser yet. No new tests added.

- [x] Resolve merging conflicts
- [ ] Remove mocked tests and use testing from latest configparser commit until new version is launched


kinow:
In GitLab by @isimo00 on Nov 5, 2024, 12:27

requested review from @dbeltrankyl and removed review request for @kinow

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 12:49

mentioned in commit 4eec259671daf0dcadb2454ac58c7a91dd9b5afe

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 12:49

added 9 commits

<ul><li>1dbf37a0...7a32f96f - 6 commits from branch <code>master</code></li><li>4eec2596 - Add all expid_dir_path functions to autosubmit.py and fix tests !509</li><li>daeeaf8f - Add expid paths to error notifications #1435</li><li>af6e9e63 - Merge branch &#39;iss1435&#39; of https://earth.bsc.es/gitlab/es/autosubmit into iss1435</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=34686&start_sha=1dbf37a0ee36932fda2ccbe39342c740290b63eb)

kinow:
In GitLab by @dbeltrankyl on Nov 5, 2024, 12:54

Commented on [autosubmit/notifications/mail_notifier.py line 76](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..1ac2fe49de7850215bb19a8756d962323844bbfe#diff-f85225ad01c05acc538c620fa1720fbcR76)



Hello @isimo00, 

Thanks for working on it! 

I think it should point directly to the exact err and exact out instead of the folder. 

Maybe.. we can attach the logs in the mail itself? and/or point to the specific log in GUI? As it can be easily read ( the latter, not sure how to do it without hardcore it FYI @LuiggiTenorioK ) 

what do you think?

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 12:58

marked the checklist item **Resolve merging conflicts** as completed

kinow:
In GitLab by @dbeltrankyl on Nov 5, 2024, 12:58

Commented on [test/unit/conftest.py line 68](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..298b0cdafe17225d0f8b500bce9b09af5872ba45#diff-93a7df2f35536e469f98fe57544195f7R68)



In general, I think that @kinow  recommended using Path for all ( per example, os.path.exists -> Path().exists() ) , os.makedirs -> Path().mkdir(Parents=True,exists_ok=True) ( didn't double check the syntax) )

edit: There are other similar cases across the changes

kinow:
In GitLab by @dbeltrankyl on Nov 5, 2024, 13:04

Commented on [test/unit/conftest.py line 197](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..867e93ff52b3583772ea6e97371c5dabf5619ea0#diff-93a7df2f35536e469f98fe57544195f7R197)



Can we use the mocker fixture there? 

Instead of using 

> from unittest import mock

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 13:04

Commented on [test/unit/conftest.py line 68](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..298b0cdafe17225d0f8b500bce9b09af5872ba45#diff-93a7df2f35536e469f98fe57544195f7R68)



The functions added here (e.g. aslogs_dir = BasicConfig.aslog_dir_mock(expid) in [line 50](https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/509/diffs#129d4ed3c5acafd36c9039bc210f3fffbb934031_50_50)) return `str` objects created with `os.path.join()` - see commit in [configparser](https://github.com/BSC-ES/autosubmit-config-parser/commit/c94fd9ea384fc37bbdeab2c800298801269b509d) -. I didn't give it much thought and carried on with `str` instead of `PosixPath` objects when possible. All of them could be switched back to Path

kinow:
In GitLab by @dbeltrankyl on Nov 5, 2024, 13:06

Thanks @isimo00! 

I comment in one stuff to change ( maybe)

and commented on some stuff about the tests. I think we should agree on what to use for the Paths with @kinow , @LuiggiTenorioK  mainly to  normalize the code

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 13:06

Commented on [autosubmit/notifications/mail_notifier.py line 76](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..1ac2fe49de7850215bb19a8756d962323844bbfe#diff-f85225ad01c05acc538c620fa1720fbcR76)



You mean adding/pasting the content of the log files in the mail?

kinow:
In GitLab by @dbeltrankyl on Nov 5, 2024, 13:15

Commented on [autosubmit/notifications/mail_notifier.py line 76](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..1ac2fe49de7850215bb19a8756d962323844bbfe#diff-f85225ad01c05acc538c620fa1720fbcR76)



I mean:

* [x] Point to the final err ( `$expid/tmp/Log_$expid/$job_name....err`) ( and out)
* [x] Adding it as an attached file ( see the screenshot ) (perhaps we can compress it before adding it )

image

kinow:
In GitLab by @LuiggiTenorioK on Nov 5, 2024, 13:34

Commented on [autosubmit/notifications/mail_notifier.py line 76](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..1ac2fe49de7850215bb19a8756d962323844bbfe#diff-f85225ad01c05acc538c620fa1720fbcR76)



> Maybe.. we can attach the logs in the mail itself? and/or point to the specific log in GUI? As it can be easily read ( the latter, not sure how to do it without hardcore it FYI @LuiggiTenorioK )

The GUI shows the latest `{expid}/tmp/ASLOGS/****_run.log` file. Then, you'll have to append the log into that file if you want to show it.

kinow:
In GitLab by @LuiggiTenorioK on Nov 5, 2024, 13:38

> and commented on some stuff about the tests. I think we should agree on what to use for the Paths with @kinow , @LuiggiTenorioK mainly to normalize the code

Agree that using `pathlib` is better than `os.path` since is a newer implementation based on OOP and has more options. However, I do find using `/` less readable than `os.path.join` so maybe we should use its `pathlib` equivalent (`Path(...).joinpath()`) instead.

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 14:35

Started switching back to `pathlib` in configparser - [github's pull request num. 51](https://github.com/BSC-ES/autosubmit-config-parser/pull/51) - with `Path(...).joinpath()`. Will extend to autosubmit.

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 15:33

mentioned in commit 867e93ff52b3583772ea6e97371c5dabf5619ea0

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 15:33

added 1 commit

<ul><li>867e93ff - Switch from os.path to pathlib.Path !509 #1435</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=34700&start_sha=af6e9e6329bb216e894608c5c8ce6c3660c268b1)

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 15:47

Commented on [test/unit/conftest.py line 197](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..867e93ff52b3583772ea6e97371c5dabf5619ea0#diff-93a7df2f35536e469f98fe57544195f7R197)



changed this line in [version 4 of the diff](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=34702&start_sha=867e93ff52b3583772ea6e97371c5dabf5619ea0#129d4ed3c5acafd36c9039bc210f3fffbb934031_197_193)

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 15:47

added 1 commit

<ul><li>298b0cda - Use mocker instead of mock from unittest</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=34702&start_sha=867e93ff52b3583772ea6e97371c5dabf5619ea0)

kinow:
In GitLab by @isimo00 on Nov 5, 2024, 15:55

added 1 commit

<ul><li>1ac2fe49 - Fix forgotten Path</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=34704&start_sha=298b0cdafe17225d0f8b500bce9b09af5872ba45)

kinow:
In GitLab by @kinow on Nov 6, 2024, 10:13

Commented on [autosubmit/notifications/mail_notifier.py line 76](https://github.com/BSC-ES/autosubmit/compare/7a32f96f65cc526774b7fa7c2999b3c3bc6a5569..1ac2fe49de7850215bb19a8756d962323844bbfe#diff-f85225ad01c05acc538c620fa1720fbcR76)



Not reviewing this one, just going through inbox emails... but maybe we could use a single f-string multiline text without concatenation?

Instead of

```py
  f'some string {vars} \n\n' \
  + f'another string being concatenated {vars} \n etc'
```

I think this should work too?

```py
f'''some string {vars} \n\n
more text {vars}
etc'''
```

kinow:
In GitLab by @isimo00 on Nov 26, 2024, 13:18

added 1 commit

<ul><li>0dab1606 - Add log file attached to mail message</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=36094&start_sha=1ac2fe49de7850215bb19a8756d962323844bbfe)

kinow:
In GitLab by @isimo00 on Dec 4, 2024, 11:46

added 1 commit

<ul><li>a776f65c - Fix message text</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=36761&start_sha=0dab16065738585890aafd6b88979e9721824930)

kinow:
In GitLab by @isimo00 on Dec 5, 2024, 10:46

mentioned in commit df8c557e5713428f0c8e1afca942081ed958e55e

kinow:
In GitLab by @isimo00 on Dec 5, 2024, 10:46

added 1 commit

<ul><li>df8c557e - Switch from os.path to pathlib.Path !509 #1435</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=36865&start_sha=a776f65c16e54bddfd2aa1de9e18188190aa7aed)

kinow:
In GitLab by @isimo00 on Dec 5, 2024, 11:19

added 1 commit

<ul><li>d4c4efc9 - Add all expid_dir_path functions  to autosubmit.py and fix tests</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=36869&start_sha=df8c557e5713428f0c8e1afca942081ed958e55e)

kinow:
In GitLab by @isimo00 on Dec 5, 2024, 11:22

mentioned in commit 6b6ffce350709e0ce23ad13c098cb3b6911e3f0a

kinow:
In GitLab by @isimo00 on Dec 5, 2024, 11:22

added 1 commit

<ul><li>6b6ffce3 - Add all expid_dir_path functions to autosubmit.py and fix tests !509</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=36874&start_sha=d4c4efc9855e0809c8771eb69f00c7316d8bbd2c)

kinow:
In GitLab by @kinow on Dec 5, 2024, 13:43

marked this merge request as **draft**

kinow:
In GitLab by @isimo00 on Dec 5, 2024, 14:20

resolved all threads

kinow:
In GitLab by @isimo00 on Dec 9, 2024, 12:34

mentioned in commit d484e645115685968a73c36b3b8ae388bd13343d

kinow:
In GitLab by @isimo00 on Dec 9, 2024, 12:34

added 1 commit

<ul><li>d484e645 - Fix tests !509</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=36986&start_sha=6b6ffce350709e0ce23ad13c098cb3b6911e3f0a)

kinow:
In GitLab by @isimo00 on Dec 9, 2024, 12:41

mentioned in commit 3dfc50ada6894e7b810054a74ebe27bfa2d67a59

kinow:
In GitLab by @isimo00 on Dec 9, 2024, 12:41

added 1 commit

<ul><li>3dfc50ad - Add all expid_dir_path functions to autosubmit.py and fix tests !509</li></ul>

[Compare with previous version](/gitlab/es/autosubmit/-/merge_requests/509/diffs?diff_id=36990&start_sha=d484e645115685968a73c36b3b8ae388bd13343d)

kinow:
In GitLab by @dbeltrankyl on Dec 10, 2024, 15:01

Hello @isimo00 , 

Sorry, I lost track of this one with all the other stuff. 

Is something pending review, or are you still working on it? 

Note that the Notify_on value seems to not be there !532, so I believe you won't be able to do a manual functional test ( launch autosubmit run and see if you receive the mail )

kinow:
In GitLab by @isimo00 on Dec 10, 2024, 15:06

The code and testing is all done as of now but it won't be merged until the newest version of the configparser (1.0.76) is released, since this has some dependencies on stuff from the configparser that has been commited but not launched as a new version yet. So maybe it's better to wait until then and I after I fix the possible merge conflicts of the future before reviewing it.
Thank for the note on the Notify_on value!

dbeltrankyl:
Hello @isimo00 

I've updated the Autosubmit config parser to 1.0.76

isimo00:
Codecov bot showing poor coverage. Although not related to this issue, tests the following autosubmit.py funcitons should be added:

- [ ] `_init_logs`. Lines [864](https://github.com/BSC-ES/autosubmit/blob/e8877c9bc25fce98437385eb2d25f7e272142cb3/autosubmit/autosubmit.py#L864) to 940 (EOF)
- [ ] run_experiment. Lines [2154](https://github.com/BSC-ES/autosubmit/blob/e8877c9bc25fce98437385eb2d25f7e272142cb3/autosubmit/autosubmit.py#L2154) to 2156
- [ ] monitor. Lines 2606  to

kinow:
Maybe we can split it, and leave the pathlib/os.path/BasicConfig changes in the other PR... then this one just adds the email thingy.

isimo00:
@kinow For some reason I can't tag you as a reviewer, but this is finally done now!

kinow:
> @kinow For some reason I can't tag you as a reviewer, but this is finally done now!

It's because I'm the creator of this pull request (due to how the migration script works -- it uses the migration user in the GitHub API calls, as owner).

isimo00:
Ah I see. Well @kinow , let me know if you think something else is missing / you can't review it or I can tag Luiggi :)

test/unit/test_catlog.py
```diff
@@ -6,25 +6,33 @@
...
-
 class TestJob(TestCase):
 
     def setUp(self):
-        self.autosubmit = Autosubmit()
-        # directories used when searching for logs to cat
         self.original_root_dir = BasicConfig.LOCAL_ROOT_DIR
         self.root_dir = TemporaryDirectory()
         BasicConfig.LOCAL_ROOT_DIR = self.root_dir.name
-        self.exp_path = Path(self.root_dir.name, 'a000')
```
kinow, line 22: Without `Path`, the import `from pathlib import Path` is not used and can be removed.

image

test/unit/test_catlog.py
```diff
@@ -6,25 +6,33 @@
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from unittest.mock import patch
+import pytest
```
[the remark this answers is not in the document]
kinow: Maybe removed `suppress` from `contextlib` too. It seems it was used before, but not needed now.

test/unit/test_describe.py
```diff
@@ -33,29 +33,35 @@ def test_describe(
         if expid not in _EXPIDS:
             continue
         exp = autosubmit_exp(expid)
-
+
```
kinow, line 36: Unnecessary blank spaces (some IDEs like VS Code & PyCharm will delete these blank spaces when the file is saved -- if any line is modified, causing Git changes)

test/unit/test_describe.py
```diff
@@ -33,29 +33,35 @@ def test_describe(
         if expid not in _EXPIDS:
             continue
         exp = autosubmit_exp(expid)
-
+    
         basic_config = mocker.MagicMock()
+        # TODO: Whenever autosubmitconfigparser gets released with BasicConfig.expid_dir() and similar functions, the following line and a lot of mocks need to be removed. This line is especially delicate because it "overmocks" the basic_config mock, thus making the last assertion of this file "assert f'Location: {exp.exp_path}' in log_result_output" a dummy assertion. The rest of the test is still useful.
```
kinow, line 38: Can you break lines so it's shorter? It adds a warning (in the IDE, or ruff, other linters), but long lines like this one if modified are also harder to see what changed in them -- https://sembr.org/

test/unit/test_describe.py
```diff
@@ -33,29 +33,35 @@ def test_describe(
...
         mocked_get_submitter = mocker.patch.object(Autosubmit, '_get_submitter')
         submitter = mocker.MagicMock()
+
         mocked_get_submitter.return_value = submitter
         submitter.platforms = [1, 2]
 
         get_experiment_descrip = mocker.patch('autosubmit.autosubmit.get_experiment_descrip')
         get_experiment_descrip.return_value = [[f'{expid} description']]
-
+
```
kinow, line 64: Unnecessary spaces...

test/unit/test_describe.py
```diff
@@ -69,6 +75,7 @@ def test_describe(
             }
         )
 
+
```
kinow, line 78: Too many blank lines now, 

image

test/unit/test_describe.py
```diff
@@ -84,4 +91,4 @@ def test_describe(
         ]
         root_dir = exp.exp_path.parent
         for expid in expids:
-            assert f'Location: {str(root_dir / expid)}' in log_result_output
+            assert f'Location: {exp.exp_path}' in log_result_output
```
kinow, line 94: With this change, `root_dir` is not used, nor `expid`. So you could simplify it with something like,

```py
    if exp and not not_described:
        log_result_output = [
            line_tuple.args[0].format(line_tuple.args[1]) for line_tuple in Log.result.mock_calls
        ]
        for _ in expids:
            assert f'Location: {exp.exp_path}' in log_result_output
```

kinow: Note that in the future we will configure a linter to fail the build with changes like this, #2040 , but you can also configure IDEs to display this (as well as coverage) locally, e.g.

image

This is displayed by default in PyCharm without any configuration. vim/emacs/etc. should have a plugin/macro that does something like this too, or you can configure linters locally.

test/unit/test_mail.py
```diff
@@ -0,0 +1,116 @@
...
+    with mock.patch('smtplib.SMTP') as mock_smtp:
+        mock_server = mock.Mock()
+        mock_smtp.return_value = mock_server
+        mock_server.sendmail = mock.Mock()
+        
+        mail_notifier._send_mail('sender@example.com', 'recipient@example.com', mock.Mock())
+
+        mock_server.sendmail.assert_called_once_with('sender@example.com', 'recipient@example.com', mock.ANY)
+
+def test_attach_files(mail_notifier):
```
kinow, line 30: The functions in this file need to follow pep8 style, especially if it's new code :pray: 

I opened your code in PyCharm, and it immediately displays the number of warnings near the top,

image

Followed by marks on the right sidebar,

image

I will comment the other ones that it found (it normally matches with ruff/pep8/pylint/etc).

kinow (commented):
@isimo00 reviewed most of the changes, pending just `mail_notifier.py` now (:coffee: break)

autosubmit/notifications/mail_notifier.py
```diff
@@ -53,26 +65,34 @@ def notify_status_change(self, exp_id, job_name, prev_status, status, mail_to):
...
         server.quit()
 
+    def _attach_files(self, message, files):
+        for f in files or []:
+            with open(f, "rb") as file:
+                part = MIMEApplication(file.read(), Name = Path(f).name)
+                part['Content-Disposition'] = 'attachment; filename="%s"' % Path(f).name
+                message.attach(part)
+
+
```
kinow, line 79: Too many blank lines, :point_up:

autosubmit/notifications/mail_notifier.py
```diff
@@ -53,26 +65,34 @@ def notify_status_change(self, exp_id, job_name, prev_status, status, mail_to):
...
-                + f'Platform affected:{str(platform.name)} using as host:{str(platform.host)} \n\n' \
-                f'[WARN] Autosubmit encountered an issue with an remote_platform.\n It will resume itself, whenever is possible\n If issue persist, you can change the host IP or put multiple hosts in the platform.yml' + '\n\n\n\n\n' \
-                f'INFO: This message was auto generated by Autosubmit, '\
-                f'remember that you can disable these messages on Autosubmit config file. \n'
+        return f'''Autosubmit notification: Remote Platform issues\n-------------------------\n
+               Experiment id: {str(exp_id)}
+               Logs and errors: {BasicConfig.expid_aslog_dir(exp_id)}
+               Attached to this message you will find the related _run.log files.
+
+               Platform affected: {str(platform.name)} using as host: {str(platform.host)}\n\n[WARN] Autosubmit encountered an issue with a remote platform.\n It will resume itself, whenever is possible\n If this issue persists, you can change the host IP or put multiple hosts in the platform.yml file' + '\n\n\n\n\nINFO: This message was auto generated by Autosubmit,
```
kinow, line 97: Can we break the text like in the example above? Otherwise this line has 370 characters, which is quite long.

autosubmit/notifications/mail_notifier.py
```diff
@@ -53,26 +65,34 @@ def notify_status_change(self, exp_id, job_name, prev_status, status, mail_to):
...
+        return f'''Autosubmit notification\n
+               -------------------------\n\n
+               Experiment id:  {str(exp_id)} \n\n
+               Job name: {str(job_name)} \n\n
+               The job status has changed from: {str(prev_status)} to {str(status)} \n\n\n\n\n
+               INFO: This message was auto generated by Autosubmit,
+               remember that you can disable these messages on Autosubmit config file. \n'''
 
     @staticmethod
     def _generate_message_experiment_status(exp_id, platform=""):
```
kinow, line 91: Maybe `platform=""` needs to change? Can you change that to the correct default value if any (e.g. `None`?), and add type hints for mypy/ide/etc, please? We are adding these to new code as we add/modify. For example, something like,

```diff
diff --git a/autosubmit/notifications/mail_notifier.py b/autosubmit/notifications/mail_notifier.py
index bfcea3c6..4bdaff49 100644
--- a/autosubmit/notifications/mail_notifier.py
+++ b/autosubmit/notifications/mail_notifier.py
@@ -25,6 +25,11 @@ from email.mime.application import MIMEApplication
 from log.log import Log
 from pathlib import Path
 from autosubmitconfigparser.config.basicconfig import BasicConfig
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from autosubmit.platforms.platform import Platform
 
 class MailNotifier:
     def __init__(self, basic_config):
@@ -88,11 +93,11 @@ class MailNotifier:
                remember that you can disable these messages on Autosubmit config file. \n'''
 
     @staticmethod
-    def _generate_message_experiment_status(exp_id, platform=""):
+    def _generate_message_experiment_status(exp_id: str, platform: Platform = None) -> str:
         return f'''Autosubmit notification: Remote Platform issues\n-------------------------\n
-               Experiment id: {str(exp_id)}
+               Experiment id: {exp_id}
                Logs and errors: {BasicConfig.expid_aslog_dir(exp_id)}
                Attached to this message you will find the related _run.log files.
 
-               Platform affected: {str(platform.name)} using as host: {str(platform.host)}\n\n[WARN] Autosubmit encountered an issue with a remote platform.\n It will resume itself, whenever is possible\n If this issue persists, you can change the host IP or put multiple hosts in the platform.yml file' + '\n\n\n\n\nINFO: This message was auto generated by Autosubmit,
+               Platform affected: {platform.name} using as host: {platform.host}\n\n[WARN] Autosubmit encountered an issue with a remote platform.\n It will resume itself, whenever is possible\n If this issue persists, you can change the host IP or put multiple hosts in the platform.yml file' + '\n\n\n\n\nINFO: This message was auto generated by Autosubmit,
                 remember that you can disable these messages on Autosubmit config file.\n'''
diff --git a/test/unit/test_mail.py b/test/unit/test_mail.py
index a87ce3b3..b4ddc811 100644
--- a/test/unit/test_mail.py
+++ b/test/unit/test_mail.py
@@ -100,7 +100,8 @@ def test_notify_experiment_status(mail_notifier, mock_glob_side_effect, send_mai
                     else:
                         mail_notifier.notify_experiment_status(exp_id, mail_to, platform)
 
-                    mail_notifier._generate_message_experiment_status.assert_called_once_with(exp_id, platform)
+                    mail_notifier._generate_message_experiment_status.assert_called_once_with(  # type: ignore
+                        exp_id, platform)
                     mock_send_mail.assert_called_once_with(mail_notifier.config.MAIL_FROM, 'recipient@example.com', mock.ANY)
 
                     if mock_glob_side_effect is None: mock_attach_files.assert_called_once_with(mock.ANY, [file1, file2]) 
```

Refs:

- https://docs.python.org/3/library/typing.html
- https://peps.python.org/pep-0484/
- https://realpython.com/python-type-checking/

autosubmit/notifications/mail_notifier.py
```diff
@@ -20,18 +20,30 @@
...
+from autosubmitconfigparser.config.basicconfig import BasicConfig
 
 class MailNotifier:
     def __init__(self, basic_config):
         self.config = basic_config
 
     def notify_experiment_status(self, exp_id,mail_to,platform):
         message_text = self._generate_message_experiment_status(exp_id, platform)
-        message = MIMEText(message_text)
+        message = MIMEMultipart()
```
kinow, line 35: We need to confirm if this is the best option. I remember the discussion, but I didn't have time to ponder about it, sorry.

@mcastril, here we have a few options, basically,

1) we send a notification as we already do, but include the location of the logs in the VM/hub/somewhere
2) we attach the complete log as-is into the email body
3) we compress and then attach the logs (and delete the compressed archive as the original logs will remain anyway) into the email body

I think I am more included towards 1) to simplify or a 1.5) option, where we let users choose whether they want the location or the attachment, but I had seen a comment about doing the option 2), which is implemented here.

WDYT, Miguel?

autosubmit/notifications/mail_notifier.py
```diff
@@ -20,18 +20,30 @@
...
-        message['Subject'] = f'[Autosubmit] Warning a remote platform is malfunctioning'
+        message['Subject'] = f'[Autosubmit] Warning: a remote platform is malfunctioning'
         message['Date'] = email.utils.formatdate(localtime=True)
+        message.attach(MIMEText(message_text))
+
+        try:
+            files = [f for f in BasicConfig.expid_aslog_dir(exp_id).glob('*_run.log') if Path(f).is_file()]
+            self._attach_files(message, files)
+        except BaseException as e:
+            Log.printlog('An error has occurred while attaching log files to a warning email about remote_platforms ', 6011)
```
kinow, line 45: So if we fail to attach any file, we just print the log? Or should we move this into the for-loop in attach_files? (i.e. log when a file fails, but keep trying the remaining ones?)

And should we just log, or fail the process?

And should we perhaps log but also add something in the email/mime-message, in the notification saying "Here's your notification, but sorry I couldn't attach file ABC.log for reasons"? :grimacing:

autosubmit/notifications/mail_notifier.py
```diff
@@ -53,26 +65,34 @@ def notify_status_change(self, exp_id, job_name, prev_status, status, mail_to):
...
+        server = smtplib.SMTP(self.config.SMTP_SERVER, timeout=60)
         server.sendmail(mail_from, mail_to, message.as_string())
         server.quit()
 
+    def _attach_files(self, message, files):
+        for f in files or []:
+            with open(f, "rb") as file:
+                part = MIMEApplication(file.read(), Name = Path(f).name)
+                part['Content-Disposition'] = 'attachment; filename="%s"' % Path(f).name
+                message.attach(part)
```
kinow, line 77: If we use type hints in the rest of the code changed here, then we would be sure that `files` is a `List[Path]` (or as sure as we can get).

If we do not specify it's `Optional` or `Union[List[Path], None]`, then we are declaring the caller must prevent `None`s, and it eliminates the need for `in files or []` (but a good idea to be defensive, IMO).

Knowing we have `Path`s in the list, then we don't need to use a context-manager, so no need for the `with open(f, "rb") as file:`, and instead we can just read the content directly with `pathlib`'s utility functions. And also no need to re-wrap the `Path` object `f` in a new `Path`. Combined with f-strings, I think the new code can be simplified (with type hints) down to something like:

```diff
    def _attach_files(self, message, files: List[Path]) -> None:
        for f in files:
            part = MIMEApplication(f.read_text(), Name=f.name)
            part['Content-Disposition'] = f'attachment; filename="{f.name}"'
            message.attach(part)
```

test/unit/conftest.py
```diff
@@ -36,17 +36,21 @@ def autosubmit_exp(autosubmit: Autosubmit, request: pytest.FixtureRequest) -> Ca
...
-        status_dir = exp_path / 'status'
-        aslogs_dir.mkdir(parents=True, exist_ok=True)
-        status_dir.mkdir(parents=True, exist_ok=True)
-
+        exp_path = BasicConfig.expid_dir(expid)
+        
+        # directories used when searching for logs to cat
+        exp_tmp_dir = BasicConfig.expid_tmp_dir(expid) 
+        aslogs_dir = BasicConfig.expid_aslog_dir(expid) 
+        status_dir =exp_path / 'status'
```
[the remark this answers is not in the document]
isimo00: I'm changing everything to the `/` operator, but should be easy to change back if we decide to use `.joinpath`. We probably should discuss this and other similar choices with the entire team; for this particular one I'm not really aware of the difference (if there's any) between these two

isimo00:
Thank you for the extensive review @kinow! I've taken it as a sign to finally start using `pylint` and `autopep8`. I've marked as resolved the comments for which I've already followed your suggestions so far, will continue on Monday :)

test/unit/conftest.py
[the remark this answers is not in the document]
kinow: >I'm changing everything to the / operator, but should be easy to change back if we decide to use .joinpath. We probably should discuss this and other similar choices with the entire team;

Agreed!

> for this particular one I'm not really aware of the difference (if there's any) between these two

No difference :-)

https://github.com/python/cpython/blob/78ffba4221dcb2e39fd5db80c297d1777588bb59/Lib/pathlib/_local.py#L146-L158

They both call the same function in the end. The `/` operator was overloaded in `pathlib` for convenience. I still use `joinpath` when I need a function (like passing it to another function) for convenience. But otherwise I normally go with `/` as it reads equally simple and it's faster for me to type (except on the Spanish keyboard, as I never know where the keys are located :laughing: )

There are other operators used in Python too, like `+` with strings concatenation, or `|` `^` etc in sets (for union, difference, etc). I think that's also part of why they decided an operator was a good idea in pathlib (but it took me a while to get used to it).

https://github.com/python/cpython/blob/78ffba4221dcb2e39fd5db80c297d1777588bb59/Lib/pathlib/_local.py#L156

kinow:
> Thank you for the extensive review @kinow! I've taken it as a sign to finally start using `pylint` and `autopep8`. I've marked as resolved the comments for which I've already followed your suggestions so far, will continue on Monday :)

I think Luiggi is using Ruff. In the end, these tools in Python tend to follow more or less the same set of rules, based on pep8 (normally pep8 is a subset of all lint rules, as these tools add more on top of it). So that's a good start! But maybe check with Luiggi if it would be better if you used ruff too. This is up to discussion too, but ruff is also used in other BSC projects (and DestinE workflow, GSV, etc.) so there is a good case for us to use that too.

autosubmit/notifications/mail_notifier.py
isimo00: I think ideally we'd want to attach as many log files as possible; so omit the ones we can't retrieve (if any) but still attach the rest. Modifying the mail body to explain this would also be nice, i.e. "This email has attached all the log files we could retrieve, there might be some missing". 
I can add another log in the `attach_files` function and write more specific log messages. I'd rather have a new one than moving this one since they can catch different exceptions - for instance, what if the ASLOGS dir doesn't exist? -.

I'm also wondering if we should compress the log files before attaching them, to delay reaching the file attachment limit (what is it?) Overall, attaching the files is a nice touch but the most relevant information for the user is the log path, which is always printed in the mail body.

autosubmit/notifications/mail_notifier.py
isimo00: Since `TYPE_CHECKING` is `False` at runtime (and testing), shouldn't I import it normally?

autosubmit/notifications/mail_notifier.py
kinow: If any of these logs include things like compilation of a model like IFS with debug info, that might be a very-very-very large file (I've seen >100MB when someone enabled some print+debug in the compilation). Even zipping it could fail to send notifications due to server size limit.

IMHO it would be safer to add location, and if we want to give the file for the user convenience, then maybe a toggle that enables the files to be attached.

autosubmit/notifications/mail_notifier.py
```diff
@@ -20,59 +20,111 @@
...
+            ('Autosubmit', self.config.MAIL_FROM))
+        message['Subject'] = f'[Autosubmit] Warning: a remote platform is malfunctioning'
         message['Date'] = email.utils.formatdate(localtime=True)
+        message.attach(MIMEText(message_text))
+
+        try:
+            files: List[Path] = [p for p in BasicConfig.expid_aslog_dir(
+                exp_id).glob('*_run.log') if p.is_file()]
+            self._attach_files(message, files)
+        except BaseException as e:
```
kinow, line 57: Can be `Exception`, I think? And as we are not using `e`, we don't need to declare it,

```suggestion
        except Exception:
```

Catching `BaseException` is normally a bad idea as it could capture a system error, or when someone presses CTRL+C (in which cases we want the exception to be raised we the program will a) end up in some weird state or b) keep running and hide the error).
I think we can replace the `BaseException` below too.

kinow: (In reality I think it'd be best to catch the specific error, like `MessageError`, from the mail package, I think... but that's for future, as at the moment in AS we catch Exceptions pretty much everywhere (for another issue).

test/unit/conftest.py
```diff
@@ -29,30 +30,36 @@ class AutosubmitExperiment:
...
+        exp_path = BasicConfig.expid_dir(expid)
+
+        # directories used when searching for logs to cat
+        exp_tmp_dir = BasicConfig.expid_tmp_dir(expid)
+        aslogs_dir = BasicConfig.expid_aslog_dir(expid)
         status_dir = exp_path / 'status'
-        aslogs_dir.mkdir(parents=True, exist_ok=True)
-        status_dir.mkdir(parents=True, exist_ok=True)
+        os.makedirs(aslogs_dir, exist_ok = True)
+        os.makedirs(status_dir, exist_ok = True)
```
kinow, line 52: Sorry, I think I said to use the `exist_ok`, but now I realize we could also drop the `os` module here.

```
        aslogs_dir.mkdir(parents=True, exist_ok=True)
        status_dir.mkdir(parents=True, exist_ok=True)
```

I think that should work the same way, and then we can also remove the `import os` at the top.

test/unit/test_job_pytest.py
```diff
@@ -143,11 +153,15 @@ def test_recover_last_log_name(tmpdir, test_with_logfiles, file_timestamp_greate
...
+def test_update_parameters_attributes(
+        autosubmit_config,
+        experiment_data,
+        attributes_to_check):
     job = create_job_and_update_parameters(autosubmit_config, experiment_data)
     for attr in attributes_to_check:
         assert hasattr(job, attr)
-        assert getattr(job, attr) == attributes_to_check[attr]
+        assert getattr(job, attr) == attributes_to_check[attr] if \
+                type(attributes_to_check[attr]) is not list else attributes_to_check[attr][0]
```
kinow, line 164: Too many spaces :grimacing: 

And missing one blank space after the function:

image

autosubmit/notifications/mail_notifier.py
```diff
@@ -20,59 +20,111 @@
...
 
-    def notify_experiment_status(self, exp_id,mail_to,platform):
-        message_text = self._generate_message_experiment_status(exp_id, platform)
-        message = MIMEText(message_text)
-        message['From'] = email.utils.formataddr(('Autosubmit', self.config.MAIL_FROM))
-        message['Subject'] = f'[Autosubmit] Warning a remote platform is malfunctioning'
+    def notify_experiment_status(
+            self,
+            exp_id: str,
+            mail_to: str,
```
kinow, line 42: There's a type error in `test_mail.py`. I thought `mail_to` was `str`, but looks like it's a `List[str]`, sorry. Can you update it here?

image

test/unit/test_mail.py
```diff
@@ -0,0 +1,148 @@
...
+        assert file2.exists()
+
+        with mock.patch.object(mail_notifier, '_generate_message_experiment_status',
+                               return_value="Test message"), \
+                mock.patch.object(mail_notifier, '_send_mail') as mock_send_mail, \
+                mock.patch.object(mail_notifier, '_attach_files') as mock_attach_files, \
+                mock.patch.object(Log, 'printlog') as mock_printlog:
+
+            mail_to = ['recipient@example.com']
+            platform = mock.Mock()
```
kinow, line 113: We have `Platform` imported at the top, but not used. If you'd like you can add it here,

```
platform: Platform = mock.Mock()
```

this will instruct mypy/IDE/etc that `platform` is a `platform`... and then auto-complete should work (and other type checking tools...). Or remove the type checking import. It's shown as not used and adds one warning to the file.

image

kinow (commented):
Finished reviewing it, @isimo00 . I think we can wait for others (especially Miguel) to have time to review the part about attaching logs to the email body. But otherwise the code looks good to me. I added a couple of comments now after reading it one last time. The part about `BaseException` was already in the code, but maybe we can improve it here. Thanks!!

autosubmit/notifications/mail_notifier.py
```diff
@@ -63,7 +63,7 @@ def notify_experiment_status(
             message['To'] = email.utils.formataddr((mail, mail))
             try:
                 self._send_mail(self.config.MAIL_FROM, mail, message)
-            except BaseException as e:
+            except BaseException:
```
kinow, line 66: We can replace the `BaseException` by `Exception` here too :+1:

kinow (commented):
There are some changes appearing here that I am not sure if they were part of your initial change. Maybe something went wrong rebasing?

There are also some conflicts...

image

isimo00:
I noticed that too and undid the rebase, thanks @kinow !

kinow:
Great! @LuiggiTenorioK merged the PR that fixes the rocrate tests. I think if you rebase it (assuming your rebase was done earlier) that should fix the build here :crossed_fingers:

autosubmit/notifications/mail_notifier.py
kinow: @isimo00 from today's AS meeting:

- Let's compress the file by default
- We don't have to worry about extremely large files as we are only attaching _run logs (shouldn't be in the GB, mayyyybe MB, more likely b/kb)
- I will create separate issues for toggle/flags that disable attachment or restrict the attachment size :+1: 

So if you add the compression here by default, we can review & merge it :-)

kinow: Added to 4.1.13 backlog, https://github.com/BSC-ES/autosubmit/issues/2050

autosubmit/notifications/mail_notifier.py
```diff
@@ -18,26 +18,52 @@
...
         message['From'] = email.utils.formataddr(('Autosubmit', self.config.MAIL_FROM))
-        message['Subject'] = f'[Autosubmit] Warning a remote platform is malfunctioning'
+        message['Subject'] = f'[Autosubmit] Warning: a remote platform is malfunctioning'
         message['Date'] = email.utils.formatdate(localtime=True)
+        message.attach(MIMEText(message_text))
+
+        files = []
+        files_compressed = []
+        try:
+            files = [f for f in BasicConfig.expid_aslog_dir(exp_id).glob('*_run.log') if Path(f).is_file()]
```
kinow, line 46: From today's meeting: get the last log (not all run logs) :+1:

isimo00: Given that we already assumed more than one file would be sent and we have to add flags/toggles to the config, I thought we could give the user the choice to select and integer _n_ and receive the _n_ most recent (and available) _run.log files - default it to 1 -.

isimo00:
@kinow I believe this is ready to merge, if you'd like to do a final revision I'll be checking the issue for further comments

kinow:
> @kinow I believe this is ready to merge, if you'd like to do a final revision I'll be checking the issue for further comments

Excellent @isimo00 ! I'm leaving it unread in my GH notifications, planning to review it next Monday! :muscle: 

Thank you!

autosubmit/notifications/mail_notifier.py
```diff
@@ -18,61 +18,127 @@
...
+                exp_id).glob('*_run.log') if Path(f).is_file()]
+            files.sort()
+            files_compressed = [self._compress_file(f) for f in files[-1:]]
+        except BaseException as e:
+            Log.printlog(
+                'An error has occurred while compressing log files for a warning email',
+                6011)
+
+        try:
+            self._attach_files(message, files_compressed, files)
```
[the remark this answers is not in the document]
isimo00: Your suggestions to handle the sequence of "try"s make sense thanks! The temporary names can also go. But regarding the initialization and sorting of `files`; since we use list comprehension inside a try to create it, it must be initialized before (will raise initialization errors if the list comprehension fails and an exception is raised), and we sort it and get the last element instead of simply retrieving only the newest to give the user the option to receive the n-latest log files.
As you also mention, test pass but I'd need someone else to manually test it - I've done it locally and received the emails with the correct attachments as expected.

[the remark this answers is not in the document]
isimo00: Your suggestions to handle the sequence of "try"s make sense thanks! The temporary names can also go. But regarding the initialization and sorting of files; since we use list comprehension inside a try to create it, it must be initialized before (will raise initialization errors if the list comprehension fails and an exception is raised), and we sort it and get the last element instead of simply retrieving only the newest to give the user the option to receive the n-latest log files.
As you also mention, test pass but I'd need someone else to manually test it - I've done it locally and received the emails with the correct attachments as expected.

autosubmit/notifications/mail_notifier.py
```diff
@@ -18,61 +18,127 @@
...
+                    'Trace:{0}\nAn error has occurred while sending a mail for the job {0}'.format(
+                        e, job_name), 6011)
 
     def _send_mail(self, mail_from, mail_to, message):
-        server = smtplib.SMTP(self.config.SMTP_SERVER,timeout=60)
+        server = smtplib.SMTP(self.config.SMTP_SERVER, timeout=60)
         server.sendmail(mail_from, mail_to, message.as_string())
         server.quit()
 
+    def _attach_files(self, message, files, original_names):
```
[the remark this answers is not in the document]
kinow: Makes sense. With the `with TemporaryDirectory()` I think all files would go inside a temporary directory that's deleted later. So that should be safe too, I hope.

autosubmit/notifications/mail_notifier.py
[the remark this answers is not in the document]
kinow: Good point! I **think** I had a `if list` just before calling the code, but I added that later, while writing the unit test. So maybe the code and what I suggested here may not match 100%. And happy to test it, I haven't configured email on Linux since forever, but it wouldn't hurt configuring it to do a thorough final test.

isimo00:
@kinow I just re-tested this and I still receive the email perfectly, with the compressed file. When uncompressed, the contents are also correct. If besides looking at the code you could also do this - I can share my very inelegant but working script -, we should be able to merge this :)

kinow (commented):
Couple of comments, but nothing important. Tests look good, CICD happy. @isimo00 if you can look at these two last comments from today, I'll test your changes in parallel with real emails, and then I think if that works then that's that! :star:

kinow (commented):
@isimo00 nearly there!

I used the script you shared via Slack, modified it a bit and used a Docker container to spin up a test SMTP server. Here's the script, and steps to run:

```py
from autosubmit.notifications.mail_notifier import MailNotifier

from typing import cast

FROM_EMAIL = 'notifier@localhost'
TO_EMAIL = ['bruno.depaulakinoshita@bsc.es']
SMTP_HOST = 'localhost'
SMTP_PORT = 1025
EXPID = 'a000'
JOB_NAME = 'SIM'

config = type('', (), {
    'MAIL_FROM': FROM_EMAIL,
    'SMTP_SERVER': f'{SMTP_HOST}:{SMTP_PORT}'
})()

MAIL_NOTIFIER = MailNotifier(config)

# Status change

MAIL_NOTIFIER.notify_status_change(
    EXPID, JOB_NAME,
    Status.VALUE_TO_KEY[Status.RUNNING],
    Status.VALUE_TO_KEY[Status.FAILED],
    TO_EMAIL
)

# Exp status


platform = cast("Platform", type('', (), {'host': 'localhost', 'name': 'fake-local'}))

MAIL_NOTIFIER.notify_experiment_status(
    EXPID,
    TO_EMAIL,
    platform
)
```

Before running it, you need to start the local SMTP server, for which I used this container https://github.com/haravich/fake-smtp-server (also in Docker Hub):

```bash
$ docker pull haravich/fake-smtp-server
$ docker run -p 1025:1025 -p 1080:1080 haravich/fake-smtp-server
```

Once you start it, the port `1025` is for SMTP messages, and `1080` to access the Web interface. Here's what it looks like after you run the script once:

image

Two emails, once for status change, and one for experiment status report. I looked at the body of the message and both look correct :clap: 

The only final feedback I have is:

* I think we should save your script somewhere in the project sources... maybe `test/manual`, or `docs/scripts`, or under a new folder;
* I accidentally used `TO_EMAIL = 'bruno.depaulakinoshita@bsc.es'` (not a list) and the code iterated the string (`for a in string` behaves the same as `for a in list` and for other iterables in Python), which resulted in multiple emails to b, r, u, n, ...; better if we can prevent this.

For the last point, I thought about something like

```diff
diff --git a/autosubmit/notifications/mail_notifier.py b/autosubmit/notifications/mail_notifier.py
index da36dada..1d8a6036 100644
--- a/autosubmit/notifications/mail_notifier.py
+++ b/autosubmit/notifications/mail_notifier.py
@@ -163,6 +163,8 @@ class MailNotifier:
         :param platform: The platform.
         :type platform: Platform
         """
+        if not isinstance(mail_to, list):
+            raise ValueError('mail_to must be a list of emails!')
         message_text = _generate_message_experiment_status(exp_id, platform)
         message = MIMEMultipart()
         message['From'] = email.utils.formataddr(
@@ -201,6 +203,8 @@ class MailNotifier:
             prev_status: str,
             status: str,
             mail_to: List[str]) -> None:
+        if not isinstance(mail_to, list):
+            raise ValueError('mail_to must be a list of emails!')
         message_text = _generate_message_text(
             exp_id, job_name, prev_status, status)
         message = MIMEText(message_text)
```

Or something similar... would still need to check coverage, but I think it needs a small accompanying test to verify thew new lines... 

Otherwise, +1 from me! Nearly there!

kinow:
Sorry, last feedback (I think! :grimacing: ), could you also replace the `BasicConfig` used in the function that attaches the files, by `self.config`, please? We already have the `BasicConfig` object, so no need to use the imported/global one.

With that done, I used this version of the script (slightly modified) after saving it in a folder with some test empty files to be attached.

```bash
test/manual/
└── notifications
    ├── log_files
    │   ├── a000_run.log
    │   ├── a000_run.log.txt
    │   ├── a001_run.log
    │   └── _run.log_run.log.txt
    └── send_notifications.py
```

In the example above, `send_notifications.py` is your script, and it must attach one file (I created `a001_run.log` after `a000`, so only `a001_run.log`).

image

image

So this looks correct to me! Great job, @isimo00 !

test/unit/test_mail.py
```diff
@@ -20,6 +20,9 @@ def mock_basic_config(mocker):
     mock_config = mocker.Mock()
     mock_config.MAIL_FROM = "test@example.com"
     mock_config.SMTP_SERVER = "smtp.example.com"
+    def mock_expid_aslog_dir(exp_id):
+        return BasicConfig.expid_aslog_dir(exp_id)
+    mock_config.expid_aslog_dir.side_effect = mock_expid_aslog_dir
```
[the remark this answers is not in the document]
isimo00: Forgot the linter :')

kinow (commented):
Impeccable! IDE happy, CICD happy, I can't see anything to change reviewing the code, and we tested it yesterday unit + manual tests :clap: Let's get it merged! Thanks @isimo00 !

kinow:
I cannot approve it, as due to migration issue appears as mine, but it's approved! :tada: I just asked GitHub got rebase it, once CICD completes, feel free to merge it

PR #1993, opened 2024-12-13
Draft:Add provenance functionality

kinow:
In GitLab by @apuiggro on Oct 28, 2024, 12:18

PR #2017, opened 2024-12-13
Draft: Run provenance

kinow:
In GitLab by @apuiggro on Dec 5, 2024, 12:18

When running an experiment, this feature detects if an rocrate.yml object exists in the config and automatically creates an rocrate object after the execution is finished. 

Pending on merge request: https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/505. This feature can only work if provenance function is inlcuded in `autosubmit.py`.


kinow (commented):
Will review this one after #1993 is done. If you remember, you can rebase this one after rebasing that other one, as that would save some time for when we are ready to review it. Thanks!

PR #2032, opened 2024-12-16
4.1.12 dev branch failed jobs intregation test

dbeltrankyl:
I've added integration tests for  run again after a sigterm and a sigkill 

This file tests is quite expensive to run so we should move it to another folder ( maybe after I come back from Holidays)


dbeltrankyl:
Well, the test in pipeline failed. 

That was the reason for doing it in a different branch. 

One can download it to test them locally

This is not blocking for the 4.1.12 release  we can fix it later

dbeltrankyl:
The target branch was squashed and rebased, so this branch is now adding changes already in the master.. so discarding it in favor of #2034

Issue #2052, opened 2025-01-02
Autosubmit & autosubmit-config-parser both contain configuration logic


kinow:
I believe the idea of separating the module was to allow multiple users to use autosubmit-config-parser, without having to import all of Autosubmit code base (which is not that large).

While working on https://github.com/BSC-ES/autosubmit/issues/2043 I realized that I could not debug the issue with only the config parser. The place where `CUSTOM_DIRECTIVES` becomes a list is on the Autosubmit side.

https://github.com/BSC-ES/autosubmit/blob/0ad5a340a9b391b9950caa10c2537f74efa60f19/autosubmit/job/job.py#L1656-L1660

After AS modifies `custom_directives` then another function in `as_conf` is called (`substitute_dynamic_variables`, I believe) where the error is raised. But it only happens if a user of config parser modifies the type of a variable, and if that variable contains placeholders. A convoluted example, but alas a real world example, that can happen again (in operational workflows, for example).

And this does not seem to be the only case. I had seen a similar "abstraction leakage" [^1][^2] some time ago while debugging another issue.

If I had to load Autosubmit experiments configs, I would not use Autosubmit Config Parser alone today, as it wouldn't really resolve configurations like that `CUSTOM_DIRECTIVES`. Instead, I'd use something like Autosubmit's `check` or JobList's `check_scripts` functions as they seem to load the configuration and also call the other functions that handle cases like this of custom directives or other configuration settings that require AS + config parser.

Other places in AS where configuration files or logic is handled, instead of using an abstraction layer like as_conf as being the point to interact with AS configuration:

- function that sets default values in AS config, https://github.com/BSC-ES/autosubmit/blob/0ad5a340a9b391b9950caa10c2537f74efa60f19/autosubmit/autosubmit.py#L1293-L1294, called in `autosubmit expid`
- Job's `update_parameters` that refreshes parameters values, many of the parameters are read from configuration files, and the order of function calls to as_conf is controlled by AS (if I needed that to generate an ecFlow experiment, for example, I'd need to call something that calls that function to get the placeholders correctly parsed, I believe) https://github.com/BSC-ES/autosubmit/blob/0ad5a340a9b391b9950caa10c2537f74efa60f19/autosubmit/job/job.py#L2038-L2082
[...]

PR #2057, opened 2025-01-07
Custom directives added letter by letter

dbeltrankyl:
Note: this requires https://github.com/BSC-ES/autosubmit-config-parser/pull/59 changes. 

The issue:

I noticed that although the custom_directive is assigned to the variable, it is not assigned in the correct order. 

Some normalization or/and preparation stuff happens while the "custom_directive" is still assigned to a placeholder. Ex. CUSTOM_DIRECTIVE: "%current_something%

So it does nothing. 

Configuration:
https://github.com/BSC-ES/autosubmit-config-parser/pull/59#issuecomment-2575118765

Bug:

https://github.com/BSC-ES/autosubmit-config-parser/pull/59#issuecomment-2575298435

Solution: 

The solution consists in:

Instead of getting the value from job_data, it is now getting from the parameters dict based on CURRENT_ values. With this, we ensure to have the most updated variable, with all the %placeholders% substituted
Move the order of calculating stuff. 
Now, it updates the platform stuff and adds the CURRENT parameters associated only with it.
Then, it adds all the %CURRENT_% parameters that are set in the YAML platforms and jobs file.
Then, it substitutes all the placeholders in the parameters dict for the appropriated value.
Finally, according to the first point, it gets the updated value from the parameters dict instead of the raw job_data file.

I think it is ready to review (while doing the changes, some tests were complaining, so I think it is covered, but I will see the coverage tomorrow). I'll add one to check the $SBATCH of the directives similar to the one we have for the reservation

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,3 @@
-4.1.12 - TODO
-
 4.1.11 - Enhancements, New Features, Documentation, and Bug Fixes
 =================================================================
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,5 @@
+4.1.12 - TODO
+
 4.1.11 - Enhancements, New Features, Documentation, and Bug Fixes
 =================================================================
 
@@ -155,4 +157,3 @@ Autosubmit supports much larger workflows in this version and has improved perfo
 - Added the previous keyword.
 - Fixed an issue with the historical db.
 - Fixed an issue with historical db logs.
-
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -1436,7 +1436,9 @@ class Job(object):
             parameters[f"CURRENT_{key.upper()}"] = value
         for key, value in as_conf.jobs_data[self.section].items():
             parameters[f"CURRENT_{key.upper()}"] = value
+
         return parameters
+
     def update_platform_parameters(self,as_conf,parameters,job_platform):
         if not job_platform:
             submitter = job_utils._get_submitter(as_conf)
@@ -1458,7 +1460,7 @@ class Job(object):
         parameters['CURRENT_LOGDIR'] = job_platform.get_files_path()
         return parameters
 
-    def process_scheduler_parameters(self,as_conf,parameters,job_platform,chunk):
+    def process_scheduler_parameters(self, job_platform, chunk):
         """
         Parsers yaml data stored in the dictionary
         and calculates the components of the heterogeneous job if any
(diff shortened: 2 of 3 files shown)


dbeltrankyl:
Hello @kinow 

It took a while, but I added a test. 

Tests the custom_directive in 8 different setups and ensures through a regex that it is well set in the slurmheader, matching the start of the line, the directive, and the end of the line.

kinow (approved):
@dbeltrankyl thank you so much for fixing this!

You are not only fixing the custom directive added letter by letter. I tested [the last issue I had in my PR](https://github.com/BSC-ES/autosubmit-config-parser/pull/59#issuecomment-2569115342), where it would remove the placeholder that didn't include `CURRENT_` when both values were used in a list (i.e. `CUSTOM_DIRECTIVES: ["CURRENT_SOMETHING", "DEFAULT.EXPID"]`, where `DEFAULT.EXPID` was discarded).

I tested with `master` first. I have my branch of the config-parser `pip install -e .`'ed in my Python environment, with `autosubmit`'s `master` branch checked out.

```bash
(autosubmit4) kinow@ranma:~/Development/python/workspace/autosubmit$ autosubmit -lc DEBUG create -f -cw -np -v a030 && autosubmit -lc DEBUG inspect a030 && grep "bruno" -A2 /home/kinow/autosubmit/a030/tmp/a030_19900101_fc0_1_31_OPA_ENERGY_ONSHORE_1.cmd
...
...
...
echo "bruno"
echo "custom directive --mem=1g"
echo "test "
```

So, as can be seen above, only one of the values used for `%I_AM_NOT_THE_CUSTOM_DIRECTIVES_YOU_ARE_LOOKING_FOR%` is being substituted by Autosubmit, and the second value is discarded.

Now, I only switched the branch of Autosubmit to this one you created, @dbeltrankyl, and executed the same test.

```bash
...
...
...
/home/kinow/autosubmit/a030/tmp/a030_19900101_fc0_1_30_OPA_ENERGY_ONSHORE_1.cmd
/home/kinow/autosubmit/a030/tmp/a030_19900101_fc0_1_31_OPA_ENERGY_ONSHORE_1.cmd

echo "bruno"
echo "custom directive ['--mem=1g', 'a030']"
echo "test "
```

This time all the variables I used in my dynamic variables/placeholders were substituted as expected.

I think we are good here! Thanks @dbeltrankyl ! :pray: :bow: 

@mcastril with this merged into the other PR we started earlier, I think @franra9 's experiments should work in the testing suite without the hard-coded values. Note too, that this fix was in the Autosubmit side. I would have spent more time looking at the config-parser code before jumping here, which is why I would urge us to review merging back the projects -- https://github.com/BSC-ES/autosubmit/issues/2052 -- it gives more trouble to devs & users than help, IMHO

@dbeltrankyl feel free to merge this as I will finish the testing in the other PR review (I will copy @franra9 's experiment that was failing to my local laptop and `create` + `inspect` to confirm our fix works).

Cheers

dbeltrankyl:
Great @kinow, that it also solved that! 

Merging then, when you merge the other branch, I'll add a new version to the climate-dt

PR #2070, opened 2025-01-20
Adjust log_path  for existing job_list

dbeltrankyl:
This branches fixes a bug related with recovery + log_path when it continues the run with a job_list of a previous version. 

Ready to review


kinow (approved):
I think we probably should include it in 4.1.12-release branch too, for our next release soon-ish.  Thank you!

Issue #1384, opened 2024-08-07
Fix deprecation warning regarding setup.py and pyproject.toml


kinow:
While reviewing @naiaraAM 's code in !459 , I created a new Micromamba env on a new laptop, and doing everything over from scratch I got this warning with Py 3.9.12 running `pip install -e .`:

>  DEPRECATION: Legacy editable install of autosubmit==4.1.9 from file:///home/kinow/Development/python/workspace/autosubmit (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

It's probably better that we fix this earlier, instead of waiting until they remove it which may prevent AS from working with newer versions of Python for a simple issue that can be fixed in under one hour.

kinow:
In GitLab by @edgano on Aug 12, 2024, 13:53

Looking for more info about these. I found this info here: https://packaging.python.org/en/latest/discussions/setup-py-deprecated/

What commands should be used instead?
| Deprecated | Recomendation |
| ------ | ------ |
|   python setup.py install     |     python -m pip install .   |
|      ...  |   ...     |

But not sure which could be the best strategy

kinow:
I think calling `python setup.py install` directly has been considered old/deprecated for a long time. Some scripts that I used in the past did not work correctly when called directly like that.

I think either `python -m pip install .` or just `pip install .` should be OK.

The error, I believe, is due to the missing `pyproject.toml`. Other projects have added or migrated to `pyproject.toml`. It would be hard for us to migrate, I think, as I don't know if the way we retrieve the project version would work.

But we can add a pyproject.toml and move part of the metadata, to see if the deprecation warning goes away. Then in the future we can move more things, and follow the PEP's and see which direction we need/want to follow.

WDYT?

kinow:
In GitLab by @edgano on Aug 12, 2024, 14:25

A ok!
yes, that sounds good ;) a good way to test without breaking anything. IDK what the others thinks about it

dbeltrankyl:
In relation to this, we should also fix this: 

https://autosubmit.readthedocs.io/en/master/installation/index.html#installation

The docs are still referring to the use

python3 setup.py install 

[...]

PR #2079, opened 2025-01-27
Create pyproject.toml for #1384

isimo00:
Fix `setup.py` deprecation warning by creating a `pyproject.toml` file

> DEPRECATION: Legacy editable install of autosubmit==4.1.9 from file:///home/kinow/Development/python/workspace/autosubmit (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

From issue #1384 

cc @kinow @dbeltrankyl

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ dependencies = [
     "packaging<=23.2",
     "typing_extensions<=4.9.0",
     "typing<=3.7.4.3",
-    "psutil<=5.6.1",
+    "psutil<=7.0.0",
     "py3dotplus==1.1.0",
     "numpy<2",
     "rocrate==0.*",


pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "autosubmit"
+dynamic = ["version", "readme"]
+description = "Autosubmit is a Python-based workflow manager to create, manage and monitor complex tasks involving different substeps, such as scientific computational experiments. These workflows may involve multiple computing systems for their completion, from HPCs to post-processing clusters or workstations. Autosubmit can orchestrate all the tasks integrating the workflow by managing their dependencies, interfacing with all the platforms involved, and handling eventual errors."
+requires-python = ">=3.8"
```
dbeltrankyl, line 28: I would set this to 3.9. Nobody is using 3.8, and our tests do not cover it. 

I would also include that the last version supported is 3.1.12 ( v3.1.13 was released last month and may or not may work)

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "autosubmitconfigparser==1.0.76",
+    "configparser",
+    "setproctitle",
+    "invoke>=2.0"
+]
+
+[project.urls]
+Homepage = "http://www.bsc.es/projects/earthscience/autosubmit/"
+Documentation = "https://readthedocs.org"
+"Download Link" = "https://earth.bsc.es/wiki/doku.php?id=tools:autosubmit"
```
dbeltrankyl, line 72: Can we add the GitHub URL? https://github.com/BSC-ES/autosubmit

Maybe mention the GUI? somewhere cc: @kinow or too out-scope?

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "rocrate==0.*",
+    "autosubmitconfigparser==1.0.76",
+    "configparser",
+    "setproctitle",
+    "invoke>=2.0"
+]
+
+[project.urls]
+Homepage = "http://www.bsc.es/projects/earthscience/autosubmit/"
+Documentation = "https://readthedocs.org"
```
dbeltrankyl, line 71: "https://readthedocs.org" -> "https://autosubmit.readthedocs.io/"

pyproject.toml
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python3
+
+# Copyright 2014 Climate Forecasting Unit, IC3
```
dbeltrankyl, line 3: This copyright.. mentions the 2014 Climate Forecasting Unit, IC3, which I believe doesn't exist anymore. 

To change for 

```python
#  Copyright 2014-2025 Barcelona Supercomputing Center (http://www.bsc.es/)
```

I think cc: @kinow

dbeltrankyl:
Hey @isimo00 

Since we're moving the metadata I commented on some stuff to update as they seem deprecated. What you touched looks good!

pyproject.toml
dbeltrankyl: Ah nevermind, I thought that "Download Link" = "https://earth.bsc.es/wiki/doku.php?id=tools:autosubmit"
was only internal ( double-checked with the phone). We can leave it as it is. Also it contains the GUI

dbeltrankyl:
Ah, Sorry I forgot. 

There was also this comment, we can use this branch or do it in a newer one as you prefer

https://github.com/BSC-ES/autosubmit/issues/1384#issuecomment-2609148845

pyproject.toml
kinow: +1, the IC3 is that unit/place where Paco and others used to work. +1 to BSC instead.

isimo00:
> Ah, Sorry, I forgot.
> 
> There was also this comment, we can use this branch or do it in a newer one as you prefer
> 
> [#1384 (comment)](https://github.com/BSC-ES/autosubmit/issues/1384#issuecomment-2609148845).
> 
> If you prefer to do it in a new branch, please open an issue because 1384 will be closed once this is merged.
> 
> What I asked there is to update the readthedocs to reflect the new pip install instructions. ( And now also, the readme as it mentions pip install . , not your newer additions)

I can do it in this issue, no problem

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2014 Climate Forecasting Unit, IC3
+# Copyright 2014-2025 Barcelona Supercomputing Center (http://www.bsc.es/)
```
kinow, line 3: https? :grimacing: Avoids a redirect and some linters look for http and then you have to whitelist the ones that are not really links used/clickable-links :pray:

.github/workflows/ci.yaml
```diff
@@ -55,7 +55,10 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e .[all]
+          pip install build
+          python -m build
+          WHEEL_FILE=$(ls dist/*.whl)
+          pip install $WHEEL_FILE[all]
```
kinow, line 61: Was there a reason to install the wheel? In Cylc, some long time ago, we ended up with some messy code in one case where (if memory serves me well) the relative imports (`from .this.bla`) worked when installing with `pip install .` or `pip install $pkg`, but failed in the editable mode. Then we used the editable version in CICD to prevent that, and I never had the case where something worked in editable, but not with wheel/source via pip... so I just defaulted to always using editable in CICD... although maybe whatever issue we had is now long fixed in the python build tooling (which is newer).

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+[project]
+name = "autosubmit"
+dynamic = ["version", "readme"]
+description = "Autosubmit is a Python-based workflow manager to create, manage and monitor complex tasks involving different substeps, such as scientific computational experiments. These workflows may involve multiple computing systems for their completion, from HPCs to post-processing clusters or workstations. Autosubmit can orchestrate all the tasks integrating the workflow by managing their dependencies, interfacing with all the platforms involved, and handling eventual errors."
+requires-python = ">=3.9,<3.13"
+authors = [
+  { name="The Autosubmit Team - Models and Workflows Team (MWT) - Barcelona Supercomputing Center (BSC)", email="support-autosubmit@bsc.es" },
+]
+maintainers = [
+  { name="The Autosubmit Team - Models and Worklows Team (MWT) - Barcelona Supercomputing Center (BSC)", email="support-autosubmit@bsc.es" },
```
kinow, line 33: Ha! My OCD :laughing: 

I noticed these two, authors and maintainers, were nicely aligned, but the line length was not the same haha. Spot the type! :timer_clock:

kinow (commented):
Looking good! Found a type, and left a command about the CI pipeline.

@isimo00 could you test deleting the duplicated info from `setup.py` now, please? We shouldn't really need to maintain both `setup.py` and `pyproject.toml`.

The new build tools behind setuptools/pip/etc. are able to handle when you have a `pyproject.toml`, and load that by default. I think `setup.py` is still used, but I don't remember if it's always loaded, or if there's a loading order -- i.e. probably safer to avoid duplication in case something picks `pyproject.toml` in one place, and `setup.py` in other.

Some example projects where I remember setup.py and pyproject.toml are mixed (in the future I think everything will go to `pyproject.toml`, unless a project needs C/C++ libs during build, maybe some obscure Cython builds, etc.)

- https://github.com/cylc/cylc-flow
- https://github.com/common-workflow-language/cwltool
- https://github.com/Textualize/rich
- https://github.com/ecmwf/pyfdb
- ...

pyproject.toml
isimo00: I guess we work so efficiently that our worklows are getting shorter! 😆

kinow: Hahaha, good one. Gonna save that one for the next time I see a similar typo!

kinow (commented):
@isimo00 sorry I couldn't reply earlier. I modified `pyproject.toml` and deleted `setup.py`. It's not yet working. Here's what I have:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 23a96312..fa71689d 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ maintainers = [
 ]
 
 keywords = ["climate", "weather", "workflow", "HPC"]
-license = {text = "GNU GPL v3"} 
+license = {text = "GNU GPL v3"}
 
 classifiers = [
     "Programming Language :: Python :: 3.8",
@@ -63,11 +63,12 @@ dependencies = [
     "autosubmitconfigparser==1.0.76",
     "configparser",
     "setproctitle",
-    "invoke>=2.0"
+    "invoke>=2.0",
+    # 'sqlalchemy[mypy]' # TODO: pending Postgres MR
 ]
 
 [project.urls]
-Homepage = "http://www.bsc.es/projects/earthscience/autosubmit/"
+Homepage = "https://www.bsc.es/projects/earthscience/autosubmit/"
 Documentation = "https://autosubmit.readthedocs.io/"
 "Download Link" = "https://earth.bsc.es/wiki/doku.php?id=tools:autosubmit"
 
@@ -81,7 +82,8 @@ tests = [
     "pytest==8.2.*",
     "pytest-cov",
     "pytest-mock",
-    "ruff==0.6.2"
+    "ruff==0.6.2",
+    # 'testcontainers'  # TODO: pending Postgres MR
 ]
 docs = [
     "livereload",
@@ -103,3 +105,22 @@ all = ["psycopg2",
     "sphinx_rtd_theme",
     "sphinx-reredirects==0.1.*"
 ]
+
+[tool.setuptools.packages.find]
+where = ["."]
+
+[tool.setuptools]
+package-data = { "autosubmit" = [
+    "autosubmit/config/files/autosubmit.conf",
+    "autosubmit/config/files/expdef.conf",
+    "autosubmit/database/data/autosubmit.sql",
+    "README",
+    "CHANGELOG",
+    "VERSION",
+    "LICENSE",
+    "docs/autosubmit.pdf"
+] }
+include-package-data = true
+
+[project.scripts]
+autosubmit = "bin.autosubmit:main"
```

And,

```bash
(autosubmit4) kinow@ranma:~/Development/python/workspace/autosubmit$ git status
On branch iss1384
Your branch is up to date with 'upstream/iss1384'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	deleted:    setup.py

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   pyproject.toml

```

Running `pip install -e .` works, but `autosubmit --version` fails:

```bash
(autosubmit4) kinow@ranma:~/Development/python/workspace/autosubmit$ autosubmit --version
  File "/home/kinow/Development/python/micromamba/envs/autosubmit4/bin/autosubmit", line 5, in <module>
    from bin.autosubmit import main
ModuleNotFoundError: No module named 'bin.autosubmit'
```

The issue here, Irene, is that we are installing a package `autosubmit`, and then including a script that's not in that package. So `setuptools` looks for an entrypoint (you can look that up in PyPA docs), but fails to locate `bin.autosubmit`. I tried adding that, which is not really the best fix, and while that worked, `python` was not able to locate that module in runtime.

So we have two options here:

1. Find out the current setting for including a console script/entrypoint that's not part of the package,
2. Tidy up our code and move the script to inside our repository

The second option is part of what I intended to do this Christmas in that refactoring to split `autosubmit.py` and its 6K lines. Basically moving the console scripts to within our main `autosubmit` package, creating scripts for the subcommands, and using a decorator and some short Python code o load subcommands automatically.

If that were done, you would not have this issue right now :disappointed_relieved: We can chat more next week, but feel free to read over what I wrote, and see if maybe you have some other ideas, or find out something else.

:wave:

kinow (commented):
@isimo00 unfortunately I don't think it'd be simple to keep the current behaviour. At the moment `setup.py` and setuptools seem to accept the script outside the package  `autosubmit`.

But if you look at the `bin/autosubmit`, it's really awkward as that's a Python script without the extension, and that uses `sys` to manipulate the PYTHONPATH so that it can use other Autosubmit Python code.

I think instead we should use this opportunity to fix it and make it into a proper script as per setuptools guidelines, and use an entrypoint (in case we want to extend/use plug-ins/etc.).

https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point

Here's one way that worked for me. Have a look if that works for you too, and if you think it'd unblock this PR.

Cheers

```diff
diff --git a/autosubmit/scripts/__init__.py b/autosubmit/scripts/__init__.py
new file mode 100644
index 00000000..78b51463
--- /dev/null
+++ b/autosubmit/scripts/__init__.py
@@ -0,0 +1,18 @@
+# Copyright 2015 Earth Sciences Department, BSC-CNS
+
+# This file is part of Autosubmit.
+
+# Autosubmit is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+# Autosubmit is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Scripts for running Autosubmit."""
diff --git a/bin/autosubmit b/autosubmit/scripts/autosubmit.py
similarity index 89%
rename from bin/autosubmit
rename to autosubmit/scripts/autosubmit.py
index 6fcb569d..1ee357bf 100755
--- a/bin/autosubmit
+++ b/autosubmit/scripts/autosubmit.py
@@ -18,12 +18,8 @@
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
 """Script for handling experiment monitoring"""
-import os
 import sys
 
-script_dir = os.path.abspath(os.path.dirname(sys.argv[0]))
-sys.path.append(script_dir)
-sys.path.append(os.path.normpath(os.path.join(script_dir, os.pardir)))
 # noinspection PyUnresolvedReferences
 from log.log import Log, AutosubmitCritical, AutosubmitError
 from autosubmit import delete_lock_file, exit_from_error
diff --git a/pyproject.toml b/pyproject.toml
index b1c96edf..4f3fd502 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,6 +122,6 @@ package-data = { "autosubmit" = [
 ] }
 include-package-data = true
 
-[project.scripts]
-autosubmit = "bin.autosubmit:main"
+[project.entry-points.console_scripts]
+autosubmit = "autosubmit.scripts.autosubmit:main"
 
diff --git a/setup.py b/setup.py
deleted file mode 100644
index 3cee1642..00000000
--- a/setup.py
+++ /dev/null
@@ -1,120 +0,0 @@
-#!/usr/bin/env python3
-
-# Copyright 2014 Climate Forecasting Unit, IC3
-
-# This file is part of Autosubmit.
-
-# Autosubmit is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-
-# Autosubmit is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-
-# You should have received a copy of the GNU General Public License
-# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
-
-from os import path
-from setuptools import setup
-from setuptools import find_packages
-
-here = path.abspath(path.dirname(__file__))
-
-# Get the version number from the relevant file
-with open(path.join(here, 'VERSION')) as f:
-    version = f.read().strip()
-
-install_requires = [
-    'xlib==0.21',
-    'setuptools<=68.2.2',
-    'bscearth.utils<=0.5.2',
-    'requests<=2.31.0',
-    'networkx<=2.6.3',
-    'portalocker<=2.7.0',
-    'paramiko>=3.5.0',
-    'pyparsing==3.1.1',
-    'matplotlib<=3.8.3',
-    'packaging<=23.2',
-    'typing_extensions<=4.9.0',
-    'typing<=3.7.4.3',
-    'psutil<=5.6.1',
-    'py3dotplus==1.1.0',
-    'numpy<2',
-    'rocrate==0.*',
-    'autosubmitconfigparser==1.0.76',
-    'configparser',
-    'setproctitle',
-    'invoke>=2.0',
-    # 'sqlalchemy[mypy]' # TODO: pending Postgres MR
-]
-
-pg_require = [
-    'psycopg2'
-]
-
-docs_require = [
-    'livereload',
-    'pydata-sphinx-theme==0.15.*',
-    'sphinx==5.*',
-    'sphinx-autobuild==2021.3.*',
-    'sphinx_rtd_theme',
-    'sphinx-reredirects==0.1.*'
-]
-
-tests_require = [
-    'pytest==8.2.*',
-    'pytest-cov',
-    'pytest-mock',
-    'ruff==0.6.2',
-    # 'testcontainers'  # TODO: pending Postgres MR
-]
-
-# You can add more groups, e.g. all_require = tests_require + graph_require, etc...
-all_require = tests_require + pg_require
-
-extras_require = {
-    'postgres': pg_require,
-    'tests': tests_require,
-    'docs': docs_require,
-    'all': all_require
-}
-
-setup(
-    name='autosubmit',
-    license='GNU GPL v3',
-    platforms=['GNU/Linux Debian'],
-    version=version,
-    description='Autosubmit is a Python-based workflow manager to create, manage and monitor complex tasks involving different substeps, such as scientific computational experiments. These workflows may involve multiple computing systems for their completion, from HPCs to post-processing clusters or workstations. Autosubmit can orchestrate all the tasks integrating the workflow by managing their dependencies, interfacing with all the platforms involved, and handling eventual errors.',
-    long_description=open('README_PIP.md').read(),
-    author='Daniel Beltran Mora',
-    author_email='daniel.beltran@bsc.es',
-    url='http://www.bsc.es/projects/earthscience/autosubmit/',
-    download_url='https://earth.bsc.es/wiki/doku.php?id=tools:autosubmit',
-    keywords=['climate', 'weather', 'workflow', 'HPC'],
-    install_requires=install_requires,
-    extras_require=extras_require,
-    classifiers=[
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
-        "License :: OSI Approved :: GNU General Public License (GPL)",
-        "Operating System :: POSIX :: Linux",
-    ],
-    packages=find_packages(),
-    include_package_data=True,
-    package_data={'autosubmit': [
-        'autosubmit/config/files/autosubmit.conf',
-        'autosubmit/config/files/expdef.conf',
-        'autosubmit/database/data/autosubmit.sql',
-        'README',
-        'CHANGELOG',
-        'VERSION',
-        'LICENSE',
-        'docs/autosubmit.pdf'
-    ]
-    },
-    scripts=['bin/autosubmit']
-)
```

(This deletes `setup.py`, adds the entry point for autosubmit.scripts.autosubmit, and moves the bin/autosubmit to this scripts location.)

.github/workflows/ci.yaml
```diff
@@ -55,7 +55,9 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e .[all]
+          pip install build
+          python -m build
```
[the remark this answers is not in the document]
isimo00: The build ones? I would say so https://github.com/cylc/release-actions/blob/49a65ae17eae27e34fa0c6a3e249f7471d70a6d3/build-python-package/action.yml#L28

isimo00:
@kinow Went in the direction you mentioned in the last message and it all seems to work well. Local installation works fine, and github actions hasn't complained either. Besides removing `setup.py` and moving `bin/autosubmit` to `autosubmit/scripts/autosubmit` I've changed its name to `bin_autosubmit.py`, because pytest complained about the module and a test file having the same name (autosubmit.py). This have also come with some modifications in other files that referenced `bin/autosubmit`

kinow (commented):
> I've changed its name to bin_autosubmit.py, because pytest complained about the module and a test file having the same name (autosubmit.py).

That doesn't look very good. I will have a look next week, but I guess we could update the code to just use `import` instead of the sorcery it's doing right now to load that file in pytest (I think it was done that way exactly because the code used to be outside the `autosubmit` package).

Thanks Irene!

test/unit/test_version.py
```diff
@@ -6,12 +6,12 @@
...
 def test_autosubmit_version():
-    bin_path = Path(__file__, '../../../bin/autosubmit').resolve()
+    bin_path = Path(__file__, '../../../autosubmit/scripts/autosubmit.py').resolve()
     exit_code, out = subprocess.getstatusoutput(' '.join([sys.executable, str(bin_path), '-v']))
     assert exit_code == 0
     assert out.strip().endswith(Autosubmit.autosubmit_version)
 
 def test_autosubmit_version_broken():
-    bin_path = Path(__file__, '../../../bin/autosubmit').resolve()
+    bin_path = Path(__file__, '../../../autosubmit/scripts/autosubmit.py').resolve()
```
kinow, line 15: Sorry, @isimo00 , I meant to simplify on this part too. This works for me, and I think it might be simpler (I hope).

```diff
diff --git a/autosubmit/scripts/autosubmit.py b/autosubmit/scripts/autosubmit.py
index 4db180fc..1de02602 100755
--- a/autosubmit/scripts/autosubmit.py
+++ b/autosubmit/scripts/autosubmit.py
@@ -20,20 +20,14 @@
 """Script for handling experiment monitoring"""
 import argparse
 import sys
-from typing import Optional
 from contextlib import suppress
-
-# noinspection PyUnresolvedReferences
-from log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
-import sys
-import os
-
-# Add the directory containing the `autosubmit` module to sys.path
-sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../../')))
+from typing import Optional
 
 from autosubmit import delete_lock_file, exit_from_error  # noqa: E402
 from autosubmit.autosubmit import Autosubmit  # noqa: E402
 from autosubmitconfigparser.config.configcommon import AutosubmitConfig  # noqa: E402
+# noinspection PyUnresolvedReferences
+from log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
 
 
 # noinspection PyProtectedMember
diff --git a/test/unit/test_version.py b/test/unit/test_version.py
index e29bf69e..654b8284 100644
--- a/test/unit/test_version.py
+++ b/test/unit/test_version.py
@@ -6,12 +6,11 @@ from autosubmit.autosubmit import Autosubmit
 
 
 def test_autosubmit_version():
-    bin_path = Path(__file__, '../../../autosubmit/scripts/autosubmit.py').resolve()
-    exit_code, out = subprocess.getstatusoutput(' '.join([sys.executable, str(bin_path), '-v']))
+    exit_code, out = subprocess.getstatusoutput('autosubmit -v')
     assert exit_code == 0
     assert out.strip().endswith(Autosubmit.autosubmit_version)
 
+
 def test_autosubmit_version_broken():
-    bin_path = Path(__file__, '../../../autosubmit/scripts/autosubmit.py').resolve()
-    exit_code, _ = subprocess.getstatusoutput(' '.join([sys.executable, str(bin_path), '-abcdefg']))
+    exit_code, _ = subprocess.getstatusoutput('autosubmit -abcdefg')
     assert exit_code == 1
```

autosubmit/scripts/autosubmit.py
```diff
@@ -19,16 +19,18 @@
...
-sys.path.append(os.path.normpath(os.path.join(script_dir, os.pardir)))
+
 # noinspection PyUnresolvedReferences
 from log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
+import sys
+import os
+
+# Add the directory containing the `autosubmit` module to sys.path
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../../')))
+
```
kinow, line 33: See comment in the test file; I think this can be removed... Manipulating the PYTHONPATH/sys.path is normally fine, but you can have unexpected results depending on what you are doing and your libraries installed... so always safer to avoid it if we can...

kinow (changes requested):

.github/workflows/ci.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e .[all]
+          pip install -e .[all]
```
kinow, line 58: Unnecessary space at end of text (which is auto-deleted in Sublime/VSCode/Pycharm unless the user configures it to disable that, and then it appears in next PRs)

autosubmit/scripts/autosubmit.py
```diff
@@ -19,16 +19,15 @@
...
-
 from contextlib import suppress
-script_dir = os.path.abspath(os.path.dirname(sys.argv[0]))
-sys.path.append(script_dir)
-sys.path.append(os.path.normpath(os.path.join(script_dir, os.pardir)))
+
 # noinspection PyUnresolvedReferences
 from log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
+import sys
+import os
```
kinow, line 29: This import can be removed :grimacing: But not required for this PR, we can fix it later.

docs/source/installation/index.rst
```diff
@@ -25,7 +25,8 @@ To install autosubmit, execute the following:
 Or download, unpack and:
 ::
 
-    python3 setup.py install
+    python3 -m build
```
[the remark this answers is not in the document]
kinow: Did you confirm if this is really needed? It is working on my environment without the `build` module, and the installation docs of that module specify you need to `pip install` it first. So we'd have to ask users to install it too. I think setuptools is now supporting it, so it may not be needed?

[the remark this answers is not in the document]
kinow: Ah, we probably should remove it, @isimo00 .

> A simple, correct Python packaging [build frontend](https://packaging.python.org/en/latest/glossary/#term-Build-Frontend).

https://packaging.python.org/en/latest/glossary/#term-Build-Frontend

>Examples of build frontends are [pip](https://packaging.python.org/en/latest/key_projects/#pip) and [build](https://packaging.python.org/en/latest/key_projects/#build).

Since we are already using `pip`, I think we don't need to ask users to install it (our landing pages already includes a `pip install` statement).

PR #2081, opened 2025-01-27
Iss#2073

VindeeR:
This issue have part of its development in this project and the other part in autosubmit-config-parser

This pull request is to treat the HPCARCH that was not being copied to a new project when when "-y expid_project" was used


kinow (commented):
@VindeeR it looks like you have some commits that are not part of your change, and there are git conflicts too. We can chat tomorrow in person to see what to do about that, in case you need some help with git (it can be complicated to resolve some conflicts, rebase, etc.). Or, if you are at the BSC, it might be easier to sit down with someone else that's also over there (in case you see Luiggi/Irene/others over there).

PR #2092, opened 2025-01-31
Iss#1469 [QoL] - Adjusting logs flagged by IDE to be grammatically correct

VindeeR:
Some of the logs or messages to the user had a couple of spelling mistakes to be fixed

### DIFF
--- a/docs/source/userguide/manage/index.rst
+++ b/docs/source/userguide/manage/index.rst
@@ -113,7 +113,7 @@ Options:
     options:
       -h, --help            show this help message and exit
       -nclean, --noclean    Avoid Cleaning of experiment folder
-      -uc, --uncompressed   Untar an uncompressed tar
+      -uc, --uncompressed   Extract files of the tar file without gzip compression
       -v, --update_version  Update experiment version
       --rocrate             Unarchive an RO-Crate file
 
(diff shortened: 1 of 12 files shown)


docs/source/userguide/manage/index.rst
```diff
@@ -113,7 +113,7 @@ Options:
     options:
       -h, --help            show this help message and exit
       -nclean, --noclean    Avoid Cleaning of experiment folder
-      -uc, --uncompressed   Untar an uncompressed tar
+      -uc, --uncompressed   Extract files of the tar file without gzip compression
```
kinow, line 116: This was the only one the I read and was a bit confused. The other version sounds easier for me, and I immediately understand it's extracting a `.tar` file. This new one is more verbose/clear grammatically, I think, but takes me a bit longer to understand what it's doing. But it might be just me.

kinow (approved):
Looks good to me, can you have a look too, @dbeltrankyl? To see if everything makes sense (many are in code comments, so those are not really important).

dbeltrankyl:
> Looks good to me, can you have a look too, @dbeltrankyl? To see if everything makes sense (many are in code comments, so those are not really important).

Good changes!

docs/source/userguide/manage/index.rst
dbeltrankyl: Honestly, I find both options fine, do you feel it will confuse the final users by being that verbose? @kinow

Issue #2073, opened 2025-01-22
Autosubmit expid -y $expid from 3.15.X is not working


dbeltrankyl:
Related to https://earth.bsc.es/gitlab/es/requests/-/issues/2731

Hello @kinow , @isimo00

There is an issue related with the ruamel update that doesn't allow to upgrade a 3.15.X experiment to 4.1.X.

Whenever you do an `autosubmit expid -y $expid` and that experiment is an 3.15, Autosubmit converts all placeholders and INI files to AS 4 format. However it crashes in something related with ruamel YAML. 

```bash
 autosubmit4 expid -H cirrus -d "Testing AS4" -y a00d
Autosubmit is running with 4.1.11
The new experiment "a00j" has been registered.
Generating folder structure...
Experiment folder: /home/admin/autosubmit/a00j
Generating config files...
Backup stored at /home/admin/autosubmit/a00j/conf/platforms_a00j.conf_AS_v3_backup
[WARNING] Error converting platforms_a00j.conf to yml: 
"dump()" has been removed, use

  yaml = YAML(typ='unsafe', pure=True)
  yaml.dump(...)

instead of file "/usr/local/lib/python3.9/site-packages/autosubmitconfigparser/config/configcommon.py", line 2481

        yaml.dump(final_dict, yaml_file, Dumper=yaml.RoundTripDumper)
```

I though I update all related code but seems that part of the files weren't in my search scope

dbeltrankyl:
Actually, I think this is a good first issue,  so maybe @VindeeR can do it.

How to reproduce the bug: 

go to bschubs:

```bash
module load autosubmit/3.15... 
autosubmit expid -dm -H "local" -d "test" # annotate the expid
module load autosubmit/4.1...
autosubmit expid -dm -H "local" -d "test" -y $expid_of_the_previous_expid
```

Possible fix? 

We updated ruamel.yaml lib and this is a side-effect of it.

So I believe that the fix is just to update the ruamel.yaml code of that function, you can revise this code ( https://github.com/search?q=repo%3ABSC-ES%2Fautosubmit-config-parser%20%20%20%20%20%20%20%20%20yaml.dump(final_dict%2C%20yaml_file%2C%20Dumper%3Dyaml.RoundTripDumper)&type=code) for reference

PR #2098, opened 2025-02-04
Iss#2085 Validating Path generation

VindeeR:
It was updated how to create a path, passing a value Path as the variable.
It was created a unit test for the function copy_as_config validating if the copy and creation of the files were done properly

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1166,20 +1166,21 @@ class Autosubmit:
         return error_message
 
     @staticmethod
-    def copy_as_config(exp_id,copy_id):
-        for conf_file in os.listdir(os.path.join(BasicConfig.LOCAL_ROOT_DIR, copy_id,"conf")):
+    def copy_as_config(exp_id, copy_id):
+        for conf_file in os.listdir(os.path.join(BasicConfig.LOCAL_ROOT_DIR, copy_id, "conf")):
             # Copy only relevant files
             if conf_file.endswith((".conf", ".yml", ".yaml")):
                 shutil.copy(os.path.join(BasicConfig.LOCAL_ROOT_DIR, copy_id, "conf", conf_file),
-                            os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id, "conf", conf_file.replace(copy_id,exp_id)))
+                            os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id, "conf",
+                                         conf_file.replace(copy_id, exp_id)))
             # if ends with .conf convert it to AS4 yaml file
             if conf_file.endswith(".conf"):
                 try:
-                    AutosubmitConfig.ini_to_yaml(os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id,"conf"),
-                                                 os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id,"conf",
-                                                              conf_file.replace(copy_id,exp_id)))
+                    AutosubmitConfig.ini_to_yaml(BasicConfig.LOCAL_ROOT_DIR / Path(exp_id + "/conf"),
+                                                 BasicConfig.LOCAL_ROOT_DIR / Path(exp_id + "/conf/" +
+                                                                                   conf_file.replace(copy_id, exp_id)))
                 except Exception as e:
-                    Log.warning(f"Error converting {conf_file.replace(copy_id,exp_id)} to yml: {str(e)}")
+                    Log.warning(f"Error converting {conf_file.replace(copy_id, exp_id)} to yml: {str(e)}")
 
     @staticmethod
     def generate_as_config(
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -36,7 +36,6 @@ from collections import defaultdict
 from configparser import ConfigParser
 from contextlib import suppress
 from importlib.metadata import version
-# import log.fd_show as fd_show
 from importlib.resources import files as read_files
 from pathlib import Path
 from time import sleep
@@ -60,28 +59,28 @@ from autosubmit.helpers.processes import process_id
 from autosubmit.helpers.utils import check_jobs_file_exists, get_rc_path
 from autosubmit.helpers.utils import strtobool
 from log.log import Log, AutosubmitError, AutosubmitCritical
-from .database.db_common import create_db
-from .database.db_common import delete_experiment, get_experiment_descrip
-from .database.db_common import get_autosubmit_version, check_experiment_exists
-from .database.db_structure import get_structure
-from .experiment.experiment_common import copy_experiment
-from .experiment.experiment_common import new_experiment
-from .git.autosubmit_git import AutosubmitGit
-from .history.experiment_history import ExperimentHistory
-from .history.experiment_status import ExperimentStatus
-from .job.job_common import Status
-from .job.job_grouping import JobGrouping
-from .job.job_list import JobList
-from .job.job_list_persistence import JobListPersistenceDb
-from .job.job_list_persistence import JobListPersistencePkl
-from .job.job_package_persistence import JobPackagePersistence
-from .job.job_packager import JobPackager
-from .job.job_utils import SubJob, SubJobManager
-from .migrate.migrate import Migrate
-from .notifications.mail_notifier import MailNotifier
-from .notifications.notifier import Notifier
-from .platforms.paramiko_submitter import ParamikoSubmitter
-from .platforms.platform import Platform
+from autosubmit.database.db_common import create_db
+from autosubmit.database.db_common import delete_experiment, get_experiment_descrip
+from autosubmit.database.db_common import get_autosubmit_version, check_experiment_exists
+from autosubmit.database.db_structure import get_structure
+from autosubmit.experiment.experiment_common import copy_experiment
+from autosubmit.experiment.experiment_common import new_experiment
+from autosubmit.git.autosubmit_git import AutosubmitGit
+from autosubmit.history.experiment_history import ExperimentHistory
+from autosubmit.history.experiment_status import ExperimentStatus
+from autosubmit.job.job_common import Status
+from autosubmit.job.job_grouping import JobGrouping
+from autosubmit.job.job_list import JobList
+from autosubmit.job.job_list_persistence import JobListPersistenceDb
+from autosubmit.job.job_list_persistence import JobListPersistencePkl
+from autosubmit.job.job_package_persistence import JobPackagePersistence
+from autosubmit.job.job_packager import JobPackager
+from autosubmit.job.job_utils import SubJob, SubJobManager
+from autosubmit.migrate.migrate import Migrate
+from autosubmit.notifications.mail_notifier import MailNotifier
+from autosubmit.notifications.notifier import Notifier
+from autosubmit.platforms.paramiko_submitter import ParamikoSubmitter
+from autosubmit.platforms.platform import Platform
 
 dialog = None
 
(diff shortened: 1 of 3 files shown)


test/unit/test_autosubmit.py
```diff
@@ -0,0 +1,41 @@
+import os
+from textwrap import dedent
+from unittest.mock import patch
+from pathlib import Path
+
+from autosubmitconfigparser.config.basicconfig import BasicConfig
+
+from autosubmit.autosubmit import Autosubmit
+
+def test_copy_as_config(tmp_path):
```
kinow, line 10: You probably want to use one of our fixtures that configure Autosubmit for you... Look at `conftest.py`.

VindeeR:
@kinow @LuiggiTenorioK and I were having a look and it's possible that the test is failing due to the code in config-parser in the function ini_to_yaml not closing the file after inserting the information on the new file.
It could be happening on gh-action and not in local because of the difference in memory allocated but we are not 100% sure

LuiggiTenorioK:
I was looking at this with @VindeeR and found issues in the config parser that might be affecting the tests in this PR.

In the tests, it is asserting if the yml file exists after `AutosubmitConfig.ini_to_yaml` is executed.
 
But then we have this on the parser: 

```python
        yaml_file = open(input_file, 'w', encoding=locale.getlocale()[1])
        yaml = YAML()
        yaml.dump(final_dict, yaml_file)
        ini_file.rename(Path(root_dir, ini_file.stem + ".yml"))
        # Method ends here
```

https://github.com/BSC-ES/autosubmit-config-parser/blob/9bf8f46de3395e7063f2096e34fdc7bd96bb3fcc/autosubmitconfigparser/config/configcommon.py#L2557-L2560

Is it possible that the tests are failing because the config parser doesn't complete writing the files since the file is never closed, making the assertion to fail. 

@kinow this might need to be reviewed.

kinow (commented):
Hi! Sorry for not being able to try the code today at the BSC, @VindeeR . Had multiple issues to review/update, and then had to rush to stop by supermarket on my way home (it's my turn to cook :man_cook: ).

I logged in again now to have a quick look at this, and checked out the branch, and activated my micromamba env, then ran the test with the IDE and got this:

```
FAILED                      [100%]BasicConfig: /tmp/tmp6kgyoszb
Backup stored at /tmp/tmp6kgyoszb/a001/conf/jobs_a001.conf_AS_v3_backup

test/unit/test_autosubmit.py:9 (test_copy_as_config)
tmp_path = PosixPath('/tmp/pytest-of-kinow/pytest-0/test_copy_as_config0')
autosubmit_config = <function autosubmit_config.<locals>._create_autosubmit_config at 0x7ab5e6b7d310>

    def test_copy_as_config(tmp_path, autosubmit_config):
    
        """Test the creation of files without damaging or causing error in the path
    
        Creates a dummy AS3 INI file, calls ``AutosubmitConfig.copy_as_config``, and
        verifies that the YAML files exists and is not empty, and a backup file was
        created. All without warnings or errors being raised (i.e. they were suppressed).
        """
        autosubmit_config('a000',{})
    
        ini_file = Path(f'{BasicConfig.LOCAL_ROOT_DIR}/a000/conf')
        new_file = Path(f'{BasicConfig.LOCAL_ROOT_DIR}/a001/conf')
        ini_file.mkdir(parents=True, exist_ok=True)
        new_file.mkdir(parents=True, exist_ok=True)
        ini_file = ini_file / 'jobs_a000.conf'
        new_file = new_file / 'jobs_a001.yml'
    
        print(f'BasicConfig: {BasicConfig.LOCAL_ROOT_DIR}')
        with open(ini_file, 'w+') as f:
            f.write(dedent('''\
                    [LOCAL_SETUP]
                    FILE = LOCAL_SETUP.sh
                    PLATFORM = LOCAL
                    '''))
            f.flush()
    
        Autosubmit().copy_as_config('a001','a000')
    
        new_yaml_file = Path(new_file.parent,new_file.stem).with_suffix('.yml')
    
>       assert new_yaml_file.exists()
E       AssertionError: assert False
E        +  where False = <bound method Path.exists of PosixPath('/tmp/tmp6kgyoszb/a001/conf/jobs_a001.yml')>()
E        +    where <bound method Path.exists of PosixPath('/tmp/tmp6kgyoszb/a001/conf/jobs_a001.yml')> = PosixPath('/tmp/tmp6kgyoszb/a001/conf/jobs_a001.yml').exists

unit/test_autosubmit.py:40: AssertionError
```

I think that's the same error from GitHub? I ran it through with the debugger, and nticed that in `copy_as_config`. That's the error that @VindeeR fixed in the config parser, https://github.com/BSC-ES/autosubmit-config-parser/pull/66

But the error is hidden, as it happens in the config parser, it is raised, but Autosubmit catches it and only logs. Pytest doesn't output the logs written. So Erick hadn't a way to know without guessing or knowing what to debug.

So I updated my `main` branch of the Autosubmit Config Parser, and then `pip install -U -e .` there, and ran the test in Autosubmit again. Now it worked.

Erick, you can leave this pull request as-is, mark it as a draft (just click the button to your right here :point_right:), and tell others that this is blocked pending a release of Config Parser. Once that's released, we will update the version in `pyproject.toml` or `setup.py` (depending on https://github.com/BSC-ES/autosubmit/pull/2079), and GitHub Actions should pass then once we trigger that again, and it uses the code you fixed :slightly_smiling_face: 

Sorry again for not having time to take a look at this code earlier today. We are discussing to move Config Parser code back to Autosubmit (https://github.com/BSC-ES/autosubmit/issues/2052), should it happen, issues like this won't happen again, as the whole development environment will exist in a single branch/environment/project.

Cheers :wave:

kinow:
Huh, hold on, let me take another look at this one (marked my previous comment as outdated).

kinow:
Ah, yeah. So as to why it wasn't happening locally, here's what I think happened, @VindeeR , @LuiggiTenorioK :detective: 

* you had an older version of as config parser installed in your venv
* you wrote the test and it either failed and you installed the dev version of as config parser, or then you already had a newer version of as config parser, with your fix
* once the test passed you pushed to github, which has a clean dev env to run the test (ephemeral), which is good

Then, even if you `pip uninstall autosubmitconfigparser`, and then reinstall it, when you rerun the tests, if you use an IDE like PyCharm that tries to help you by default adding source code to the path of the tests, then your pytests would be using the version from autosubmitconfigparser you had opened. In that case, you have to uncheck the options to use code from content root and from project in PyCharm. VS Code/etc. should have a similar option.

image

If downgrade autosubmitconfigparser and you run the test locally but from the command line then you get the same error as in GitHub.

image

But not from the IDE :grimacing: 

This is a case where the tool is trying to be helpful, but it ends up getting in the way (although once you know this behaviour you just need to remember to watch for that).

So again, I think once we release config parser your test should just pass here and then be ready for review & merge :muscle: :crossed_fingers: 

And here's the image of how I found it in the debugger (it's always good to try to run the code through a debugger first whenever you have issues like this):

image

Ok, time to cook now :wave:

dbeltrankyl:
I see this is a draft from 2 months ago. Any update here?

kinow (changes requested):
GH ACtions broken :grimacing:

kinow (commented):
Check if it was merged, if so, we can close it! :tada:

VindeeR:
@kinow this was an issue related to the [issue 2073](https://github.com/BSC-ES/autosubmit/issues/2073) where if a project had itself converted from v3.x.x to the v4.x.x and had a .conf as its configuration files it would pass a string instead of a path as required, so I went back to this issue today and finished the fix with a test.

I took the opportunity to fix the import order, done by pycharm, in the autosubmit.py file too

autosubmit/autosubmit.py
```diff
@@ -39,6 +67,8 @@
 from .experiment.experiment_common import copy_experiment
 from .experiment.experiment_common import new_experiment
 from .git.autosubmit_git import AutosubmitGit
+from .history.experiment_history import ExperimentHistory
```
kinow, line 70: I think if we will update/fix the imports here, we might as well make them uniform. At the moment we have relative and absolute imports. Let's make everything absolute (there are some corner cases where it may fail to start a debugger or run as a module in IDE/cmdline when you have these imports that start with a dot).

kinow (changes requested):

kinow:
The IDE configuration is working, and after the few past PRs your code already improved a lot, @VindeeR !!! Left some comments mainly about a bug I think I introduced in conftest, and there is just one minor IDE warning in the test and missing license header :grimacing: Thanks!

autosubmit/autosubmit.py
```diff
@@ -1168,20 +1166,21 @@ def _perform_deletion(experiment_path: Path, structure_db_path: Path, job_data_d
...
-                            os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id, "conf", conf_file.replace(copy_id,exp_id)))
+                            os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id, "conf",
+                                         conf_file.replace(copy_id, exp_id)))
             # if ends with .conf convert it to AS4 yaml file
             if conf_file.endswith(".conf"):
                 try:
-                    AutosubmitConfig.ini_to_yaml(os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id,"conf"),
-                                                 os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id,"conf",
-                                                              conf_file.replace(copy_id,exp_id)))
+                    AutosubmitConfig.ini_to_yaml(BasicConfig.LOCAL_ROOT_DIR / Path(exp_id + "/conf"),
```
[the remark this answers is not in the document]
VindeeR: yes it would be a Path, I'll convert it inside my test for now and them we can create a fix for that

autosubmit/autosubmit.py
```diff
@@ -1166,20 +1165,21 @@ def _perform_deletion(experiment_path: Path, structure_db_path: Path, job_data_d
...
-                            os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id, "conf", conf_file.replace(copy_id,exp_id)))
+                            os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id, "conf",
+                                         conf_file.replace(copy_id, exp_id)))
             # if ends with .conf convert it to AS4 yaml file
             if conf_file.endswith(".conf"):
                 try:
-                    AutosubmitConfig.ini_to_yaml(os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id,"conf"),
-                                                 os.path.join(BasicConfig.LOCAL_ROOT_DIR, exp_id,"conf",
-                                                              conf_file.replace(copy_id,exp_id)))
+                    AutosubmitConfig.ini_to_yaml(Path(BasicConfig.LOCAL_ROOT_DIR + "/" + exp_id + "/conf"),
```
kinow, line 1178: `Path(BasicConfig.LOCAL_ROOT_DIR, exp_id, 'conf')` the constructor of `Path` should act like `os.path.join`, so we don't need to worry if concatenating strings will produce a valid path, and the constructor of `Path` will do that for us :grimacing:

autosubmit/autosubmit.py
```diff
@@ -60,28 +59,28 @@
...
+from autosubmit.job.job_list_persistence import JobListPersistenceDb
+from autosubmit.job.job_list_persistence import JobListPersistencePkl
+from autosubmit.job.job_package_persistence import JobPackagePersistence
+from autosubmit.job.job_packager import JobPackager
+from autosubmit.job.job_utils import SubJob, SubJobManager
+from autosubmit.migrate.migrate import Migrate
+from autosubmit.notifications.mail_notifier import MailNotifier
+from autosubmit.notifications.notifier import Notifier
+from autosubmit.platforms.paramiko_submitter import ParamikoSubmitter
+from autosubmit.platforms.platform import Platform
```
kinow, line 83: :heart: thank you!!!

kinow (approved):
:clap: :clap: :clap: we are always fixing something, but it's really good when we can also improve little by little :slightly_smiling_face: Thank you, Erick!

PR #1914, opened 2024-12-12
Postgres Support

kinow:
In GitLab by @LuiggiTenorioK on May 6, 2024, 15:59

Merge request to support Postgres #1285 

- [x] update db_common.py
- [x] update db_manager.py
- [x] update db_structure.py
- [x] Add pytest tests
- [x] update database_manager.py
- [x] update experiment_status_db_manager.py
- [ ] update experiment_history_db_manager.py
- [ ] check nullable values that are not nullable (sqlite does not enforce & AS code is writing nulls!)
- [ ] perform tests listed in the issue #1285

Issue #1285, opened 2024-04-30
Postgres layer option


kinow:
In GitLab by @LuiggiTenorioK on Apr 30, 2024, 11:07

To deploy Autosubmit in EDITO, we will need to have the option to save to a different DBMS like Postgres.

- [x] This option could be available inside the `.autosubmitrc` file and it will persist the DDBB data inside the Postgres instance we will have in EDITO.
- [x] As a scope, we aim to implement a middle layer to decide which engine to use (SQLite/Postgres) keeping the same table schemas for now.
- [ ] Update the AS code
  - [x] `autosubmit.database` package, already done by @LuiggiTenorioK !)
  - [ ] `autosubmit.history` :scream:, @kinow will take a look if we can remove things first
  - [ ] Tests!
    - [x] @LuiggiTenorioK will check out pytest + unittest (use pytest to run existing tests) and add a `conftest.py`
    - [x] Create a fixtures to overwrite the BasicConfig and prepare the DDBB for testing
    - [x] Create temporary sqlite for each tests as needed
    - [x] Create a Docker postgres DB for each tests as needed
    - [x] Run sqlite tests on GitLab worker, Docker+pg locally

Tests that need to be performed (@LuiggiTenorioK feel free to add more to this list):

- [x] Install AS from scratch
- [x] Create a workflow
- [x] Run a dummy experiment
- [x] Set status
- [ ] Use wrappers
- [ ] Archive a workflow (what's going to happen here?
- [x] Delete experiments https://earth.bsc.es/gitlab/es/autosubmit/-/issues/1317#note_284520

@mcastril @kinow @dbeltrankyl

kinow:
In GitLab by @LuiggiTenorioK on Apr 30, 2024, 11:07

changed due date to June 30, 2024

kinow:
In GitLab by @LuiggiTenorioK on May 2, 2024, 09:33

mentioned in issue autosubmit_api#74

kinow:
In GitLab by @LuiggiTenorioK on May 8, 2024, 11:21

marked the checklist item **`autosubmit.database` package, 50%+ done by @LuiggiTenorioK !)** as completed

kinow:
In GitLab by @LuiggiTenorioK on May 9, 2024, 09:42

marked the checklist item **Tests! @LuiggiTenorioK will check out pytest + unittest (use pytest to run existing tests) and add a `conftest.py`** as completed

kinow:
[...]

PR #2107, opened 2025-02-05
Add DbManager factory function

LuiggiTenorioK:
Related to #1285 and PR #1914 

The idea of this PR is to include the code organization changes related to #1914. In this case:

* Adding the DbManager factory function
* Update JobListPersistenceDb and JobPackagePersistence `__init__` arguments
* Fix tests to mock the classes correctly changed above

These changes will make it easier to implement and review the changes needed to enable the support to Postgres.

### DIFF
--- a/autosubmit/database/db_manager.py
+++ b/autosubmit/database/db_manager.py
@@ -19,14 +19,14 @@
 
 import sqlite3
 import os
-from typing import Dict, Iterable, List, Protocol, Union, cast
+from typing import Any, Dict, Iterable, List, Protocol, Union, cast
 
 class DbManager(object):
     """
     Class to manage an SQLite database.
     """
 
-    def __init__(self, root_path, db_name, db_version):
+    def __init__(self, root_path: str, db_name: str, db_version: int):
         self.root_path = root_path
         self.db_name = db_name
         self.db_version = db_version
@@ -153,7 +153,7 @@ class DbManager(object):
         if os.path.exists(self._get_db_filepath()):
             os.remove(self._get_db_filepath())
 
-    def _get_db_filepath(self):
+    def _get_db_filepath(self) -> str:
         """
         Returns the path of the .db file
         :return path: int
@@ -173,7 +173,7 @@ class DbManager(object):
         self.insert(options_table_name, columns, ['name', self.db_name])
         self.insert(options_table_name, columns, ['version', self.db_version])
 
-    def _select_with_all_fields(self, table_name, where=[]):
+    def _select_with_all_fields(self, table_name: str, where: List[str] = []) -> sqlite3.Cursor:
         """
         Returns the cursor of the select command with the given parameters
         :param table_name: str
@@ -190,7 +190,7 @@ class DbManager(object):
     """
 
     @staticmethod
-    def generate_create_table_command(table_name, fields):
+    def generate_create_table_command(table_name: str, fields: List[str]) -> str:
         create_command = 'CREATE TABLE IF NOT EXISTS ' + table_name + ' (' + fields.pop(0)
         for field in fields:
             create_command += (', ' + field)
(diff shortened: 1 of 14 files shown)


LuiggiTenorioK:
@kinow FYI as some of this code is yours

autosubmit/database/db_manager.py
```diff
@@ -94,6 +94,17 @@ def insertMany(self, table_name, data):
         cursor.executemany(insert_many_command, data)
         self.connection.commit()
 
+    def delete_where(self, table_name: str, where: list[str]):
```
isimo00, line 97: To be consistent with type hints, should we add the `None` type to the output of the function?

`def delete_where(self, table_name: str, where: list[str]) -> None:`

autosubmit/database/db_manager.py
```diff
@@ -234,3 +245,67 @@ def generate_select_command(table_name, where=[]):
         for condition in where:
             select_command += ' AND ' + condition
         return select_command
+    
+    @staticmethod
+    def generate_delete_command(table_name: str, where: list[str] = []):
```
isimo00, line 250: Type hint return string - see other comment -?

autosubmit/database/db_manager.py
LuiggiTenorioK: I think adding the type when it only returns `None` is not mandatory.

If it may return any other type or `None` (`Union[<OTHER_TYPE>, None]`), it will be useful to have that. So, when someone else calls that function should also handle the `None` case.

autosubmit/database/db_manager.py
LuiggiTenorioK: I'm sure that @kinow didn't put the hints there to avoid making more changes and keep it consistent which is valid when the PR is big.

I'll add the return types for the generate commands now because this PR is not that big.

kinow:
Rebased, git conflicts fixed. Now waiting for GH Actions :+1:

kinow:
Done, and GH Actions passing now, cc @LuiggiTenorioK .

kinow:
Luiggi and I cannot review this one, can you do a final review whenever you have time, please, @dbeltrankyl ? Thanks!

kinow:
Fixing conflicts...

dbeltrankyl:
Looks good! 

Sorry for not looking at it earlier; I didn't notice I was assigned. 

It has now conflicts likely due to the https://github.com/BSC-ES/autosubmit/pull/2238 😓

kinow:
Thanks @dbeltrankyl ! Conflicts fixed, had to adjust some of the tests after the conflicts were resolved as the `JobListPersistence` class will create a sqlite file that will fail unless the `autosubmit_config` fixture is loaded. But I believe the tests will pass. @LuiggiTenorioK can you have a look once GH Actions pass if everything looks in order, please? (After fixing conflicts a couple of times I stop being able to see the wood for the trees :sweat_smile: )

autosubmit/job/job_list_persistence.py
```diff
@@ -140,8 +142,14 @@ class JobListPersistenceDb(JobListPersistence):
         "wrapper_type",
     ]
 
-    def __init__(self, persistence_path, persistence_file):
-        self.db_manager = DbManager(persistence_path, persistence_file, self.VERSION)
+    def __init__(self, expid: str):
+        options = {
+            "root_path": str(Path(BasicConfig.LOCAL_ROOT_DIR, expid, "pkl")),
+            "db_name": f"job_list_{expid}",
+            "db_version": self.VERSION
```
LuiggiTenorioK, line 149:
```suggestion
            "db_version": self.VERSION,
            "schema": expid
```

I think we forgot to add the schema here :point_up: 

About the changes in the tests, they look way simpler than before. Nice change! :raised_hands:

kinow:
Thanks for the final review, @LuiggiTenorioK . I accepted your suggestion. I'm about to log off for today. Feel free to take over and merge if GH Actions pass :+1: Otherwise, I can merge it tomorrow morning. Cheers

LuiggiTenorioK:
Thanks @kinow! Everything seems sorted now. Merging :rocket:

Issue #1403, opened 2024-08-27
Solving lint rule issues


kinow:
In GitLab by @LuiggiTenorioK on Aug 27, 2024, 14:49

The current source code of Autosubmit has many underlying issues that can be easily detected by linters.

Just running Ruff, we have the following rules that are violated:

```bash
$ ruff check --statistics autosubmit
191     F841    [*] unused-variable
 50     F401    [*] unused-import
 39     E721    [ ] type-comparison
 37     E402    [ ] module-import-not-at-top-of-file
 31     E722    [ ] bare-except
 18     F523    [ ] string-dot-format-extra-positional-arguments
 15     F541    [*] f-string-missing-placeholders
 15     F811    [ ] redefined-while-unused
 12     F901    [*] raise-not-implemented
  9     F821    [ ] undefined-name
  2     E712    [*] true-false-comparison
  2     F402    [ ] import-shadowed-by-loop-var
  1     E713    [*] not-in-test
  1     E731    [*] lambda-assignment
```

It will be nice to solve them to prevent future bugs and keep tracking them in future releases as part of our development process.

@kinow @dbeltrankyl @edgano

kinow:
In GitLab by @dbeltrankyl on Aug 28, 2024, 15:15

mentioned in merge request autosubmit4-config-parser!9

kinow:
In GitLab by @LuiggiTenorioK on Aug 28, 2024, 15:41

> Maybe we should start by using that tool in a separate merge request, merge it, and we can include a procedure to run it in each new merge requests as it will only fix the new code added and doesn't complicate the code review

+1

> The IDE (pycharm) can solve a few of these issues with some inbuilt tools, I've used it a couple of times. I usually don't use it in a merge request because it makes modifications to all code and complicates the code review.

Indeed, Ruff also can fix some of them "safely". I prefer to use Ruff since it also includes the code formatter and rules from other linters. Also, it is used in other projects in ES (like the one presented by Bruno and Aina) and works on different IDEs (VSCode and Pycharm as well).

kinow:
In GitLab by @LuiggiTenorioK on Sep 2, 2024, 14:12

mentioned in issue autosubmit-api#87

kinow:
This is something we can enlist Erick to help with. :+1:

dbeltrankyl:
To close and reference in a new issue

PR #2117, opened 2025-02-10
Iss#1403 fix f841 variables

VindeeR:
Related to #1403 
commenting all the variable that were not used and had not being removed from the code

### DIFF
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -1366,9 +1366,8 @@ class Job(object):
         if new_status == Status.COMPLETED:
             Log.debug(
                 "{0} job seems to have completed: checking...".format(self.name))
-            if not self._platform.get_completed_files(self.name, wrapper_failed=self.packed):
-                log_name = os.path.join(
-                    self._tmp_path, self.name + '_COMPLETED')
+            # if not self._platform.get_completed_files(self.name, wrapper_failed=self.packed):
+                # log_name = os.path.join(self._tmp_path, self.name + '_COMPLETED')
 
             self.check_completion()
         else:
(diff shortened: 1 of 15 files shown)


autosubmit/job/job.py
```diff
@@ -2479,7 +2478,7 @@ def write_total_stat_by_retries(self, total_stats, first_retrial = False):
         else:
             f.write('\n' + date2str(datetime.datetime.fromtimestamp(total_stats[0]), 'S') + ' ' + date2str(datetime.datetime.fromtimestamp(total_stats[0]), 'S') + ' ' + date2str(datetime.datetime.fromtimestamp(total_stats[1]), 'S') + ' ' + total_stats[2])
         out, err = self.local_logs
-        path_out = os.path.join(self._tmp_path, 'LOG_' + str(self.expid), out)
+        # path_out = os.path.join(self._tmp_path, 'LOG_' + str(self.expid), out)
```
kinow, line 2481: If a variable is not used, I think it's better to remove it, instead of leaving a comment (as we can always use git to look at the previous versions of the file where the variable was used).

test/unit/test_run_command_intregation.py
```diff
@@ -389,7 +389,7 @@ def init_run(run_tmpdir, jobs_data):
 def test_run_uninterrupted(run_tmpdir, prepare_run, jobs_data, expected_db_entries, final_status):
     log_dir = init_run(run_tmpdir, jobs_data)
     # Run the experiment
-    exit_code = Autosubmit.run_experiment(expid='t000')
+    Autosubmit.run_experiment(expid='t000')
```
kinow, line 392: Or maybe the test here could `assert exit_code == something_something`?

PR #2120, opened 2025-02-10
Iss#2060 adjust path generation to make use of new standard of our code

### LINKED ISSUE #2060: Update code in _get_from_total_stats to use pathlib, close the file after reading it
See code in https://github.com/BSC-ES/autosubmit/blob/120c335394fa087b50be69ac9a999e439bb60bed/autosubmit/job/job.py#L1030-L1048

While reviewing the code for @lauherrero , I noticed that the code is still using `os.path` where it could be simplified by `pathlib`, and also that it looks like we are not closing the file descriptor after opening it, which may create some memory leaks or unnecessary resources being left open in the system.

It should be easy to update the code to use a context manager, and then write a short test (either with pytest, or a doctest?).

VindeeR:
closes #2060 
Changing the way the path is detected, closing the file after being used, and test to check if file is detect correctly with new method of getting path


autosubmit/job/job.py
```diff
@@ -1059,15 +1059,16 @@ def _get_from_total_stats(self, index):
...
         lst = []
-        if os.path.exists(log_name):
+        if log_name.exists() and log_name.stat().st_size > 0:
             f = open(log_name)
             lines = f.readlines()
             for line in lines:
                 fields = line.split()
                 if len(fields) >= index + 1:
                     lst.append(parse_date(fields[index]))
+            f.close()
```
kinow, line 1071: Probably simpler/better to use a context manager, https://docs.python.org/3/reference/compound_stmts.html#with, like `with open ... as f:`

This way we don't have to worry about closing it... as there could be an error before this new `f.close()` is called too…

test/unit/test_job.py
```diff
@@ -6,6 +6,7 @@
 from pathlib import Path
 
 import pytest
+from typing import Callable
```
kinow, line 9: Not used? Or maybe you were going to add it to a function, but realized this was an older test?

test/unit/test_job.py
```diff
@@ -1085,12 +1085,36 @@ def test_job_script_checking_contains_the_right_variables(self):
...
+                    DEFAULT:
+                        DATE: 1998
+                        EXPID: 199803
+                        HPCARCH: 19980324
+                    '''))
+                f.flush()
+
+            lst = self.job._get_from_total_stats(1)
+
+        assert len(lst) > 0
```
kinow, line 1117: Maybe you can write a test for the case when the `_TOTAL_STATS` file is not found? That'd bring the patch coverage here to 100%.

Also, the test can be more specific, to test that `lst` is not only empty, but contains what's expected.

```diff
diff --git a/test/unit/test_job.py b/test/unit/test_job.py
index 117de8d8..77b837bf 100644
--- a/test/unit/test_job.py
+++ b/test/unit/test_job.py
@@ -20,7 +20,6 @@ from autosubmitconfigparser.config.configcommon import BasicConfig, YAMLParserFa
 from mock import Mock, MagicMock
 from mock import patch
 
-import log.log
 from autosubmit.autosubmit import Autosubmit
 from autosubmit.job.job import Job
 from autosubmit.job.job_common import Status
@@ -1090,7 +1089,6 @@ CONFIG:
         self.assertEqual("FALSE", parameters['CHUNK_FIRST'])
         self.assertEqual("FALSE", parameters['CHUNK_LAST'])
 
-
     def test_get_from_total_stats(self):
         # arrange
         with tempfile.TemporaryDirectory() as temp_dir:
@@ -1114,7 +1112,16 @@ CONFIG:
 
             lst = self.job._get_from_total_stats(1)
 
-        assert len(lst) > 0
+        assert len(lst) == 3
+        # down to the minute
+        fmt = '%Y-%m-%d %H:%M'
+        expected = [
+            datetime.datetime(1998, 1, 1, 0, 0),
+            datetime.datetime(1998, 3, 1, 0, 0),
+            datetime.datetime(1998, 3, 24, 0, 0)
+        ]
+        for left, right in zip(lst, expected):
+            assert left.strftime(fmt) == right.strftime(fmt)
 
     def test_sdate(self):
         """Test that the property getter for ``sdate`` works as expected."""

```

kinow (approved):
:clap: Excellent. Just remember to squash the commits :+1: Thanks!

PR #2125, opened 2025-02-11
Fix exit code in autosubmit commands that return bool

### LINKED ISSUE #2124: autosubmit create returning 1 even though it succeeds
Found while testing #2123 for @orioltinto. I had an `autosubmit create a000 && autosubmit inspect a000`, and the RHS part was not being evaluated.

I checked and `autosubmit create` was not logging any errors but it was returning `1`. Looking at the code, it's because `def create(...)` returns a bool instead of an int. Enabling mypy + type hints might be the cheapest way to find issues like this, or we review everything or write tests for the missing parts.

I have the fix done locally, but will prepare a small test later and submit a pull request (will focus on #2104 as that has higher priority).

kinow:
Closes #2124

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -158,7 +158,7 @@ class Autosubmit:
         script_dir = os.path.join(script_dir, os.path.pardir)
 
     version_path = os.path.join(script_dir, 'VERSION')
-    readme_path = os.path.join(script_dir, 'README')
+    readme_path = os.path.join(script_dir, 'README.md')
     changes_path = os.path.join(script_dir, 'CHANGELOG')
     if os.path.isfile(version_path):
         with open(version_path) as f:
--- a/autosubmit/scripts/autosubmit.py
+++ b/autosubmit/scripts/autosubmit.py
@@ -1,41 +1,35 @@
 #!/usr/bin/env python
 
 # Copyright 2015 Earth Sciences Department, BSC-CNS
-
+#
 # This file is part of Autosubmit.
-
+#
 # Autosubmit is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
-
+#
 # Autosubmit is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
-
+#
 # You should have received a copy of the GNU General Public License
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
 """Script for handling experiment monitoring"""
 import argparse
-import sys
-from typing import Optional, Union
-from pathlib import Path
-from contextlib import suppress
 import traceback
 from contextlib import suppress
 from os import _exit  # type: ignore
-
-# noinspection PyUnresolvedReferences
-from log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
-
-from autosubmit.autosubmit import Autosubmit  # noqa: E402
-from autosubmitconfigparser.config.configcommon import AutosubmitConfig  # noqa: E402
+from pathlib import Path
+from typing import Optional, Union
 
 from portalocker.exceptions import BaseLockException
 
-from log.log import Log, AutosubmitCritical, AutosubmitError
+from autosubmit.autosubmit import Autosubmit  # noqa: E402
+from autosubmitconfigparser.config.configcommon import AutosubmitConfig  # noqa: E402
+from log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
 
 
 def delete_lock_file(base_path: str = Log.file_path, lock_file: str = 'autosubmit.lock') -> None:
@@ -101,9 +95,10 @@ def exit_from_error(e: BaseException) -> int:
     Log.info("More info at https://autosubmit.readthedocs.io/en/master/troubleshooting/error-codes.html")
     return err_code
 
+
 # noinspection PyProtectedMember
 def main():
-    args = Optional[argparse.Namespace]
+    args: Optional[argparse.Namespace] = None
     try:
         return_value, args = Autosubmit.parse_args()
         if args:
@@ -126,9 +121,7 @@ def main():
         Log.error(f"Arguments provided: {str(args)}")
         Log.error(f"This is the experiment: {expid} which had an issue with the command: {command} and it is currently using the Autosubmit Version: {version}.")
         return_value = exit_from_error(e)
+    # TODO: we need to define whether the function called here will return an int or bool
+    if type(return_value) is bool:
+        return_value = 0 if return_value else 1
     return return_value
-
-
-if __name__ == "__main__":
-    exit_code = main()
-    sys.exit(exit_code)  # Sys.exit ensures a proper cleanup of the program, while os._exit() does not.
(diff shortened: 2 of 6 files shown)


autosubmit/autosubmit.py
```diff
@@ -155,7 +155,7 @@ def experiment_data(self):
         script_dir = os.path.join(script_dir, os.path.pardir)
 
     version_path = os.path.join(script_dir, 'VERSION')
-    readme_path = os.path.join(script_dir, 'README')
+    readme_path = os.path.join(script_dir, 'README.md')
```
kinow, line 158: Picked the `readme` command to test, and it was broken :disappointed_relieved: . This fixes it.

autosubmit/scripts/autosubmit.py
```diff
@@ -101,9 +95,10 @@ def exit_from_error(e: BaseException) -> int:
     Log.info("More info at https://autosubmit.readthedocs.io/en/master/troubleshooting/error-codes.html")
     return err_code
 
+
 # noinspection PyProtectedMember
 def main():
-    args = Optional[argparse.Namespace]
+    args: Optional[argparse.Namespace] = None
```
kinow, line 101: This was wrong, sorry. Now declared as `None`.

autosubmit/scripts/autosubmit.py
```diff
@@ -126,9 +121,7 @@ def main():
...
         return_value = exit_from_error(e)
+    # TODO: we need to define whether the function called here will return an int or bool
+    if type(return_value) is bool:
+        return_value = 0 if return_value else 1
     return return_value
-
-
-if __name__ == "__main__":
-    exit_code = main()
-    sys.exit(exit_code)  # Sys.exit ensures a proper cleanup of the program, while os._exit() does not.
```
kinow, line 134: This is never called. If you `pip install -e .`, then do a `which autosubmit, and open that file. You will see that Python automatically creates the entry point script, which does pretty much the same above, but done automatically by Python (which is why you cannot execute the file directly, as it's an entry point installed/managed by setuptools, not a simple Python script).

log/log.py
```diff
@@ -243,7 +242,7 @@ def set_file(file_path, type='out', level="WARNING"):
                     Log.log.addHandler(status_file_handler)
                 os.chmod(file_path, 509)
             except Exception: # retry again
-                pass
+                sleep(timeout * retries)
```
kinow, line 245: Writing the test for `autosubmit readme` it worked locally, and then failed on GitHub Actions.

That's because I had an `.autosubmitrc` in my local computer, and on GitHub we didn't have one. I used an existing fixture that sets the `CONFIG_FILE_FOUND` but I changed that to `True` (it was `False`, but was not used).

Still, the test started taking a long time. When I debugged it, I noticed this function that was creating the log files. However, the logic was 1) sleep, 2) then do the work, 3) if it fails continue, and then back to 1. This change here does 1) try to do the work, 2) if it fails sleep, then go back to 1, which looks harmless and is probably faster for users and for our tests?

test/unit/conftest.py
```diff
@@ -1,14 +1,14 @@
...
 from dataclasses import dataclass
 from pathlib import Path
-from ruamel.yaml import YAML
 from shutil import rmtree
 from tempfile import TemporaryDirectory
 from typing import Any, Dict, Callable, List, Protocol, Optional
-import os
+
+import pytest
+from ruamel.yaml import YAML
```
kinow, line 11: Optimized imports in PyCharm

test/unit/conftest.py
```diff
@@ -121,6 +121,7 @@ def _create_as_conf(autosubmit_exp: AutosubmitExperiment, yaml_files: List[Path]
 
     return _create_as_conf
 
+
```
kinow, line 124: Fixed a couple of code style warnings for lint/pep/etc :point_up:

test/unit/conftest.py
```diff
@@ -200,5 +203,6 @@ def prepare_basic_config(tmpdir):
     BasicConfig.MAIL_FROM = ""
     BasicConfig.ALLOWED_HOSTS = ""
     BasicConfig.DENIED_HOSTS = ""
-    BasicConfig.CONFIG_FILE_FOUND = False
+    BasicConfig.CONFIG_FILE_FOUND = True
```
kinow, line 206: For now setting this as `True`, but I think in the future we might need a fixture that creates the config and also prepares the DB and `.autosubmitrc` file.

kinow:
Thought it'd be easier to fix this one. Ended up finding a few loose ends in our code. Added tests, left comments explaining the changes I did. Assigned @dbeltrankyl as I touched some parts of the code that could be a bit dangerous :+1: Now ready for review.

image

kinow:
Rebased.

dbeltrankyl:
I haven't revised it yet. Is it necessary for 4.1.12? @kinow I can do it if so if not, I'll move it to 4.1.13

autosubmit/scripts/autosubmit.py
dbeltrankyl: if __name__ == "__main__":
exit_code = main()
sys.exit(exit_code)  # Sys.exit ensures a proper cleanup of the program, while os._exit() does not.

 I wonder if we have zombies as the sys.exit was avoiding that 🤔

dbeltrankyl: Seems not, atleast with 4.1.12-mem-fixes

Censured output:

```bash
[dbeltran@climatedt-wf ~]$ ps auxww | grep 'Z'      
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
x 1 0.0  0.0      0     0 ?        Z    05:14   0:02 [autosubmit log ] <defunct> # 4.1.12-dev-branch (prob killed)
x 1 0.0  0.0      0     0 ?        Z    05:14   0:01 [autosubmit log ] <defunct># 4.1.12-dev-branch(prob killed)
x 1  0.0  0.0      0     0 ?        Z    05:14   0:00 [autosubmit log ] <defunct># 4.1.12-dev-branch(prob killed)
x 1  0.0  0.0      0     0 ?        Z    05:20   0:01 [autosubmit log ] <defunct># 4.1.12-dev-branch(prob killed)
x 1  0.0  0.0      0     0 ?        Z    05:20   0:01 [autosubmit log ] <defunct># 4.1.12-dev-branch(prob killed)
x   1  0.0  0.0      0     0 ?        Z    05:20   0:00 [autosubmit log ] <defunct># 4.1.12-dev-branch (prob killed)
y 1  0.0  0.0      0     0 ?        Z    13:57   0:00 [id] <defunct># gui
y 1  0.0  0.0      0     0 ?        Z    14:08   0:00 [id] <defunct># gui
y 1 0.0  0.0      0     0 ?        Z    14:14   0:00 [id] <defunct># gui
a   1  0.0  0.0      0     0 ?        Z    Mar06   0:00 [python3] <defunct> # mar06, prev dev
a 1  0.0  0.0      0     0 ?        Z    Mar06   0:00 [python3] <defunct>  # mar06, prev dev
y  1  0.0  0.0      0     0 ?        Z    Mar17   0:00 [id] <defunct> # gui
````

File: test/unit/conftest.py [hunk 8]
```diff
@@ -146,22 +147,24 @@ def autosubmit_config(
     """
 
     original_root_dir = BasicConfig.LOCAL_ROOT_DIR
-    tmp_dir = TemporaryDirectory()
```
dbeltrankyl, line 149: I also removed this in the autosubmit config parser, probably this setup should match again ( that with your changes, it may not )

dbeltrankyl (approved):

dbeltrankyl:
Some comments, but seems good. Thanks!

kinow (commented):
@dbeltrankyl rebased, and replied the comments (did not change any code). Let me know if these still look good and can be merged, or if you prefer to wait, or maybe change it to 4.1.13. Thanks for the review!

kinow:
Merged! Thanks @dbeltrankyl !

Issue #2030, opened 2024-12-13
Migrate Gitlab templates to Github


dbeltrankyl:
Hello @kinow, @isimo00.  cc @LuiggiTenorioK 

We had these templates in gitlab.

https://github.com/BSC-ES/autosubmit/tree/master/.gitlab/issue_templates

We need to update/add them

PR #2127, opened 2025-02-12
Removed GitLab files, and updated some references to GitLab in our do…

kinow:
…cs and code.

Related to #1380

Was about to log off when I saw the .gitlab-ci.yml file in my IDE. Spent some 10 minutes~ searching and replacing some values in our docs/files.

The templates were deleted but we can easily grab the again by browsing the code in our latest tags in GitHub, then adjust the gitlab templates to github (for #2030).


docs/source/conf.py
```diff
@@ -163,10 +163,9 @@
 }
 
 html_context = {
-    'gitlab_url': 'https://earth.bsc.es/gitlab',
```
isimo00, line 166: Why no `github_url`?

LuiggiTenorioK (commented):
- This is minor but we will probably need to change the GitLab icon on the home page.

image

- Also, I saw that there is a GitLab issue referenced in the Installation page:

```diff
-.. warning:: This procedure is still WIP. You can follow the process at `issue #864 <https://earth.bsc.es/gitlab/es/autosubmit/-/issues/886>`_. We strongly recommend using the pip procedure.
+.. warning:: This procedure is still WIP. You can follow the process at `issue #864 <https://github.com/BSC-ES/autosubmit/issues/886>`_. We strongly recommend using the pip procedure.
```

- Then, I found some GitLab URLs in the EasyBuild recipe but not sure if that should be changed.

docs/source/conf.py
kinow: This is the way it's documented in the theme docs: https://github.com/pydata/pydata-sphinx-theme/blob/afc87beed0327d64083676d0fe355cd494a61f4c/docs/user_guide/source-buttons.rst#github

image

Looks like the generated GitHub URLs are correct, e.g. https://autosubmit--2127.org.readthedocs.build/en/2127/ (this PR generated page -- expand Checks below to find the link in the RTD check).

image

image

kinow:
@LuiggiTenorioK & @isimo00 I addressed your feedback (thanks for that!), could you please have another look. I haven't resolved the conversations so that it'd be easier for you both to review it.

docs/source/conf.py
isimo00: Ah, great! Approved then :)

isimo00 (approved):

docs/source/conf.py
kinow: Thanks Irene!

LuiggiTenorioK (approved):
Looks good to me! :100:  Thanks @kinow!

PR #2130, opened 2025-02-14
Gh 2122 mem issues

dbeltrankyl:
Better mem management and fixed an issue with recovery exiting prematurely.

Deleted unused stuff

Also added tests and separated the wait_for_work function into three different functions. 

This doesn't solve the memory leak ( still debugging it in a different branch )

I was also checking the FD.. but it doesn't happen anymore after the queue changes? I'm debugging what could be the cause... because it is weird

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -19,7 +19,7 @@ import locale
 import platform
 # You should have received a copy of the GNU General Public License
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
-import threading
+
 from bscearth.utils.date import date2str
 from configparser import ConfigParser
 from pathlib import Path
@@ -27,7 +27,6 @@ from ruamel.yaml import YAML
 from typing import Dict, Set, Tuple, Union, Any, List, Optional
 
 from autosubmit.database.db_common import update_experiment_descrip_version
-from autosubmit.helpers.parameters import PARAMETERS
 from autosubmit.helpers.utils import strtobool
 from autosubmitconfigparser.config.basicconfig import BasicConfig
 from autosubmitconfigparser.config.configcommon import AutosubmitConfig
@@ -147,7 +146,6 @@ class Autosubmit:
         """Get the current voltage."""
         return self.experiment_data
 
-    sys.setrecursionlimit(500000)
     # Get the version number from the relevant file. If not, from autosubmit package
     script_dir = os.path.abspath(os.path.dirname(__file__))
 
@@ -1142,8 +1140,7 @@ class Autosubmit:
 
         Log.info("Removing Structure db...")
         try:
-            if structure_db_path.exists():
-                os.remove(structure_db_path)
+            os.remove(structure_db_path)
         except BaseException as e:
             error_message += f"Cannot delete structure: {e}\n"
 
@@ -1235,7 +1232,7 @@ class Autosubmit:
                             yield key, value
         template_files = [file.name for file in (read_files('autosubmitconfigparser.config')/'files').iterdir() if file.is_file()]
         if parameters is None:
-            parameters = PARAMETERS
+            parameters = {}
         parameter_comments = dict(_recurse_into_parameters(parameters))
 
         for as_conf_file in template_files:
@@ -1488,25 +1485,17 @@ class Autosubmit:
         :return: Nothing, modifies input.
         """
 
-        Log.debug("Loading parameters...")
-        parameters = as_conf.load_parameters()
-        Log.debug("Parameters load.")
-        for platform_name in platforms:
-            platform = platforms[platform_name]
-            # Call method from platform.py parent object
-            platform.add_parameters(parameters)
+        Log.debug("Loading HPC parameters...")
         # Platform = from DEFAULT.HPCARCH, e.g. marenostrum4
         if as_conf.get_platform() not in platforms.keys():
             Log.warning("Main platform is not defined in platforms.yml")
         else:
             platform = platforms[as_conf.get_platform()]
-            platform.add_parameters(parameters, True)
+            platform.add_parameters(as_conf)
         # Attach parameters to JobList
-        parameters['STARTDATES'] = []
+        as_conf.experiment_data['STARTDATES'] = []
         for date in job_list._date_list:
-            parameters['STARTDATES'].append(date2str(date, job_list.get_date_format()))
-
-        job_list.parameters = parameters
+            as_conf.experiment_data['STARTDATES'].append(date2str(date, job_list.get_date_format()))
 
     @staticmethod
     def inspect(expid, lst, filter_chunks, filter_status, filter_section, notransitive=False, force=False,
@@ -1551,8 +1540,6 @@ class Autosubmit:
                                   "pkl", "job_packages_" + expid + ".db"), 0o644)
 
             packages_persistence.reset_table(True)
-            job_list_original = Autosubmit.load_job_list(
-                expid, as_conf, notransitive=notransitive)
             job_list = Autosubmit.load_job_list(
                 expid, as_conf, notransitive=notransitive)
             job_list.packages_dict = {}
@@ -1791,8 +1778,6 @@ class Autosubmit:
                 wrapper_job.new_status)
             # Erase from packages if the wrapper failed to be queued ( Hold Admin bug )
             if wrapper_job.status == Status.WAITING:
-                for inner_job in wrapper_job.job_list:
-                    inner_job.packed = False
                 job_list.job_package_map.pop(
                     wrapper_id, None)
                 job_list.packages_dict.pop(
@@ -2005,6 +1990,7 @@ class Autosubmit:
             raise AutosubmitCritical(
                 "Corrupted job_list, backup couldn't be restored", 7040, e.message)
         except BaseException as e:
+            Log.debug(f"Error while loading job_list: {str(e)}")
             raise AutosubmitCritical(
                 "Corrupted job_list, backup couldn't be restored", 7040, str(e))
         Log.debug("Length of the jobs list: {0}", len(job_list))
@@ -2076,6 +2062,7 @@ class Autosubmit:
 
         Log.debug("Checking job_list current status")
         job_list.update_list(as_conf, first_time=True)
+        job_list.clear_generate()
         job_list.save()
         as_conf.save()
         if not recover:
@@ -2129,6 +2116,9 @@ class Autosubmit:
     @staticmethod
     def check_logs_status(job_list, as_conf, new_run):
         for job in job_list.get_completed_failed_without_logs():
+            if new_run:
+                job.platform.spawn_log_retrieval_process(as_conf)
+
             job_list.update_log_status(job, as_conf, new_run)
 
     @staticmethod
@@ -2159,7 +2149,7 @@ class Autosubmit:
             profiler = Profiler(expid)
             profiler.start()
 
-        # Initialize common folders
+        # Initialize common folders'
         try:
             exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, expid)
             tmp_path = os.path.join(exp_path, BasicConfig.LOCAL_TMP_DIR)
@@ -2176,7 +2166,7 @@ class Autosubmit:
                     job_list, submitter , exp_history, host , as_conf, platforms_to_test, packages_persistence, _ = Autosubmit.prepare_run(expid, notransitive, start_time, start_after, run_only_members)
                 except AutosubmitCritical as e:
                     #e.message += " HINT: check the CUSTOM_DIRECTIVE syntax in your jobs configuration files."
-                    raise AutosubmitCritical(e.message, 7014, e.trace)
+                    raise
                 except Exception as e:
                     raise AutosubmitCritical("Error in run initialization", 7014, str(e))  # Changing default to 7014
                 Log.debug("Running main running loop")
@@ -2201,16 +2191,16 @@ class Autosubmit:
                 recovery_retrials = 0
                 Autosubmit.check_logs_status(job_list, as_conf, new_run=True)
                 while job_list.get_active():
-                    Autosubmit.refresh_log_recovery_process(platforms_to_test, as_conf)
-                    for job in [job for job in job_list.get_job_list() if job.status == Status.READY]:
-                        job.update_parameters(as_conf, {})
-                    did_run = True
                     try:
                         if Autosubmit.exit:
                             Autosubmit.check_logs_status(job_list, as_conf, new_run=False)
                             if job_list.get_failed():
                                 return 1
                             return 0
+                        Autosubmit.refresh_log_recovery_process(platforms_to_test, as_conf)
+                        for job in job_list.get_ready():
+                            job.update_parameters(as_conf, set_attributes=True)
+                        did_run = True
                         # reload parameters changes
                         Log.debug("Reloading parameters...")
                         try:
@@ -2435,6 +2425,7 @@ class Autosubmit:
         platform_issues = ""
         ssh_config_issues = ""
         private_key_error = "Please, add your private key to the ssh-agent ( ssh-add <path_to_key> ) or use a non-encrypted key\nIf ssh agent is not initialized, prompt first eval `ssh-agent -s`"
+
         for platform in platform_to_test:
             platform_issues = ""
             try:
@@ -2972,7 +2963,8 @@ class Autosubmit:
             if len(current_active_jobs) > 0:
                 if force and save:
                     for job in current_active_jobs:
-                        if job.platform_name is None:
+                        job.platform_name = as_conf.jobs_data.get(job.section, {}).get("PLATFORM", "").upper()
+                        if not job.platform_name:
                             job.platform_name = hpcarch
                         job.platform = submitter.platforms[job.platform_name]
                         platforms_to_test.add(job.platform)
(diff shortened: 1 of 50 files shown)


dbeltrankyl:
I'll start another branch from this one to review this part.

kinow:
I think I can start reviewing this one tomorrow :+1:

autosubmit/platforms/platform.py
```diff
@@ -17,23 +18,68 @@
...
+        return f"UniqueDeque({list(self.deque)})"
+
+
 class UniqueQueue(Queue):
     """
     A queue that avoids retrieves the same job and retrial during the same run.
     """
 
-    def __init__(self, maxsize: int = -1, block: bool = True, timeout: float = None):
+    def __init__(self, maxsize: int = -1, block: bool = True, timeout: float = None, max_items: int = 200):
```
kinow, line 69: Would limiting the number of elements in the queue (`maxsize`) be another way to solve this issue?

autosubmit/platforms/platform.py
```diff
@@ -45,13 +91,16 @@ def put(self, job: Any, block: bool = True, timeout: float = None) -> None:
...
         else:
-            unique_name = job.name+str(job.fail_count) # We gather retrial per retrial
+            unique_name = job.name+str(job.fail_count)  # We gather retrial per retrial
+
         if unique_name not in self.all_items:
-            self.all_items.add(unique_name)
-            super().put(copy(job), block, timeout)  # Without copy, the process seems to modify the job for other retrials.. My guess is that the object is not serialized until it is get from the queue.
+            self.all_items.append(unique_name)
+            # Without copy, the process seems to modify the job for other retrials.. My guess is that the object is not serialized until it is get from the queue.
+            super().put(copy(job), block, timeout)
```
kinow, line 103: I think I got the gist of the change here. We add things to `UniqueQueue` as we did before, but the `UniqueQueue` uses the `UniqueDeque` (backed by a `set` and a `deque`) to then limit the number of elements added.

I think `UniqueDeque` will consume more memory then a `set`, but it will limit the number of elements added, assuming `maxlen` is exceeded by some considerable amount.

Wouldn't it work too if we used just a `deque`? The memory consumption difference wouldn't be major. Access to elements int he middle of the queue are `O(n)`, which is not that bad, given that we are also limiting it to `200` by default. So it shouldn't really matter `set` vs. `deque` for performance I think.

```diff
diff --git a/autosubmit/platforms/platform.py b/autosubmit/platforms/platform.py
index 16b7fc89..46b58d75 100644
--- a/autosubmit/platforms/platform.py
+++ b/autosubmit/platforms/platform.py
@@ -9,7 +9,7 @@ import locale
 import os
 import traceback
 from autosubmit.job.job_common import Status
-from typing import List, Union, Callable, Set, Any
+from typing import List, Union, Set, Any
 from autosubmit.helpers.parameters import autosubmit_parameter
 from autosubmitconfigparser.config.configcommon import AutosubmitConfig
 from log.log import AutosubmitCritical, AutosubmitError, Log
@@ -18,49 +18,6 @@ from multiprocessing.queues import Queue
 import time
 
 
-class UniqueDeque:
-    def __init__(self, maxlen=None):
-        self.maxlen = maxlen
-        self.deque = deque(maxlen=maxlen)
-        self.set = set()
-
-    def append(self, item):
-        if item not in self.set:
-            if len(self.set) >= self.maxlen:
-                self.set.remove(self.deque.popleft())
-            self.deque.append(item)
-            self.set.add(item)
-
-    def appendleft(self, item):
-        if item not in self.set:
-            if len(self.set) >= self.maxlen:
-                self.set.remove(self.deque.popleft())
-            self.deque.appendleft(item)
-            self.set.add(item)
-
-    def pop(self):
-        item = self.deque.pop()
-        self.set.remove(item)
-        return item
-
-    def popleft(self):
-        item = self.deque.popleft()
-        self.set.remove(item)
-        return item
-
-    def __contains__(self, item):
-        return item in self.set
-
-    def __len__(self):
-        return len(self.deque)
-
-    def __iter__(self):
-        return iter(self.deque)
-
-    def __repr__(self):
-        return f"UniqueDeque({list(self.deque)})"
-
-
 class UniqueQueue(Queue):
     """
     A queue that avoids retrieves the same job and retrial during the same run.
@@ -79,7 +36,7 @@ class UniqueQueue(Queue):
         """
         self.block = block
         self.timeout = timeout
-        self.all_items = UniqueDeque(maxlen=max_items)  # Won't be popped, so even if it is being processed by the log retrieval process, it won't be added again.
+        self.all_items = deque(maxlen=max_items)
         super().__init__(maxsize, ctx=multiprocessing.get_context())
 
     def put(self, job: Any, block: bool = True, timeout: float = None) -> None:
diff --git a/test/unit/test_log_recovery.py b/test/unit/test_log_recovery.py
index 6f33eb9a..12c8344d 100644
--- a/test/unit/test_log_recovery.py
+++ b/test/unit/test_log_recovery.py
@@ -1,14 +1,15 @@
-import time
-import pytest
-from pathlib import Path
 import os
 import pwd
+import time
+from pathlib import Path
+
+import pytest
 
 from autosubmit.autosubmit import Autosubmit
+from autosubmit.job.job import Job
 from autosubmit.job.job_common import Status
-from autosubmit.platforms.platform import UniqueQueue, UniqueDeque
+from autosubmit.platforms.platform import UniqueQueue
 from autosubmitconfigparser.config.configcommon import AutosubmitConfig
-from autosubmit.job.job import Job
 
 
 def _get_script_files_path() -> Path:
@@ -263,48 +264,3 @@ def test_unique_elements(local):
     for i in range(max_items):
         local.recovery_queue.put(Job(f'rng2{i}', f'000{i}', Status.COMPLETED, 0))
     assert len(local.recovery_queue.all_items) == max_items
-
-
-def test_unique_deque():
-    unique_deque = UniqueDeque(maxlen=5)
-
-    # Test append and uniqueness
-    unique_deque.append(1)
-    unique_deque.append(2)
-    unique_deque.append(1)
-    assert list(unique_deque) == [1, 2]
-
-    # Test appendleft and uniqueness
-    unique_deque.appendleft(0)
-    unique_deque.appendleft(2)  # Duplicated
-    assert list(unique_deque) == [0, 1, 2]
-
-    # Test pop
-    assert unique_deque.pop() == 2
-    assert list(unique_deque) == [0, 1]
-
-    # Test popleft
-    assert unique_deque.popleft() == 0
-    assert list(unique_deque) == [1]
-
-    # Test maxlen
-    unique_deque.append(2)
-    unique_deque.append(3)
-    unique_deque.append(4)
-    unique_deque.append(5)
-    unique_deque.append(6)
-    assert list(unique_deque) == [2, 3, 4, 5, 6]
-
-    # test maxlen appendleft
-    unique_deque.appendleft(1)
-    unique_deque.appendleft(0)
-    unique_deque.appendleft(-1)
-    unique_deque.appendleft(-2)
-    unique_deque.appendleft(-3)
-    assert list(unique_deque) == [-3, 3, 4, 5, 6]
-
-    # Test contain
-    assert -3 in unique_deque
-
-    # test repr
-    assert repr(unique_deque) == "UniqueDeque([-3, 3, 4, 5, 6])"
```

I think the rest should work the same?

docs/source/userguide/configure/index.rst
```diff
@@ -352,6 +352,8 @@ There are some other parameters that you may need to specify:
 
 * TOTAL_JOBS: maximum number of jobs to be running at the same time in this platform.
 
+* LOG_RECOVERY_QUEUE_SIZE: A memory-consumption optimization for the recovery of logs. If not specified, defaults to min(100,TOTAL_JOBS) * 2, in case of issues with the recovery of logs, you can increase this value.
```
kinow, line 355: I think the default is `200`.

`TOTAL_JOBS` doesn't have a default (above this line), so I would guess it to be `0` or `-1`... and then `min(100, TOTAL_JOBS`) would return `-1` instead of `100`?

(found the default, I think)

image

image

autosubmit/platforms/paramiko_submitter.py
```diff
@@ -151,7 +151,6 @@ def load_platforms(self, asconf, retries=5, auth_password = None, local_auth_pas
                 else:
                     platform_type_value = platform_type or "<not defined>"
                     raise AutosubmitCritical(f"PLATFORMS.{section.upper()}.TYPE: {platform_type_value} for {section.upper()} is not supported", 7012)
-                remote_platform.main_process_id = os.getpid()
```
kinow, line 154: :+1: looks like it had been removed and we missed this one. Thanks for fixing it!

kinow (commented):
Looking good, left some comments about the code, but a more general question: Am I right assuming that if the default is left at `200`, and for some member/chunk/split level we have 1000 jobs running simultaneously, then we would only collect the logs of 200 files? And will the other files still be collected later?

Thanks Dani!

autosubmit/platforms/platform.py
dbeltrankyl: The set is to ensure that the element is unique. 

Autosubmit relies on the updated_log variable set to False and the job is completed or failed to send the job to the queue. 

Autosubmit uses the update_status function to call

                self.platform.add_job_to_log_recover(self)

But that is also called in 
                Autosubmit.check_logs_status(job_list, as_conf, new_run=False)

I think that this situation should happen only when all jobs are finished. 

With the deque only we could have the same job twice in the queue.

Maybe we can just avoid to do the call in  Autosubmit.check_logs_status(job_list, as_conf, new_run=False)
 if it is not called in a new run?

autosubmit/platforms/platform.py
dbeltrankyl: > Am I right assuming that if the default is left at 200, and for some member/chunk/split level we have 1000 jobs running simultaneously, then we would only collect the logs of 200 files? And will the other files still be collected later?

The logs will always be recovered, but in case all 1000 jobs end simultaneously ( in that platform) and they're not inside a wrapper, Autosubmit may try to recover them more than once ( and in that case, it will say that he couldn't download them ) . But that case is an extreme case

docs/source/userguide/configure/index.rst
kinow: That makes sense then :+1:

autosubmit/platforms/platform.py
dbeltrankyl: Ah, okay. Omitted that if

 Yes, I'll apply the change.

dbeltrankyl:
Feedback resolved, #2141  is blocking the merge of this one

dbeltrankyl:
Thanks @kinow 

I see it approved, but #2141 is based on this one. If we squash the commits, #2141 will still reference the old commits, and it will be messed up

dbeltrankyl:
Master branch: ( Note that my expdef configuration seems not to be compatible with the master branch; the log process is failing and starting again over and over, which could mess with the metric.) 

Mem starts at 67.84~ which quickly raises

image

77 mem after just 200 seconds
image


-----------------------------------

Mem starts at 65.525~
image

Zoom in: 65.6 ~ after 1600 seconds still leaking a bit
image

dbeltrankyl:
> Looks a lot better, @dbeltrankyl ! Looks like there are still 24 unit tests failing. Are you planning to fix them before we deploy to the VM?

Yes

dbeltrankyl:
Test with 27k jobs

image

I had to change one thing about the update parameters tho.. I'll have to redo the small tests. 

When I increased the number of chunks, the init of the run was taking a lot of time

dbeltrankyl:
After the changes due to the big experiment (100 chunks), I had leaks again in the small case (1 chunk).

So I made more changes, and I think it is solved now, but I will check the big experiment tomorrow. I also want to test that the  wrappers are working and fix the current tests

a Full run (273 jobs, although I had some already completed ) 

image

dbeltrankyl:
Ok, tests passing ( with the configparser branch ) , no mem leaks. 

I'll create a module called 4.1.12-mem-fixes on the climate-dt @kinow. 

I would want to increase the coverage, but I think it is not feasible.. at least not now

I also want to add typing, etc, but I want to do a last update ( in a different PR ) to fix the CPU issue with the logs after that I'll try to add them

dbeltrankyl:
Module added 4.1.12-dev-mem-issues, the last commits are to re enable the safety timer and delete unused imports related to debug tools

dbeltrankyl:
New updated, and revised mem

image

Still seems good

dbeltrankyl:
Tested in the climate-dt.

Waiting for the merge until the workflow developers manage to run the workflow with this version.

dbeltrankyl:
The coverage is 70.76023% I expected less, so this is a bit of relief

autosubmit/autosubmit.py
```diff
@@ -2137,6 +2127,15 @@ def refresh_log_recovery_process(platforms: List[Platform], as_conf: AutosubmitC
                 p.spawn_log_retrieval_process(as_conf)
             p.work_event.set()
 
+    # @staticmethod
+    # def show_differences(snapshot, snapshot2, limit=10):
+    #     top_stats = snapshot2.compare_to(snapshot, 'lineno')
+    #
+    #     print("[ Top 30 differences ]")
+    #     for stat in top_stats[:limit]:
+    #         print(stat)
```
kinow, line 2136: Should we delete it?

autosubmit/autosubmit.py
```diff
@@ -2197,17 +2196,24 @@ def run_experiment(expid, notransitive=False, start_time=None, start_after=None,
...
-                    Autosubmit.refresh_log_recovery_process(platforms_to_test, as_conf)
-                    for job in [job for job in job_list.get_job_list() if job.status == Status.READY]:
-                        job.update_parameters(as_conf, {})
-                    did_run = True
+                    # if i == 5:
+                    #     snapshot = tracemalloc.take_snapshot()
+                    # if i > 5:
+                    #     snapshot2 = tracemalloc.take_snapshot()
+                    #     Autosubmit.show_differences(snapshot, snapshot2, limit=30)
+                    # i += 1
```
kinow, line 2206: Delete it now?

autosubmit/job/job_list.py
```diff
@@ -321,17 +295,24 @@ def generate(self, as_conf, date_list, member_list, num_chunks, chunk_ini, param
...
                 job_platform = None
                 if first:
                     if not job.platform and hasattr(job, "platform_name") and job.platform_name:
                         submitter = _get_submitter(as_conf)
                         submitter.load_platforms(as_conf)
-                        if job.platform_name not in submitter.platforms:
-                            job.update_parameters(as_conf,{})
                         job_platform = submitter.platforms[job.platform_name]
-                    first = False
+                        first = False
```
kinow, line 306: The previous code was toggling the value of `first` whenever `first` was `True`. Now, it will change it only when that's `True` **and** the `if not job.platform and hasattr(job, "platform_name") and job.platform_name` condition is `True`. Just checking if that's intentional?

kinow (approved):
@dbeltrankyl I did a speedy review, just quickly browsing the code, skipped tests, and commented only in parts that I wasn't sure if the code was doing really what it was intended to do, or if there was code to be deleted.

I think it now needs to be rebased too. Marking it as approved assuming it's rebased and CICD passes. Thanks!!

autosubmit/job/job.py
```diff
@@ -261,48 +296,107 @@ def __init__(self, name, job_id, status, priority):
         self.wrapper_name = None
         self.is_wrapper = False
         self._wallclock_in_seconds = None
-
-    def _adjust_new_parameters(self) -> None:
```
kinow, line 265: @dbeltrankyl , @LuiggiTenorioK just warned me about this change.

I forgot about the `_adjust_new_parameters` with slots. I guess we cannot really remove it, as we will want it to handle cases like loading data from older code.

And the current code needs the `default_parameters` added too.

LuiggiTenorioK: I'm not sure if it is only `default_parameters`.

To be sure I'll suggest grabbing a 4.1.11 pkl and doing a `pickle.load` having these PR changes.

dbeltrankyl: Hello @kinow 

> I forgot about the _adjust_new_parameters with slots. I guess we cannot really remove it, as we will want it to handle cases like loading data from older code.

This function was made originally because the loaded jobs weren't using the __init__ function. This is not longer the case as the job is recreated through Job(loaded_data=job_from_disk) 

So, we don't need that function anymore

> And the current code needs the default_parameters added too.

Why do you think so? It is now in the as_conf parser

LuiggiTenorioK: I see. So, the temporary solution will be to lock the Autosubmit dependency of the API to version 4.1.11 so it can load the pkl from 4.1.X to 4.1.11.

> However, it would work between 4.1.12 and 4.1.13 ( if we don't change the pickle functions there).

I'm worried about it. Does the new pkl have code inside? Ideally, the pkl should only have data to avoid breaking retro-compatibility and, if it only has data, why don't we store it on a JSON file?

dbeltrankyl: I didn't do that in this merge, because it will break the code for sure in multiple places not enough time to implement it

dbeltrankyl: > The idea is to store it as a yaml or/and disk as mentioned two weeks ago. And ideally we only want to have the job_status  ( not the job object) in memory

Also, in this version, we only store the job data, but ideally, we want to store the dependencies in a yaml format not sure if in the same file.. stored in the db ... that's for determinate. I'll create an issue for this 

That would be good to: 

* Don't call to add_dependencies if there are no changes 
* Fine-tune the dependencies directly in that yaml in case it is needed for some corner case

autosubmit/job/job_list.py
dbeltrankyl: It was intentional this first thing.. it doesn't make much sense.

This intends to set the platform, as it is not stored in the pkl. So we must load all platforms once, not load the platforms once per platform

```yaml
        submitter = _get_submitter(as_conf)
        submitter.load_platforms(as_conf)

        for platform in job_list_per_platform:
            for job in job_list_per_platform[platform]:
                if create or new:
                    job.reset_logs()
                # Maybe the platform doesn't exist. (It should be controlled in the config parser, tho.
                if job.platform_name in submitter.platforms:  
                    job.platform = submitter.platforms[job.platform_name]
```

dbeltrankyl:
Squashed before Rebasing + rebase

dbeltrankyl:
I rebased it again. The last time, I chose origin/master using Pycharm; however, it was rebased on my outdated local master 🤔.

Seems that some tests are not passing after rebasing, I'm looking at it

dbeltrankyl:
Should we merge this branch, and if there are issues, with the DestinE workflow, do a new PR based on master? @kinow

kinow (approved):

autosubmit/job/job.py
kinow: > Between 4.1.11 and 4.1.12, it is impossible to recover the jobs as the pickle functions changed. This is intended.

I think that means users will have to `autosubmit create -f` if they have 4.1.11.

Wouldn't it be possible to move the old logic that handled missing attributes to `__getstate__` so that it reads the pickle and fills what is available from the Digraph if that's a Digraph, leaving parameters empty?

And that way maybe 4.1.12 would be able to load 4.1.11 data, and in the next call to save a Job that would replace the DiGraph by a dictionary in the disk?

PR #2132, opened 2025-02-17
creating function to test if profile is running correctly and generating the file

VindeeR:
This commit is merely focused on creating a wider coverage of a function already tested but with blind spots

### DIFF
--- a/test/unit/test_run_command_intregation.py
+++ b/test/unit/test_run_command_intregation.py
@@ -414,24 +414,23 @@ def test_run_uninterrupted(run_tmpdir, prepare_run, jobs_data, expected_db_entri
     # assert_exit_code(final_status, exit_code)
 
 
-@pytest.mark.parametrize("jobs_data, profiler", [
-                                # Success
-                                ("""
-                                    EXPERIMENT:
-                                        NUMCHUNKS: '3'
-                                    JOBS:
-                                        job:
-                                            SCRIPT: |
-                                                echo "Hello World with id=Success"
-                                            PLATFORM: local
-                                            RUNNING: chunk
-                                            wallclock: 00:01
-                                    """,
-                                    True
-                                ),
-                            ]
-                         )
+@pytest.mark.parametrize("jobs_data, profiler",
+                        [
+                            ("""EXPERIMENT:
+                                NUMCHUNKS: '3'
+                            JOBS:
+                                job:
+                            SCRIPT: |
+                                echo "Hello World with id=Success"
+                            PLATFORM: local
+                            RUNNING: chunk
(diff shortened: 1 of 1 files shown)


test/unit/test_run_command_intregation.py
```diff
@@ -318,6 +318,20 @@ def init_run(run_tmpdir, jobs_data):
     return log_dir
 
 
+def check_profile(run_tmpdir) -> bool:
```
isimo00, line 321: - To concatenate the path I believe the `/` operator is preferred over `f"{}`, e.g. `str(run_tmpdir) / "t000/tmp/profile"`
- If `profile_path` doesn't exist, this function doesn't return anything - I guess it should return `false`?-
- What is this function testing? Each time you loop in the `for` (line 328) you are sure already that the `profile_file` exists, the conditionals inside will always return `true`. It returns `true or `false` as soon as it finds the first file inside `profile_path` (via `profile_file.exists()`)

test/unit/test_run_command_intregation.py
```diff
@@ -400,3 +414,28 @@ def test_run_uninterrupted(run_tmpdir, prepare_run, jobs_data, expected_db_entri
...
+            RUNNING: chunk
+            wallclock: 00:01
+    """, 3, "COMPLETED"),  # Number of jobs
+], ids=["Success"])
+def test_run_profile(run_tmpdir, prepare_run, jobs_data, expected_db_entries, final_status):
+    log_dir = init_run(run_tmpdir, jobs_data)
+    # Run the experiment
+    try:
+        exit_code = Autosubmit.run_experiment(expid='t000', profile=True)
+        check_profile(run_tmpdir)
```
isimo00, line 438: You can remove this line, `check_profile(run_tmpdir)` is already called in the following one; `assert check_profile(run_tmpdir)`

isimo00 (commented):
Sorry to chime in unsolicitedly, I commented a few things I've seen

test/unit/test_run_command_intregation.py
```diff
@@ -400,3 +412,30 @@ def test_run_uninterrupted(run_tmpdir, prepare_run, jobs_data, expected_db_entri
     assert_files_recovered(files_check_list)
     # TODO: GITLAB pipeline is not returning 0 or 1 for check_exit_code(final_status, exit_code)
     # assert_exit_code(final_status, exit_code)
+
+
+@pytest.mark.parametrize("jobs_data, profiler", [
```
isimo00, line 417: Linting :dizzy_face:

isimo00 (approved):

PR #2135, opened 2025-02-18
Add script to test SMTP notifications with Docker

### LINKED ISSUE #2077: Add the script to test SMTP notifications with Docker somewhere in our source tree
See https://github.com/BSC-ES/autosubmit/pull/1997#pullrequestreview-2564678607

It's based on a script written by @isimo00 that can be used to test Autosubmit notifications without the need to use a real SMTP server. We should probably document it somewhere, maybe store in our Git repo, etc., so that we have it in the future if/when needed.

isimo00:
Add the STMP script we used to test email notifications in #1997 to project tree plus additional documentation

Closes #2077

### DIFF
--- a/autosubmit/notifications/mail_notifier.py
+++ b/autosubmit/notifications/mail_notifier.py
@@ -217,7 +217,7 @@ class MailNotifier:
 
         self._send_message(mail_to, self.config.MAIL_FROM, message)
 
-    def _send_message(self, mail_to: str, mail_from: str, message) -> None:
+    def _send_message(self, mail_to: list[str], mail_from: str, message) -> None:
         formatted_addresses = [
                 email.utils.formataddr(
                     (mail, mail)) for mail in mail_to]


dbeltrankyl:
Hello @kinow , @isimo00 

This seems to have been ready for review for a long time. Any update?

kinow (changes requested):
@isimo00 I think instead of adding just the script, we should be able to write a test for it.

You can write a integration test in `test/integration` using pytest. You won't have to use the fixtures, I think.

You can import `testcontainers.DockerContainer` and use it (I think we already have a test with that), passing that container I used to review your other PR. Then copy parts of your `SMTP_notifications.py` script in the pytest test, and that should run fine inside the container (the container has that fake SMTP server, and a command to check for emails, I think), and should increase the coverage, and prevent future issues as it'll be part of the CI jobs.

kinow:
> Retaking this issue now. Should take SMTP server container info from this comment [#1997 (review)](https://github.com/BSC-ES/autosubmit/pull/1997#pullrequestreview-2564678607) and following ones in the discussion

Great! Thanks Irene!!!

test/integration/test_mail.py
```diff
@@ -0,0 +1,166 @@
...
+        smtp_host = container.get_container_host_ip()
+        api_base = f"http://{smtp_host}:{api_port}"
+
+        for _ in range(10):
+            try:
+                requests.get(f"{api_base}/api/v2/messages")
+                break
+            except requests.ConnectionError:
+                import time
+                time.sleep(1)
```
kinow, line 30: Ah, an interesting way to make sure the container is running!

TestContainers provide a similar feature already. Check the `DockerContainer` source for an example. You can give it an expected string. Once that string appears in the logs, then the container is considered ready to be used. You can also find a example at `test/integration/test_paramiko_platform.py`. Just check what the container used here outputs when it's ready to receive requests.

With that I think we can remove the `requests` dependency from this test :+1:

test/integration/test_mail.py
```diff
@@ -0,0 +1,166 @@
+import pytest
+import requests
+from typing import List
+from testcontainers.core.container import DockerContainer
+from autosubmit.notifications.mail_notifier import MailNotifier
+from autosubmit.job.job_common import Status
+from typing import cast
+
+SMTP_PORT = 1025
+API_PORT = 8025
```
kinow, line 10: I think we should make these random. Wouldn't it make the test fail if it's executed in parallel in the same host, trying to use the same ports? If so, you can use a similar approach to `test_paramiko_platform.py`, using a different range to get some ports for your tests.

test/integration/test_mail.py
```diff
@@ -0,0 +1,166 @@
...
+
+def test_experiment_status(mail_notifier, fake_smtp_server):
+    api_base = fake_smtp_server["api_base"]
+    expid = 'a000'
+    job_name = 'SIM'
+    to_email = ['test@example.com']
+    requests.delete(f"{api_base}/api/v1/messages")
+
+    platform = cast(
+        "Platform", type(
```
kinow, line 105: I **think** we can import the `Platform` here (see comment in the test file).

test/integration/test_mail.py
```diff
@@ -0,0 +1,166 @@
...
+
+    config = type('Config', (), {
+        'MAIL_FROM': 'notifier@localhost',
+        'SMTP_SERVER': f'{smtp_host}:{smtp_port}',
+        'expid_aslog_dir': staticmethod(expid_aslog_dir),
+    })()
+    return MailNotifier(config)
+
+
+def check_metadata(
```
kinow, line 59: Not important, but safer to use "private" (real hard air-quotes as in Python nothing is really private) functions whenever these are not expected to be used elsewhere. So `_check_metadata` for example here.

test/integration/test_mail.py
```diff
@@ -0,0 +1,166 @@
...
+        emails: List[str],
+        expected_subject: str,
+        expid: str,
+        sender: str,
+        recipients: List[str]) -> None:
+    subject = [email["Content"]["Headers"]["Subject"][0] for email in emails]
```
kinow, line 65: I'm getting some type warnings here.

image

I think that's because `emails` is `list[str]` (can use `list` instead of `List`), but we use `email` as an object or dictionary, accessing things like `email["Content"]["Headers"]["Subject"][0]`.

I think the type is incorrect in the function signature.

kinow (commented):
Coverage going up, nice! Have a look at the comments about warnings, small code formatting/style, and a point about using the testcontainers feature to check when a container is ready :+1: 

Thanks Irene!!

isimo00:
@kinow @LuiggiTenorioK Do you have any idea how I can fix this in the integration test? I've randomized the smtp and api ports and even included a funciton that (supposedly) ensures the random ports are free and not equal to each other, but it doesn't seem to work:

```
docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.48/containers/4e839e39fc8e82a5e694fc8d237d095c789bb2287008bd96c2972e690961f7ea/start: Internal Server Error ("failed to set up container networking: driver failed programming external connectivity on endpoint happy_dewdney (22271629274bebd5489d45a5a248046413dcae5524bf1cbe0d79385f6dbf2629): Bind for 0.0.0.0:2996 failed: port is already allocated")
```

LuiggiTenorioK:
> @kinow @LuiggiTenorioK Do you have any idea how I can fix this in the integration test? I've randomized the smtp and api ports and even included a funciton that (supposedly) ensures the random ports are free and not equal to each other, but it doesn't seem to work:

My guess is that it is because the tests are distributed among subprocesses, making the `fake_smtp_server` fixture run multiple times per worker.

The usual way I check for free ports is with `netstat`, which seems a little messy to plug with Python.

However, something that you can try is to distribute all those tests to the same worker using `@pytest.mark.xdist_group(name="mail_tests")`, so you can make sure the fixture is run only once. Here is some documentation on how to do it: https://pytest-xdist.readthedocs.io/en/stable/distribution.html

As far as I remember, `--dist loadgroup` is already set in the pytest command. So, no change there. Right, @kinow?

test/integration/test_mail.py
kinow: The issue here, is that if we change anything in `Platform`, this line probably will not appear in any IDE refactoring tool (PyCharm/VSCode index the code to search it and indicate where the code could be used). So we run the risk that this test may get out of sync with the real code.

If you only need to call name and host, maybe we could use `mocker` to create a `Mock` or `MagicMock` of a `Platform` object. You can also try the `spec=Platform` in the constructor. That would mock the specification, raising an error if any attribute that does not exist is accessed (there are still risks if `Platform` has some behaviour hidden with `@property`, etc., but at least the object should reflect something closer to what the actual `Platform` does.

test/integration/test_mail.py
```diff
@@ -0,0 +1,166 @@
...
+    assert resp.json()["count"] == 1
+    emails = resp.json()["items"]
+    check_metadata(emails, "status has changed to FAILED",
+                   expid, 'notifier@localhost', to_email)
+
+
+def test_experiment_status(mail_notifier, fake_smtp_server):
+    api_base = fake_smtp_server["api_base"]
+    expid = 'a000'
+    job_name = 'SIM'
```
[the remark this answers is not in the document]
isimo00: You're right, I mixed up `test_experiment_status` and `test_recipients_list` 😅 Thanks!

kinow:
Oh, and I also rebased your branch locally when writing this test as I noticed a package was missing in the tests dir. So it might be good to rebase this PR too :+1:

autosubmit/notifications/mail_notifier.py
```diff
@@ -1,23 +1,22 @@
...
+ # the Free Software Foundation, either version 3 of the License, or 
+ # (at your option) any later version. 
+ # 
+ # Autosubmit is distributed in the hope that it will be useful, 
+ # but WITHOUT ANY WARRANTY; without even the implied warranty of 
+ # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the 
+ # GNU General Public License for more details. 
+ # 
+ # You should have received a copy of the GNU General Public License 
+ # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
```
kinow, line 16: Every line of the header has a space before the `#` sign.

image

autosubmit/notifications/mail_notifier.py
```diff
@@ -187,39 +196,50 @@ def notify_experiment_status(
...
+        message["To"] = ", ".join(formatted_addresses)
+        try:
+            self._send_mail(
+                self.config.MAIL_FROM,
+                formatted_addresses,
+                message)
+        except Exception as e:
+            Log.printlog(
+                f'Trace:{str(e)}\nAn error has occurred while sending a warning mail '
+                f'about remote_platform', 6011)
```
kinow, line 242: IDE warned about code duplication here, and I think this could also become a function, similar to what you did with `_check_mail_address`

image

test/integration/test_mail.py
```diff
@@ -0,0 +1,178 @@
...
+# 
+# Autosubmit is distributed in the hope that it will be useful, 
+# but WITHOUT ANY WARRANTY; without even the implied warranty of 
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the 
+# GNU General Public License for more details. 
+# 
+# You should have received a copy of the GNU General Public License 
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>. 
+
+from random import randint, sample
```
kinow, line 18: `randint` is now not used.

test/integration/test_mail.py
```diff
@@ -0,0 +1,178 @@
...
+from typing import Tuple, cast
```
kinow, line 19: `cast` is now also not used,

image

test/integration/test_mail.py
```diff
@@ -0,0 +1,178 @@
...
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.waiting_utils import wait_for_logs
+
+from autosubmit.job.job_common import Status
+from autosubmit.notifications.mail_notifier import MailNotifier
+from autosubmit.platforms.platform import Platform
+
+
+@pytest.fixture(scope="module")
+def fake_smtp_server() -> Tuple[int, str]:
```
kinow, line 33: My bad, @isimo00 , in reality it's a generator of tuples (because of the `yield`).

```diff
diff --git a/test/integration/test_mail.py b/test/integration/test_mail.py
index b07b5c52..558b0cf5 100644
--- a/test/integration/test_mail.py
+++ b/test/integration/test_mail.py
@@ -16,7 +16,7 @@
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>. 
 
 from random import randint, sample
-from typing import Tuple, cast
+from typing import Generator, Tuple, cast
 from unittest.mock import Mock
 
 import pytest
@@ -30,7 +30,7 @@ from autosubmit.platforms.platform import Platform
 
 
 @pytest.fixture(scope="module")
-def fake_smtp_server() -> Tuple[int, str]:
+def fake_smtp_server() -> Generator[Tuple[int, str], None, None]:
     """Start fake SMTP server container.
     :return: A tuple with the SMTP port, and the SMTP test server API base URL """
     smtp_port, api_port = sample(range(3500, 4000), 2)
```

test/integration/test_mail.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+@pytest.mark.docker
+def test_experiment_status(mail_notifier, fake_smtp_server, mock_platform):
+    _, api_base = fake_smtp_server
+    expid = 'a000'
+    to_email = ['test@example.com']
+    requests.delete(f"{api_base}/api/v1/messages")
+
+    platform = mock_platform()
```
kinow, line 118: I think `mock_platform` is used only in this test, but it doesn't hurt to have that as a fixture.

However, I got confused when the IDE alerted for this line. Then, after staring at it for some seconds (haven't had :coffee: yet), I realized it's complaining because there's a `mock_platform` given via fixture/argument of the function, but I think you are calling the function here, instead of getting the `Platform` object injected into the function as fixture?

Surprised it's working fine :dizzy: It should work if you use the `mock_platform` that you passed in the function (just drop the `()`)

test/scripts/SMTP_notifications.py
kinow, line 1: minor nit-pick, can we make the file name all lower case? :grimacing:

test/scripts/SMTP_notifications.py
```diff
@@ -0,0 +1,56 @@
...
+ # the Free Software Foundation, either version 3 of the License, or 
+ # (at your option) any later version. 
+ # 
+ # Autosubmit is distributed in the hope that it will be useful, 
+ # but WITHOUT ANY WARRANTY; without even the implied warranty of 
+ # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the 
+ # GNU General Public License for more details. 
+ # 
+ # You should have received a copy of the GNU General Public License 
+ # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
```
kinow, line 16: Spaces before the `#`

isimo00:
I think it's ready now @kinow 👀

PR #2138, opened 2025-02-18
Wrapper deadlock fix

dbeltrankyl:
I had a bug deadlock because this was set to True when it shouldn't.

Changed all setters and getters related to packed. 

Added a new property to check and set this to false/true whenever needed (during the submission process ).

I think the pytests will be done only for the additions, not the deletions


dbeltrankyl:
Closing branch, reason:

https://github.com/BSC-ES/autosubmit/pull/2157#issuecomment-2687188438

PR #2139, opened 2025-02-18
This commit is focused on improving coverage and basic tests

VindeeR:
It was developed a couple extra targeting functions existent in the job/job_utils file focused on smoke test and making sure the objects works as intended

### DIFF
--- a/test/unit/test_job_pytest.py
+++ b/test/unit/test_job_pytest.py
@@ -327,7 +327,7 @@ def test_sub_job_instantiation(tmp_path, autosubmit_config):
                                  (
                                          {
                                              'dummy2':
-                                                 {'dummy','dummy','dummy'},
+                                                 {'dummy','dummy1','dummy4'},
                                              'dummy3':
                                                  'dummy'
                                          }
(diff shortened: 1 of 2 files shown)


test/unit/test_job_pytest.py
```diff
@@ -282,3 +284,97 @@ def test_no_start_time(autosubmit_config, experiment_data):
...
+    assert job.name == "dummy"
+    assert job.package is None
+    assert job.queue == 0
+    assert job.run == 0
+    assert job.total == 0
+    assert job.status == "UNKNOWN"
+
+
+#TODO this test is mostly a smoke test, which can be improved to expect specific results and specifics inputs
+@pytest.mark.parametrize("jobs, job_to_package, package_to_job, current_structure",
```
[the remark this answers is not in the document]
VindeeR: The idea for doing this way is that I would expand more on this test once I understand the data better and knew exactly what each input would generate, and for that reason I decided to tackle the internal functions first so I could build more knowledge before going back.

For the sake of keeping the code more readable and organized I'll simplify it for now, since I'm not sure when I'll be able to go back to this function.

the Id thing is GOAT thanks for the tip

test/unit/test_job_pytest.py
```diff
@@ -282,3 +284,89 @@ def test_no_start_time(autosubmit_config, experiment_data):
...
+    assert job.name == "dummy"
+    assert job.package is None
+    assert job.queue == 0
+    assert job.run == 0
+    assert job.total == 0
+    assert job.status == "UNKNOWN"
+
+
+#TODO this test is mostly a smoke test, which can be improved to expect specific results and specifics inputs
+@pytest.mark.parametrize("current_structure",
```
isimo00, line 325: Sorry to be picky, promise this is the last thing on this PR; linting :)

VindeeR: being picky is a good thing not a bad thing, you´re being extremely helpful 🙌

isimo00 (approved):

PR #2140, opened 2025-02-19
targeting improvement in coverage and making to increase the amount of targetting done by test_job_list

VindeeR:
Coverage of the system focusing on improving the failure detection of the system


autosubmit/job/job_list.py
```diff
@@ -456,7 +456,7 @@ def _add_dependencies(self, date_list, member_list, chunk_list, dic_jobs, option
             self.find_and_delete_redundant_relations(problematic_jobs)
         self._add_all_jobs_edge_info(dic_jobs, option)
 
-    def find_and_delete_redundant_relations(self, problematic_jobs):
+    def find_and_delete_redundant_relations(self, problematic_jobs: {}) -> None:
```
isimo00, line 459: As said in #2139, please add type hints for all function arguments:

`def find_and_delete_redundant_relations(self, problematic_jobs: dict = {}) -> None`

test/unit/test_job_list.py
```diff
@@ -657,11 +649,18 @@ def test_generate_job_list_from_monitor_run(self):
...
+                assert self.job_list.find_and_delete_redundant_relations(problematic_jobs) is None
+                mock_job_list.return_value = False
+                assert self.job_list.find_and_delete_redundant_relations(problematic_jobs) is None
 
-
-
-
-
+            except Exception as e:
+                assert f'function failed to execute exception Raised: {e}'
```
isimo00, line 663: Improve exception message

isimo00 (commented):
Two very easy fixes here :)

autosubmit/job/job_list.py
```diff
@@ -456,15 +456,15 @@ def _add_dependencies(self, date_list, member_list, chunk_list, dic_jobs, option
             self.find_and_delete_redundant_relations(problematic_jobs)
         self._add_all_jobs_edge_info(dic_jobs, option)
 
-    def find_and_delete_redundant_relations(self, problematic_jobs: {}) -> None:
-        '''
+    def find_and_delete_redundant_relations(self, problematic_jobs: dict) -> None:
```
isimo00, line 459: You can have both the type hint and default value:

`def find_and_delete_redundant_relations(self, problematic_jobs: dict = {}) -> None:`

VindeeR: It's not necessarily needed, since it didn't have it before.
To keep a consistent style as the others I think it would be better to let just dict, that is the proper and standard way to add the type hint

isimo00: ah, okay! Approved then :)

isimo00 (approved):

PR #2144, opened 2025-02-20
job_list _calculate_splits_dependencies improving coverage od the system

VindeeR:
improving coverage of the system


VindeeR:
The function is not used on the system and it'll be removed hence no need to test it and push the test

PR #2146, opened 2025-02-20
Gh 2122 mem issues reduce child mem leak

dbeltrankyl:
@kinow 

Testing mem leak. 

I've tested a very simple change, to set self.parameters = {} whenever the job is completed or waiting

Using: 1 job with 40000 chunks and 2 members ( 80k)

Before:
image

now:
image

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2128,10 +2128,10 @@ class Autosubmit:
 
     @staticmethod
     def check_logs_status(job_list, as_conf, new_run):
-        track_unused_platforms = []
         for job in job_list.get_completed_failed_without_logs():
-            if job.platform not in track_unused_platforms and new_run:
+            if new_run:
                 job.platform.spawn_log_retrieval_process(as_conf)
+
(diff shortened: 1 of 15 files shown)


dbeltrankyl:
Changed the base to #2146

kinow:
> Changed the base to #2146

#2141 I think?

kinow (commented):
The graph pattern looks better! Less memory ups-and-downs, it looks like we are saving some allocations with this change :+1: Will look at the other PR now.

dbeltrankyl:
> https://github.com/BSC-ES/autosubmit/pull/2141 I think?

yes

dbeltrankyl:
Hm, I have a bit of confusion if this branch is still useful, 

Too many branches working at the same time and the internet issues didn't help

I think the good one is #2157

dbeltrankyl:
Yes, closing this one, this is the dirty test before applying the real changes

dbeltrankyl:
( Don't remove the branch yet, just in case something is needed but I think not)

PR #2151, opened 2025-02-24
coverage improvment in function _check_dates

VindeeR:
As I was following the workflow of the test apply jobs edge, I notice a few of the functions being used were not full of coverage and trying to get a better understanding on how to properly create a test case for that function I've been working on the functions inside of it to expand the coverage and get more insight of the system

### DIFF
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -729,7 +729,7 @@ class JobList(object):
         # {DATES_TO: "20020201", MEMBERS_TO: "fc2", CHUNKS_TO: "ALL", SPLITS_TO: "2"}
         return filters_to_apply
 
-    def _check_members(self, relationships, current_job):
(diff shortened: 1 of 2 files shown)


autosubmit/job/job_list.py
```diff
@@ -699,7 +699,7 @@ def _check_relationship(self, relationships, level_to_check, value_to_check):
             filters = [{}]
         return filters
 
-    def _check_dates(self, relationships, current_job):
+    def _check_dates(self, relationships, current_job) -> {}:
```
isimo00, line 702: Improve type hints, see reviews in #2139 and #2155

isimo00, line 702: Improve type hints, see reviews in #2139 and #2155

isimo00 (approved):

PR #2153, opened 2025-02-24
improving coverage of the test_check_chunk and adding appropriate return

VindeeR:
As I was following the workflow of the test apply jobs edge, I notice a few of the functions being used were not full of coverage and trying to get a better understanding on how to properly create a test case for that function I've been working on the functions inside of it to expand the coverage and get more insight of the system

### DIFF
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -749,7 +749,7 @@ class JobList(object):
         filters_to_apply = self._unify_to_filters(filters_to_apply)
         return filters_to_apply
 
-    def _check_chunks(self, relationships, current_job) -> {}:
+    def _check_chunks(self, relationships: Dict, current_job: Job) -> {}:
         """
         Check if the current_job_value is included in the filter_from and retrieve filter_to value
         :param relationships: Remaining filters to apply.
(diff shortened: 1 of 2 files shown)


autosubmit/job/job_list.py
```diff
@@ -749,7 +749,7 @@ def _check_members(self, relationships, current_job):
         filters_to_apply = self._unify_to_filters(filters_to_apply)
         return filters_to_apply
 
-    def _check_chunks(self, relationships, current_job):
+    def _check_chunks(self, relationships, current_job) -> {}:
```
isimo00, line 752: Improve type hints, #2139 and #2155

test/unit/test_dependencies.py
```diff
@@ -329,17 +329,33 @@ def test_check_chunks(self):
...
 
         self.mock_job.chunk = 1
-        result = self.JobList._check_chunks(self.relationships_chunks, self.mock_job)
-        expected_output = {
-            "DATES_TO": "20020201",
-            "MEMBERS_TO": "fc2",
-            "CHUNKS_TO": "all",
-            "SPLITS_TO": "1"
+
+        chunks = {
```
isimo00, line 333: Improve linting (ruff, autopep8...). I think Luiggi uses ruff

VindeeR: I´m using both pylint and ruff and there are no issues there for me

isimo00 (approved):

PR #2155, opened 2025-02-25
creating testing case for _normalize_to_filters function

VindeeR:
Expanding coverage and adding return to the function

### DIFF
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -839,7 +839,7 @@ class JobList(object):
         return unified_filter
 
     @staticmethod
-    def _normalize_to_filters(filter_to, filter_type):
+    def _normalize_to_filters(filter_to, filter_type) -> None:
         """
         Normalize filter_to filters to a single string or "all"
         :param filter_to: Unified filter_to dictionary
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -839,7 +839,7 @@ class JobList(object):
         return unified_filter
 
     @staticmethod
-    def _normalize_to_filters(filter_to, filter_type) -> None:
+    def _normalize_to_filters(filter_to: dict, filter_type: str) -> None:
         """
         Normalize filter_to filters to a single string or "all"
         :param filter_to: Unified filter_to dictionary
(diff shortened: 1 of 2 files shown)


test/unit/test_job_list.py
```diff
@@ -657,8 +659,21 @@ def test_generate_job_list_from_monitor_run(self):
...
+        filter_to = [
+            {"DATES_TO": ""},
+            {"DATES_TO": "all"},
+            {"DATES_TO": ",20020205,[20020207:20020208],"}
+            # ,{"DATES_TO": 123} # Error Case
+        ]
+        filter_type = "DATES_TO"
 
-
+        for i in range(len(filter_to)):
```
isimo00, line 672: Why not:

```
for directory in filter_to:
   try:
       self.job_list._normalize_to_filters(directory, filter_type)
```

test/unit/test_job_list.py
```diff
@@ -657,8 +659,21 @@ def test_generate_job_list_from_monitor_run(self):
...
+            try:
+                self.job_list._normalize_to_filters(filter_to[i], filter_type)
+            except Exception as e:
+                print(f'Found an error: {e}')
```
isimo00, line 676: Improve error message

test/unit/test_job_list.py
```diff
@@ -657,8 +659,21 @@ def test_generate_job_list_from_monitor_run(self):
...
+                assert not bool(e)
```
isimo00, line 677: If you want to assert an exception is not raised, you can simply use: 

```
except Exception as e:
        pytest.fail(f"Unexpected exception raised: {e}")
```

autosubmit/job/job_list.py
```diff
@@ -839,7 +839,7 @@ def _unify_to_filter(self, unified_filter, filter_to, filter_type, splits = None
         return unified_filter
 
     @staticmethod
-    def _normalize_to_filters(filter_to, filter_type):
+    def _normalize_to_filters(filter_to, filter_type) -> None:
```
isimo00:
```
from typing import Dict, Any
from typing_extensions import Literal

def _normalize_to_filters(filter_to: Dict[str, Any], filter_type: Literal["DATES_TO", "MEMBERS_TO", "CHUNKS_TO", "SPLITS_TO"]) -> None:
```

test/unit/test_job_list.py
VindeeR: The problem is that this test is done with unit test and not pytest, since pytest doesn´t work inside the unitest pipeline but the message can still be used

PR #2162, opened 2025-02-27
test handle_start_after and refactoring code

VindeeR:
- Testing function handle_start_after
- File lint in accordance
- Refactoring code

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1994,7 +1994,7 @@ class Autosubmit:
                 # Handling starting time
                 AutosubmitHelper.handle_start_time(start_time)
                 # Start after completion trigger block
-                AutosubmitHelper.handle_start_after(start_after, expid, BasicConfig())
+                AutosubmitHelper.handle_start_after(start_after, expid)
                 # Handling run_only_members
             except AutosubmitCritical as e:
                 raise
--- a/autosubmit/helpers/autosubmit_helper.py
+++ b/autosubmit/helpers/autosubmit_helper.py
@@ -20,7 +20,7 @@
 import datetime
 import sys
 from time import sleep
-from typing import List
+from typing import Union, Any
 
 from autosubmit.database.db_common import check_experiment_exists
 from autosubmit.history.experiment_history import ExperimentHistory
@@ -35,7 +35,6 @@ def handle_start_time(start_time: str) -> None:
         Log.info("User provided starting time has been detected.")
         # current_time = time()
         datetime_now = datetime.datetime.now()
-        target_date = parsed_time = None
         try:
             # Trying first parse H:M:S
             parsed_time = datetime.datetime.strptime(start_time, "%H:%M:%S")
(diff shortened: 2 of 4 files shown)


kinow:
Git conflicts :point_up:

autosubmit/autosubmit.py
```diff
@@ -1965,7 +1965,7 @@ def prepare_run(expid, notransitive=False, start_time=None, start_after=None,
                 # Handling starting time
                 AutosubmitHelper.handle_start_time(start_time)
                 # Start after completion trigger block
-                AutosubmitHelper.handle_start_after(start_after, expid, BasicConfig())
+                AutosubmitHelper.handle_start_after(start_after, expid)
```
kinow, line 1968: Hmmm, the old code looks messy. We passed a new instance of `BasicConfig()` here. Then, in the function, the argument was named `BasicConfig`.

There is an import `BasicConfig` too that accesses a singleton. Now, which one was supposed to be used? An empty/new instance of `BasicConfig`? Or the global `BasicConfig`?

I'd need to check what's in `BasicConfig` before hitting the approve here, :bow: so won't be able to review it now (doing a 30 minutes round of reviews)

But thanks for spotting this and trying to fix! And with a test :muscle:

test/unit/test_run_command_intregation.py
```diff
@@ -400,90 +400,3 @@ def test_run_uninterrupted(run_tmpdir, prepare_run, jobs_data, expected_db_entri
...
-    Autosubmit.run_experiment(expid='t000')
-    # Check and display results
-    db_check_list = check_db_fields(run_tmpdir, expected_db_entries, final_status)
-    files_check_list = check_files_recovered(run_tmpdir, log_dir, expected_files=expected_db_entries*2)
-
-    # Assert
-    assert_db_fields(db_check_list)
-    assert_files_recovered(files_check_list)
-    # TODO: GITLAB pipeline is not returning 0 or 1 for check_exit_code(final_status, exit_code)
-    # assert_exit_code(final_status, exit_code)
```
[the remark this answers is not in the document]
VindeeR: Confusion during rebase, returning file to how it was before

test/unit/test_autosubmit_helper.py
```diff
@@ -71,3 +70,44 @@ def test_get_allowed_members(as_member_list, ids, return_list_value, result,
...
+def teste_handle_start_after(autosubmit_helper, mock_experiment_history,
+                             mocked_sleep, autosubmit_config: Callable, time: str,
+                             header_skip: bool, experiment_exists: bool):
+    """
+    function to test the function handle_start_time inside autosubmit_helper
+    """
+    expid = 'a000'
+    autosubmit_config(expid, experiment_data={})
+
+    experiment_history = experiment_history2 = MagicMock(spec='experiment_history')
```
kinow, line 95: This is creating a single mock object. Both instances are pointing to the same object.

In general, multiple assignments in a single line is not very common. I myself have trouble when I read code with that as I sometimes skip the assignments.

test/unit/test_autosubmit_helper.py
```diff
@@ -71,3 +70,44 @@ def test_get_allowed_members(as_member_list, ids, return_list_value, result,
...
+
+    experiment_history.total = experiment_history2.total = 2
+
+    mock_experiment_history.return_value.manager.get_experiment_run_dc_with_max_id.side_effect = \
+        [experiment_history, experiment_history2]
+    mock_experiment_history.return_value.is_header_ready.return_value = header_skip
+    autosubmit_helper.return_value = experiment_exists
+    mocked_sleep.return_value = 0
+
+    assert helper.handle_start_after(time, expid) is None
```
kinow, line 113: @VindeeR I think we can use the walrus (like in Pica-Pau/woodpecker) operator to simplify the code. 

The issue with `is None` is that the boolean operation and the while loop are not working as expected, which is why with mocks you need to check for None.

You can have some fun searching for articles on how Python treats truthy values. But from what I remember, Python uses either `__bool__` or `__len__`. And if none exist, then the object is considered `True` (Python objects fall in this last category). And your MagicMock's has no `__bool__` and `__len__` that returns `0`. Which means using MagicMock objects in an `if` condition return `False`. Which is why you had to use `is None`.

But changing the if statement to accommodate changes in tests is not correct, in most cases.

Take a look at this version (it's late already, it may have :bug: :sleepy: :sleeping_bed: )

```diff
diff --git a/autosubmit/helpers/autosubmit_helper.py b/autosubmit/helpers/autosubmit_helper.py
index 7d54c164..b9dafbce 100644
--- a/autosubmit/helpers/autosubmit_helper.py
+++ b/autosubmit/helpers/autosubmit_helper.py
@@ -89,17 +89,15 @@ def handle_start_after(start_after: str, expid: str) -> None:
         Log.info(
             "Autosubmit will start monitoring experiment {0}. When the number of completed jobs plus suspended jobs becomes equal to the total number of jobs of experiment {0}, experiment {1} will start. Querying every 60 seconds. Status format Completed/Queuing/Running/Suspended/Failed.".format(
                 start_after, expid))
-        while True:
+        while current_run := exp_history.manager.get_experiment_run_dc_with_max_id():
             # Query current run
-            current_run = exp_history.manager.get_experiment_run_dc_with_max_id()
-            if (current_run is not None and current_run.finish > 0 and
-                    0 < current_run.total == current_run.completed + current_run.suspended):
+            if current_run.finish > 0 and 0 < current_run.total == current_run.completed + current_run.suspended:
                 break
-            else:
-                sys.stdout.write(
-                    f"\rExperiment {start_after} ({current_run.total} total jobs) status {current_run.completed}/"
-                    f"{current_run.queuing}/{current_run.running}/{current_run.suspended}/{current_run.failed}")
-                sys.stdout.flush()
+
+            sys.stdout.write(
+                f"\rExperiment {start_after} ({current_run.total} total jobs) status {current_run.completed}/"
+                f"{current_run.queuing}/{current_run.running}/{current_run.suspended}/{current_run.failed}")
+            sys.stdout.flush()
             # Update every 60 seconds
             sleep(60)
 
diff --git a/test/unit/test_autosubmit_helper.py b/test/unit/test_autosubmit_helper.py
index 7f974e59..e91694e7 100644
--- a/test/unit/test_autosubmit_helper.py
+++ b/test/unit/test_autosubmit_helper.py
@@ -23,6 +23,8 @@ import datetime
 from datetime import timedelta
 
 from typing import Callable
+
+import mock
 import pytest
 from mock.mock import patch, MagicMock
 
@@ -92,21 +94,26 @@ def teste_handle_start_after(autosubmit_helper, mock_experiment_history,
     expid = 'a000'
     autosubmit_config(expid, experiment_data={})
 
-    experiment_history = experiment_history2 = MagicMock(spec='experiment_history')
+    experiment_history = MagicMock()
+    experiment_history2 = MagicMock()
+
+    for attr in ['finish', 'total', 'completed', 'suspended', 'queuing', 'running', 'failed']:
+        setattr(experiment_history, attr, 0)
+
+    experiment_history.total = 2
+    experiment_history.__bool__ = lambda _: True
 
-    experiment_history.finish = experiment_history.total = experiment_history.completed = (
-        experiment_history).suspended = experiment_history.queuing = experiment_history.running = (
-        experiment_history).failed = 0
+    for attr in ['finish', 'total', 'completed', 'suspended', 'queuing', 'running', 'failed']:
+        setattr(experiment_history2, attr, 1)
 
-    experiment_history2.finish = experiment_history2.total = experiment_history2.completed = (
-        experiment_history2).suspended = experiment_history2.queuing = experiment_history2.running \
-        = experiment_history2.failed = 1
+    experiment_history2.total = 2
+    experiment_history2.__bool__ = lambda _: True
 
-    experiment_history.total = experiment_history2.total = 2
+    mocked_exp_history = MagicMock()
+    mocked_exp_history.manager.get_experiment_run_dc_with_max_id.side_effect = [experiment_history, experiment_history2, None]
+    mocked_exp_history.is_header_ready.return_value = header_skip
+    mock_experiment_history.return_value = mocked_exp_history
 
-    mock_experiment_history.return_value.manager.get_experiment_run_dc_with_max_id.side_effect = \
-        [experiment_history, experiment_history2]
-    mock_experiment_history.return_value.is_header_ready.return_value = header_skip
     autosubmit_helper.return_value = experiment_exists
     mocked_sleep.return_value = 0

```

As an extra step, it would be nice if the test failed if the `while` loop was never called (at the moment if that happens, the test could succeed without alerting the dev, which is dangerous).

autosubmit/helpers/autosubmit_helper.py
```diff
@@ -84,29 +83,25 @@ def handle_start_after(start_after, expid, BasicConfig):
...
-                                                                                          current_run.suspended,
-                                                                                          current_run.failed))
-                sys.stdout.flush()
+            sys.stdout.write(
+                f"\rExperiment {start_after} ({current_run.total} total jobs) status {current_run.completed}/"
+                f"{current_run.queuing}/{current_run.running}/{current_run.suspended}/{current_run.failed}")
+            sys.stdout.flush()
             # Update every 60 seconds
             sleep(60)
+    return None
```
kinow, line 104: Not needed. A function that returns `None` can be just left without a return statement.

test/unit/test_autosubmit_helper.py
```diff
@@ -18,10 +18,11 @@
 """Test file for ``autosubmit_helper.py``."""
 
 import datetime
-import pytest
 from datetime import timedelta
 from typing import Callable
+import pytest
 
+from mock import MagicMock
```
kinow, line 25: You can use the `mocker` fixture directly and let pytest-mock control mocks (useful if they ever switch lib, then we don't have to chase what's using `pytest-mock` and what's using `mock`).

test/unit/test_autosubmit_helper.py
```diff
@@ -68,3 +69,51 @@ def test_get_allowed_members(mocker, ids, return_list_value, result,
     as_member_list.return_value = return_list_value
 
     assert helper.get_allowed_members(ids, as_config) == result
+
```
kinow, line 72: Two blank lines.

test/unit/test_autosubmit_helper.py
```diff
@@ -68,3 +69,51 @@ def test_get_allowed_members(mocker, ids, return_list_value, result,
...
+],ids=['expid instead of time','wrong format hours','right format hours','fulldate wrong format',
+       'fulldate wrong format false']
+)
+def teste_handle_start_after(mocker, autosubmit_config: Callable, time: str,
+                             header_skip: bool, experiment_exists: bool):
+    """
+    function to test the function handle_start_time inside autosubmit_helper
+    """
+    expid = 'a000'
+    autosubmit_config(expid, experiment_data = {})
```
kinow, line 88: No space between variable and dict here (only when declared as a variable out of a function call... except when there are types, I think).

autosubmit/helpers/autosubmit_helper.py
kinow: And can you check this again? This is not needed, and at least my IDE is not complaining. There is, however, in `test_autosubmit_helper.py` an assert checking that it's `None`. That assertion needs to be removed (a function that returns only `None` will return `None` forever, no need to check if in that test).

test/unit/test_autosubmit_helper.py
```diff
@@ -18,9 +18,9 @@
 """Test file for ``autosubmit_helper.py``."""
 
 import datetime
-import pytest
 from datetime import timedelta
 from typing import Callable
+import pytest
```
kinow, line 23: `pytest` is not stdlib, so I think it would be placed in one more line below (i.e. separated from stdlib imports and from autosubmit imports). I checked the other places where you changed imports, and my IDE produced no change when I optimized imports :clap: . Except for this one.

kinow (commented):
GH UI is showing the file `test/unit/test_run_command_intregation.py`. This was renamed because there was a typo in the name, and later moved to test/integration. I think it got added back here by accident during a git rebase, perhaps?

kinow (commented):
@VindeeR looking good! I left one comment about the variable that's already not `None` in the `while`, so that can be removed. After that, I can have a final look and approve it.

kinow (approved):

PR #2169, opened 2025-03-03
diagram seq and populate_statistics test

VindeeR:
- test for the function_seq, populate_statistics in diagram

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -19,21 +19,25 @@
 import os
 import pwd
 from dataclasses import dataclass
+from datetime import datetime
 from fileinput import FileInput
 from pathlib import Path
+from random import seed, randint, choice
 from re import sub
 from textwrap import dedent
 from time import time
-from typing import TYPE_CHECKING, Any, Dict, Callable, Protocol, Optional, Type
+from typing import TYPE_CHECKING, Any, Dict, Callable, Protocol, Optional, Type, List
 
 import pytest
+from autosubmitconfigparser.config.basicconfig import BasicConfig
+from autosubmitconfigparser.config.configcommon import AutosubmitConfig
 from pytest_mock import MockerFixture
 from ruamel.yaml import YAML
 
 from autosubmit.autosubmit import Autosubmit
+from autosubmit.job.job import Job
+from autosubmit.job.job_common import Status
 from autosubmit.platforms.slurmplatform import SlurmPlatform, ParamikoPlatform
-from autosubmitconfigparser.config.basicconfig import BasicConfig
-from autosubmitconfigparser.config.configcommon import AutosubmitConfig
 
 if TYPE_CHECKING:
     from py._path.local import LocalPath  # type: ignore
@@ -118,6 +122,7 @@ def autosubmit_exp(
 
     TODO: Use minimal to avoid having to juggle with the configuration files.
     """
+
     def _create_autosubmit_exp(
             expid: Optional[str] = None,
             experiment_data: Optional[Dict] = None,
@@ -438,7 +443,7 @@ PLATFORMS:
         """)
     # add a job of each platform type
     with jobs_path.open('w') as f:
-        f.write(f"""
+        f.write("""
 JOBS:
     base:
         SCRIPT: |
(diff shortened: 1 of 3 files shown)


test/unit/test_diagram.py
```diff
@@ -18,10 +18,34 @@
...
+from autosubmit.job.job import Job
+from autosubmit.monitor import diagram
 from autosubmit.monitor.diagram import JobData, JobAggData
 
+def test_seq():
+    """ function to test the Class JobData inside autosubmit/monitor/diagram.py """
+    for value in diagram._seq(100, 2, 10):
+        assert value is not None
+        assert isinstance(value, int)
+
```
kinow, line 33: Two empty lines

test/unit/test_diagram.py
```diff
@@ -18,10 +18,34 @@
...
+        Job('test', "a27z", "COMPLETED", 200),
+        Job('test', "a26z", "FAILED", 10)
+    ]
+
+    date_ini = datetime.datetime.now()
+    date_fin = date_ini + datetime.timedelta(0.10)
+    queue_time_fixes = ['test', 5]
+
+    statistics = diagram.populate_statistics(jobs_data,date_ini,date_fin, queue_time_fixes)
+    assert statistics is not None
```
kinow, line 48: I think we could test a bit more. Like whether the statistics were populated (i.e. calculated) correctly.

test/unit/test_diagram.py
```diff
@@ -46,6 +48,17 @@ def test_job_agg_data():
...
+def test_seq():
+    """
+    function to test the Class JobData inside autosubmit/monitor/diagram.py
+    """
+    for value in diagram._seq(100, 2, 10):
+        assert value is not None
+        assert isinstance(value, int)
+
+def test_populate_statistics(jobs):
+    """ function to test the Class JobData inside autosubmit/monitor/diagram.py """
```
kinow, line 60: This function is now empty? (I checked out the branch as this wasn't a draft PR :grimacing:

test/unit/test_diagram.py
```diff
@@ -46,6 +48,32 @@ def test_job_agg_data():
     assert job_agg.number_of_columns() == 6
 
 
+def test_seq():
+    """
+    function to test the Class JobData inside autosubmit/monitor/diagram.py
+    """
+    for value in diagram._seq(100, 2, 10):
```
kinow, line 55: Maybe test the `len` of the seq? In this case it must have `9` elements, `100, 110, 120, 130, 140, 150, 160, 170, 180`. This way, if `_seq` is broken and returns less/none elements, this test would fail (otherwise, the for would not enter, and the `assert`s would never be called -- so test wouldn't fail).

kinow (commented):
Needs to be rebased, comments about testing the length.

test/unit/test_diagram.py
```diff
@@ -46,6 +48,32 @@ def test_job_agg_data():
...
+
+def test_populate_statistics(jobs):
+    """ function to test the Class JobData inside autosubmit/monitor/diagram.py """
+
+    date_ini = datetime.datetime.now()
+    date_fin = date_ini + datetime.timedelta(10.10)
+    queue_time_fixes = {'test': 5,'test1': 50,'test2': 500,'test3': 5000,'test4': 50000}
+
+    statistics = diagram.populate_statistics(jobs,date_ini,date_fin, queue_time_fixes)
+    for job_stat in statistics.jobs_stat:
```
kinow, line 68: Check the length, so that we know the assert is called within the for loop.

test/unit/test_diagram.py
```diff
@@ -16,9 +16,11 @@
 
 # You should have received a copy of the GNU General Public License
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
-
-""" Test file for autosubmit/monitor/diagram.py """
+"""
+Test file for autosubmit/monitor/diagram.py
+"""
 import datetime
+from test.unit.test_statistics import jobs
```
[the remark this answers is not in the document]
VindeeR: Should we create a new issue for that?

test/unit/test_diagram.py
```diff
@@ -52,22 +52,30 @@ def test_seq():
     """
     function to test the Class JobData inside autosubmit/monitor/diagram.py
     """
-    for value in diagram._seq(100, 2, 10):
+    arr = diagram._seq(100, 2, 10)
+    length = 0
+    for value in arr:
+        len+=1
```
kinow: :point_up: 

```python
assert 9 == len([x for x in diagram._seq(100, 2, 10)])
```

test/unit/test_diagram.py
```diff
@@ -52,22 +52,30 @@ def test_seq():
...
 
 def test_populate_statistics(jobs):
     """ function to test the Class JobData inside autosubmit/monitor/diagram.py """
 
+    length = 0
     date_ini = datetime.datetime.now()
     date_fin = date_ini + datetime.timedelta(10.10)
     queue_time_fixes = {'test': 5,'test1': 50,'test2': 500,'test3': 5000,'test4': 50000}
 
     statistics = diagram.populate_statistics(jobs,date_ini,date_fin, queue_time_fixes)
```
kinow, line 72: Something like `assert len(statistics.jobs_stat) == 5`?

VindeeR:
@kinow I believe I have implemented the change you suggested to move the jobs from the `test_statistics` & `test_diagram` to the `test/conftest.py` Once this commit is approved I'll rebase and change the #2170

pytest.ini
```diff
@@ -3,7 +3,6 @@ addopts =
     --verbose
     --strict-markers
     --doctest-modules
-    -n auto --max-worker-restart 0 --dist loadgroup
```
kinow, line 6: Why did you change the pytest settings? I only ever change it when I need to debug, and I do it directly in pycharm (if you pass any setting there that overrides these)

test/unit/test_diagram.py
```diff
@@ -45,6 +47,37 @@ def test_job_agg_data():
     assert job_agg.number_of_columns() == 6
 
 
+def test_seq():
+    """
+    function to test the Class JobData inside autosubmit/monitor/diagram.py
+    """
+    arr = diagram._seq(100, 2, 10)
+    length = 0
```
kinow, line 55: You can drop the `length` variable and just use a list comprehension to convert the iterator into a list.

test/unit/test_diagram.py
```diff
@@ -45,6 +47,37 @@ def test_job_agg_data():
...
+    for value in arr:
+        length+=1
+        assert value is not None
+        assert isinstance(value, int)
```
kinow, line 59: Just `assert isinstance(value, int)` is enough. It verifies if the value is not `None`, and also instance of `int`, `$ python3 -c 'print(isinstance(None, int))'`.

test/unit/test_diagram.py
```diff
@@ -45,6 +47,37 @@ def test_job_agg_data():
...
+    assert length == 9
```
kinow, line 60: This `test_seq` can be simplified to:

```python
def test_seq():
    """Test for the ``_seq`` function."""
    seq = [x for x in _seq(100, 2, 10)]
    assert len(seq) == 9
    assert all([x for x in seq if isinstance(x, int)])
```

kinow (commented):
Sorry, will finish the review in the morning, only had time to have a quick look at a couple of PRs but now I have to look at some emails.

test/conftest.py
```diff
@@ -497,3 +503,68 @@ def as_db_sqlite(monkeypatch: pytest.MonkeyPatch, tmp_path: "LocalPath") -> Type
     monkeypatch.setattr(BasicConfig, "DB_PATH", str(tmp_path / "autosubmit.db"))
 
     return BasicConfig
+
+
+@pytest.fixture()
+def jobs(mocker, request) -> List[Job]:
```
kinow, line 509: If this returns a [factory fixture](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#factories-as-fixtures), then you won't have to hardcode the number of jobs. Useful if you have a test that is checking for `5`, and then someone else changes the number of jobs to `20`, and now they have to update the `assert len(jobs) == 5` to `== 20` in a few places (i.e. this makes the tests and their data/state independent from each other)

But this is going in the right direction!! A reusable fixture for jobs in our tests, something that was missing (quite sure later we will be able to simplify some tests using your fixture).

pytest.ini
VindeeR: oooopps committed this one by mistake, I always run the pytest, but this line breaks in my local so I have to remove it and restore before commit, I'll change it back

kinow:
Reviewed everything but this one :sweat_smile: Going to check about linters & test_expid with @VindeeR now, have lunch, but afternoons I'm switching to postgres as the deadline is approaching :+1:

test/conftest.py
kinow: @VindeeR I see your marked this as resolved, but the fixture is still not using the factory fixture pattern from pytest (e.g. autosubmit_config and autosubmit_exp). Did you look at this review feedback?

VindeeR: I had seen it but I thought it was to stop using hard coded numbers, and I thought it was just that, now I see I should apply the factory fixture too, sorry for the confusion

kinow (approved):
Have a look if you can rename that function, and then if GH Actions pass, LGTM.

kinow (approved):

PR #2170, opened 2025-03-03
Test create stats report

VindeeR:
Create a test case for the function create_stats_report asserting if functions is creating files properly with the information from the jobs

### DIFF
--- a/test/unit/test_diagram.py
+++ b/test/unit/test_diagram.py
@@ -17,7 +17,6 @@
 import datetime
 
 import pytest
-from mock.mock import patch
 
 from autosubmit.job.job import Job
 from autosubmit.monitor import diagram
@@ -70,6 +69,24 @@ def test_populate_statistics(create_jobs):
     assert statistics.failed_jobs_dict == {'example_name_0': 1, 'example_name_1': 1,
                                            'example_name_2': 1, 'example_name_3': 1,
                                            'example_name_4': 1}
+
+
+@pytest.mark.parametrize("create_jobs", [[5, 20]], indirect=True)
+def test_create_stats_report(create_jobs, tmp_path, mocker):
+    """ function to test the function create_stats_report inside autosubmit/monitor/diagram.py """
+
+    expid = "a000"
+    period_ini = datetime.datetime.now()
+    period_fi = period_ini + datetime.timedelta(10)
+    tmp_path_pdf = tmp_path / "report.pdf"
+    tmp_path_csv = tmp_path / "report.csv"
+
+    mocker.patch('autosubmit.monitor.diagram._create_table')
+    diagram.create_stats_report(expid, create_jobs, [], str(tmp_path_pdf), True, True, False, period_ini, period_fi, {
+        'test': 1, 'test1': 5, 'test2': 50, 'test3': 500, 'test4': 5000})
+    assert tmp_path.exists()
+    assert tmp_path_pdf.exists()
+    assert tmp_path_csv.exists()
 
 
 def test_create_csv_stats(tmpdir):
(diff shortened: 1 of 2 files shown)


test/unit/test_diagram.py
```diff
@@ -18,7 +18,15 @@
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
 """ Test file for autosubmit/monitor/diagram.py """
+
+import tempfile
 import datetime
+from pathlib import Path
+
+from mock.mock import patch
```
kinow, line 26: `tepmfile` and `patch` can be replaced by the built-in fixtures that come with pytest: https://docs.pytest.org/en/6.2.x/fixture.html#pytest-fixtures-explicit-modular-scalable

For `tempfile` there's `tmp_path` which returns a `Path` already for you (I think you may not need to import Path` here, maybe.

And for `patch`, the `pytest_mock` lib we use has `mocker`, https://pytest-mock.readthedocs.io/en/latest/usage.html

Both work the same, but with pytest fixture the code is a bit shorter (one of the main selling points of pytest over unittest was to write less to have the same -- no classes, just functions, no need to import a lot of things, just use fixtures, etc...)

test/unit/test_diagram.py
```diff
@@ -30,14 +38,33 @@ def test_job_data():
...
+    ]
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        if Path(temp_dir).exists() is False:
+            Path(temp_dir).mkdir(parents=True, mode=0o777)
+
+    jobs_by_association.return_value = [JobData(),JobAggData()]
+    create_status = diagram.create_stats_report(expid, jobs_data, {}, temp_dir,True,True,False,None,
+                                                None,None)
+    assert create_status is False
```
kinow, line 70: :laughing: this is not really fair. This is the function you mentioned that receives a `bool` and returns that same `bool`. So this test will pass whether we break the `create_stats_report` or not (the focus when writing tests must be to prevent bugs, not only coverage).

I think a better test would be one that either produces a plot somewhere in a temp dir, and then we assert the plot was created, has more than 0 bytes, and maybe that's that. Or then mock just the matplotlib part, and confirm the mock was called as expected.

isimo00 (commented):
I agree with Bruno's review and haven't seen anything else worth mentioning :)

test/unit/test_diagram.py
```diff
@@ -20,6 +20,7 @@
 """ Test file for autosubmit/monitor/diagram.py """
 import datetime
 
+from test.unit.test_statistics import jobs
```
kinow, line 23: Is `jobs` a fixture? If so, could it be moved to a common place (conftest.py)?

VindeeR: Yes, its a fixture in test_statistcs, there's a few other fixtures there, should all of them be moved to config?

test/unit/test_diagram.py
kinow: I've recently written a few tests for `stats` and `monitor` in #2322 , and there you can see the `mocker.patch` fixture being used.

test/unit/test_diagram.py
kinow: If the fixture will be shared by the two files, it's probably better to move it to either `test/conftest.py` or ` new `test/unit/conftest.py` (I think we can use the existing one).

The fixture uses the `NUM_JOBS` from `test_statistics`, set to `5`. Instead, you could return a function that receives a number `N` of jobs, and then returns an array filled with `N` jobs, then use it in statistics with `5` as before, and use it here too.

VindeeR:
Waiting for approval of the issue #2169 to refine this issue

test/unit/csv
kinow, line 1: Delete me :grimacing:

kinow (commented):
Remove CSV, use pytest-mock, and figure out what's creating the CSV file. Otherwise, looks good. Will approve once these tests are updated and GH Actions is passing.

VindeeR:
@kinow I used the new fixture to develop the test and a bit of tidiness of the code too

Thanks for the master class yesterday, I'm not sure I would manage this branch this easily without it :heart:

kinow (approved):
:boom:

kinow:
> @kinow I used the new fixture to develop the test and a bit of tidiness of the code too
> 
> Thanks for the master class yesterday, I'm not sure I would manage this branch this easily without it ❤️

It was fun!!! Thanks for the different Thursday!

Issue #2160, opened 2025-02-26
Memory leak in 4.1.11 and 4.1.12


kinow:
Breaking down the issue into a sub issue so that's easier to distinguish which PRs are fixing what.

See parent issue for more details.

kinow:
The example workflow we are using:

```yaml
TEST:
  JOB: &job
    SCRIPT: sleep 2
  PLATFORM: &platform
    USER: kinow
    PROJECT: ehpc01
# mkdir -p /tmp/scratch/ehpc01/kinow/

CONFIG:
  AUTOSUBMIT_VERSION: "4.1.12"
  TOTALJOBS: 20
  MAXWAITINGJOBS: 20
DEFAULT:
  EXPID: "a034"
  HPCARCH: "MARENOSTRUM5"
EXPERIMENT:
  DATELIST: 19900101
  MEMBERS: fc0
  CHUNKSIZEUNIT: month
  SPLITSIZEUNIT: day
  CHUNKSIZE: 1
  NUMCHUNKS: 1
  CALENDAR: standard
PROJECT:
  PROJECT_TYPE: local
  PROJECT_DESTINATION: local_project
LOCAL:
  PROJECT_PATH: ~/autosubmit/project
PLATFORMS:
  MARENOSTRUM5:
    <<: *platform
    TYPE: ps
    HOST: localhost
    QUEUE: gp_debug
    SCRATCH_DIR: /tmp/scratch
    ADD_PROJECT_TO_HOST: false
    MAX_WALLCLOCK: 02:00
    TEMP_DIR: ''
    MAX_PROCESSORS: 99999
    PROCESSORS_PER_NODE: 112
  MARENOSTRUM5-login:
    <<: *platform
    TYPE: ps
    HOST: localhost
    SCRATCH_DIR: /tmp/scratch
    ADD_PROJECT_TO_HOST: false
    MAX_WALLCLOCK: 02:00
    TEMP_DIR: ''
    MAX_PROCESSORS: 99999
JOBS:
  LOCAL_SETUP:
    <<: *job
    CHECK: on_submission
    PLATFORM: LOCAL
    RUNNING: once
  SYNCHRONIZE:
    <<: *job
    CHECK: on_submission
    DEPENDENCIES:
      LOCAL_SETUP: {}
    PLATFORM: LOCAL
    RUNNING: once
  REMOTE_SETUP:
    <<: *job
    CHECK: on_submission
    DEPENDENCIES:
[...]

PR #2157, opened 2025-02-26
GH-2122-Mem-issues-parameters-leak-soft-changes

dbeltrankyl:
This PR aims to reduce memory consumption and memory leaks. 

* Includes __slots__ branch and fixed all the tests related. 
* Packed PR was remade here ( need to move to another pr but is needed for the e-suite anyway) and fixed related tests
* Removes parameters from the job and changes according to calls
* TODO checks multiple loops for mem consumption and also leaks
* Changes job list and pickle load/save.

Still WIP

PR #2173, opened 2025-03-04
Import certain packages only when necessary to consume less memory (was 78.7 MB, is now 37.3 MB)

kinow:
I  used [Sciagraph](https://www.sciagraph.com/) to proflie Autosubmit for #2160 and found that that profiler was helpful actually in providing the detailed memory usage of the imports in our code.

To install and configure it, I ran `pip install sciagraph` and then did the account set up at https://account.sciagraph.com/ui/ (for the free version, they have paid plans too).

Sciagraph uses pydantic, which relies on typing_extensions >= 4.12, but we had a fixed range for ==4.9.0. I replaced it by `==4.*` as I think the typing extensions shouldn't break between major/minor updates for us.

Then I launched sciagraph with this command:

```bash
$ autosubmit create -v -f -np...
$ python -m sciagraph run /home/bdepaula/mambaforge/envs/autosubmit4/bin/autosubmit -lc DEBUG run a00g
```

I let it run a few iterations, like up to `90 out of 96 jobs` or earlier, and then `CTRL+C`'ed it. See below some screenshots from Sciagraph.

Screenshot from 2025-03-04 12-38-22

Screenshot from 2025-03-04 12-39-24

As you can see, `autosubmit run` had a peak of 78.7 MB of memory (internal). Within that amount, `20MB` of memory allocated when `autosubmit.monitor` is imported. `10MB` are from matplotlib.

In the image below, you can see `PathPatch` from `matplotlib` bringing another `13MB`. This is from the `locplatform`, for local jobs. 

Screenshot from 2025-03-04 12-42-22

But the interesting part is that `PathPatch` is never actually used in `locplatform`. An unused import brings `13MB` of memory unnecessarily -- this could have been fixed by fixing our unused imports with a linter like pylint/ruff/pycharm/etc..

I separated in commits what I did, so that it'd be easier to review it (I can squash everything to a single commit, or two if we want to leave the dependency change in a separate commit). To recap, here's how the profiler found `master`:

image

Removing the imports for `Monitor`, `requests`, and `matplotlib` and moving them to only the functions that actually where `Monitor` is used, went from 78.7 to 43.8 (-34.9):

image

Then some other smaller imports that were bringing `distutils` and importing all `networkx as nx` and also `from networkx.nx import DiGraph` (can be simplified) it went down to 37.3 (-6.5 MB).

image

Overall that's 37.3 MB instead of 78.7 MB, so less than half of the initial memory.

I used `memory_profiler` to run a test experiment #2160 for `240` seconds. This is the plog using `master` early this morning for #2157 :

image

And this is running the same test with this branch:

Figure_1

Plotted together to make comparing them easier. Black is `master`, blue this branch:

Figure_1

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -22,13 +22,13 @@ import platform
 import threading
 from bscearth.utils.date import date2str
 from configparser import ConfigParser
-from distutils.util import strtobool
 from pathlib import Path
 from ruamel.yaml import YAML
 from typing import Dict, Set, Tuple, Union, Any, List, Optional
 
 from autosubmit.database.db_common import update_experiment_descrip_version
 from autosubmit.helpers.parameters import PARAMETERS
+from autosubmit.helpers.utils import strtobool
 from autosubmitconfigparser.config.basicconfig import BasicConfig
 from autosubmitconfigparser.config.configcommon import AutosubmitConfig
 from autosubmitconfigparser.config.yamlparser import YAMLParserFactory
@@ -48,7 +48,6 @@ from .job.job_list_persistence import JobListPersistencePkl
 from .job.job_package_persistence import JobPackagePersistence
 from .job.job_packager import JobPackager
 from .job.job_utils import SubJob, SubJobManager
-from .profiler.profiler import Profiler
 from .notifications.mail_notifier import MailNotifier
 from .notifications.notifier import Notifier
 from .platforms.paramiko_submitter import ParamikoSubmitter
@@ -2153,6 +2152,7 @@ class Autosubmit:
         """
         # Start profiling if the flag has been used
         if profile:
+            from .profiler.profiler import Profiler
             profiler = Profiler(expid)
             profiler.start()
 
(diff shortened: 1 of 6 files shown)


kinow:
@dbeltrankyl the PR I mentioned in your pull request. I **think** this could improve the numbers in the spawn in case any of these imports are included when it spawns the new interpreter. Got the idea from one of your comments from last week, and by chance found this info super accessible when running Sciagraph (which I'm now testing for your PR too).

Will wait for GH Actions to see whether the build passes, and the coverage too.

dbeltrankyl:
Hello @kinow 

Thanks for the improvement; there is not much to comment on, so feel free to merge it for me.

 About the coverage, I mean, they're only imports, and some functions here are touched by other PR in progress

kinow:
> Hello @kinow
> 
> Thanks for the improvement; there is not much to comment on, so feel free to merge it for me.
> 
> About the coverage, I mean, they're only imports, and some functions here are touched by other PR in progress

I almost started writing a test for `autosubmit create` but the plot part looks a bit tricky to test. For now I added a test to the `strtobool` that I copied from `distutils` to save memory. If GH Actions build succeeds it should be ready for review. Thanks!

kinow:
was 

image

is now

image

dbeltrankyl (approved):
All good

kinow:
Thank you, Dani!

PR #2183, opened 2025-03-06
Autosubmit 100% trusted vulture

VindeeR:
all tests were executed and approved
too many lint errors to adjust in a single commit

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -97,23 +97,17 @@ sys.path.insert(0, os.path.abspath('.'))
 # noinspection PyUnusedLocal
 
 
-def signal_handler(signal_received, frame):
+def signal_handler():
     """
     Used to handle interrupt signals, allowing autosubmit to clean before exit
-
-    :param signal_received:
-    :param frame:
     """
     Log.info('Autosubmit will interrupt at the next safe occasion')
     Autosubmit.exit = True
 
 
-def signal_handler_create(signal_received, frame):
+def signal_handler_create():
     """
     Used to handle KeyboardInterrupt signals while the create method is being executed
-
-    :param signal_received:
-    :param frame:
     """
     raise AutosubmitCritical(
         'Autosubmit has been closed in an unexpected way. Killed or control + c.', 7010)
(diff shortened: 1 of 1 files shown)


autosubmit/autosubmit.py
```diff
@@ -97,23 +97,17 @@
...
-
-    :param signal_received:
-    :param frame:
     """
     Log.info('Autosubmit will interrupt at the next safe occasion')
     Autosubmit.exit = True
 
 
-def signal_handler_create(signal_received, frame):
+def signal_handler_create():
```
kinow, line 108: These args need to be here. They are used by the `signal.signal`, https://docs.python.org/3/library/signal.html#signal.signal

image

This means that if we remove it, we have something akin to:

image

Which causes a runtime error. So this needs to stay. We need to mark it as ignored by linters :+1: (PyCharm warns about it...)

autosubmit/autosubmit.py
```diff
@@ -97,23 +97,17 @@
 # noinspection PyUnusedLocal
 
 
-def signal_handler(signal_received, frame):
```
kinow, line 100: This needs to stay.

VindeeR:
As vulture was accusing 100% not being used, I removed the variables, but now I'm adding an fix to ignore this variables, we can create a whitelist as well to keep all the unused variable in the same place, and avoiding the pollution of the code.

I'll create an issue for the creation of a whitelist as recommended by vulture

autosubmit/autosubmit.py
```diff
@@ -97,23 +97,29 @@
...
+def signal_handler(signal_received, frame): # noqa: F841
```
kinow, line 100: Two spaces between `:` and the `#` :grimacing:

autosubmit/autosubmit.py
```diff
@@ -97,23 +97,29 @@
...
     :param signal_received:
     :param frame:
+    :return:
     """
     Log.info('Autosubmit will interrupt at the next safe occasion')
     Autosubmit.exit = True
 
 
-def signal_handler_create(signal_received, frame):
+def signal_handler_create(signal_received, frame): # noqa: F841
```
kinow, line 114: Ditto here

kinow (commented):
Nearly there! Two warnings added in PyCharm with these changes... but the exclusion you added seems to fix the rest of the warnings in PyCharm! TIL how to exclude those warnings, thanks!

PR #2186, opened 2025-03-06
Adjusting to remove unused varible in history_db_manager

VindeeR:


autosubmit/history/database_managers/experiment_history_db_manager.py
```diff
@@ -450,7 +450,7 @@ def delete_experiment_run(self, run_id):...
 
 
 def create_experiment_history_db_manager(
-    db_engine: str, **options: Any
```
kinow, line 453: Ditto here, I believe this was intentional. We need to ignore this.

Issue #1352, opened 2024-07-08
[traceability] Fully versioned workflow configuration


kinow:
User feedback from DestinE, but I had also noticed this after working on the traceability deliverables for DestinE.

We explain that users can version-control their workflow configuration, but that's not entirely true. If you use `minimal.yml`, you can have configuration there that's not under version control.

Or, if you use a `minimal.yml` and then load a flag-toggling file (like `main.yml` in DestinE) that bootstraps files in `proj/.../conf`, then that flag-toggling file is also not under version control.

Compare this to Cylc or ecFlow:

- In Cylc you have a Git project that you clone somewhere, and then perform a `cylc install $folder`. The Cylc workflow configuration matches what you have in your Git repository.
- In ecFlow you have a Git project that you clone somewhere and contains the scripts to create the suite, or a PyFlow script that creates the suite. Here, also, you have the configuration created from the Git repository.

The Autosubmit configuration templates from autosubmit4-config-parser introduce a break in traceability, and we must document it, and try to improve this.

PR #2187, opened 2025-03-06
WIP: New Postgres support

LuiggiTenorioK:


kinow:
@LuiggiTenorioK I closed the other branch, without deleting the branch. That was the idea with this new PR, right? Just waiting for you to confirm before I jump in and start looking at the git conflicts here. Cheers

LuiggiTenorioK:
> @LuiggiTenorioK I closed the other branch, without deleting the branch. That was the idea with this new PR, right? Just waiting for you to confirm before I jump in and start looking at the git conflicts here. Cheers

That's right! Good luck with the conflicts :four_leaf_clover:

kinow:
Thanks @LuiggiTenorioK !

kinow:
Once #2107 is merged I will rebase this one and start working on the conflicts, if any. :+1:

LuiggiTenorioK:
Conflicts solved! However, I think there will be concurrency issues by running the tests on many workers :face_with_spiral_eyes:

kinow:
I can take a look at these test issues tomorrow :+1: thanks for fixing the conflicts!

kinow:
Fixed one test, then the integration test run command started failing locally, so I started reviewing what was going on and merging the `autosubmit_config` fixture with the fixture/setup code in that integration code. Work-in-progress here: https://github.com/BSC-ES/autosubmit/pull/2250

kinow:
Rebased after #2307 , carefully dropping 100+ commits. If GH Actions succeed then I'll rebase the sibling #2250

kinow:
The tests were passing locally when executed individually. I think @LuiggiTenorioK mentioned some time ago that there was a possibility of some race condition between the two PG tests we have at the moment. So I bit the bullet and implemented part of #2250 here, using TestContainers' Postgres context-manager in a fixture, launching a container with Postgres for each test, using a random port, updating the connection string URL, and running in parallel.

The tests in test_db_common are a bit slow, so after fixing the pipeline I'll investigate if we can speed that up in #2250 .

kinow:
I think we have all branches merged, tests updated. If GH Actions build pass, the next steps here will be to restart the code review (to see if we are missing tests for any part of it, or if we forgot any table), and manual tests with real experiments. cc @mcastril & @f-macchia

LuiggiTenorioK:
Seems that the parallel execution of the tests makes the `multiprocessing.Queue` stale longer due to CPU overload. GitHub workers seem to have only 2 physical cores, and since we distribute the tests, maybe one worker is consuming the resources of the other one when using `multiprocessing.Process`.

I increased the timeout for now to mitigate the issue.

LuiggiTenorioK:
Added the support to the recent code I merged to master to update the experiment details on `autosubmit expid` and `autosubmit delete`.

Unfortunately, the timeout is still persisting, so I fixed the number of pytest xdist workers to 1. If this problem also happens in other branches, then we should apply that patch to master directly :cry:

autosubmit/experiment/detail_updater.py
```diff
@@ -120,14 +165,80 @@ def delete_details(self, exp_id: int):
...
+class ExperimentDetailsSQLAlchemyRepository(ExperimentDetailsRepository):
+    """
+    Class to manage the experiment details in a SQLAlchemy database.
+    This class is responsible for creating the database, creating the
+    table, and providing methods to insert, update, delete, and retrieve
+    experiment details.
+    """
+
+    def __init__(self):
+        self.table: Table = get_table_from_name(schema='', table_name='details')
```
kinow, line 177: Note: the `schema` was missing here. Need to test that an empty schema works too (or does it need to be `None`?)

LuiggiTenorioK: It needs to be `None` or SQLAlchemy will append it wrong (I think it from looking at the source code that it does a ìf schema is not None` instead of `if schema`)

kinow:
Still going through the other branch, and comparing `master` with the current code. By the end of the week I hope to have all changes from `master` & from the old PG branch either incorporated or at least mapped what's missing. Then over next days work will be to add what's missing and test, test, test.

kinow:
List of files to be updated is complete now, I think. Just need to go through the pending ones looking at `master` and then start over looking at the old branch. Maaaaybe it will be over tomorrow? Or by Wed :crossed_fingers:

kinow:
Finished the list of what's missing from the old branch `postgres-support`:

- autosubmit/history/database_managers/experiment_history_db_manager.py
- autosubmit/history/database_managers/experiment_status_db_manager.py
- autosubmit/job/job_list_persistence.py
- test/unit/test_history_db_manager.py
- test/unit/test_job_list_persistence.py
- test/unit/test_status_db_manager.py

Now :coffee: and then bring that old code to this new branch.

autosubmit/experiment/detail_updater.py
kinow: Done! Set the type to `Optional[str]` so we/others will know that can be `None` :+1:

kinow:
Finished comparing the current code with `master` and with the old `postgres-support` branch, bringing the code & tests for history package and some for other parts of the code like `autosubmit.py` (some boring code, thanks for writing it, @LuiggiTenorioK !).

Not sure if the tests are passing & stable, but I will continue trudging through the remaining pending files, implementing the code in sqlalchemy, and then will start running some manual tests, to then start ramping up the coverage here with more pytests being added.

:wave:

test/integration/experiment/test_detail_updater.py
```diff
@@ -1,11 +1,35 @@
...
-    ExperimentDetailsRepository,
+    create_experiment_details_repository,
 )
 
 
-def test_details_properties():
+def test_details_properties(mocker):
+    # NOTE: mocked create_experiment_details_repository as it fails intermittently with
+    #       sqlite3.OperationalError: unable to open database file
+    mocker.patch('autosubmit.experiment.detail_updater.create_experiment_details_repository')
```
kinow, line 32: Note to self: this bug is hitting others. @VindeeR found this in some of his builds. I fixed it by mocking the class, although I am not 100% sure why that's happening. FYI @dbeltrankyl , @LuiggiTenorioK , @isimo00

kinow: Will create a small PR for this later.

LuiggiTenorioK: Thanks @kinow for the fix! I didn't think about that in that moment.

It makes sense that it could trigger DB locks since the tests run in parallel. I won't be surprised that some other tests that do some side processing like this one will fail in the future :sweat_smile:

kinow:
Huh, something interesting. The fixtures check the DB_PATH and then calls configure/install. But with Postgres we won't have the db_path, so better to do it some other way (DB_PATH with sqlite, but if it's postgres then just configure + install)

kinow:
Fixed conflicts, rebased, organized commits, added one more test, fixed something small in the sqlalchemy code... and then the tests stopped working :grimacing: Went to have lunch, came back, and couldn't find what's wrong. Will continue on Monday, aiming at fixing the build and then restarting the work to write the remaining tests :+1:

kinow:
Build fixed. Our conftest had a few issues that only appeared in the most recent integration tests. The reason why I couldn't fix it last week was that it was passing locally but not on GH Actions, and that was caused due to the test not loading the properties correctly and (only that one test :angry:) creating `~/autosubmit`. After some debugging, managed to fix the conftest settings, and now 6 out of 10 files pending to be tested :slightly_smiling_face: Will continue going through them tomorrow, hoping to finish by the start of next week or earlier :running_man:

kinow:
Five files to go now! Now I need a break :sweat_smile: :coffee: :sleeping:

kinow:
I started writing tests for `db_structure` and `db_manager` this morning. Tests that would call the real code (no mocking).

Doing so, I realized many of the functoins in `db_manager` were not used. Not sure if this has always been like that, but could be due to the cleanup that @VindeeR did some days ago, allowing us to see what's not being used.

I started removing those functions, and what was left were functions to create tables, delete tables, add data, delete data. These functions had no business logic (i.e. they were really just doing DB operations).

SQLAlchemy can do all that was being done in that class. We had two functions for each function. The old one, that I was writing tests for, and the new SQLAlchemy, which I would have to write tests. So I decided to keep just the SQLAlchemy since that works with both (fixing some TODO's and bad code design comments in tests :tada:).

I think the unit tests should pass, but the integration tests may require some further work.

dbeltrankyl:
> SQLAlchemy can do all that was being done in that class. We had two functions for each function. The old one, that I was writing tests for, and the new SQLAlchemy, which I would have to write tests. So I decided to keep just the SQLAlchemy since that works with both (fixing some TODO's and bad code design comments in tests 🎉).

Yup, I started the $expid/pkl/job_list.db  based on this branch, and I'm also using the SQLAlchemy for that ( doing the sqlite part only tho)

Are your changes safe to rebase? 

I also have a question: What is this table for?


```
JobPklTable = Table(
    "job_pkl",
    metadata_obj,
    Column("expid", String, primary_key=True),
    Column("pkl", LargeBinary),
    Column("modified", String),
)
```

dbeltrankyl:
> I think this one was to store the pkl as a BLOB inside the DB. We agree not to move that implementation to this branch since we are also working on the pkl to db https://github.com/BSC-ES/autosubmit/issues/2211, so we should remove that table from tables.py.

Ok, so I'll remove it in my branch ( that is doing that issue)

kinow:
> I also have a question: What is this table for?

I think this was created by @LuiggiTenorioK some time ago, to persist the pickle into the Postgres DB.

> I think this one was to store the pkl as a BLOB inside the DB. We agree not to move that implementation to this branch since we are also working on the pkl to db https://github.com/BSC-ES/autosubmit/issues/2211, so we should remove that table from tables.py.

I missed that discussion, sorry. No worries removing it :+1: 

I'll fix the tests either today (in the next 10 mins before calling it a day) or tomorrow morning. Then you can rebase it.

dbeltrankyl:
> I'll fix the tests either today (in the next 10 mins before calling it a day) or tomorrow morning. Then you can rebase it.

Thanks no rush tho, I have maaany things to modify 😓 not strictly related to db files

kinow:
Unit and integration tests passed locally for me, but they may still fail in the CICD pipeline, although I think if it needs any more changes and kicks to the build server, these should be minor changes. Tomorrow I will continue with the files missing tests, probably no more commits from me today :+1: :wave:

kinow:
We had a PRIMARY KEY on package name and expid in the new code, but the old code didn't (thanks for the tip on what was the problem, @LuiggiTenorioK). Quickly removed it and some tests that were failing passed locally. Let's see if GH Actions build passes.

Also, after removing some mocks with the latest ditch/removal of db_manager, a part of the code now failed (it was working!). That's because we were missing `wallclock` in a table :grimacing: We still have a few more tests to write before this is ready for manual testing...

dbeltrankyl:
> Some time ago I saw an issue with dark theme with some UI buttons. Maybe test the light theme to see if the same happens?

The same happens

> Or could you check if that happens in every function double backtick? We have that in a few more places... but if this is the only function is could be some weird character somewhere, or Git/linux identifying the file with a different encoding, etc.


It is only in these files. Older functions don't have the issue

kinow:
Four to go, db_manager pretty much covered :tada: 

image

kinow:
Build passing, I'm writing tests for `db_common.py` (yesterday after meetings/today). Will continue on Monday as I'm already over my weekly hours :wave:

kinow:
Whew, I was getting different coverage running locally. Searched our code, couldn't find anything strange, nor in our settings.

Searched ned's coverage repository for issues and changelog announcements, nada.

Then looked at pytest-cov, and they had `--dist load` in the page about xdist + coverage, https://pytest-cov.readthedocs.io/en/latest/xdist.html

Enabling that option and running everything locally gave me the correct coverage. This became an issue now as `db_common.py` uses multiprocessing in a lot of functions, and as I progressed writing tests, and debugging things locally to confirm coverage, the numbers on GH+codecov didn't match.

Should be fixed now :tada: We are looking quite good. Fixed a couple of issues today while writing tests, that would be annoying to fix with real experiments (mentioned that to Luiggi too). So quite happy to have some time to catch these bugs now, instead of after deploying it in EDITO -- where debugging is more challenging as we may need to have a similar environment, copy experiments, have access to models, etc...

test/unit/job/test_user_metrics.py
```diff
@@ -173,101 +192,78 @@ def test_metric_spec_load_invalid(metric_specs: Any):
...
 
-    assert metric_specs[1].name == "metric2"
-    assert metric_specs[1].filename == "file2"
-    assert metric_specs[1].selector.type == MetricSpecSelectorType.JSON
-    assert metric_specs[1].selector.key == ["key1", "key2", "key3"]
+    assert 'Invalid or missing metrics section' in str(cm.value)
 
 
-def test_process_metrics(disable_metric_repository):
+def test_process_metrics(disable_metric_repository, mocker):
```
LuiggiTenorioK, line 210:
```suggestion
def test_process_metrics(disable_metric_repository, mocker: MockerFixture):
```

I might add this to have the type hints

kinow: Applied, thanks!

kinow: Huh,

```
ERROR test/unit/job/test_user_metrics.py - NameError: name 'MockerFixture' is not defined
```

I think I will have to import that first. Let me fix that and squash the commits...

kinow:
> Need to port this change from @dbeltrankyl in the experiment manager https://github.com/BSC-ES/autosubmit/pull/2345

Done while writing the unit tests for experiment_history_db_manager. Test written for SQLAlchemy too. Now onto the final 5 unit tests for that db manager. What will be left is autosubmit.py, then the unit & integration tests for this branch will be done! :tada:

autosubmit/autosubmit.py
```diff
@@ -887,7 +889,10 @@ def _init_logs(args, console_level='INFO', log_level='DEBUG', expid='None'):
                 tmp_path = os.path.join(exp_path, BasicConfig.LOCAL_TMP_DIR)
                 aslogs_path = os.path.join(tmp_path, BasicConfig.LOCAL_ASLOG_DIR)
                 if not os.path.exists(exp_path):
-                    raise AutosubmitCritical("Experiment does not exist", 7012)
+                    if BasicConfig.DATABASE_BACKEND == 'sqlite':
+                        raise AutosubmitCritical("Experiment does not exist", 7012)
+                    else:
+                        os.mkdir(exp_path)
```
kinow, line 895: @LuiggiTenorioK it's really hard to write tests for this part of the code, but when I started doing this, I read the code in this for loop, and noticed (above, you have to click to expand the code :point_up: ) that we call `as_conf = AutosubmitConfig(expid, BasicConfig, YAMLParserFactory())` before, to create the `exp_path` used here.

That call to `AutosubmitConfig` fails before this one :grimacing: I'm finishing the test, and once that happens, I will have move the code around.

This is not due to Postgres. This is just a bug in the code, as this check would never ever happen, since `AutosubmitConfig` would fail first, so impossible for this to happen anyway.

https://github.com/BSC-ES/autosubmit-config-parser/blob/545bedac35707777f32d46be4972afaf6753da5d/autosubmitconfigparser/config/configcommon.py#L44-L81

image

(So even if annoying to write these tests, this is being really useful to fix issues like this in the code :+1: )

kinow: Confirmed...

```
        self.basic_config.read()
        if not Path(BasicConfig.LOCAL_ROOT_DIR, expid).exists():
>           raise IOError(f"Experiment {expid} does not exist")
E           OSError: Experiment 0000 does not exist

../../../micromamba/envs/autosubmit4/lib/python3.9/site-packages/autosubmitconfigparser/config/configcommon.py:60: OSError

Process finished with exit code 1
```

LuiggiTenorioK: Seems quite hard to test! It definitely will be easier to test and read that code if it is well modularized :sweat_smile:

kinow: The test failed again, as only having the experiment folder is not enough. `AutosubmitConfig` tries to load files from `<EXPID>/conf`, which is not found.

I isolated the code from `Autosubmit.expid` that creates the required folders, and moved it to `experiment_common` in a function `create_required_folders(expid: str, exp_path: Path)`.

That function is already fully tested by other tests, but now it's easy to write a test for that one too.

The code here now calls that function when the DB is Postgres, effectively replicating what happens in `autosubmit expid`. It's not perfect since AS configs are not 100% under version control #1352 , but it should be enough for now.

kinow: > Seems quite hard to test! It definitely will be easier to test and read that code if it is well modularized 😅

Definitely! Once postgres is out, I plan to finally break down that big file into entrypoint/console and move things to separate packages/modules. I think I can do that in August while people are travelling/on vacation. :crossed_fingers:

autosubmit/autosubmit.py
```diff
@@ -2892,7 +2899,7 @@ def statistics(expid, filter_type, filter_period, file_format, section_summary,
                                                                                                    job_list.get_job_list()])
             queue_time_fixes = {}
             if job_to_package:
-                current_table_structure = get_structure(expid, BasicConfig.STRUCTURES_DIR)
+                current_table_structure = get_structure(expid, Path(BasicConfig.STRUCTURES_DIR))
```
kinow, line 2902: Tried to cover this change, which is pretty simple, but I couldn't reach this part of the code. Dani can confirm later, but it looks like during the memory profiling changes, we removed the `.name` from job packages.

Everywhere that `job_package_persistence` would call `save`, which would get to this line in the code, is now protected by a `hasattr(obj, 'name')` which is `False` for all the job packages we have. This will be left without tests :+1:

kinow:
Waiting for CICD now, but I think unit & integration tests are done. It doesn't cover 100% of the code, but the parts not covered were either not covered before, or are code for error handling, or just too complicated to reach.

Writing the tests, existing and new code had to be fixed (due to bugs written while writing the SQLAlchemy port, or due to bugs that we had but were not being reported by users or not covered by tests).

I need a short break from coding just to reset my brain, so if the CICD pipeline passes, I will work on cleaning up my email/slack/mattermost inboxes, and check something that's pending on me for Severo Ochoa (and there's also a query from Outi about DestinE A1...) :wave:

autosubmit/job/job_list_persistence.py
```diff
@@ -93,94 +110,109 @@ def load(self, persistence_path, persistence_file):
...
-        options = {
-            "root_path": str(Path(BasicConfig.LOCAL_ROOT_DIR, expid, "pkl")),
-            "db_name": f"job_list_{expid}",
-            "db_version": self.VERSION,
-            "schema": expid
-        }
+    def __init__(self, expid):
         self.expid = expid
-        self.db_manager = create_db_manager(BasicConfig.DATABASE_BACKEND, **options)
+        database_file = Path(BasicConfig.LOCAL_ROOT_DIR, expid, 'pkl', f'job_list_{expid}.db')
```
LuiggiTenorioK, line 155: I think creating the DB file should not be possible in this case, since this class should be only used when using Postgres

dbeltrankyl: All related to the job_list will change in my branch anyway. ( in fact that file won't even exists)

dbeltrankyl: https://github.com/BSC-ES/autosubmit/blob/as-postgres-joblist/autosubmit/database/db_manager_job_list.py

I still didn't make the PR, as I may change some core stuff but you can take a glance there

LuiggiTenorioK:
@kinow I was doing some manual tests using Autosubmit and the API together. From what I saw, everything looks smooth  when: 

* Doing `autosubmit expid` a dummy experiment
* `autosubmit create`
* `autosubmit run`
* Same as above, building the user-defined metrics

But I saw some issues:

* Wrappers were built outside the experiment schema (fixed :heavy_check_mark: )
* The job pkl is in the public schema, but I think it is better to stay this way. wdyt?

kinow:
Yup, fixed. Also added a test to cover the only missing part in db_manager. Thanks @LuiggiTenorioK !

kinow:
Adding one more commit as I'm having issues installing the DB, and in the logs this is all I see:

```bash
(base) bdepaula@bsces107921:~/Development/python/workspace/autosubmit$ autosubmit -lc DEBUG install
Creating autosubmit Postgres database...
[ERROR] Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='DEBUG', command='install')
[ERROR] This is the experiment: <no expid provided> which had an issue with the command: <install> and it is currently using the Autosubmit Version: <no version found>.
 [CRITICAL] Database can not be created [eCode=7004]
```
With the new commit the exception message/text will be added, as the `-lc DEBUG` didn't bring anything else.

```bash
(base) bdepaula@bsces107921:~/Development/python/workspace/autosubmit$ autosubmit -lc DEBUG install
Creating autosubmit Postgres database...
[ERROR] Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='DEBUG', command='install')
[ERROR] This is the experiment: <no expid provided> which had an issue with the command: <install> and it is currently using the Autosubmit Version: <no version found>.
 [CRITICAL] Database can not be created: Can't load plugin: sqlalchemy.dialects:postgres [eCode=7004]
```

After seeing the error, quickly fixed s/postgres/postgresql.

kinow:
`configure` and `install` working as expected.

`configure` gives you the default configuration file, which you then have to modify to set the DB backend and connection URL. It also creates the required folders, all empty, and now SQLite file.

image

`install` creates the new tables in the public schema of the DB configured.

image

kinow:
To investigate later, I can see a few logs with errors for `details`,

```bash
2025-05-27 08:06:43.255 UTC [62] LOG:  checkpoint complete: wrote 4 buffers (0.0%); 0 WAL file(s) added, 0 removed, 0 recycled; write=0.308 s, sync=0.017 s, total=0.344 s; sync files=4, longest=0.015 s, average=0.005 s; distance=1 kB, estimate=213 kB; lsn=0/1979CE8, redo lsn=0/1979C90
2025-05-27 08:08:51.143 UTC [159] ERROR:  relation "details" does not exist at character 13
2025-05-27 08:08:51.143 UTC [159] STATEMENT:  DELETE FROM details WHERE details.exp_id = 1
2025-05-27 08:11:42.331 UTC [62] LOG:  checkpoint starting: time
```

autosubmit/database/db_structure.py
```diff
@@ -1,177 +1,107 @@
...
+    except Exception as exp:
+        Log.printlog("Get structure error: {0}".format(str(exp)), 6014)
         Log.debug(traceback.format_exc())
-        Log.warning("Error on Delete : {0}".format(str(type(e).__name__)))
+    return None
+
+
+def save_structure(graph: DiGraph, expid: str, structures_path: Optional[Path]):
+    """Save the experiment structure into the database."""
+    _check_structures_path(structures_path)
```
LuiggiTenorioK, line 89: It doesn't make sense to call `_check_structures_path` when the backend is Postgres, since the file shouldn't exist.

kinow:
Today I tested a dummy workflow, and auto-mhm using MN5, both using SQLite -- to confirm we didn't break anything. Haven't found anything caused by this branch, but found two issues (bug/enhancement), and a third one that had already been created by a user.

Tomorrow I will continue with a climatedt workflow experiment with SQLite, and then will switch to testing the examples from the docs. Planning to start on Kubernetes this week too with Luiggi.

kinow:
Update: ClimateDT apps (based on t061) worked locally just fine too :slightly_smiling_face: No regressions added in SQLite (so far) with the Postgres/SQLAlchemy changes.

For tomorrow: test the docs :tada:

kinow (commented):
Tested a few examples from the docs locally, using Postgres, and had no issues. Graphs created correctly, workflow ran (for a really long time in some cases) fine too. Left a review for @VindeeR as I used his branch from #2366 for the docs. Tomorrow I start looking at containers (@LuiggiTenorioK will deploy the AS-Demo updated image tomorrow), and checking with Luiggi how I can help with the helm chart.

dbeltrankyl:
Added my changes to a squashed commit so it is easy to identify what comes from my changes (5d662c543188295e686c695539bf75573659a6ad)

I'll now be working on this branch. Rebasing it ...

dbeltrankyl:
I unified my branch and performed a rebase to master.

I fixed issues with regression and integration.

There is one failing from autosubmitconfigparser that was previously not checked.  I think it is derived from a change from TYPE: sqlite/postgres to TYPE: pkl/db. It should be easy to fix.

There are currently 27 unit tests that are failing, but I have not investigated them yet.

I will continue working on this on Monday.

dbeltrankyl:
Also cc: @VindeeR maybe you see something in the logs since you worked on them ( the slurm ones)

dbeltrankyl:
* [ ] Check TODO'S* [ ] See what else can be covered https://app.codecov.io/github/BSC-ES/autosubmit/commit/a2cc5b67bf37c2df5ab57bf960d84271635d35a7 
* [ ]  Test run with wrappers, while fixing tests I touch some parts of the code, ( maybe the bug I had is fixed now) 
* [ ] Manual test of all commands

dbeltrankyl:
(at least vertical) Wrappers working under dynamic jobs loading.

That is possible through 

```def load_inner_jobs_by_section(self, inner_sections: list[str]) -> List[Job]:```

There are some improvements to be made there... but for now maybe it is good enough? 

```python
            # Temporally shallow load all jobs of this section
            # This is the basic working version, but it can be optimized
            # TODO: This can (temporally) potentially load a lot of jobs, consider optimizing
            # TODO: use MIN/MAX, WALLCLOCK SUM, etc to load only the necessary jobs
            # TODO: Another alternative is to store the "vertical level" of the database and do calculations based on that
```

Basically, the current approach what does is to shallow ( the db row in dict format) load every job of the WRAPPERS.WRAPPER.JOBS_IN_WRAPPER sections  of the wrapper section in order to check if it is wrapprable ( parents completed ). Right now I need to shallow load them to check the "section". 

* [ ] I'm not sure if this will work for all wrappers types, I'll see it when I add the pytest regarding that...

dbeltrankyl:
With that last part, working tests etc I think I have a functional branch. 

Next steps are:

* [ ] Add missing unit tests. 
* [ ] check other commands not related to create, monitor, run .
* [ ] Add new integration run tests that uses wrappers to check the dynamic load behavior https://github.com/BSC-ES/autosubmit/pull/2550 I'll wait for this one to be merged cc : @VindeeR

dbeltrankyl:
The rebase (from yesterday afternoon on master) is taking a while; it broke many tests, mainly due to conflicting changes to conftest.py. 

Some time ago (about a month ago), I noticed we had an issue with the BasicConfig not pointing to the correct file. Today, I'm fixing that because it was closely related to the issues I was facing.

I had a lot of issues with some tests mocking read() another not, others using an isolated function, some tests pointing to one name and others to a different name, postgres vs sqlite differences...

So, after the commit ( still on works), for every test that calls `_create_autosubmit_exp `or `_create_autosubmit_config `, you will be able to use `BasicConfig` directly to access things like DB paths, metadata paths, and everything else you can find in the `autosubmitrc` generated with `_initialize_autosubmitrc`

Also created a new method ( BasicConfig.generate_dirs() ) which is right now only used in the tests

Examples 

       ```
 db_manager = _create_db_manager(Path(tmp_path, 'tests.db'))
 -> db_manager = _create_db_manager(Path(BasicConfig.DB_PATH))
```

```
    job_data_db = run_tmpdir / f'metadata/data/job_data_{_EXPID}.db'
    autosubmit_db = Path(run_tmpdir, "tests.db")
```
    
    -> 
```
        job_data_db = Path(BasicConfig.JOBDATA_DIR) / f"job_data_{_EXPID}.db"
    autosubmit_db = Path(BasicConfig.DB_PATH)
    
```

**Comment by kinow** [comment]
@dbeltrankyl 

> Also created a new method ( BasicConfig.generate_dirs() ) which is right now only used in the tests

Maybe we can prepare a PR for `master` with this change, or maybe with other changes that are not specifically to Postgres, and get them into master/4.1.16?

That could simplify future rebases too, I think. Just in case you are aware of parts of the new code that could be included now (a lot of the improvements in the fixture code in autosubmit_exp/autosubmit_config was written while working on this PR).

**Comment by dbeltrankyl** [comment]
Sure, but there is stuff related to postgres / sqlite not sure if they will be 1:1 compatible

**Comment by dbeltrankyl** [comment]
27/08
* Fixed conflicts of 26/08 
* Improving test coverage... the main focus of today was job_list.db -related code should be covered 
* Also fixed an issue with dynamic loading of the wrapper not working properly ( detected when writing the test )

**Comment by dbeltrankyl** [comment]
Squashed, fixed an implementation issue with reload functionality, added more tests

File: autosubmit/database/db_manager_job_list.py [hunk 8]
```diff
@@ -0,0 +1,462 @@
...
+        self._FINAL_STATUSES = ['COMPLETED', 'FAILED']
+
+    def save_jobs(self, job_list: List[Job]) -> None:
+        """Save the job list to the database. Normally this will save the current active jobs."""
+        table: Table = get_table_from_name(schema=self.schema, table_name=JobsTable.name)
+        self.create_table(table.name)
+        persistent_data = [job.__getstate__(log_process=False) for job in job_list]
+
+        pkeys = ['name']
+        self.upsert_many(table.name, persistent_data, pkeys)
```
LuiggiTenorioK, line 55: @dbeltrankyl is it possible to also save the `created` and `modified` times for each job? The API uses this value sometimes, especially to know if the experiment is running. Also, having these values could significantly help us debug in the future.

dbeltrankyl: Hello @LuiggiTenorioK , 

The create and modified times are untouched, they're still on the job_data table

dbeltrankyl: this save_jobs is for the job_list itself, it never had create and modified but I can add it

LuiggiTenorioK: You are right! I don't know why I was thinking those values of the `job_data` were updated only at the end of the experiment run :sweat_smile:

dbeltrankyl: And about the edges, I'll rename some of the fields to be clearer but I don't like the new names either so if you have a better idea..

experiment_structure fields: 

e_from # edge_from
e_to # edge_to
status - rename_to -> target_status/ expected status ( this is when you put the STATUS: RUNNING) 
completed - rename_to-> is_edge_completed? 
from_step 
optional - rename_to-> allow_failure?... this is for running even if the expected_status of the edge is not fulfilled   

wdyt

LuiggiTenorioK: > this save_jobs is for the job_list itself ( which is now the jobs table, it never had create and modified but I can add it

Now that I think about it, even if we could use the ones in the `job_data` table, it could be handy to have them in the jobs, sections, and edges tables if it is not too much trouble. Eventually, we will also need a historical table for the sections and edges tables, like we have for the jobs. So, there is no problem adding it later.

dbeltrankyl: > ble, it could be handy to have them in the jobs, sections, and edges table

sure

dbeltrankyl: > I don't fully understand this feature. This means that the job in edge_to runs when the job from edge_from changes to STATUS: RUNNING, like in the example?

It is the https://autosubmit.readthedocs.io/en/latest/userguide/defining_workflows/index.html#start-conditions 

It had a rework in this branch, because the old way of doing it was incompatible with the pkl->db changes. 

> This means that the job in edge_to runs when the job from edge_from changes to STATUS: RUNNING, like in the example?

Yes, to clarify, it means the minimum status possible in which we consider the edge fulfilled ( so the child consider that parent "complete"

> completion_status

sounds good

> fail_ok 

I think this one is better

Thanks for your input!

LuiggiTenorioK: > Yes, to clarify, it means the minimum status possible in which we consider the edge fulfilled ( so the child considers that parent "completed")


So, it could be `min_trigger_status` maybe. Wdyt?

autosubmit/database/tables.py
```diff
@@ -0,0 +1,309 @@
...
+        Column("package_id", Integer, ForeignKey("{name}_info.id"), nullable=False, primary_key=True),
+        Column("package_name", String, ForeignKey(f"{name}_info.name"), nullable=False, primary_key=True),
+        Column("job_name", String, ForeignKey("jobs.name"), nullable=False, primary_key=True),
+        Column("timestamp", String, nullable=True),
+    )
+    return table_package_info, table_jobs_inside_wrapper
+
+
+WrapperInfoTable, WrapperJobsTable = create_wrapper_tables("wrappers", metadata_obj)
+PreviewWrapperInfoTable, PreviewWrapperJobsTable = create_wrapper_tables("preview_wrappers", metadata_obj)
```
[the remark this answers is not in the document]
dbeltrankyl: Yes, the columns are different tho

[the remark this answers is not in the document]
LuiggiTenorioK: Perfect! Just wanted to know to handle it correctly in the API

dbeltrankyl:
Can you add your experiment_data.yml? thanks I'll check

autosubmit/database/tables.py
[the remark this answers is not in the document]
dbeltrankyl: In fact, the preview_wrapper_jobs maybe shouldn't even exists but to don't complicate the changes I replicated the old behavior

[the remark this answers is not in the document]
LuiggiTenorioK: It could be a good option. Before, the GUI used to use the old `wrapper_job_package` table to render the Graph view visualization of the wrappers. Now, these wrappers will only be visible when the new `wrappers_jobs` is written as it doesn't differentiate a preview from the real wrappers as you mentioned.

dbeltrankyl:
> Great @dbeltrankyl ! Just to confirm, the plan is that this is the PR with postgres, to be reviewed, tested and merged for 4.1.16. Then another PR will have the joblist changes right?

https://github.com/BSC-ES/autosubmit/pull/2580#issuecomment-3248009641

autosubmit/job/job_utils.py
```diff
@@ -232,20 +234,18 @@ def get_job_package_code(expid: str, job_name: str) -> int:
...
-            packages = packages_wrapper if len(packages_wrapper) > len(packages_wrapper_plus) else packages_wrapper_plus
-            for exp, package_name, _job_name in packages:
-                if job_name == _job_name:
-                    code = int(package_name.split("_")[-3])
-                    return code
-    except Exception as e:
-        pass
+    basic_conf = BasicConfig()
+    basic_conf.read()
+    # TODO this is used to store something for the GUI/API? not sure what is this "code" @Luiggi
```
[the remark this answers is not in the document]
dbeltrankyl: If yes, why that code? because it looks random

If not, I'll remove it instead of adapting it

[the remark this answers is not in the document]
LuiggiTenorioK: From the meeting: It is not used.

dbeltrankyl:
Hello @kinow 

I've made changes to recovery to take advantage of the DB job list( still WIP). The thing is that it also overlaps with #2577, #2563

Now there are two modes of running recovery: 

online mode (default): 

- calls to the new added   `def get_completed_job_names(self) -> List[str]:` ( paramiko_platform)
 - Uses  `cmd = f"find {self.remote_log_dir} -maxdepth 1 -name '*_COMPLETED' -type f"`
 - and returns the job_names set
 - [wip] Handle compressed logs ( if they include _COMPLETED files (@LuiggiTenorioK ) ) 
 
- ( todo ) Update jobdata table if some job don't figure there # But I think this will be part of another PR 
- ( todo ) fallbacks to offline mode if no-force flag
 
 Offline mode:
 
  - [wip ] Uses jobdata table to get which jobs have been completed 
  - and returns the job_names set
  
 - After that, it loads all the job_list and sets them to COMPLETED if they figure in the job_names set
 - Finally, it checks if the COMPLETED jobs can be really completed ( any parent to waiting) and set them back to WAITING if that's the case
 - Updates ready jobs, etc.
 - [wip] Checks if any ready job is part of a wrapper or if it is running ( only online mode) 
 
 
- Moved the grouped stuff into the noplot ( this part is unnecessary but was easy to do)

dbeltrankyl:
* Now setstatus , recovery and stats are compatible with the db
* Added modified and created times (UTC) for jobstable.

I think that I'm only missing the report command.  ( and add unit tests for these commands)

PSA: Stats with wrappers is bugged in master

dbeltrankyl:
Hello @VindeeR , @kinow 

About the last rebase I have two questions ( slurm-container and docs) 

Slurm-container:

Apparently, I can't reserve more than 1 processor, do you know how we could add more than one processor? ( horizontal wrappers needs at least one for job ) 

The sbatch command fails due that ( -n 2 ) 


Second:

Docs build is falling with this message: 

sphinx.errors.SphinxWarning: Could not find latest image in /home/runner/work/autosubmit/autosubmit/docs/build/autosubmit/autosubmit/a000/plot

Which is strange because I don't recall conflicts in the docs 

 <details> 
 <summary> full log </summary>
 
```python
reading sources... [ 43%] troubleshooting/changelog
[app] emitting event: 'env-purge-doc'(<sphinx.environment.BuildEnvironment object at 0x7d5a7bcac2b0>, 'troubleshooting/changelog')
[app] emitting event: 'source-read'('troubleshooting/changelog', ['#########\nChangelog\n#########\n\nThis page shows the main changes
Searching for YAML configuration for experiment a000 in /home/dbeltran/AS/docs/source/troubleshooting
Copying /home/dbeltran/AS/docs/source/troubleshooting/code/jobs_new_dependencies.yml to /home/dbeltran/AS/docs/autosubmit/autosubmit/a000/conf/jobs_a000.yml
[app] emitting event: 'build-finished'(SphinxWarning('Could not find latest image in /home/dbeltran/AS/docs/autosubmit/autosubmit/a000/plo

  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/util/logging.py", line 266, in pending_warnings
    yield memhandler
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 318, in build
    updated_docnames = set(self.read())
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 425, in read
    self._read_serial(docnames)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 477, in _read_serial
    self.read_doc(docname)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 536, in read_doc
    publisher.publish()
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/core.py", line 234, in publish
    self.document = self.reader.read(self.source, self.parser,
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/io.py", line 106, in read
    self.parse()
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/readers/__init__.py", line 76, in parse
    self.parser.parse(self.input, document)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/parsers.py", line 83, in parse
    self.statemachine.run(inputlines, document, inliner=self.inliner)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 169, in run
    results = StateMachineWS.run(self, input_lines, input_offset,
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/statemachine.py", line 233, in run
    context, next_state, result = self.check_line(
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/statemachine.py", line 445, in check_line
    return method(match, context, next_state)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 3034, in text
    self.section(title.lstrip(), source, style, lineno + 1, messages)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 325, in section
    self.new_subsection(title, lineno, messages)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 391, in new_subsection
    newabsoffset = self.nested_parse(
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 279, in nested_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 195, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/statemachine.py", line 233, in run
    context, next_state, result = self.check_line(
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/statemachine.py", line 445, in check_line
    return method(match, context, next_state)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 2790, in underline
    self.section(title, source, style, lineno - 1, messages)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 325, in section
    self.new_subsection(title, lineno, messages)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 391, in new_subsection
    newabsoffset = self.nested_parse(
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 279, in nested_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 195, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/statemachine.py", line 233, in run
    context, next_state, result = self.check_line(
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/statemachine.py", line 445, in check_line
    return method(match, context, next_state)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 2359, in explicit_markup
    self.explicit_list(blank_finish)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 2384, in explicit_list
    newline_offset, blank_finish = self.nested_list_parse(
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 316, in nested_list_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 195, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/statemachine.py", line 233, in run
    context, next_state, result = self.check_line(
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/statemachine.py", line 445, in check_line
    return method(match, context, next_state)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 2662, in explicit_markup
    nodelist, blank_finish = self.explicit_construct(match)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 2369, in explicit_construct
    return method(self, expmatch)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 2106, in directive
    return self.run_directive(
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/docutils/parsers/rst/states.py", line 2156, in run_directive
    result = directive_instance.run()
  File "/home/dbeltran/AS/docs/source/ext/autosubmitfigure.py", line 61, in run
    _copy_rst_yaml_files_to_as_experiment(source_path, self.env.app.outdir.parent, configuration_name)
  File "/home/dbeltran/AS/docs/source/ext/autosubmitfigure.py", line 178, in _copy_rst_yaml_files_to_as_experiment
    copy(path_from, path_to)
  File "/usr/lib/python3.9/shutil.py", line 427, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.9/shutil.py", line 266, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/home/dbeltran/AS/docs/autosubmit/autosubmit/a000/conf/jobs_a000.yml'

  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/cmd/build.py", line 337, in build_main
    app.build(args.force_all, args.filenames)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/application.py", line 378, in build
    self.builder.build_update()
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 297, in build_update
    self.build(to_build,
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 318, in build
    updated_docnames = set(self.read())
  File "/usr/lib/python3.9/contextlib.py", line 137, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/util/logging.py", line 273, in pending_warnings
    memhandler.flushTo(logger)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/util/logging.py", line 238, in flushTo
    logger.handle(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 1599, in handle
    self.callHandlers(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 1661, in callHandlers
    hdlr.handle(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 948, in handle
    rv = self.filter(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 806, in filter
    result = f.filter(record)
  File "/home/dbeltran/.venv39/lib/python3.9/site-packages/sphinx/util/logging.py", line 473, in filter
    raise exc
sphinx.errors.SphinxWarning: Could not find latest image in /home/dbeltran/AS/docs/autosubmit/autosubmit/a000/plot
```

</details> 

Any clue of what the issue may be?

manuel-g-castro:
Hi @dbeltrankyl ,

> Apparently, I can't reserve more than 1 processor, do you know how we could add more than one processor? ( horizontal wrappers needs at least one for job )

You have two options here. 

1) You alter the file `/etc/slurm/slurm.conf`, line 36, where it is defined the characteristics of the node. Then restart the docker container. 

I have myself changed it to:

```
35 AccountingStorageType=accounting_storage/slurmdbd
36 NodeName=slurmctld CPUs=4 RealMemory=8192 State=UNKNOWN
37 SlurmdParameters=config_overrides
```

2) Or you build the image that I have worked to [add MPI support](https://github.com/manuel-g-castro/slurm-openssh-container/tree/change-to-debian-trixie). In this branch, besides MPI, the virtual node is defined with more cores. But it was also reported that it was using more resources for the same workload, hence it got blocked until we figure out why.

dbeltrankyl:
> Any clue? in my home PC it works flawlessly

Apparently, it was an issue with the Docker installed through Snap. I've installed it now through apt (docker.io) and it is working

dbeltrankyl:
Regarding the docs, I've rebased it now, and I'm still facing the same issue. @kinow, @VindeeR could you help me? Thanks!

kinow:
> Regarding the docs, I've rebased it now, and I'm still facing the same issue. @kinow, @VindeeR could you help me? Thanks!

On it! (just gotta get some coffee first)

kinow:
@dbeltrankyl looking at the logs, I think there are some errors hidden in the logs (not sure why they are not failing the build.

https://app.readthedocs.org/api/v2/build/29639481.txt

e.g. 

```bash
Building packages for WRAPPER_V
[ERROR] Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='DEBUG', command='create', expid='a000', noplot=False, hide=True, detail=False, output='png', group_by=None, expand=None, expand_status=None, notransitive=False, check_wrapper=True, update_version=True, profile=False, force=False)
[ERROR] This is the experiment: <a000> which had an issue with the command: <create> and it is currently using the Autosubmit Version: 4.1.15.
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/scripts/autosubmit.py", line 105, in main
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 789, in run_command
    return Autosubmit.create(args.expid, args.noplot, args.hide, args.output, args.group_by, args.expand,
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 4751, in create
    Autosubmit.generate_scripts_andor_wrappers(
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 1944, in generate_scripts_andor_wrappers
    Autosubmit.submit_ready_jobs(as_conf, job_list, platforms_to_test, True,
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 2713, in submit_ready_jobs
    packages_to_submit = packager.build_packages()
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/job/job_packager.py", line 619, in build_packages
    built_packages_tmp = self._build_vertical_packages(jobs, wrapper_limits, wrapper_info=current_info)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/job/job_packager.py", line 754, in _build_vertical_packages
    job_vertical_packager = JobPackagerVerticalMixed(dict_jobs, job, [job], "00:00",
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/job/job_packager.py", line 971, in __init__
    self.sorted_jobs = dict_jobs[date][member]
KeyError: datetime.datetime(2012, 1, 1, 0, 0)

 [CRITICAL] Unexpected error: datetime.datetime(2012, 1, 1, 0, 0).
More info at https://autosubmit.readthedocs.io/en/master/troubleshooting/error-codes.html</literal_block>
```

dbeltrankyl:
Thanks @kinow 

I'll check, is strange that error tho, I would expect to find it running AS normally or in the integrations test

dbeltrankyl:
* Added all wrapper tests

While adding these tests, I identified some issues with the overall check logic. I also noticed some inconsistencies when reloading from the database, which have now been fixed.

I believe we will no longer encounter the no `job.id` issue (cc @kinow) in this branch.

I've changed:

* Removed `get_complete_files`.

Replaced it with get_completed_job_names(self, job_names: Optional[list[str]] = None), which only fetches the remote.

* `check_job `(from WrapperJob)

Complete rework that reduced complexity and removed some no longer necessary functions.

* Changed the main loop to reduce complexity.

Now, `Autosubmit.check_wrappers(as_conf, job_list, platforms_to_test, expid)` only checks wrappers instead of also returning the non_wrapped jobs.

Additionally, now it relies on the` job_list.job_package_map` and `job_list.packages_dict` from `job_list `to obtain what is wrapped ( and running, since it only contains active wrappers) instead of looking for grouped job_ids 

* `check_alljobs`: For some reason, it was receiving a job_list that included `prev_status`, but the `prev_status `was never used. Now, it receives a job_list containing only the job objects.

* Db changes: 
 * Now wrapper.status is stored as "COMPLETED, FAILED..." instead of "0, 1, 2". 
 * Added a function to obtain all wrapper_ids.
 * Fixed an issue when loading/saving multiple wrappers at once.
 
* Fixed an issue with prev_status not taking into account the ready -> submitted change 

* TODO: Fix doc test

All these changes are tested through integration tests ( may not be shown as covered) 

@LuiggiTenorioK 

I want to ask one thing: 

```python
                        try:
                            # Track all jobs change for GUI
                            job_changes_tracker = dict()
                            for job in [job for job in job_list.get_job_list() if job.prev_status is not None and job.prev_status != job.status]:
                                job_changes_tracker[job.name] = (Status.VALUE_TO_KEY[job.prev_status], Status.VALUE_TO_KEY[job.status])

                            exp_history = Autosubmit.process_historical_data_iteration(job_list, job_changes_tracker,
                                                                                       expid)
``` 

                                                                                     
Do you know for what is this used?

kinow:
JobList work (i.e. move of pickle file from disk to DB, and improvements in performance, wrappers, more tests, etc.) moved to https://github.com/BSC-ES/autosubmit/pull/2619. I've dropped the other commits, leaving just the oldest.

Now will work through the tests and conflicts...

kinow:
It's rebased and building and passing unit tests locally for me. Tomorrow I will go through any integration or regression errors (build still running, but it's late now -- just wanted to have this ready for tomorrow's meeting :sleeping_bed: ).

dbeltrankyl:
Hello @kinow 

Thanks for doing it. Is it too late to request that the first commit remain untouched, right? :sweat_smile: since I also have in my branch that commit, they will now differ. 

It would be good if, just before and after doing the merge into master, I can rebase my branch and discard the changes made for the conflict resolution as they will differ in my branch

kinow:
I changed the scope of Docker containers (postgres, git, slurm) to be session scoped. Moved fixtures made for integration tests to `test/integration/contest.py`. Fixtures for unit tests only are now in `test/unit/conftest.py`. And the fixtures shared are in `test/conftest.py`.

Some fixtures as loaded automatically, like one that sets up the autosubmitrc for all tests, and others for postgres+docker in integration tests (thus, the separated `conftest.py`, to prevent integration test fixtures from starting containers when running just unit tests).

I noticed the integration tests were also triggering slurm tests, so I changed the mark filters in GitHub Actions, and now test-integration should really be running only postgres and other integration tests, without slurm/docker (which are executed in the test-slurm job).

On my laptop, everything runs **a lot faster**, integration tests complete in under 1 minute. In GH Actions, I think there's some Docker caching that's not happening, so it's taking between 3 and 6 minutes. The slurm too took 6 minutes in the last 2 builds I triggered here.

Tests seem stable.

I think if we find nothing wrong in a final review, this branch should be ready to either be merged to `master`, or at least to be used on Monday when we work together with @tiggi and @sigurbrynjar at the BSC (I believe we'll start after lunch, and after the session at the auditorium, by 2PM?).

@LuiggiTenorioK , @dbeltrankyl this is now ready for a review. I will send my approval, as the PR if from @LuiggiTenorioK . @LuiggiTenorioK you can just reply here saying it's OK to merge.

@mcastril @f-macchia, FYI.

LuiggiTenorioK:
Bug found when trying workflow with vertical wrapper :bug: :

```
  File "/home/ltenorio/projects/autosubmit/autosubmit/autosubmit.py", line 2302, in run_experiment
    job_list, submitter , exp_history, host , as_conf, platforms_to_test, packages_persistence, _ = Autosubmit.prepare_run(expid, notransitive, start_time, start_after, run_only_members)
  File "/home/ltenorio/projects/autosubmit/autosubmit/autosubmit.py", line 2175, in prepare_run
    os.chmod(os.path.join(BasicConfig.LOCAL_ROOT_DIR,
FileNotFoundError: [Errno 2] No such file or directory: '/home/ltenorio/autosubmit_pg/a000/pkl/job_packages_a000.db'

  File "/home/ltenorio/projects/autosubmit/autosubmit/scripts/autosubmit.py", line 105, in main
  File "/home/ltenorio/projects/autosubmit/autosubmit/autosubmit.py", line 747, in run_command
  File "/home/ltenorio/projects/autosubmit/autosubmit/autosubmit.py", line 2307, in run_experiment
    raise AutosubmitCritical("Error in run initialization", 7014, str(e))  # Changing default to 7014
autosubmit.log.log.AutosubmitCritical:  

Trace: [Errno 2] No such file or directory: '/home/ltenorio/autosubmit_pg/a000/pkl/job_packages_a000.db'
 [CRITICAL] Error in run initialization [eCode=7014]
```

It is in the `prepare_run`, and it seems that could be solved with an if. It is here https://github.com/BSC-ES/autosubmit/blob/1a2fcf481567ca61f14c2dddd042ff7349aac759/autosubmit/autosubmit.py#L2173-L2176

dbeltrankyl:
I see this warning

/appl/AS/4.1.16-dev-postgres-1a2fcf4/lib64/python3.9/site-packages/networkx/utils/backends.py:135: RuntimeWarning: networkx backend defined more than once: nx-loopback
  backends.update(_get_backends("networkx.backends"))
  
  Aside of it, the sqlite database run seems to work in the climate-dt

kinow:
Huh, hopefully that warning isn't anything serious. I've fixed the issues reported by @LuiggiTenorioK . Luiggi was right about the `if`, I think. We had another chmod in the same file but that was already protected with the if BasicConfig..... == 'sqlite': chmod`. I've replicated it to prevent the call when using Postgres.

LuiggiTenorioK:
From https://github.com/BSC-ES/autosubmit/issues/2627#issuecomment-3375991846, I added the details table to be created when using autosubmit install.

kinow:
Linter failed, then I executed this on my machine `ruff check $(git diff --name-only --cached -- '*.py')`. Fixed each issue, and I think it should pass now (BTW, some test-integration builds are finishing in under 3 minutes :D)

kinow:
I'm adding the `ExperimentStructureTable` to the list of tables to be created when the database is created. Same as @LuiggiTenorioK did with the `DetailsTable`. I'm also removing the call to create the table from `save_structure` and `get_structure` in the `db_structure.py` file, as that won't have to be performed (a `CREATE IF NOT EXISTS`) anymore.

PR #2189, opened 2025-03-06
status_db: Fixing indentation, lines too long, and import order

VindeeR:
- Import order fixed
- Indentation fixed
- Type hint fixed
- Docstring Fixed


kinow (approved):
Looks great! I saw `print_current_table` and remembered we noticed during the postgres work that that was not needed. I can't recall if it was used but could be removed, or if it was just not used.

After this gets merged, this class is a good candidate for a vulture with 100/90 threshold.

Thanks @VindeeR !

PR #2190, opened 2025-03-06
job_list: Adjusting to remove unused variable

VindeeR:
- Vulture not used variables fixed

### DIFF
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -22,7 +22,7 @@ import re
 import traceback
 from contextlib import suppress
 from shutil import move
-from typing import List, Dict, Tuple, Any, Optional
+from typing import List, Dict, Tuple, Any, Optional, Union
 from pathlib import Path
 
 from time import strftime, localtime, mktime
@@ -33,6 +33,7 @@ from networkx import DiGraph
 
 from autosubmitconfigparser.config.basicconfig import BasicConfig
 from autosubmitconfigparser.config.configcommon import AutosubmitConfig
+
 from log.log import AutosubmitCritical, AutosubmitError, Log
 from autosubmit.job.job_utils import transitive_reduction
 from autosubmit.helpers.data_transfer import JobRow
@@ -415,7 +416,7 @@ class JobList(object):
 
     def find_and_delete_redundant_relations(self, problematic_jobs: dict) -> None:
         """
-        Jobs with intrisic rules than can't be safelty not added without messing other workflows.
+        Jobs with intrinsic rules than can't be safely not added without messing other workflows.
         The graph will have the least amount of edges added as much as safely possible
         before this function.
         Structure:
@@ -575,7 +576,7 @@ class JobList(object):
         else:
             # value
             return [value_to_check]
-        ## values to return
+        # values to return
         if len(value_list) > 0:
             if start is None:
                 start = value_list[0]
@@ -883,7 +884,7 @@ class JobList(object):
         filters_to_apply = {}
         # Check if filter_from-filter_to relationship is set
         if relationships is not None and len(relationships) > 0:
-            # Look for a starting point, this can be if else becasue they're exclusive as a
+            # Look for a starting point, this can be if else because they're exclusive as a
             # DATE_FROM can't be in a MEMBER_FROM and so on
             if "DATES_FROM" in relationships:
                 filters_to_apply = self._check_dates(relationships, current_job)
@@ -1000,11 +1001,9 @@ class JobList(object):
         return distance
 
     def _calculate_natural_dependencies(self, dic_jobs, job, dependency, date, member, chunk, graph,
-                                        dependencies_keys_without_special_chars, distances_of_current_section,
-                                        distances_of_current_section_members,
-                                        key, dependencies_of_that_section,
-                                        chunk_list, date_list, member_list, special_dependencies, max_distance,
-                                        problematic_dependencies):
+                                        distances_of_current_section, key, dependencies_of_that_section,
+                                        chunk_list, date_list, member_list, special_dependencies,
+                                        max_distance, problematic_dependencies):
         """
         Calculate natural dependencies and add them to the graph if they're necessary.
         :param dic_jobs: JobList
@@ -1014,9 +1013,7 @@ class JobList(object):
         :param member: Member
         :param chunk: Chunk
         :param graph: Graph
-        :param dependencies_keys_without_special_chars: Dependencies of current job without special chars ( without SIM-10 -> SIM )
         :param distances_of_current_section: Distances of current section
-        :param distances_of_current_section_members: Distances of current section members
         :param key: Key
         :param dependencies_of_that_section: Dependencies of that section ( Dependencies of target parent )
         :param chunk_list: Chunk list
@@ -1117,12 +1114,10 @@ class JobList(object):
                 graph.add_edge(parent.name, job.name)
         return problematic_dependencies
 
-    def _calculate_filter_dependencies(self, filters_to_apply, dic_jobs, job, dependency, date, member, chunk, graph,
-                                       dependencies_keys_without_special_chars, distances_of_current_section,
-                                       distances_of_current_section_members,
-                                       key, dependencies_of_that_section,
-                                       chunk_list, date_list, member_list, special_dependencies,
-                                       problematic_dependencies):
+    def _calculate_filter_dependencies(self, filters_to_apply, dic_jobs, job, dependency, date,
+                                       member, chunk, graph, dependencies_keys_without_special_chars,
+                                       dependencies_of_that_section, chunk_list, date_list, member_list,
+                                       special_dependencies, problematic_dependencies):
         """
         Calculate dependencies that has any kind of filter set and add them to the graph if they're necessary.
         :param filters_to_apply: Filters to apply
@@ -1134,9 +1129,6 @@ class JobList(object):
         :param chunk: Chunk
         :param graph: Graph
         :param dependencies_keys_without_special_chars: Dependencies keys without special chars
-        :param distances_of_current_section: Distances of current section
-        :param distances_of_current_section_members: Distances of current section members
-        :param key: Key
         :param dependencies_of_that_section: Dependencies of that section
         :param chunk_list: Chunk list
         :param date_list: Date list
@@ -1144,8 +1136,8 @@ class JobList(object):
         :param special_dependencies: Special dependencies
         :param problematic_dependencies: Problematic dependencies
         :return:
-
         """
+
         all_none = True
         for filter_value in filters_to_apply.values():
             if str(filter_value).lower() != "none":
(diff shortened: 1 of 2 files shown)


autosubmit/job/job_list.py
```diff
@@ -2798,7 +2789,7 @@ def update_list(self, as_conf, store_change=True, fromSetStatus=False, submitter
                             parent.status == Status.COMPLETED or parent.status == Status.SKIPPED or parent.status == Status.FAILED]
                     if len(tmp2) == len(job.parents):
                         for parent in job.parents:
-                            if () and parent.status != Status.COMPLETED:
+                            if parent.status != Status.COMPLETED:
```
kinow, line 2792: Huh! Interesting! 

```python
kinow@ranma:~$ python3
Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] on linux
>>> if () and 1 > 0:
...   print('ok')
... 
>>> 
```

So `()` is treated as false. I guess that makes sense, as `()` is an empty tuple,

```py
>>> type(())
<class 'tuple'>
```

As this is changing the condition in this line, do you think we could get a test added to this PR, @VindeeR ? I think we discussed simplifying the pull requests, but this one looks a bit dangerous to merge without the test (not sure if the code is working as expected or will cause some issue to surface now that the `False` is out of the `if` statement).

VindeeR: I'll start working on it as soon as I finish my current fix

kinow (commented):
Can you rebase it, please, @VindeeR ?

kinow:
More conflicts :grimacing: 

image

test/unit/test_job_list.py
```diff
@@ -1,46 +1,42 @@
...
 from autosubmit.job.job_dict import DicJobs
 from autosubmit.job.job_list import JobList
 from autosubmit.job.job_list_persistence import JobListPersistencePkl
-from autosubmitconfigparser.config.yamlparser import YAMLParserFactory
-from log.log import AutosubmitCritical
 
 
-class TestJobList(TestCase):
-    def setUp(self):
+class TestJobList(TestCase) :
```
kinow, line 24: It's odd that we need to add spaces here. I thought it'd be the other way around? No space between `)` and `:`. Was it found via ruff?

test/unit/test_job_list.py
```diff
@@ -1,46 +1,42 @@
...
         self.experiment_id = 'random-id'
         self.as_conf = MagicMock()
         self.as_conf.expid = self.experiment_id
         self.as_conf.experiment_data = dict()
         self.as_conf.experiment_data["JOBS"] = dict()
         self.as_conf.jobs_data = self.as_conf.experiment_data["JOBS"]
         self.as_conf.experiment_data["PLATFORMS"] = dict()
-        parameters = {'fake-key': 'fake-value',
-                      'fake-key2': 'fake-value2'}
+        parameters = {'fake-key' : 'fake-value',
```
kinow, line 33: Same here, odd to see the space added, not removed.

VindeeR: I was having a problem with indentation which I sorted by using the PyCharm stylizing option and the space before the : was selected by mistake, issue fixed

autosubmit/job/job_list.py
```diff
@@ -1049,10 +1049,11 @@ def find_current_section(self, job_section, section, dic_jobs, distance, visited
...
 
-    def _calculate_natural_dependencies(self, dic_jobs, job, dependency, date, member, chunk,
-            graph, dependencies_keys_without_special_chars, distances_of_current_section,
-            distances_of_current_section_members, key, dependencies_of_that_section, chunk_list,
-            date_list, member_list, special_dependencies, max_distance, problematic_dependencies):
+
+    def _calculate_natural_dependencies(self, dic_jobs, job, dependency, date, member, chunk, graph,
+                                distances_of_current_section, key, dependencies_of_that_section,
+                                chunk_list, date_list, member_list, special_dependencies,
+                                max_distance, problematic_dependencies):
```
kinow, line 1056: Was the indentation fixed with a linter? PyCharm now complains about this. Please take a look at what PyCharm is showing in this PR (if you are not seeing the warnings we need to either adjust your PyCharm settings, or others have to adjust their settings to match yours).

image

autosubmit/job/job_list.py
```diff
@@ -1487,11 +1481,11 @@ def _manage_job_dependencies(self, dic_jobs, job, date_list, member_list, chunk_
                 dependencies_of_that_section.append(key_aux_stripped)
             # Adds the dependencies to the job, and if not possible, adds the job to the
             # problematic_dependencies
+            ## Adds the dependencies to the job, and if not possible, adds the job to the problematic_dependencies
```
kinow, line 1484: Single `#` instead of `##`. PyCharm warns about it... Please check that your PyCharm is properly configured :+1: 

image

autosubmit/job/job_list.py
```diff
@@ -1487,11 +1481,11 @@ def _manage_job_dependencies(self, dic_jobs, job, date_list, member_list, chunk_
...
             problematic_dependencies = self._calculate_natural_dependencies(dic_jobs, job,
-                dependency, date, member, chunk, graph, dependencies_keys_without_special_chars,
-                distances_of_current_section, distances_of_current_section_member, key,
-                dependencies_of_that_section, chunk_list, date_list, member_list,
-                special_dependencies, max_distance, problematic_dependencies)
+                        dependency, date, member, chunk, graph, distances_of_current_section, key,
+                        dependencies_of_that_section, chunk_list, date_list, member_list,
+                        special_dependencies, max_distance, problematic_dependencies)
```
kinow, line 1488: Same warning here about overly indentation. If it's ruff or pylint, we will need to prepare these changes later telling others to adjust their IDEs and use the linter :+1:

dbeltrankyl:
Not sure what is missing here, It is assigned to me but it was reviewed by @kinow :thinking:

autosubmit/job/job_list.py
```diff
@@ -3253,12 +3237,9 @@ def retrieve_times(status_code, name, tmp_path, make_exception=False, job_times=
         energy = 0
         seconds_queued = 0
         seconds_running = 0
-        queue_time = running_time = 0
         submit_time = datetime.timedelta()
         start_time = datetime.timedelta()
         finish_time = datetime.timedelta()
-        running_for_min = datetime.timedelta()
-        queuing_for_min = datetime.timedelta()
```
kinow, line 3261: I think these values have already been deleted elsewhere. Can you rebase it, please, @VindeeR ?

VindeeR:
I did a rebase and now everything works fine.
I too took it as an opportunity to learn how to manage git better leaving a clean branch without many commits, sorry if it looks ugly here @kinow but I think I have a good idea on how to manage, organize, and keep branches, and the rebase ended up breaking a few things 🤣

autosubmit/job/job_list.py
```diff
@@ -1134,18 +1129,16 @@ def _calculate_filter_dependencies(self, filters_to_apply, dic_jobs, job, depend
...
-        :param key: Key
         :param dependencies_of_that_section: Dependencies of that section
         :param chunk_list: Chunk list
         :param date_list: Date list
         :param member_list: Member list
         :param special_dependencies: Special dependencies
         :param problematic_dependencies: Problematic dependencies
-        :return:
 
+        :return:
```
kinow, line 1139: AFAIK there's no need to separate return statements (I think in Java that was a thing?). 

e.g. https://github.com/python/cpython/blob/08d7687094c6acb8c2ea1925a292a94ce1246c82/Lib/email/quoprimime.py#L84-L94

autosubmit/job/job_list.py
```diff
@@ -1377,14 +1370,17 @@ def _manage_job_dependencies(self, dic_jobs, job, date_list, member_list, chunk_
             if dic_jobs.as_conf.jobs_data.get(aux_key, {}).get("RUNNING", "once") == "chunk":
                 if aux_key in distances_of_current_section and distance > distances_of_current_section[aux_key]:
                     distances_of_current_section[aux_key] = distance
-            elif dic_jobs.as_conf.jobs_data.get(aux_key, {}).get("RUNNING", "once") == "member":
-                if aux_key in distances_of_current_section_member and distance > distances_of_current_section_member[aux_key]:
-                    distances_of_current_section_member[aux_key] = distance
+
+            elif dic_jobs.as_conf.jobs_data.get(aux_key, { }).get("RUNNING", "once") == "member":
```
kinow, line 1374: No need to have a blank space between `{}`. This actually adds a warning in PyCharm, 

image

autosubmit/job/job_list.py
```diff
@@ -1377,14 +1370,17 @@ def _manage_job_dependencies(self, dic_jobs, job, date_list, member_list, chunk_
...
+                if aux_key in distances_of_current_section_member and distance > distances_of_current_section_member[
+                    aux_key]:
+                    distances_of_current_section_member[aux_key]=distance
```
kinow, line 1377: This introduced a couple of warnings in PyCharm. First, the `aux_key]` in the next line shows as wrong indentation (the `[` normally implies the next item will come in the same column).

Maybe it'd be easier to simplify the `if` statement into a single one. Maybe something like `if distances_of_current_section_member.get(aux_key, False):`

And below, it's missing a space around `=`.

autosubmit/job/job_list.py
```diff
@@ -1377,14 +1370,17 @@ def _manage_job_dependencies(self, dic_jobs, job, date_list, member_list, chunk_
...
 
         # Process sections with special filters
-        sections_to_calculate = [key for key in dependencies_keys.keys() if key not in dependencies_to_del]
-        natural_sections = []
+        sections_to_calculate=[key for key in dependencies_keys.keys() if key not in dependencies_to_del]
+        natural_sections=[]
```
kinow, line 1381: Missing space. No space in args of functions without types, and spaces everywhere else, I think. Easier to trust IDE/linter. These two can be reverted.

autosubmit/job/job_list.py
kinow: Ah, I think I saw this code being changed somewhere else and got confused. Sorry.

kinow (commented):
Great job with Git!!! Some comments about code style. Thanks!

autosubmit/job/job_list.py
```diff
@@ -1377,14 +1369,16 @@ def _manage_job_dependencies(self, dic_jobs, job, date_list, member_list, chunk_
             if dic_jobs.as_conf.jobs_data.get(aux_key, {}).get("RUNNING", "once") == "chunk":
                 if aux_key in distances_of_current_section and distance > distances_of_current_section[aux_key]:
                     distances_of_current_section[aux_key] = distance
+
             elif dic_jobs.as_conf.jobs_data.get(aux_key, {}).get("RUNNING", "once") == "member":
-                if aux_key in distances_of_current_section_member and distance > distances_of_current_section_member[aux_key]:
+                if distances_of_current_section_member.get(aux_key, False) < distance:
```
kinow, line 1374: Ah, my bad! This would introduce a bug. I hadn't seen the `< distance` yesterday (risk of reviewing code in the afternoon without :coffee: , sorry!). We will have to keep the code as was. If you want to fix the column length, then this should do.

```py
                if (aux_key in distances_of_current_section_member and 
                        distance > distances_of_current_section_member[aux_key]):
```

Or breaking it at some other point.

kinow (commented):
Sorry, I almost introduced a bug :grimacing: :bug: 

The bot is not complaining about coverage because there are tests for the lines modified, and the other lines that did not have tests still do not have tests.

The function you modified, `retrieve_times`, is never tested. So we can't be sure if it will work after this change. So it might be safer to look at what's changed here and not covered and see if we can write a test or two for them.

The line where I almost introduced a bug is covered (line), but not fully (branch). So this bug would only appear during runtime. It'd be better if we could cover that with a test too :grimacing:

autosubmit/job/job_list.py
```diff
@@ -1000,11 +1000,10 @@ def find_current_section(self, job_section, section, dic_jobs, distance, visited
...
     def _calculate_natural_dependencies(self, dic_jobs, job, dependency, date, member, chunk, graph,
-                                        dependencies_keys_without_special_chars, distances_of_current_section,
-                                        distances_of_current_section_members,
-                                        key, dependencies_of_that_section,
-                                        chunk_list, date_list, member_list, special_dependencies, max_distance,
-                                        problematic_dependencies):
+                                        distances_of_current_section, key, dependencies_of_that_section,
+                                        chunk_list, date_list, member_list, special_dependencies,
+                                        max_distance, problematic_dependencies):
+
```
kinow, line 1006: Extra blank line (not important, docstrings will be rendered fine)

kinow (commented):
No more IDE warnings :clap: :clap: 

Missing a test (if doable), see previous comment: https://github.com/BSC-ES/autosubmit/pull/2190#pullrequestreview-2817212195

kinow (approved):
Looks good! Just need to wait for GH Actions to pass.

PR #2191, opened 2025-03-07
Functions from .__init__.py to scripts/autosubmit.py

### LINKED ISSUE #2182: Move autosubmit.__init__.py functions to autosubmit/scripts/autosubmit.py
Those functions appear to be used only for the entrypoint console script `autosubmit.scripts.autosubmit.py`. Moving them to that function reduces clutter in the code, avoid possible cyclic dependencies, and also issues importing `autosubmit` when that could conflict with a package or module.

isimo00:
We restructure by moving contents from `__init__.py` to `scripts/autosubmit.py`. This will also avoid circular imports.
Closes #2182


kinow (approved):
:boom: thank you!!

PR #2199, opened 2025-03-10
Add expid evaluation option

### LINKED ISSUE #2194: Add e-suite experiments as a new reserved experiment class
Autosubmit's EXPIDs are made up of 4 alphanumeric characters. The `expid` command generates a new EXPID, taking into account the last one created, starting by a000

However, there are some reserved EXPIDs for testing and operations:

```
  -op, --operational    creates a new experiment with operational experiment
                        id

  -t, --testcase        creates a new experiment with testcase experiment id
```

https://autosubmit.readthedocs.io/en/master/userguide/create/index.html

Operational EXPIDs start with `o` and test EXPIDs start with `t`. This has to be taken into account at the time of creating an operational or testing EXPID and at the time of creating any other EXPID, which should never start with `o` or `t`.

Now we need to add a new type of experiment to the family, for evaluation purposes (e-suite). 

The idea is to implement a new flag in the `expid` command, like (-ev, --evaluation) that would lead to an experiment starting by `e`. Therefore, non-evaluation experiments cannot start by `e`.

As a funny note, at the BSC the last experiment created is `a8pq` after many years... so we shouldn't break anything in production.

@isimo00 @kinow @dbeltrankyl

isimo00:
Add new type of experiment: evaluation

Closes #2194

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -234,6 +234,8 @@ class Autosubmit:
 
             group.add_argument('-op', '--operational', action='store_true',
                                help='creates a new experiment with operational experiment id')
+            group.add_argument('-ev', '--evalutation', action='store_true',
+                               help='creates a new experiment with evaluation experiment id')
             subparser.add_argument('-H', '--HPC', required=False, default="local",
                                    help='specifies the HPC to use for the experiment')
             subparser.add_argument('-d', '--description', type=str, required=True,
@@ -1337,7 +1339,7 @@ class Autosubmit:
                     f.write(content)
 
     @staticmethod
-    def expid(description, hpc="", copy_id='', dummy=False, minimal_configuration=False, git_repo="", git_branch="", git_as_conf="", operational=False,  testcase = False,use_local_minimal=False):
+    def expid(description, hpc="", copy_id='', dummy=False, minimal_configuration=False, git_repo="", git_branch="", git_as_conf="", operational=False,  testcase = False, evaluation = False, use_local_minimal=False):
         """
         Creates a new experiment for given HPC
         description: description of the experiment
@@ -1349,6 +1351,7 @@ class Autosubmit:
         git_branch: git branch to clone
         git_as_conf: path to as_conf file in git repository
         operational: if true, creates an operational experiment
+        evaluation: if true, creates an evaluation experiment
         local: Gets local minimal instead of git minimal
         """
         if use_local_minimal:
@@ -1372,10 +1375,10 @@ class Autosubmit:
                 if not os.path.exists(copy_id_folder):
                     raise AutosubmitCritical(
                         "Experiment {0} doesn't exists".format(copy_id), 7011)
-                exp_id = copy_experiment(copy_id, description, Autosubmit.autosubmit_version, testcase, operational)
+                exp_id = copy_experiment(copy_id, description, Autosubmit.autosubmit_version, testcase, operational, evaluation)
             else:
                 # Create a new experiment from scratch
-                exp_id = new_experiment(description, Autosubmit.autosubmit_version, testcase, operational)
+                exp_id = new_experiment(description, Autosubmit.autosubmit_version, testcase, operational, evaluation)
 
             if exp_id == '':
                 raise AutosubmitCritical("No expid", 7011)
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -234,7 +234,7 @@ class Autosubmit:
 
             group.add_argument('-op', '--operational', action='store_true',
                                help='creates a new experiment with operational experiment id')
-            group.add_argument('-ev', '--evalutation', action='store_true',
+            group.add_argument('-ev', '--evaluation', action='store_true',
                                help='creates a new experiment with evaluation experiment id')
             subparser.add_argument('-H', '--HPC', required=False, default="local",
                                    help='specifies the HPC to use for the experiment')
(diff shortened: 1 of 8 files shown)


autosubmit/autosubmit.py
```diff
@@ -234,6 +234,8 @@ def parse_args() -> Tuple[int, Optional[argparse.Namespace]]:
 
             group.add_argument('-op', '--operational', action='store_true',
                                help='creates a new experiment with operational experiment id')
+            group.add_argument('-ev', '--evalutation', action='store_true',
```
kinow, line 237: s/evalutation/evaluation

kinow (commented):
Code looking really good. Waiting for CICD and a typo before having another look and maybe trying the code.

isimo00:
Thanks @kinow! Did some very **minimal** testing but seems like it's working. Feel free to test and review!

```console
(as-dev) isimomun@bsces110916:~/repos/autosubmit$ autosubmit expid -H local -d "eval test" -ev
Autosubmit is running with 4.1.12
The new experiment "e000" has been registered.
Generating folder structure...
Experiment folder: /home/isimomun/autosubmit/e000
Generating config files...
Experiment e000 created
```

test/unit/test_expid.py
```diff
@@ -1,9 +1,10 @@
 import tempfile
 import os
+import pytest
```
kinow, line 3: Unused import.

test/unit/test_expid.py
```diff
@@ -57,6 +65,20 @@ def test_create_new_operational_experiment_with_previous_one(self, db_common_moc
...
+        self._build_db_mock(current_experiment_id, db_common_mock)
+        experiment_id = new_experiment(self.description, self.version, False, False, True)
+        self.assertEqual("e113", experiment_id)
+
+    @patch('autosubmit.experiment.experiment_common.db_common')
+    def test_copy_experiment_new(self, db_common_mock):
+        current_experiment_id = "empty"
+        self._build_db_mock(current_experiment_id, db_common_mock)
+        experiment_id = copy_experiment(current_experiment_id, self.description, self.version, False, False, True)
+        self.assertEqual("", experiment_id)
```
kinow, line 80: This one doesn't seem to be copying the experiment. And with `empty`, but passing `evaluation=True`, I'd expect the new experiment to be `e000`? Maybe something like this,

```diff
diff --git a/test/unit/test_expid.py b/test/unit/test_expid.py
index 2cdb3a7c..b556b366 100644
--- a/test/unit/test_expid.py
+++ b/test/unit/test_expid.py
@@ -1,6 +1,5 @@
 import tempfile
 import os
-import pytest
 from unittest import TestCase
 from mock import Mock, patch
 from autosubmit.autosubmit import Autosubmit
@@ -9,6 +8,7 @@ from textwrap import dedent
 from pathlib import Path
 from autosubmitconfigparser.config.basicconfig import BasicConfig
 from itertools import permutations, product
+from typing import Union
 
 
 class TestExpid(TestCase):
@@ -75,14 +75,16 @@ class TestExpid(TestCase):
     @patch('autosubmit.experiment.experiment_common.db_common')
     def test_copy_experiment_new(self, db_common_mock):
         current_experiment_id = "empty"
-        self._build_db_mock(current_experiment_id, db_common_mock)
+        self._build_db_mock(current_experiment_id, db_common_mock, experiment_exists=[True, False])
         experiment_id = copy_experiment(current_experiment_id, self.description, self.version, False, False, True)
-        self.assertEqual("", experiment_id)
+        self.assertEqual("e000", experiment_id)
     
     @staticmethod
-    def _build_db_mock(current_experiment_id, mock_db_common):
+    def _build_db_mock(current_experiment_id, mock_db_common, experiment_exists: Union[bool, list[bool]] = False):
+        if type(experiment_exists) != list:
+            experiment_exists = [experiment_exists]
         mock_db_common.last_name_used = Mock(return_value=current_experiment_id)
-        mock_db_common.check_experiment_exists = Mock(return_value=False)
+        mock_db_common.check_experiment_exists = Mock(side_effect=experiment_exists)
 
     @patch('autosubmit.autosubmit.read_files')
     def test_autosubmit_generate_config(self, read_files_mock):
```

Found this looking at the codecov missing lines. Two more files to go :slightly_smiling_face:

isimo00: Before my last push I actually tried making the function test both the cases where the file exists and doesn't, but for some reason the case in which the file exists never stops running. I've tried investigating it but no errors or messages pop up. I'm working on it!

kinow: Maybe this side effects could be helfpul? A list with the return value for each call made to the mock.

kinow: I will review the final 2 warnings in codecov in a bit (taking a break now)

kinow: Did you try the side_effects for the test, @isimo00 ? Testing the patch I sent above the code that's missing on Codecov was covered locally when I ran it.

test/unit/test_expid.py
```diff
@@ -57,6 +64,20 @@ def test_create_new_operational_experiment_with_previous_one(self, db_common_moc
...
+    
     @staticmethod
     def _build_db_mock(current_experiment_id, mock_db_common):
         mock_db_common.last_name_used = Mock(return_value=current_experiment_id)
```
kinow, line 83: There's another partial covered statement (screenshot), which you can cover by testing that the empty last expid returns an empty one too.

image


```python
    @patch('autosubmit.experiment.experiment_common.db_common')
    def test_create_new_evaluation_experiment_with_empty_current(self, db_common_mock):
        current_experiment_id = ""
        self._build_db_mock(current_experiment_id, db_common_mock)
        experiment_id = new_experiment(self.description, self.version, False, False, True)
        self.assertEqual("", experiment_id)
```

kinow (approved):
I **think** there are two statements we can cover. One looks more important as that allows us to test that it returns `e000` in one test. The other is not important. The other coverage warning is for `expid()` when eval true, but that was tested manually, and writing unit tests for that function probably requires a fair amount of mocking (i.e. we will probably end up testing mocks, not code).

@isimo00 take a look if my comments about the two statements that could be tested make sense (and actually help to test something). If so, we can update the test and check CICD passes. Otherwise, assuming CICD is green and it's rebased, +1! Thanks!

PR #2200, opened 2025-03-10
Run_profile test bad formatting

VindeeR:
From around 1:50m to about 50s, it still take some time, so for now at least, I'll be moving the file to the integration folder as we avoid super long tests, I don't believe the profiler to be changed all that often


isimo00:
Runs a tad faster now :) thanks!

PR #2206, opened 2025-03-11
Job: first commit vulture fix

VindeeR:
Class wrapper_job -> there would be too many changes

- Vulture Fix


kinow (approved):
These are really not used! Thank you!

PR #2209, opened 2025-03-12
Platform: removing unused variable/functions

VindeeR:
- Vulture platform fix


autosubmit/platforms/platform.py
```diff
@@ -109,25 +108,13 @@ def __init__(self, expid, name, config, auth_password=None):
...
         self.two_factor_method = self.config.get("PLATFORMS", {}).get(self.name.upper(), {}).get("2FA_METHOD", "token")
-        if not self.two_factor_auth:
-            self.pw = None
-        elif auth_password is not None and self.two_factor_auth:
-            if type(auth_password) == list:
-                self.pw = auth_password[0]
-            else:
-                self.pw = auth_password
-        else:
-            self.pw = None
```
kinow, line 130: Is this really not used? I would think the `self.pw` and the `two_factor_auth` code to be used/important.

dbeltrankyl: It has been a long time since it was used, but this should be set whenever the TWO_FACTOR is activated

autosubmit/platforms/platform.py
```diff
@@ -281,20 +270,6 @@ def root_dir(self):
...
-        :type job: Job
-        :return: exclusive directive
-        :rtype: str
-        """
-        # only implemented for slurm
-        return ""
-
-    def get_multiple_jobids(self, job_list, valid_packages_to_submit, failed_packages, error_message="", hold=False):
-        return False, valid_packages_to_submit
-        # raise NotImplementedError
```
kinow, line 297: Ah, I'm removing these two in #2337

autosubmit/platforms/platform.py
```diff
@@ -532,41 +507,6 @@ def add_parameters(self, as_conf):
...
-        :param filename: file name
-        :type filename: str
-        :param must_exist: If True, raises an exception if file can not be copied
-        :type must_exist: bool
-        :param relative_path: relative path inside tmp folder
-        :type relative_path: str
-        :return: True if file is copied successfully, false otherwise
-        :rtype: bool
-        """
-        raise NotImplementedError
```
kinow, line 569: I think these need to stay? The idea was to provide a default impl that classes implementing this would need to include, or the error would be raised?

kinow (commented):
Some conflicts, and left a comment.

autosubmit/platforms/platform.py
```diff
@@ -710,8 +685,6 @@ def remove_checkpoint_file(self, filename):
         if self.check_file_exists(filename):
             self.delete_file(filename)
 
-    def check_file_exists(self, src, wrapper_failed=False, sleeptime=5, max_retries=3):
-        return True
```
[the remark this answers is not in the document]
kinow: Actually, this definitely stays. If you look at the code, now there are new warnings in PyCharm added because of this change. That's because `Platform` is kinda like an abstract class, without the abstract part from Python's abc classes.

`check_file_exists` is used in several methods in this file, so this will probably cause some OOP bugs in the future if an implementation doesn't have this function (thus, the parent must specify this function is needed -- somehow)

autosubmit/platforms/platform.py
```diff
@@ -659,15 +637,12 @@ def get_completed_files(self, job_name, retries=0, recovery=False, wrapper_faile
...
+            for _ in range(retries):
                 if self.get_file('{0}_COMPLETED'.format(job_name), False, ignore_log=recovery):
                     return True
             return False
-        if self.check_file_exists('{0}_COMPLETED'.format(job_name), wrapper_failed=wrapper_failed):
-            if self.get_file('{0}_COMPLETED'.format(job_name), True, wrapper_failed=wrapper_failed):
-                return True
-            else:
-                return False
+        if self.get_file('{0}_COMPLETED'.format(job_name), True, wrapper_failed=wrapper_failed):
```
kinow, line 644: Hmmm, below I commented that `check_file_exists` must stay. But reading this part of the code :thinking: I would think the same manner. But maybe the code was written this way to prevent unnecessary serialization?

If the platform is a remote one, running in an HPC and the file is really large, then I guess it'd make sense to first check it exists, and then to transfer the file. This, assuming the contract of the function `get_file` does not specify how implementations must handle when the file of `get_file` does not exist.

Probably safer to leave as-was.

autosubmit/platforms/platform.py
```diff
@@ -601,17 +590,6 @@ def get_files(self, files, must_exist=True, relative_path=''):
...
-    def delete_file(self, filename):
-        """
-        Deletes a file from this platform
-
-        :param filename: file name
-        :type filename: str
-        :return: True if successful or file does not exist
-        :rtype: bool
-        """
-        raise NotImplementedError
```
kinow, line 613: This needs to stay...

autosubmit/platforms/platform.py
kinow: So I think this stays too... It's used in Paramiko platform, but I think it needs to exist regardless of whether an implementation is paramiko-based or not.

autosubmit/platforms/platform.py
```diff
@@ -126,25 +125,15 @@ def __init__(self, expid, name, config, auth_password=None):
...
         self.put_cmd = None
         self._submit_hold_cmd = None
-        self._submit_command_name = None
         self._submit_cmd = None
         self._submit_cmd_x11 = None
         self._checkhost_cmd = None
         self.cancel_cmd = None
-        self.otp_timeout = None
         self.two_factor_auth = None
-        self.otp_timeout = self.config.get("PLATFORMS", {}).get(self.name.upper(), {}).get("2FA_TIMEOUT", 60 * 5)
```
kinow, line 136: These properties removed above all stay, I believe. They are declared here in the parent, to be used by implementations like Paramiko and grand-children classes (slurm, pbs, etc.).

kinow (changes requested):

PR #2213, opened 2025-03-12
Track the workflow commit

### LINKED ISSUE #2179: [traceability] Missing project commit from workflow configuration and logs
Hi @dbeltrankyl , @mcastril ,

I'm looking at one of my testing suite experiments, `t02z`, which I used for some months running it against the workflow's `main` branch.

Aina/Francesc/Leo/Ivan/others fixed things in the workflow, creating new commits in `main`, then I would refresh it with the testing suite, and run it again. Similar to what should happen in e/o suite envs, and in normal development.

The Autosubmit GUI has access to all the workflow runs. And you can view the Git configuration used, including repository URL, branch, and commit. But I think the commit is only being populated if the user points to a specific commit? i.e. it appears that using `main` with commit `123`, then running again `main` with commit `456` for Autosubmit these two runs are equivalent, as they both have branch `main` and the commit is ignored?

Image

I'm searching the logs, and database now, but I am not finding the specific commit of the Git branch used. Do you know where this is stored, @dbeltrankyl ?

### LINKED ISSUE #2212: Track the workflow commit
Current working plan:

1) Store the commit on as_conf.experiment_data["AUTOSUBMIT"]["WORKFLOW_COMMIT"]

If $expid/proj/$proj_destination has no .git folder, set it as "not-versioned"

2) Store the commit to the job. This must be done whenever the log reset function is called.

3) Print it whenever the job  submitted 

4) Store it on the job_data_$expid.db # <- I'm here

dbeltrankyl:
Closes #2179  
Needs: https://github.com/BSC-ES/autosubmit-config-parser/pull/74  #  Pending last review
Status: Working, some lines are not covered

### DIFF
--- a/autosubmit/history/data_classes/job_data.py
+++ b/autosubmit/history/data_classes/job_data.py
@@ -32,7 +32,7 @@ class JobData(object):
                  status="UNKNOWN", rowtype=0, ncpus=0, wallclock="00:00", qos="debug", energy=0, date="", section="",
                  member="", chunk=0, last=1, platform="NA", job_id=0, extra_data="", nnodes=0, run_id=None, MaxRSS=0.0,
                  AveRSS=0.0, out="", err="", rowstatus=Models.RowStatus.INITIAL, children="", platform_output="",
-                 commit=""):
+                 workflow_commit=""):
         """
         """
         self._id = _id
@@ -76,7 +76,7 @@ class JobData(object):
         self.rowstatus = rowstatus
         self.children = children  # DB 17
         self.platform_output = platform_output  # DB 17
-        self.commit = commit
+        self.workflow_commit = workflow_commit
 
     @classmethod
     def from_model(cls, row):
@@ -112,7 +112,7 @@ class JobData(object):
                        row.rowstatus,
                        row.children,
                        row.platform_output,
-                       row.commit)
+                       row.workflow_commit)
         return job_data
 
     @property
--- a/autosubmit/history/data_classes/job_data.py
+++ b/autosubmit/history/data_classes/job_data.py
@@ -20,96 +20,100 @@ import collections
 import autosubmit.history.utils as HUtils
 import autosubmit.history.database_managers.database_models as Models
 from datetime import datetime, timedelta
-from json import dumps , loads
+from json import dumps, loads
+
 
 class JobData(object):
     """
     Robust representation of a row in the job_data table of the experiment history database.
     """
 
-    def __init__(self, _id, counter=1, job_name="None", created=None, modified=None, submit=0, start=0, finish=0, 
-                status="UNKNOWN", rowtype=0, ncpus=0, wallclock="00:00", qos="debug", energy=0, date="", section="", 
-                member="", chunk=0, last=1, platform="NA", job_id=0, extra_data="", nnodes=0, run_id=None, MaxRSS=0.0, 
-                AveRSS=0.0, out="", err="", rowstatus=Models.RowStatus.INITIAL, children="", platform_output=""):
-      """
-      """
-      self._id = _id
-      self.counter = counter
-      self.job_name = job_name
-      self.created = HUtils.get_current_datetime_if_none(created)
-      self.modified = HUtils.get_current_datetime_if_none(modified)
-      self._submit = int(submit)
-      self._start = int(start)
-      self._finish = int(finish)
-      self.status = status
-      self.rowtype = rowtype
-      self.ncpus = ncpus
-      self.wallclock = wallclock
-      self.qos = qos if qos else "debug"
-      self._energy = round(energy, 2) if energy else 0
-      self.date = date if date else ""
-      self.section = section if section else ""
-      self.member = member if member else ""
-      self.chunk = chunk if chunk else 0
-      self.last = last
-      self._platform = platform if platform and len(
-          platform) > 0 else "NA"
-      self.job_id = job_id if job_id else 0
-      # TODO jobs_data wilmer did this part... need to check that is loading yaml.
-      self.extra_data_parsed = {} # Fail fast
-      try:
-          if extra_data != "":
-            self.extra_data_parsed = loads(extra_data)
-      except Exception as exp:
-          self.extra_data_parsed = {} # Fail fast
-      self.extra_data = extra_data
-      self.nnodes = nnodes
-      self.run_id = run_id
-      self.require_update = False
-      # DB VERSION 15 attributes
-      self.MaxRSS = MaxRSS
-      self.AveRSS = AveRSS
-      self.out = out
-      self.err = err
-      self.rowstatus = rowstatus      
-      self.children = children # DB 17
-      self.platform_output = platform_output # DB 17
+    def __init__(self, _id, counter=1, job_name="None", created=None, modified=None, submit=0, start=0, finish=0,
+                 status="UNKNOWN", rowtype=0, ncpus=0, wallclock="00:00", qos="debug", energy=0, date="", section="",
+                 member="", chunk=0, last=1, platform="NA", job_id=0, extra_data="", nnodes=0, run_id=None, MaxRSS=0.0,
+                 AveRSS=0.0, out="", err="", rowstatus=Models.RowStatus.INITIAL, children="", platform_output="",
+                 commit=""):
+        """
+        """
+        self._id = _id
+        self.counter = counter
+        self.job_name = job_name
+        self.created = HUtils.get_current_datetime_if_none(created)
+        self.modified = HUtils.get_current_datetime_if_none(modified)
+        self._submit = int(submit)
+        self._start = int(start)
+        self._finish = int(finish)
+        self.status = status
+        self.rowtype = rowtype
+        self.ncpus = ncpus
+        self.wallclock = wallclock
+        self.qos = qos if qos else "debug"
+        self._energy = round(energy, 2) if energy else 0
+        self.date = date if date else ""
+        self.section = section if section else ""
+        self.member = member if member else ""
+        self.chunk = chunk if chunk else 0
+        self.last = last
+        self._platform = platform if platform and len(
+            platform) > 0 else "NA"
+        self.job_id = job_id if job_id else 0
+        # TODO jobs_data wilmer did this part... need to check that is loading yaml.
+        self.extra_data_parsed = {}  # Fail fast
+        try:
+            if extra_data != "":
+                self.extra_data_parsed = loads(extra_data)
+        except Exception as exp:
+            self.extra_data_parsed = {}  # Fail fast
+        self.extra_data = extra_data
+        self.nnodes = nnodes
+        self.run_id = run_id
+        self.require_update = False
+        # DB VERSION 15 attributes
+        self.MaxRSS = MaxRSS
+        self.AveRSS = AveRSS
+        self.out = out
+        self.err = err
+        self.rowstatus = rowstatus
+        self.children = children  # DB 17
+        self.platform_output = platform_output  # DB 17
+        self.commit = commit
 
     @classmethod
     def from_model(cls, row):
-      """ Build JobData from JobDataRow. """
-      job_data = cls(row.id, 
-                      row.counter,
-                      row.job_name,
-                      row.created,
-                      row.modified,
-                      row.submit,
-                      row.start,
-                      row.finish,
-                      row.status,
-                      row.rowtype,
-                      row.ncpus,
-                      row.wallclock,
-                      row.qos,
-                      row.energy,
-                      row.date,
-                      row.section,
-                      row.member,
-                      row.chunk,
-                      row.last,
-                      row.platform,
-                      row.job_id,
-                      row.extra_data,
-                      row.nnodes,
-                      row.run_id,
-                      row.MaxRSS,
-                      row.AveRSS,
-                      row.out,
-                      row.err,
-                      row.rowstatus,
-                      row.children,
-                      row.platform_output)
-      return job_data
+        """ Build JobData from JobDataRow. """
+        job_data = cls(row.id,
+                       row.counter,
+                       row.job_name,
+                       row.created,
+                       row.modified,
+                       row.submit,
+                       row.start,
+                       row.finish,
+                       row.status,
+                       row.rowtype,
+                       row.ncpus,
+                       row.wallclock,
+                       row.qos,
+                       row.energy,
+                       row.date,
+                       row.section,
+                       row.member,
+                       row.chunk,
+                       row.last,
+                       row.platform,
+                       row.job_id,
+                       row.extra_data,
+                       row.nnodes,
+                       row.run_id,
+                       row.MaxRSS,
+                       row.AveRSS,
+                       row.out,
+                       row.err,
+                       row.rowstatus,
+                       row.children,
+                       row.platform_output,
+                       row.commit)
+        return job_data
 
     @property
     def children_list(self):
@@ -119,7 +123,7 @@ class JobData(object):
 
     @property
     def computational_weight(self):
-        return round(float(self.running_time * self.ncpus),4)
+        return round(float(self.running_time * self.ncpus), 4)
 
     @property
     def submit(self):
@@ -155,13 +159,13 @@ class JobData(object):
         Returns the energy spent value (JOULES) as an integer.
         """
         return self._energy
-    
+
     @property
     def wrapper_code(self):
         """ 
         Another name for rowtype
         """
-        if self.rowtype > 2:        
+        if self.rowtype > 2:
             return self.rowtype
         else:
             return None
@@ -245,6 +249,7 @@ class JobData(object):
             return o_datetime.strftime(HUtils.DATETIME_FORMAT)
         else:
             return None
+
     @property
     def start_datetime_str(self):
         """
@@ -255,6 +260,7 @@ class JobData(object):
             return o_datetime.strftime(HUtils.DATETIME_FORMAT)
         else:
             return None
+
     @property
     def finish_datetime_str(self):
         """
@@ -275,7 +281,7 @@ class JobData(object):
         :rtype: int
         """
         if self.status in ["RUNNING", "COMPLETED", "FAILED"]:
-            return HUtils.calculate_run_time_in_seconds(self.start, self.finish)                        
+            return HUtils.calculate_run_time_in_seconds(self.start, self.finish)
         return 1
 
     @property
@@ -287,7 +293,7 @@ class JobData(object):
         :rtype: int
         """
         if self.status in ["SUBMITTED", "QUEUING", "RUNNING", "COMPLETED", "HELD", "PREPARED", "FAILED", "SKIPPED"]:
-            return HUtils.calculate_queue_time_in_seconds(self.submit, self.start)            
+            return HUtils.calculate_queue_time_in_seconds(self.submit, self.start)
         return 0
 
     def get_hdata(self):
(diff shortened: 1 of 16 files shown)


dbeltrankyl:
Code working, tried 

* 4.1.11 -> loads 4.1.12 db -> works.
* 4.1.12 -> loads 4.1.11 db -> works.
* 4.1.12 -> creates a new db -> works.

Pytests:

Most of the lines are on the DB side. 

* I added a debug commit to the run integration tests that checks that the workflow_commit field is not empty. ( Missing to add a check that ensures that it exists)
* There are a lot of lines that say it is not covered, but it is because the tests are "skipped," and most of it is just fixing a bad indentation.

* Missing job and platforms pytests 

You can start reviewing it meanwhile if you wish @kinow

autosubmit/history/data_classes/job_data.py
```diff
@@ -20,96 +20,100 @@
...
+        self.require_update = False
+        # DB VERSION 15 attributes
+        self.MaxRSS = MaxRSS
+        self.AveRSS = AveRSS
+        self.out = out
+        self.err = err
+        self.rowstatus = rowstatus
+        self.children = children  # DB 17
+        self.platform_output = platform_output  # DB 17
+        self.workflow_commit = workflow_commit
```
kinow, line 79: I think soon this may have to be changed if we review our DB schema.

We have the commit, but not the branch, nor the URL. And if we normalize the database, I believe the commit should not belong to the job, but to a table like "Project" with relationship similar to `experiment 1 <—> 0|1 projects`, and `experiment 1 <—> 0|* jobs` (the commit belongs to a project & experiment, not the workflow task/job). And a local project, for instance, wouldn't have a commit, so we would have to be careful how to model & normalize this info. This is something @LuiggiTenorioK spotted in other tables, where data is duplicated, with different formats (like diff time formats) or put together even if they don't belong to the same entity.

@dbeltrankyl, @mcastril  what if we simplify here, and drop the workflow_commit from the database, and leave it to be added later, along with other information about the project?

If we do that, in reality, the workflow commit will be in the DB anyway :grimacing: 

I just executed the mHM workflow I used for RO-Crate, and that AiiDA devs were using too. Here's my jobs_data DB row with the commit (the workflow worked with previous versions of AS, but now it's failing to generate the RO-Crate due to `ZeroDivisionError`, I will report it later):

image

But, it's also available when we archive the experiment metadata YAML. So we can retrieve the same information in two ways from the DB, using this branch. In one we get this from the job, and the other from the experiment:

```sql
> SELECT j.workflow_commit FROM job_data j WHERE j.job_name = 'a035_LOCAL_SETUP';
c2672568fb14706c6e66481bc7ceac467db8261e
> SELECT json_extract(r.metadata, "$.AUTOSUBMIT.WORKFLOW_COMMIT") FROM experiment_run r ORDER BY run_id DESC LIMIT 1;
c2672568fb14706c6e66481bc7ceac467db8261e
```

So since we will have it already in the experiment, I think we can drop it from the `Job` class & DB, simplifying it a bit?

kinow (commented):
@dbeltrankyl I found an issue running my auto-mhm workflow with 4.1.12, but I think it might be something else not in this PR (it's in the statistics functions).

I left a comment about the Job changes in DB.

Another thing I noticed, is that everything ran fine, but I don't have a `metadata/experiment_data.yml` file, nor the `metadata` folder. I will check `master` to verify if my workflow works there, and if I get the metadata folder created.

autosubmit/history/data_classes/job_data.py
dbeltrankyl: Hello @kinow 

The reason for being there is because right now, we don't detect historical changes when the run -> run. 

> > SELECT json_extract(r.metadata, "$.AUTOSUBMIT.WORKFLOW_COMMIT") FROM experiment_run r ORDER BY run_id DESC LIMIT 1;
c2672568fb14706c6e66481bc7ceac467db8261e


For Autosubmit right now, a run -> run is the same run. So if we don't store it here with our current db structures, we will only save the last commit, not the ones in between

dbeltrankyl: This is also the safest version I could do because we can modify the experiment_run database to detect run->run. As a different run, I'm not sure how the GUI/api would tackle it and also if it really should be the default behavior, etc. Also, it could be that it is not a different run but rather the autosubmit was stopped for whatever reason

dbeltrankyl: Lastly, autosubmit also detects changes in the workflow during running time. How do you "mark" as a new full run while in running time to keep track of these changes? Because we want to register the changes once the local changes are committed or pulled from git, as otherwise, we won't track the lastest changes. (Also, maybe for the operational, this setting has to set off. )

dbeltrankyl: > I think soon this may have to be changed if we review our DB schema.

I agree with this, but that could be rethought in 4.2 with postgres. Here we don't have much time

mcastril: Same as Dani

I agree with the concern. We need something more structured, probably when we move to PostGreSQL

However, we need something for now. We always trusted that we had this information in Autosubmit but we didn't realize that it was only per run

I fully agree with the idea of creating an additional table for this. I had  a similar thought today about the configuration. We would like to log configuration changes too.

In an ideal world, we would have a configuration table with one entry for each time Autosubmit is run with a different configuration, and pointers from the job level to these entries. This way we could log changes in the config even if they are done in the middle of a run

kinow: Yeah, I ran the workflow a few times now, create+run, observing when the experiment_run table is created and populated. Looks like the order is:

- create -> creates an experiment_run, sets the `start` time to now (which causes issues in the GUI as that date is used in some fields, https://github.com/BSC-ES/autosubmit-gui/issues/251#issuecomment-2633235447).
- run -> runs, updates job list I think, and at the end sets the `finish` time.

Setstatus and other commands also update the experiment run.

The experiment_run starts with a create. But if you have a blank experiment and call other expid commands (e.g. `inspect`) the command runs fine, its data is written to disk & logs, but no experiment history DB is created.

@mcastril @dbeltrankyl , just thinking other alternatives to avoid adding the table & just to remove it soon (and we already have accumulated technical debt over years, so the least we add now, the better IMHO).

What if we changed and `create` behaved like `inspect`? It populates the pickle, checks out the git project on its first run, and creates the blank experiment history DB? Without creating an experiment run.

Then, every `autosubmit run` would create a new experiment run in the DB?

I did a quick test, with this diff (10 mins):

```diff
diff --git a/autosubmit/autosubmit.py b/autosubmit/autosubmit.py
index 99fbbd68..7705f4d2 100644
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -53,7 +53,7 @@ from .platforms.paramiko_submitter import ParamikoSubmitter
 from .platforms.platform import Platform
 from .migrate.migrate import Migrate
 
-from time import sleep
+from time import time, sleep
 import argparse
 import subprocess
 import json
@@ -1892,8 +1892,7 @@ class Autosubmit:
                                             historiclog_dir_path=BasicConfig.HISTORICAL_LOG_DIR)
             exp_history.initialize_database()
             exp_history.process_status_changes(job_list.get_job_list(), as_conf.get_chunk_size_unit(),
-                                               as_conf.get_chunk_size(),
-                                               current_config=as_conf.get_full_config_as_json())
+                                               as_conf.get_chunk_size())
             Autosubmit.database_backup(expid)
         except Exception as e:
             try:
@@ -2164,6 +2163,12 @@ class Autosubmit:
                     # This function is called only once, when the experiment is started. It is used to initialize the experiment and to check the correctness of the configuration files.
                     # If there are issues while running, this function will be called again to reinitialize the experiment.
                     job_list, submitter , exp_history, host , as_conf, platforms_to_test, packages_persistence, _ = Autosubmit.prepare_run(expid, notransitive, start_time, start_after, run_only_members)
+
+                    exp_history.create_new_experiment_run(
+                        as_conf.get_chunk_size_unit(), as_conf.get_chunk_size(), job_list.get_job_list())
+                    Autosubmit.database_backup(expid)
+                    if not exp_history.start_current_experiment_run():
+                        raise AutosubmitCritical(f'Failed to set start time of experiment {expid} in the history DB!')
                 except AutosubmitCritical as e:
                     #e.message += " HINT: check the CUSTOM_DIRECTIVE syntax in your jobs configuration files."
                     raise
@@ -4591,14 +4596,6 @@ class Autosubmit:
                         exp_history = ExperimentHistory(expid, jobdata_dir_path=BasicConfig.JOBDATA_DIR,
                                                         historiclog_dir_path=BasicConfig.HISTORICAL_LOG_DIR)
                         exp_history.initialize_database()
-
-                        # exp_history.create_new_experiment_run(as_conf.get_chunk_size_unit(), as_conf.get_chunk_size(), as_conf.get_full_config_as_json(), job_list.get_job_list())
-                        exp_history.process_status_changes(job_list.get_job_list(),
-                                                           chunk_unit=as_conf.get_chunk_size_unit(),
-                                                           chunk_size=as_conf.get_chunk_size(),
-                                                           current_config=as_conf.get_full_config_as_json(),
-                                                           create=True)
-                        Autosubmit.database_backup(expid)
                     except BaseException as e:
                         Log.printlog("Historic database seems corrupted, AS will repair it and resume the run",
                                      Log.INFO)
@@ -5384,8 +5381,7 @@ class Autosubmit:
                     exp_history.initialize_database()
                     exp_history.process_status_changes(job_list.get_job_list(),
                                                        chunk_unit=as_conf.get_chunk_size_unit(),
-                                                       chunk_size=as_conf.get_chunk_size(),
-                                                       current_config=as_conf.get_full_config_as_json())
+                                                       chunk_size=as_conf.get_chunk_size())
                     Autosubmit.database_backup(expid)
                 else:
                     Log.printlog(
diff --git a/autosubmit/history/experiment_history.py b/autosubmit/history/experiment_history.py
index 806f6add..823d941d 100644
--- a/autosubmit/history/experiment_history.py
+++ b/autosubmit/history/experiment_history.py
@@ -216,7 +216,7 @@ class ExperimentHistory:
             self._log.log(str(exp), traceback.format_exc())
             Log.debug(f'Historical Database error: {str(exp)} {traceback.format_exc()}')
 
-    def process_status_changes(self, job_list=None, chunk_unit="NA", chunk_size=0, current_config="", create=False):
+    def process_status_changes(self, job_list=None, chunk_unit="NA", chunk_size=0, create=False):
         """ Detect status differences between job_list and current job_data rows, and update. Creates a new run if necessary. """
         try:
             try:
@@ -232,7 +232,7 @@ class ExperimentHistory:
             if len(update_these_changes) > 0 and not should_create_new_run:
                 self.manager.update_many_job_data_change_status(update_these_changes)
             if should_create_new_run:
-                return self.create_new_experiment_run(chunk_unit, chunk_size, current_config, job_list)
+                return self.create_new_experiment_run(chunk_unit, chunk_size, job_list)
             return self.update_counts_on_experiment_run_dc(current_experiment_run_dc, job_list)
         except Exception as exp:
             self._log.log(str(exp), traceback.format_exc())
@@ -282,6 +282,13 @@ class ExperimentHistory:
         experiment_run_dc.total = status_counts["TOTAL"]
         return self.manager.update_experiment_run_dc_by_id(experiment_run_dc)
 
+    def start_current_experiment_run(self):
+        if self.manager.is_there_a_last_experiment_run():
+            current_experiment_run_dc = self.manager.get_experiment_run_dc_with_max_id()
+            current_experiment_run_dc.start = int(time())
+            return self.manager.update_experiment_run_dc_by_id(current_experiment_run_dc)
+        return None
+
     def finish_current_experiment_run(self):
         if self.manager.is_there_a_last_experiment_run():
             current_experiment_run_dc = self.manager.get_experiment_run_dc_with_max_id()
@@ -289,19 +296,19 @@ class ExperimentHistory:
             return self.manager.update_experiment_run_dc_by_id(current_experiment_run_dc)
         return None
 
-    def create_new_experiment_run(self, chunk_unit="NA", chunk_size=0, current_config="", job_list=None):
+    def create_new_experiment_run(self, chunk_unit="NA", chunk_size=0, job_list=None):
         """ Also writes the finish timestamp of the previous run.  """
         self.finish_current_experiment_run()
         return self._create_new_experiment_run_dc_with_counts(chunk_unit=chunk_unit, chunk_size=chunk_size,
-                                                              current_config=current_config, job_list=job_list)
+                                                              job_list=job_list)
 
-    def _create_new_experiment_run_dc_with_counts(self, chunk_unit, chunk_size, current_config="", job_list=None):
+    def _create_new_experiment_run_dc_with_counts(self, chunk_unit, chunk_size, job_list=None):
         """ Create new experiment_run row and return the new Models.ExperimentRun data class from database. """
         status_counts = self.get_status_counts_from_job_list(job_list)
         experiment_run_dc = ExperimentRun(0,
                                           chunk_unit=chunk_unit,
                                           chunk_size=chunk_size,
-                                          metadata=current_config,
+                                          metadata=None,
                                           start=int(time()),
                                           completed=status_counts[HUtils.SupportedStatus.COMPLETED],
                                           total=status_counts["TOTAL"],
```

And at least the happy-path scenario works, `autosubmit create` runs, and I get this.

image

image

The pickle file is created, so are the other DB & files.

Finally, `autosubmit run` prepares the run, and immediately creates an experiment run, setting the `start` time to when the simulation started (which would fix the bug we have in the GUI I think). And also creates the backup.

image

image

Couldn't we solve it this way? I think that way we get an `experiment_run` for every time we actually run the experiment, and we can store the `commit` used for that run in the experiment.

kinow: And about the idea for the new table, etc., that's just to explain my concern in merging it quickly now. I think we can finish this on Monday, so we have a few more hours to think in other alternatives, I guess. Fully agree that we need this now, and once it's implemented I'll think how it will look in the RO-Crate archives and check if the GUI is displaying the information correctly (the current RO-Crates are probably using the wrong start time too, the same bug from the GUI).

dbeltrankyl: This approach can also work; instead of adding the workflow_commit, I can add it in that experiment_run table. Or I can generate a new one with run_id | workflow commit

I did some testing, and the only stuff we miss is this metadata

image

My only concern is that the field of "TOTAL" will be weird following this strategy because that run could be cut at any time. The total field should be the total number of jobs submitted, no?

Also, I'm unsure if the experiment_run is used for other stuff. Weren't we had some metrics about an experiment run?  @mcastril 

With your changes, the run_id of the jobs will change if the experiment is stopped and resumed, but it is not necessary for there to be any change in the workflow not sure if it will affect any metric etc

image

dbeltrankyl: Another concern is about what happens with the run_id associated with a job when autosubmit is stopped but the job is running. 

Does that job have the new run_id once it is written in the database? I think that's the case, and it shouldn't be because it could have been run with another commit.

This can be solved by adding a field "run_id" to the job class

dbeltrankyl: So in summary, 

What do we consider that is a new run? 

Current master: A new run is only when the joblist is recreated and the run starts from the first job

Your proposed changes: Any time we prompt autosubmit run regardless of the previous status.

kinow: >Does that job have the new run_id once it is written in the database? I think that's the case, and it shouldn't be because it could have been run with another commit.

Agreed. Does this happen with these changes? Or does it need to be tested to confirm? 

> This can be solved by adding a field "run_id" to the job class

Which sounds OK, I think. But not sure if it'd be much more effort than adding the commit to job (i.e. if that's going to take really long, then there's nothing we can do and we will have to just go with the temporary solution, I think).

kinow: >With your changes, the run_id of the jobs will change any time the experiment is stopped and resumed. Even if there are no workflow changes or job changes. I'm not sure if it will affect any metric etc

Good point. I guess it'd be OK to have that, semantically -- an `autosubmit run` always produces an `experiment_run`. I think that way we would have two experiment runs, with two configurations that may be the same or not. Which could be helpful, to compare what changed when the user restarted it. But if it affects metrics, then it could be a problem.

dbeltrankyl: Also, I have a jump from 5 to 7, which raises another question. 

Do we consider an empty run ( no jobs processed ) a new run?

dbeltrankyl: > Which sounds OK, I think. But not sure if it'd be much more effort than adding the commit to job (i.e. if that's going to take really long, then there's nothing we can do and we will have to just go with the temporary solution, I think).

At first glance it will be harder ( not sure how much ) because the update methods are like this: 

>  def delete_experiment_run(self, run_id): ...
>     def get_job_data_dcs_last_by_run_id(self, run_id): ...

So.. all our job_data methods relies on first gather the last run_id from the experiment_run table

kinow: Yeah, your points also added a few more questions to what I had in mind. I guess we will have to go with the current approach then, and then discuss later what to do about the commit in the job, drop it, move it, keep it, etc.

Thanks for the tests, @dbeltrankyl ! 

I guess what's pending now is just the tests to mark it as ready for review? I tested it again this morning with that comment I wrote in the AS config parser issue (where I compared what happens with PROJECT_COMMIT, changing files, commiting in the proj/git_project folder, etc.) and found no issues :slightly_smiling_face:

dbeltrankyl: I don't know @kinow 

Honestly, I think that If we approach this solution, I see many unanswered questions that we should have more time to dedicate to instead of being rushed to avoid unexpected issues. 

My solution I already knew (before coding it) that my solution is not good in the long term, but was better in the short term ( that is today or tomorrow). Autosubmit will behave the same except for having this new field in the DB, and I don't think it will be hard to undo it 

I can do one or the other, but we should decide fast so it can be tested asap

What do you think now that I've further explained my concerns?

kinow: We crossed each other :slightly_smiling_face: my response is above (TL;DR: let's do it the way you implemented and think about it later). I think we could do it the other way now, but it would be **a lot riskier**, so a no-no as we need the release out soon and we want to avoid issues in esuite if possible. Thanks!!

dbeltrankyl: Yeah, I saw it. 

Thanks @kinow !

dbeltrankyl: > I guess what's pending now is just the tests to mark it as ready for review? I tested it again this morning with that comment I wrote in the AS config parser issue (where I compared what happens with PROJECT_COMMIT, changing files, commiting in the proj/git_project folder, etc.) and found no issue


The db are tested with the regression one.. not sure how to do that part with unit ones, but the run integration test, is covering that part

I need to add this ( Missing to add a check that ensures that it exists) to that test

I'll do it now

dbeltrankyl:
rebase

kinow (approved):
Code looks good, and the summary on coverage issues was really helpful to review it faster. Tested last week and earlier this morning, and in both cases I got the commit ID in the database, in the newly created column. LGTM. Thanks @dbeltrankyl !

dbeltrankyl:
> Code looks good, and the summary on coverage issues was really helpful to review it faster. Tested last week and earlier this morning, and in both cases I got the commit ID in the database, in the newly created column. LGTM. Thanks @dbeltrankyl !

Missed this comment, I'll rebase and merge then

thanks!

PR #2218, opened 2025-03-13
Change scheduler is not installed exception.

### LINKED ISSUE #2102: Scheduler not installed raised as an Autosubmit Critical error in the middle of the run. ( Should be an Autosubmit Error )
Hello,

I'm not sure if this issue is related to the workflow or the AS, but I suspect it is more related to AS.

I'm using AS 4.1.11 and WF 4.2.0 on MN5.

In my experiment a1yv, the last chunk was apparently successfully COMPLETED I can see this from the model logs and also from the file:
`/gpfs/scratch/ehpc01/bsc998159/a1yv/LOG_a1yv/a1yv_19900101_fc0_332_SIM_COMPLETED`.

However, when I run autosubmit monitor a1yv or check the AS GUI, this same SIM job still appears as RUNNING. The experiment crashed, outputting the following in the nohup:
```
[CRITICAL] Scheduler is not installed. [eCode=7052]  
```
I would like to continue the experiment ASAP. Would it be okay if I change the job status from RUNNING to WAITING and re-submit? Or would this prevent you from investigating the root cause of the issue?

dbeltrankyl:
This fixes #2102. 

For a month, there has been an issue with marenostrum5 in which the sacct, sbatch, or squeue commands are not found. 

I changed the exception to trigger the recovery platform code instead of stopping Autosubmit when it happens.

Initially, it was critical because this could be a user mistake in the platforms.yml.

Per example, %marenostrum5.type% = pjm"

We can just make this change for 4.1.12 and, in other versions ( if we don't have more time), add a new function ( for example, when checking the platform's connectivity) that checks if the command exists. 

So, if marenostrum5 is slurm, the commands of pjm wouldn't be found, raising a critical.

### DIFF
--- a/autosubmit/platforms/paramiko_platform.py
+++ b/autosubmit/platforms/paramiko_platform.py
@@ -1179,7 +1179,12 @@ class ParamikoPlatform(Platform):
                     raise AutosubmitError(
                         'SSH Session not active, will restart the platforms', 6005)
                 if errorLine.find("command not found") != -1:
-                    raise AutosubmitCritical("scheduler is not installed.",7052,self._ssh_output_err)
+                    raise AutosubmitError(
+                        f"A platform command was not found. This may be a temporary issue. "
+                        f"Please verify that the correct scheduler is specified for this platform: '{self.name}.{self.type}'.",
+                        7052,
+                        self._ssh_output_err
+                    )
                 elif errorLine.find("syntax error") != -1:
                     raise AutosubmitCritical("Syntax error",7052,self._ssh_output_err)
                 elif errorLine.find("refused") != -1 or errorLine.find("slurm_persist_conn_open_without_init") != -1 or errorLine.find("slurmdbd") != -1 or errorLine.find("submission failed") != -1 or errorLine.find("git clone") != -1 or errorLine.find("sbatch: error: ") != -1 or errorLine.find("not submitted") != -1 or errorLine.find("invalid") != -1 or "[ERR.] PJM".lower() in errorLine:


kinow (approved):
Looks like it's not Slurm commands, but something with the `module`... or a "Stale file handle" :dizzy: 

```
2025-03-12 17:32:19,229 /gpfs/scratch/ehpc01/bsc032155/a278/LOG_a278/a278_19900101_fc22_3_SIM.cmd.err.0 File does not exist.. waiting 0s for a new retry (retries left: 1)
2025-03-12 17:32:19,298 Traceback (most recent call last):
  File "/appl/AS/4.1.11/lib64/python3.9/site-packages/autosubmit/autosubmit.py", line 2260, in run_experiment
    platform.check_Alljobs(platform_jobs, as_conf)
  File "/appl/AS/4.1.11/lib64/python3.9/site-packages/autosubmit/platforms/paramiko_platform.py", line 722, in check_Alljobs
    self.send_command(cmd)
  File "/appl/AS/4.1.11/lib64/python3.9/site-packages/autosubmit/platforms/paramiko_platform.py", line 1116, in send_command
    raise AutosubmitCritical("scheduler is not installed.",7052,self._ssh_output_err)
log.log.AutosubmitCritical:

  File "/appl/AS/4.1.11/bin/autosubmit", line 36, in main
    return_value = Autosubmit.parse_args()
  File "/appl/AS/4.1.11/lib64/python3.9/site-packages/autosubmit/autosubmit.py", line 711, in parse_args
  File "/appl/AS/4.1.11/lib64/python3.9/site-packages/autosubmit/autosubmit.py", line 2382, in run_experiment
    raise AutosubmitCritical(e.message, e.code, e.trace)
log.log.AutosubmitCritical:

2025-03-12 17:32:19,299 Trace: /etc/profile.d/01-module.sh: line 11: /apps/modules/LMOD/8.7/lmod/8.7/init/profile: Stale file handle
/etc/profile.d/01-module.sh: line 30: module: command not found
/etc/profile.d/01-module.sh: line 31: module: command not found

[CRITICAL] 2025-03-12 17:32:19,299 scheduler is not installed. [eCode=7052]
2025-03-12 17:32:19,299 More info at https://autosubmit.readthedocs.io/en/master/troubleshooting/error-codes.html
```

This is from the experiment of @malerany. For Linux, that appears to be a normal error if the file you have the handle is deleted.

But for GPFS [it appears more serious](https://www.ibm.com/support/pages/gpfs-internal-and-external-mount-states),

image

I think this could be due to GPFS / IO errors in MN5, which would make sense as others (e.g. @ainagaya) had issues running in MN5 in the past few days. If something is happening with GPFS and its components like the metadata server I think this error could happen (where we try to access `module` and suddenly the file `module` we had access to disappears... then reappears later on, with another inode).

@dbeltrankyl I think even if it's in the IO, it might be worth raising the `AutosubmitError` to allow retries to happen? If so, lets get it merged :tada: Leaving it approved.

And I spent 10 minutes looking at the code to see if this could be tested... but it uses SSH, `select` function, needs iterators... :sweat_smile: yeah, it would take a very long time to write a test, and it wouldn't look very good.

Re-reviewed the code, with calm and :coffee: , and it looks correct.

dbeltrankyl:
Thanks @kinow !

Issue #2165, opened 2025-02-28
Create Maestro Platform based on Slurm


kinow:
This task is for adding a new platform in Autosubmit. This platform will be called Maestro (see https://www.maestro-data.eu/), and it should extend the existing [`SlurmPlatform`](https://github.com/BSC-ES/autosubmit/blob/05381cd9e5ea1f82377c73a98d1821a02129fed5/autosubmit/platforms/slurmplatform.py).

## Steps

Steps (BSC, I believe):

1. Read about Maestro, check out the [CSC Wiki](https://wiki.eduuni.fi/spaces/cscRDIcollaboration/pages/288471456/DestinE+Climate+DT), and [this issue](https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/issues/434) in the ClimateDT workflow;
2. Review the documentation that we have for platforms in our Sphinx docs: https://github.com/BSC-ES/autosubmit/tree/05381cd9e5ea1f82377c73a98d1821a02129fed5/docs;
3. Review the Python API documentation in the `slurmplatform.py` linked above;
4. Write documentation explaining how platforms work in AS, how one can be extended, and document the functions that we have (e.g. find a function that will be needed for Maestro but doesn't have pydocs, and write `"""This function does...`, etc.)
5. If possible include examples in the Python docs, e.g. https://github.com/pydantic/pydantic/blob/b9fb3f110d6653653fb48a44430e7d2fca886ae9/pydantic/networks.py#L562-L590, and in the Sphinx docs
6. Send documentation and examples to HPE

And remaining steps (for HPE, I think):

1. Extend the `SlurmPlatform` using the instructions provided by BSC
2. Create a sample workflow that uses the platform, but with simpler tasks (something like a hello world but launched in the Maestro platform, using a Slurm job)
3. Verify that the DATA_DEPENDENCIES is working as expected (not sure if we will need any change in AS)

If HPE needs support during any of these tasks, or if they require more information, we can use this issue. @chhaine you tag @isimo00 and @VindeeR. They will be working on this task.

## How to develop this

The development can be done in a local machine, with Linux and Python 3.9. I would create a virtual environment, install Autosubmit from the branch in editable mode (`pip install -e .`), and follow the instructions from @VindeeR & @isimo00 on how to extend `SlurmPlatform`, writing a new module like `autosubmit.platforms.maestro.py` with the required code.

[...]

Issue #977, opened 2023-03-16
[testing] Write tests for platforms


kinow:
From https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/324#note_203977

> We currently don't have any test for platform code aside of this new one

Probably having functional/integration tests would be helpful, as it might be hard to cover the platforms code with just unit tests (or the mocking needed will make the tests less useful, IMO).

kinow:
mentioned in issue digital-twins/de_340/workflow#16

kinow:
This is a bit bigger than other issues, but we could solve it by having a small set of functional tests that submit one or two experiments and use,

* local platform
* ps platform (launching a remote docker ssh container like https://docs.linuxserver.io/images/docker-openssh-server/
* slurm platform (maybe with an existing docker/docker-compose setup, or the simulator that @manuel-g-castro used for his master/phd)

PR #2233, opened 2025-03-18
Add unit and integration tests for Slurm

kinow:
Partially addresses #977

Temporary branch, that I will either update or delete in the next days.

@sergipalomas commented about an issue on how energy is reported by the API, but looking at AS code, it seems that AS is the one actually storing `0` in the `energy` column of job data.

https://github.com/BSC-ES/autosubmit-api/issues/175#issuecomment-2732826290

I started writing a unit test for that, but I realized that the `sacct` output I have is missing some columns. I have jobs with energy consumption, but I'm interested in using Sergi's example to check if I can reproduce the case where `energy=0` (then I can get that and write a test, and get another one that doesn't return `0`, covering both -- this is, if we confirm that it is indeed `0` and not another value that should be returned).

@VindeeR & @isimo00 will also work on the Slurm platform soon for Maestro (#2165), and we will have to increase our test coverage for that module, so it'll be a win-win for all if we can get this test written and working.

### DIFF
--- a/test/unit/platforms/test_slurmplatform.py
+++ b/test/unit/platforms/test_slurmplatform.py
@@ -1,0 +1,43 @@
+# Copyright 2015-2025 Earth Sciences Department, BSC-CNS
+#
+# This file is part of Autosubmit.
+#
+# Autosubmit is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Autosubmit is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+"""Tests for the Autosubmit Slurm platform."""
+
+import pytest
+
+from textwrap import dedent
+
+from autosubmit.platforms.slurmplatform import SlurmPlatform
+
+
+@pytest.fixture
+def slurm_platform(autosubmit_config) -> SlurmPlatform:
+    """Fixture to create a Slurm platform object."""
+    as_conf = autosubmit_config('a000', {})
+    return SlurmPlatform(as_conf.expid, 'TestSlurmPlatform', as_conf.experiment_data)
+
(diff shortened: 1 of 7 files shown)


test/unit/platforms/test_slurmplatform.py
```diff
@@ -0,0 +1,43 @@
...
+        JobID                     Submit               Start                 End    Elapsed ConsumedEnergy
+        ------------ ------------------- ------------------- ------------------- ---------- --------------
+        15994954     2025-02-24T16:11:33 2025-02-24T16:11:42 2025-02-24T16:21:30   00:09:48        883.55K
+        15994954.ba+ 2025-02-24T16:11:42 2025-02-24T16:11:42 2025-02-24T16:21:30   00:09:48        497.36K
+        15994954.ex+ 2025-02-24T16:11:42 2025-02-24T16:11:42 2025-02-24T16:21:30   00:09:48        883.55K
+        15994954.0   2025-02-24T16:11:47 2025-02-24T16:11:47 2025-02-24T16:11:52   00:00:05              0
+        15994954.1   2025-02-24T16:12:17 2025-02-24T16:12:17 2025-02-24T16:21:22   00:09:05        844.90K
+    ''')
+    tuple_data = slurm_platform.parse_job_finish_data(output, packed=False)
+    assert tuple_data is not None
```
kinow, line 43: Note to self, get the correct output from Sergi, then update the string above, and continue writing the test to actually test something, not `is not None` :grimacing:

dbeltrankyl: To push local changes

dbeltrankyl:
I see this assigned to myself, but I also see that this is a draft .. so what is the real status of this one?

kinow:
I will update the unit test that I wrote using performance data. I had written it in some branch, in some laptop, but I cannot find it :sweat_smile: So I will have to rewrite it tomorrow or over next days when I take a break from Postgres.

kinow:
@VindeeR I think if you can copy the unit tests from this branch into yours, #2362 , then we can close this as superseded by your PR.

kinow:
Superseded by #2362, thanks @manuel-g-castro & @VindeeR for the work on the image, GH Actions & tests ! :star:

PR #2237, opened 2025-03-19
4.1.12 recovery is slow

dbeltrankyl:
Removed unnecessary update_parameters calls


dbeltrankyl:
I only removed stuff, and we don't have a proper way of automatically testing the recovery command. Do you think it is necessary to do the pytest here? @kinow 

It should be safe to remove (it was also removed from create in the mem-branch) due:

 # Update parameters with attributes is called in: build_packages, submit wrappers, and for every ready job. It shouldn't be necessary here.

The recovery in the current master is very slow for larger workflows ( 0.3 seconds per job) this one is (Time spent: '0:00:00.088198' per job)

kinow (approved):
The change looks good, but testing it would be hard. We don't have all the tooling needed for writing tests for functions like recovery, I am afraid. Let's merge it and test with the testing suite. Francesc can probably try again if he's still transferring data. Thanks Dani!

PR #2238, opened 2025-03-19
Use pytestify to migrate tests to pytest, use xdist for faster tests, move integration tests to another folder, format/add headers

### LINKED ISSUE #961: [enhancement] Use pytest instead of unittest for writing tests for Autosubmit
I'm starting to look into writing tests for the RO-Crate and `cat-log` merge requests, but I really miss the [fixtures](https://docs.pytest.org/en/7.2.x/how-to/fixtures.html) in pytest, and also not having to write a class, then methods, etc.

For me one of the biggest wins of using pytest  is adding fixtures that can be reused in several other tests. With unitest we can create objects needed for the tests in the setUp function, but then if we have another file with another test that needs something similar (or parts of it) then we have to design some OO way to share that to avoid duplication.For example, see these two different test files.

image

image

There is a lot of boilerplate code that is duplicated (I think there are other test files with similar code). With pytest we could have a conftest.py file at the root of the test directory, with something like:

```py
import pytest
import random
from autosubmit.experiment.experiment_common import base36encode

@ pytest.fixture
def expid():
    """Autosubmit expid is a base-36 radix value, with at least
    4 characters.

    We first pick a random number between 0 and the last possible
    number to produce a 4-character expid ``(36 ^ 4) - 1``, so
    ``1_679_615`` (we do not need to test with len > 4)."""
    random_int = random.randint(0, 1_679_615)
    return base36encode(random_int)

@ pytest.fixture
def job_list(expid, mocker):
    as_conf = mocker.Mock()
    as_conf.experiment_data = dict()
    as_conf.experiment_data["JOBS"] = dict()
    as_conf.jobs_data = as_conf.experiment_data["JOBS"]
    as_conf.experiment_data["PLATFORMS"] = dict()
    return JobList(expid, FakeBasicConfig, YAMLParserFactory(...), JobListPersistenceDb(...), as_conf)  
```

And I could use it in any test with:

```py
def test_job_graph(job_list):
    ...

def test_synchronize_date_group_date(job_list):
    ...
```

kinow:
Closes #961 

While working on the esuite release, I noticed writing tests and waiting for it to run was super slow (@isimo00 raised a similar issue too some days/weeks ago).

The changes in this branch should be mainly (if not only) in the tests:

- [x] Executed `pytestify` to convert tests from unittest to pytest
- [x] Uses pytest-xdist to run tests in parallel (unit tests use 8 processes, runs in 6 seconds on my laptop)
- [x] Moves unit tests from integration to unit, and integration tests from unit to integration
- [x] Executes integration tests with GH Actions :tada: 
- [x] Adds updated (2025) license header to all test files
- [x] Removes unused imports and formats code (quickly with PyCharm, @VindeeR is working on a more definitive solution with linters, and configuration to be shared with others)
- [x] Git ignores .coverage.$host.... files that are now created with coverage (didn't happen before)
- [x] Disables coverage by default so unit tests are faster, and debugger works without any modification in PyCharm (otherwise, the debugger won't stop due to instrumented code, and you have to 1) run, 2) edit configuration adding `--no-cov`, 3) now run again -- that's not needed now)
- [x] Fix a bug in our `pyproject.toml` as the optional group `all` was repeating dependencies instead of linking to existing optional groups
- [x] Removed almost all uses of `FakeBasicConfig`, please use `autosubmit_config` and avoid classes (`autosubmit_config` is widely used now, and it is the default type, `scope=function`, so if you write pytest with classes that fixture is not available as it's != scope). Two places where it'd take too long to remove were kept with a `TODO`

Hoping this can be merged soon so that I can use this in the postgres development, for writing new tests (unit tests for sure, but also possibly some integration tests).


kinow:
> Out of curiosity, what are the specific tests that are taking long for you?

Those test_run_command_intregation.py. Moving they out, and parallellizing runs in ~7 seconds now :slightly_smiling_face:

dbeltrankyl:
> EDIT: also, running those tests in parallel fail sometimes... so those tests are slow and brittle. Definitely need some rework in the future.

Yes, I was about to comment on that because I tried to use the xdist as well and had that issue. I think I also had that issue in the 

>  so those tests are slow and brittle

In the end, they're running a normal experiment and just checking if everything is okay ( db, logs etc )  but if you add a safetysleeptime in their conf to 0 seconds, it may be faster ( similar to what we did for the profile)

kinow:
> > EDIT: also, running those tests in parallel fail sometimes... so those tests are slow and brittle. Definitely need some rework in the future.
> 
> Yes, I was about to comment on that because I tried to use the xdist as well and had that issue. I think I also had that issue in the
> 
> > so those tests are slow and brittle
> 
> In the end, they're running a normal experiment and just checking if everything is okay ( db, logs etc ) but if you add a safetysleeptime in their conf to 0 seconds, it may be faster ( similar to what we did for the profile)

:+1: in this branch they are grouped together and assigned to a single worker of xdist. This way everything runs in parallel but those tests.

kinow:
Interesting. This branch is now failing in `test_paramiko_platform.test_send_file`. That test is actually opening an SSH connection to localhost. No idea how that's working on GH Action workers (as far as I know there's no SSH in those workers?).

dbeltrankyl:
> test_send_file

In master they're skipped: @pytest.mark.skip(reason="Skipping this test until Github transition is complete")

kinow:
> > test_send_file
> 
> In master they're skipped: @pytest.mark.skip(reason="Skipping this test until Github transition is complete")

Aaahh! D'oh! I removed that skip marker locally, and they worked because I had SSH running :laughing: :laughing: :laughing: 

I need more :coffee: already. Thanks @dbeltrankyl !

kinow:
Less than 10 files now. Let's see if it passes the GH Actions checks, and then wrap up removing the last few places where `unittest` is still used.

.gitignore
```diff
@@ -17,6 +17,7 @@ build/
 # Coverage files
 coverage/
 .coverage
+.coverage.*
```
kinow, line 20: Sometimes `pytest-cov`/`coverage` create multiple `.coverage.$host.$uid` files (when running tests in parallel, I think). So they may appear in the git diff without this. `.coveragerc` is not affected (tested).

autosubmit/autosubmit.py
```diff
@@ -143,7 +143,6 @@ class Autosubmit:
     """
 
     def __init__(self):
-        self._experiment_data = {}
```
kinow, line 146: Not used code was removed in this file.

autosubmit/job/job_utils.py
```diff
@@ -293,17 +293,6 @@ def __init__(self, section, distance=None, running=None, sign=None, delay=-1, sp
...
-    """
-    A simple replacement for jobs
-    """
-
-    def __init__(self, name, tmppath, statuscode) -> None:
-        self.name = name
-        self._tmp_path = tmppath
-        self.status = statuscode
-
-
```
kinow, line 306: Not used.

autosubmit/platforms/wrappers/wrapper_factory.py
```diff
@@ -193,57 +193,6 @@ def header_directives(self, **kwargs):
         return ""
 
 
-class LocalWrapperFactory(WrapperFactory):
```
kinow, line 196: Duplicate (look above this lines in `master`, to see the exact copy of this class re-declared).

pyproject.toml
```diff
@@ -84,8 +84,9 @@ tests = [
     "pytest>=8.2,<8.4",
     "pytest-cov",
     "pytest-mock",
-    "ruff==0.11.0"
-    # 'testcontainers'  # TODO: pending Postgres MR
+    "ruff==0.11.0",
+    "pytest-xdist",
+    "testcontainers"
```
kinow, line 89: Testcontainers will be used in Postgres too. We can add it now to launch the SSH server for the integration tests.

pyproject.toml
```diff
@@ -95,17 +96,10 @@ docs = [
...
-    "pydata-sphinx-theme==0.15.*",
-    "sphinx>=5,<9",
-    "sphinx-autobuild>=2021.3,<2024.11",
-    "sphinx_rtd_theme",
-    "sphinx-reredirects==0.1.*"
+all = [
+    "autosubmit[tests]",
+    "autosubmit[docs]"
+    # "autosubmit[postgres]"
 ]
```
kinow, line 103: To avoid duplicating the groups/optional dependencies here again.

pytest.ini
```diff
@@ -1,10 +1,9 @@
 [pytest]
 addopts =
     --verbose --ignore=regression
-    --cov=autosubmit --cov-config=.coveragerc
-    --cov-report=html:test/htmlcov --cov-report=xml:test/coverage.xml
```
kinow, line 5: Disabled coverage by default as that slows down the unit tests, and the debugger is not able to locate the lines with break points (as the code is modified by `coverage`).

pytest.ini
```diff
@@ -1,10 +1,9 @@
...
     --strict-markers
     --doctest-modules
+    -n auto --max-worker-restart 0 --dist loadgroup
```
kinow, line 6: Run in parallel!

test/regression/tests_utils.py
```diff
@@ -70,48 +71,3 @@ def get_default_copy_cmd(db_path, filename, experiment_id):
...
-def base36decode(number):
-    """
-    Converts a base36 string to a positive integer
-
-    :param number: base36 string to convert
-    :type number: str
-    :return: number's integer value
-    :rtype: int
-    """
-    return int(number, 36)
```
kinow, line 117: Duplicate from `experiment_common`. Deleted the copy in the test file, used the real code to test (not sure why it was done the other way).

kinow:
@dbeltrankyl ready for review.

The current coverage warning is for changes in joblist where I added types to the functions to resolve IDE warnings in the test code.

I re-enabled integration tests. So the `conftest.py` was moved higher to the `test` folder to be shared by both test types.

Even then, the coverage increase is minimal (what shows @VindeeR has been doing tremendous work as he increased +2% over last days), but better than nothing :slightly_smiling_face: poco a poco.

image

A second commit is deleting code that's not used. @dbeltrankyl to review this, I think a simpler approach is:

- quickly look over test code to see if you spot anything strange
- use coverage on `master` and this branch to verify things look about the same (hopefully, better here)
- look at the source code modified and confirm with IDE everything looks in order

If the code removal changes are too much, I can edit the second commit and add back parts of the code, or everything that was removed, and move to a separate commit.

Also, please try running `pytest` or `pytest test/unit` and `pytest test/integration`. The latter may fail as it's relying on Docker and I think there's something else missing with the way I'm binding directories (for now at least it's working fine on GH Actions). With xdist I expect the tests to run much faster.

:vulcan_salute:

kinow:
I've added a few more tests, for https://github.com/BSC-ES/autosubmit-api/issues/175, using the output provided by @sergipalomas in that issue.  If there are no issues in CICD, I expect the coverage to increase (even if a tiny bit).

kinow:
It was increasing ~0.8%, now the coverage is increased in 1.08%. Little by little :slightly_smiling_face:

dbeltrankyl:
Hello @kinow 

> Also, please try running pytest or pytest test/unit and pytest test/integration. The latter may fail as it's relying on Docker and I think there's something else missing with the way I'm binding directories (for now at least it's working fine on GH Actions). With xdist I expect the tests to run much faster.


I wanted to test this before reviewing anything ( because it will be easier for me ) 

Before,
 I was running them just with Pycharm and right click -> run
image

```python
/home/dbeltran/.venv39/bin/python /home/dbeltran/pycharm/plugins/python-ce/helpers/pycharm/_jb_pytest_runner.py --path /home/dbeltran/GIT_autosubmit/test/unit/test_autosubmit_git.py -- --no-cov 
Testing started at 2:09 PM ...
Launching pytest with arguments --no-cov /home/dbeltran/GIT_autosubmit/test/unit/test_autosubmit_git.py --no-header --no-summary -q in /home/dbeltran/GIT_autosubmit/test/unit

ERROR: usage: _jb_pytest_runner.py [options] [file_or_dir] [file_or_dir] [...]
_jb_pytest_runner.py: error: unrecognized arguments: -n --max-worker-restart 0 --dist loadgroup /home/dbeltran/GIT_autosubmit/test/unit/test_autosubmit_git.py
  inifile: /home/dbeltran/GIT_autosubmit/pytest.ini
  rootdir: /home/dbeltran/GIT_autosubmit


Process finished with exit code 4

Empty suite
```

But in this branch it is complaining. Do I have to configure something before?

dbeltrankyl:
> Or maybe it is some pytest package that I'm missing...

Yes, it was it. It seems that Pycharm only detects as missing the mandatory packages, not the optional ones for test/doc

dbeltrankyl:
@kinow 

Only test_send_file not passing:


```python
self = <docker.api.client.APIClient object at 0x7f175d00b790>

    def _retrieve_server_version(self):
        try:
            return self.version(api_version=False)["ApiVersion"]
        except KeyError as ke:
            raise DockerException(
                'Invalid response from docker daemon: key "ApiVersion"'
                ' is missing.'
            ) from ke
        except Exception as e:
          raise DockerException(
                f'Error while fetching server API version: {e}'
            ) from e
          docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```

kinow:
> Btw not related, but I've uninstalled and installed autosubmit again with pip install -e .[all] just in case..
> 
> Without adding this to the [docs] section "docutils>=0.21"
> 
> It complains about:
> 
> ```
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> readme-renderer 44.0 requires docutils>=0.21.2, but you have docutils 0.19 which is incompatible.
> ```

That's news to me. Maybe it's something new in setuptools/pip. But from the error message looks like it's complaining about inconsistency in our dependencies (transitive, doesn't seem to be something we are using directly). Let me add that to our docs group.

kinow:
Couldn't reproduce this error after doing,

```bash
$ pip uninstall autosubmit autosubmitconfigparser
$ pip install -U packaging pip twine
$ pip install -U -e .[all]
```

With

```
(autosubmit4) kinow@ranma:~/Development/python/workspace/autosubmit$ python -V
Python 3.9.0
```

The only place where `readme_renderer` is used, according to `pipdeptree`:

```bash
twine==6.1.0
├── readme_renderer [required: >=35.0, installed: 44.0]
│   ├── nh3 [required: >=0.2.14, installed: 0.2.21]
│   ├── docutils [required: >=0.21.2, installed: 0.21.2]
```

@dbeltrankyl what happens if you try this again

```bash
$ pip uninstall autosubmit autosubmitconfigparser
$ pip install -U packaging twine pip
$ pip install -U -e .[all]
```

kinow:
Ok, looking at the logs of Docker on my system when running the test:

```bash

```

This can be solved by something like

```bash
sudo setfacl --modify user:$USER:rw /var/run/docker.sock
```

to give my user access to the Docker socket, so testcontainers is able to communicate with the Docker API and launch the containers. But this should be optional, IMHO, as not all developers will want to run tests with Dockers, which may download some MB/GB of data.

Let me create a mark/group for docker, and enable so that GitHub will run unit + integration + integration[docker]. I'll add docs too to developers that want to run the integration tests with Docker.

dbeltrankyl:
> $ pip uninstall autosubmit autosubmitconfigparser
> $ pip install -U packaging twine pip
> $ pip install -U -e .[all]

I did everything from 0, and it works, sorry for the false alarm

autosubmit/history/platform_monitor/slurm_monitor.py
```diff
@@ -75,8 +75,4 @@ def extern(self):
   def steps_plus_extern_approximate_header_energy(self):
     return abs(self.steps_energy + self.extern.energy - self.header.energy) <= 0.01*self.header.energy
 
-  def print_items(self):
-    for item in self.input_items:
-      print(item)  
-
```
kinow, line 81: Not used.

kinow:
Pytest marker added, tests working locally after I enabled access to the Docker socket. But now it's failing with the same error you had earlier today, @dbeltrankyl .

I tried removing packaging/setuptools from our list of dependencies, and also updating setuptools/packaging/pip. But at least in GH Actions environment it doesn't seem to fix the build problem.

I suspect other PRs should be affected by that over next days too... will keep looking at this issue a bit more today, but if I cannot fix it I will leave for tomorrow morning + :coffee: 

image

pyproject.toml
```diff
@@ -47,17 +45,14 @@ classifiers = [
...
     "bscearth.utils<=0.5.2",
     "requests<=2.32.3",
     "networkx<=2.6.3",
     "portalocker<=2.7.0",
     "paramiko>=3.5.0",
     "pyparsing==3.2.1",
     "matplotlib<=3.8.3",
-    "packaging<=24.2",
     "typing_extensions==4.*",
-    "typing<=3.7.4.3",
```
kinow, line 60: Typing is for compatibility with older versions of Python, safe to remove as we are using Python 3.9 now, and typing has been part of stdlib for a while.

pyproject.toml
```diff
@@ -47,17 +45,14 @@ classifiers = [
...
     "xlib==0.21",
-    "setuptools<=77.0.3",
     "bscearth.utils<=0.5.2",
     "requests<=2.32.3",
     "networkx<=2.6.3",
     "portalocker<=2.7.0",
     "paramiko>=3.5.0",
     "pyparsing==3.2.1",
     "matplotlib<=3.8.3",
-    "packaging<=24.2",
```
kinow, line 58: Build dependency, not a runtime dependency.

pyproject.toml
```diff
@@ -47,17 +45,14 @@ classifiers = [
 
 dependencies = [
     "xlib==0.21",
-    "setuptools<=77.0.3",
```
kinow, line 50: Build dependency, not a runtime dependency.

kinow:
Alright, tried removing setuptools & packaging, nada. Then adding license_file in setup.cfg, also nothing. Finally, tried everything, adding license_file in setup.cfg, and removing the runtime dependencies from setup.py (packaging/setuptools and also the outdated typing).

Finally it passed GH Actions build! :sleepy: :wave:

kinow:
Ready for review again, @dbeltrankyl . Have a look at `CONTRIBUTING.md` for instructions on how to run the tests, or just try

- `pytest` unit tests
- `pytest test/integration` for integration tests
- `pytest test/integration -m ''` if you want to run the docker tests (you will have to `setfacl` for the Docker socket as documented in `CONTRIBUTING.md`

kinow:
Reading the night feed of news: "Setuptools 78.0.1 breaks the Internet" on [reddit](https://old.reddit.com/r/Python/comments/1jiy2sm/setuptools_7801_breaks_the_internet/).

Here's the issue so we have the link to it here too: https://github.com/pypa/setuptools/issues/4910

dbeltrankyl:
Thanks @kinow

First of all, it is very fast! Impressive.

Testing first without containers:

test_catlog->test_is_workflow_not_found

```python
mocker = <pytest_mock.plugin.MockerFixture object at 0x7f6efa7be550>
autosubmit = <autosubmit.autosubmit.Autosubmit object at 0x7f6efa7be130>

    def test_is_workflow_not_found(mocker, autosubmit):
        mocked_log = mocker.patch('autosubmit.autosubmit.Log')
        autosubmit.cat_log(_EXPID, 'o', 'c')
>       assert mocked_log.info.called
E       AssertionError: assert False
E        +  where False = <MagicMock name='Log.info' id='140114625616144'>.called
E        +    where <MagicMock name='Log.info' id='140114625616144'> = <MagicMock name='Log' id='140114625552736'>.info

```

This doesn't always fail when running all the unit or test folder. However, it always happens if I only run the test_catlog.py or only that test 


image

On the master branch, it is always working

--------------------------------------------------

Another bizarre stuff is that, under this branch, I'm seeing visions of logs ( of my a000 experiment )

```
2025-03-18 12:51:55,265 
Jobs ready for MARENOSTRUM5-LOGIN: 0
2025-03-18 12:51:55,267 Number of jobs available: 235
2025-
2025-03-18 12:51:55,265 
Jobs ready for MARENOSTRUM5-LOGIN: 0
2025-03-18 12:51:55,267 Number of jobs available: 235
2025-
test/unit/test_catlog.py:78 (test_is_workflow_not_found)
mocker = <pytest_mock.plugin.MockerFixture object at 0x7fe06e68c8b0>
autosubmit = <autosubmit.autosubmit.Autosubmit object at 0x7fe099a53ee0>

    def test_is_workflow_not_found(mocker, autosubmit):
        mocked_log = mocker.patch('autosubmit.autosubmit.Log')
        autosubmit.cat_log(_EXPID, 'o', 'c')
>       assert mocked_log.info.called
E       AssertionError: assert False
E        +  where False = <MagicMock name='Log.info' id='140601902111568'>.called
E        +    where <MagicMock name='Log.info' id='140601902111568'> = <MagicMock name='Log' id='140601900740912'>.info

unit/test_catlog.py:82: AssertionError
[CRITICAL] Chunk unit not valid: other

```

Again, this on master doesn't happen 🤔; I just saw that the test below is giving me issues. Perhaps it is looking at real experiments?

kinow:
@dbeltrankyl 

>Perhaps it is looking at real experiments?

Yes, it is! Yuck! Let me fix that. Great catch, Dani! Not sure how that's been working fine so far, my bad.

CONTRIBUTING.md
```diff
@@ -2,10 +2,100 @@
...
+preferably being able to run `docker` without `sudo` (i.e. adding your `$USER`
+to the `docker` group), and you also need access to the Docker socket.
+
+> NOTE: the access to the Docker socket is required as the test library
+>       Testcontainers will query the Docker API to instantiate/edit/destroy
+>       containers on-the-fly, and without that access the tests will fail.
+
+```bash
+# Grant permission to your `$USER` to read and write to the Docker socket
+$ sudo setfacl --remove user:$USER:rw /var/run/docker.sock
```
dbeltrankyl, line 86: The docker works! But there is a mistake here

sudo setfacl --remove user:$USER:rw /var/run/docker.sock

should be

sudo setfacl -m user:$USER:rw /var/run/docker.sock

From docs: 

       The  -m  (--modify)  and  -M (--modify-file) options modify the ACL of a file or directory.  ACL entries for
       this operation must include permissions.

       The -x (--remove) and -X (--remove-file) options remove ACL entries. It is not an error to remove  an  entry
       which  does  not  exist.   Only  ACL  entries  without  the  perms  field are accepted as parameters, unless
       POSIXLY_CORRECT is defined.

.gitignore
dbeltrankyl: Thanks, I kept adding --no-cov everywhere

autosubmit/autosubmit.py
```diff
@@ -3171,14 +3170,6 @@ def check(experiment_id, notransitive=False):
 
         return job_list.check_scripts(as_conf)
 
-    @staticmethod
```
dbeltrankyl, line 3174: This was ( long ago) moved to config parser ( and I think it doesn't even exists as it were )

autosubmit/history/data_classes/job_data.py
```diff
@@ -295,29 +262,3 @@ def queuing_time(self):
         if self.status in ["SUBMITTED", "QUEUING", "RUNNING", "COMPLETED", "HELD", "PREPARED", "FAILED", "SKIPPED"]:
             return HUtils.calculate_queue_time_in_seconds(self.submit, self.start)
         return 0
-
-    def get_hdata(self):
```
dbeltrankyl, line 299: I didn't know we had json function for this, tbh. I guess this was used before I arrived

autosubmit/history/platform_monitor/slurm_monitor.py
dbeltrankyl: This looks like a debug function that somebody was using whenever he/she wanted to debug that part. It is okay to remove from my side

autosubmit/job/job_list.py
```diff
@@ -2235,19 +2148,6 @@ def get_delayed(self, platform=None):
                     job.platform.name == platform.name) and job.status == Status.DELAYED]
         return delayed
 
-    def get_skipped(self, platform=None):
```
dbeltrankyl, line 2238: These get_{job_status} of lists are not used right now.. but maybe it is necessary when we load only the status of the jobs.. anyway I can just look to older commits to recover it if needed.

autosubmit/platforms/headers/pjm_header.py
```diff
@@ -111,19 +111,6 @@ def get_custom_directives(self, job, parameters):
             return '\n'.join(str(s) for s in parameters['CUSTOM_DIRECTIVES'])
         return ""
 
-    def get_tasks_directive(self, job, het=-1):
```
dbeltrankyl, line 114: Hmm, is this not used? strange. 

Ah, right, pjm uses shape instead of tasks, I think.

autosubmit/platforms/slurmplatform.py
```diff
@@ -382,150 +382,6 @@ def get_remote_log_dir(self):
     def parse_job_output(self, output):
         return output.strip().split(' ')[0].strip()
 
-    def parse_job_finish_data(self, output, packed):
```
dbeltrankyl, line 385: Weren't there reports about the energy metric not being extracted?

dbeltrankyl (commented):
Great work @kinow !

The code removed looks safe to remove some small comments, and that's it.

I briefly reviewed the code related to the tests, but I've also tested them, and aside from one that I've already commented on, they're swift and working. 

There is a mistake in the contibutions.md, and once it is fixed and all tests are passing, we can merge it.

The next step for our tests ( and other PRs) is cleaning. We have a lot of integration ones that I believe we never run, and in the regression tests, some only work under my computer ( that I expect to adapt them in the near future )

CONTRIBUTING.md
kinow: Ops! Eagle eyes! Fixed. Thanks, Dani.

autosubmit/job/job_list.py
kinow: Agreed. I thought the same, and the bonus point is that if we re-add it, we will have the perfect excuse to write tests for ti too.

autosubmit/platforms/headers/pjm_header.py
kinow: Not anywhere in our current code. I removed the code that was not used in `./autosubmit`, including the tests for some (i.e. we had code tested, but never used; can't recall if this one if an example of that).

autosubmit/platforms/slurmplatform.py
kinow: Yeah, but I think those are computed in `slurm_monitor.py`.

This function has this code (which I cannot find used anywhere, but as example):

image

But from working a bit on https://github.com/BSC-ES/autosubmit-api/issues/175, looks like the code in AS that reports energy metric is actually the code from slurm_monitor/slurm_monitor_item,

image

kinow:
Thank you for the thorough review, @dbeltrankyl . I fixed the CONTRIBUTING.md mistake you found, and adjusted the `test_catlog` to always use the fake as config (`autouse=True` in the as_conf function to always have a fake `AutosubmitConfig` for each test function).

I also resolved some comments, but left unresolved a few ones where I replied (in case you want to have a look).

Thanks Dani!

kinow:
And now removed the `setup.cfg` as I **hope** the setuptools team fixed their issue, and this new version (reported by dependabot) should fix that `license-files` / `license_files` build problem. Just waiting for GH Actions.

kinow:
All good, ready for review @dbeltrankyl

autosubmit/platforms/slurmplatform.py
dbeltrankyl: Ok I wasn't sure. 

I was wondering that when I moved the stat function code to work within the log _process I missed something about  the energy ( as this function is the left out from that time, I think)

Thanks for checking

dbeltrankyl:
Thanks @kinow 

All good! all tests are passing now ( 50 seconds), and no more strange logs

dbeltrankyl (approved):

kinow:
Thanks, @dbeltrankyl ! Let's get it merged then (and have faster tests!).

Issue #2231, opened 2025-03-18
Fail or warn pip install if Python 3.8 is used


manuel-g-castro:
Hello,

Could you add a deprecation or constraint to the setup phase of the pipy package to avoid installing it with python 3.8, which is unsupported?

I was creating a test environment for Autosubmit within a Docker image (`ubuntu:20.04`), which comes with python 3.8 by default. 

The installation went smooth, but running an `autosubmit --version` would fail with this nasty error:

```
root@b49b35c1975c:/# autosubmit --version
Traceback (most recent call last):
  File "/usr/local/bin/autosubmit", line 30, in <module>
    from autosubmit.autosubmit import Autosubmit
  File "/usr/local/lib/python3.8/dist-packages/autosubmit/autosubmit.py", line 76, in <module>
    from importlib.resources import files as read_files
ImportError: cannot import name 'files' from 'importlib.resources' (/usr/lib/python3.8/importlib/resources.py)
```

For reference this is my environment:

```
root@b49b35c1975c:/# pip list     
Package                Version           
---------------------- ------------------
aiohappyeyeballs       2.4.4             
aiohttp                3.10.11           
aiosignal              1.3.1             
arcp                   0.2.1             
async-timeout          5.0.1             
attrs                  25.3.0            
autosubmit             4.1.11            
autosubmitconfigparser 1.0.73            
bcrypt                 4.3.0             
bioblend               1.4.0             
bscearth.utils         0.5.2             
cachecontrol           0.14.2            
certifi                2025.1.31         
cffi                   1.17.1            
charset-normalizer     3.4.1             
click                  8.1.8             
configobj              5.0.9             
configparser           7.1.0             
contourpy              1.1.1             
cryptography           44.0.2            
cycler                 0.12.1            
filelock               3.16.1            
files                  1.1.1             
fonttools              4.56.0            
frozenlist             1.5.0             
galaxy2cwl             0.1.4             
gxformat2              0.20.0            
idna                   3.10              
importlib              1.0.4             
importlib-resources    6.4.5             
invoke                 2.2.0             
isodate                0.7.2             
jinja2                 3.1.6             
kiwisolver             1.4.7             
MarkupSafe             2.1.5             
matplotlib             3.7.5             
mistune                3.0.2             
mock                   5.2.0             
msgpack                1.1.0             
multidict              6.1.0             
mypy-extensions        1.0.0             
networkx               2.6.3             
numpy                  1.24.4            
packaging              23.2              
[...]

PR #2240, opened 2025-03-20
Specify python version and update command line

manuel-g-castro:
This solves #2231.


kinow (approved):
We should automate it someday. But for now I think this reduces the chance of users running it with 3.8 after reading just that installation page. Thanks @manuel-g-castro !

PR #2241, opened 2025-03-21
Fixes an issue with some jobs being parentless 

dbeltrankyl:
This PR does two things:

Removes "changed" related code, as the dependencies are remade each time the autosubmit needs the job_list. 

And, if there is a parent_less job after all the dependencies addition, autosubmit will add all possible parents as an edge and remove it in an old-fashioned way ( check redundant dependencies ). 

The calculation of the "distance" between sections has some flaws, but they are hard to fix. This workaround will cause the graph dependencies addition to be a bit slow in these cases, but it will at least generate the correct graph. 

I'll make some tests that check the graph generated

PR #2255, opened 2025-03-31
Update CI to parallel test execution

LuiggiTenorioK:
Update to make the CI run unit, integration, and regression tests in parallel.

This PR supports #2241.

To avoid creating too many jobs, tests have been set to only Python 3.12 for integration and regression tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,6 +64,43 @@ jobs:
             --cov-report=xml:test/coverage.xml --cov-append \
             test/unit
 
+      - name: Coverage report
+        run: |
+          coverage xml
+          coverage report
+
+      - name: Upload coverage artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage_unit_py-${{ matrix.python-version }}
+          path: coverage.xml
+          retention-days: 7
+
+  test-integration:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    strategy:
+      matrix:
+        python-version: ["3.12"]
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


LuiggiTenorioK:
@dbeltrankyl here is the CI updated to run in parallel.

Please tell me if something is wrong with the coverage or execution :pray:

dbeltrankyl:
Hello @LuiggiTenorioK 

Thanks!

image

I think they're only running for python 3.12

LuiggiTenorioK:
> I think they're only running for python 3.12.

Yes, I set it on purpose. I'll set this up to use the 4 Python versions.

LuiggiTenorioK:
It is now ready and using all Python versions :heavy_check_mark:

dbeltrankyl:
> It is now ready and using all Python versions ✔️

Thanks @LuiggiTenorioK !

We can merge then. 

I have one additional question but I guess that this is not possible: 

Is this schema possible?

```mermaid
graph TD
    subgraph Python 3.9
        A1[setup python] --> A2[install system dependencies]
        A2 --> A3[tests]
        A3 --> A3a[unit]
        A3 --> A3b[integration]
        A3 --> A3c[regression]
        A3a --> A4[coverage report]
        A3b --> A4[coverage report]
        A3c --> A4[coverage report]
        A4 --> A5[upload coverage]
        A5 --> A6[post set up]
        A6 --> A7[post checkout]
        A7 --> A8[complete job]
    end
```

dbeltrankyl (approved):

LuiggiTenorioK:
> Is this schema possible? ( so only the part of the test code is running in parallel)

I think it is not easy to do it since the job environments are isolated without an option to have a shared one. Probably, we can cache some dependencies, but it doesn't seem feasible at first sight.

PR #2256, opened 2025-04-01
Removing unused function from parent class

VindeeR:
- Cleaning code 
- More readable

### DIFF
--- a/autosubmit/platforms/paramiko_platform.py
+++ b/autosubmit/platforms/paramiko_platform.py
@@ -230,12 +230,6 @@ class ParamikoPlatform(Platform):
                     if twofactor_nonpush is None:
                         twofactor_nonpush = input("Please type the 2FA/OTP/token code: ")
                     answers.append(twofactor_nonpush)
-        # This is done from the server
-        # if self.two_factor_method == "push":
-        #     try:
-        #         inputimeout(prompt='Press enter to complete the 2FA PUSH authentication', timeout=self.otp_timeout)
-        #     except:
-        #         pass
         return tuple(answers)
 
     def map_user_config_file(self, as_conf) -> None:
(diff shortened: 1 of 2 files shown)


autosubmit/platforms/ecplatform.py
```diff
@@ -37,18 +37,6 @@ class EcPlatform(ParamikoPlatform):
...
-
-    def parse_queue_reason(self, output, job_id):
-        pass
-
-    def get_checkAlljobs_cmd(self, jobs_id):
-        pass
-
-    def submit_Script(self, hold=False):
-        pass
-
```
kinow, line 51: I believe the idea here is exactly to have these functions so that calling them won't call the parent class. If `ECPlatform` doesn't have `parse_Alljobs_output`, for instance, I think if `ecplatform_instance.parse_Alljobs_output()` is called, then it would call `parse_Alljobs_output` from the `ParamikoPlatform`. And instead of `pass`, that function would raise a `NotImplementedError`, which could cause AS to crash.

Can you confirm that, @VindeeR ? If that's really the case, then I believe we don't want to remove these functions...

kinow: It is how it was designed :shrug: I think it's safer to leave as it was before.

autosubmit/platforms/pbsplatform.py
```diff
@@ -78,9 +69,6 @@ def __init__(self, expid, name, config, version):
     def create_a_new_copy(self):
         return PBSPlatform(self.expid, self.name, self.config, self.config[self.name.upper()].get('VERSION', ''))
 
-    def parse_queue_reason(self, output, job_id):
-        pass
```
kinow, line 82: It's implemented in the parent. Not sure why they had a pass here... but probably safer to leave it as it was (PBS is broken anyway, this whole file can be reverted)

PR #2259, opened 2025-04-02
padronize vulture and pylint

VindeeR:
Creating files to standardize the vulture and pylint configurations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ build/
 # Coverage files
 coverage/
 .coverage
+.coverage.*
 test/coverage.xml
 coverage.xml
 test/htmlcov
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,650 @@
+[MAIN]
+
+# Analyse import fallback blocks. This can be used to support both Python 2 and
+# 3 compatible code, which means that the block might have code that exists
+# only in one or another interpreter, leading to false positives when analysed.
+analyse-fallback-blocks=no
+
+# Clear in-memory caches upon conclusion of linting. Useful if running pylint
+# in a server-like mode.
+clear-cache-post-run=no
+
+# Load and enable all available extensions. Use --list-extensions to see a list
+# all available extensions.
+#enable-all-extensions=
+
+# In error mode, messages with a category besides ERROR or FATAL are
+# suppressed, and no reports are done by default. Error mode is compatible with
+# disabling specific errors.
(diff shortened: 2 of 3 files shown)


isimo00:
Can you add a description please? :)

.pylintrc
```diff
@@ -0,0 +1,650 @@
...
+check-protected-access-in-special-methods=no
+
+# List of method names used to declare (i.e. assign) instance attributes.
+defining-attr-methods=__init__,
+                      __new__,
+                      setUp,
+                      asyncSetUp,
+                      __post_init__
+
+# List of member names, which should be excluded from the protected access
```
dbeltrankyl, line 274: Related to this commented on #2284
fyi @kinow 
I think that we shouldn't have complaints about protected stuff in test methods, as they may require us to modify where they're pointing to ( or we just need to test them )

kinow: Normally private functions have to be tested via the public functions that call them. When that's not doable, it's normally an indicator of code that can be redesigned/simplified - e.g. (From 2023, but I first read about it some time ago) https://www.baeldung.com/cs/testing-public-vs-private-methods

> Well-designed public methods should exercise all the critical paths of the private methods, ensuring that the private logic is validated through the public interface.

Testing private code directly is fine too, but in general it's better to do that as an exception to the rule; and I believe that the rationale for linters screaming when they see it first, unless devs have explicitly told it to ignore them (like in this list).

.gitignore
```diff
@@ -26,3 +26,5 @@ dockerfiles/id_rsa*
 dockerfiles/authorized_keys
 
 venv/
+
+whitelist.py
```
kinow, line 30: This file is not in this PR. Is it generated when the linter runs?

kinow (commented):
Haven't tested it yet, but looking good. Looks like you've put some great effort on the linter RC file. :clap: Thanks!

kinow (approved):
Looks good to me! @dbeltrankyl has a question about protected, so maybe Erick can discuss it with him before merging it.

Dani, the changes here won't break the build. Vulture and PyLint are not running in GH Actions yet. So we could merge it and refine it later. I think Erick created this PR as he was investigating these tools (we could even switch to ruff/etc later). So it's not set on stone yet :+1:

dbeltrankyl (approved):
Fine for me. Didn't read the Bruno answers until now. Thanks!

PR #2260, opened 2025-04-02
Remove --init from docker README

LuiggiTenorioK:
There is a warning when following the README file of Docker:

```
[WARN  tini (7)] Tini is not running as PID 1 and isn't registered as a child subreaper.
Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.
```

This is because when you run `docker run --init`, Docker will use `docker-init` as the init process instead of the one defined by the image, which in our case is `tini`. We don't have to use the `-init` flag to avoid that warning.

This is not critical since `docker-init` is backed up by `tini` but it is nice to not have this warning anymore.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -52,7 +52,7 @@ can be used to submit jobs to the other nodes (as it is normally
 configured with VM's submitting jobs to remote Slurm, PBS, etc.).
 
 ```bash
-$ docker run --rm --init ${USER}/autosubmit:latest \
+$ docker run --rm ${USER}/autosubmit:latest \
   autosubmit --version
 ```
 
@@ -113,7 +113,7 @@ or remote hosts (always via `ssh`).
 To start with SSH (see prerequisites note about SSH keys):
 
 ```bash
-$ docker run --rm --init \
+$ docker run --rm \
   -ti \
   -p 2222:22 \
   -e DISPLAY=$DISPLAY \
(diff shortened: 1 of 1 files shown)


VindeeR:
I just run a test following the `readme` instruction and I ran into the same problem

image

is there anything else that needs to be done?

LuiggiTenorioK:
> I just run a test following the `readme` instruction and I ran into the same problem
> 
> is there anything else that needs to be done?

The idea of this PR is that you have to run it without the `--init` flag. For example, with the flag:

```bash
$ docker run --rm --init -v /home/ltenorio/autosubmit:/app/autosubmit/database -v /home/ltenorio/autosubmit:/app/autosubmit/experiments autosubmit:4.1.13 ps -p 1
[WARN  tini (7)] Tini is not running as PID 1 and isn't registered as a child subreaper.
Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.
Restarting OpenBSD Secure Shell server: sshd.
  PID TTY          TIME CMD
    1 ?        00:00:00 docker-init
```

And without the flag:

```bash
$ docker run --rm -v /home/ltenorio/autosubmit:/app/autosubmit/database -v /home/ltenorio/autosubmit:/app/autosubmit/experiments autosubmit:4.1.13 ps -p 1
Restarting OpenBSD Secure Shell server: sshd.
  PID TTY          TIME CMD
    1 ?        00:00:00 tini
```

Without the flag, `tini` is the process with PID 1 (init process) instead of `docker-init`.

VindeeR:
OOOPPPSS 

image

that was just me running the wrong command it's all good

VindeeR (approved):

PR #2262, opened 2025-04-02
fixing f-string the error

VindeeR:
pylint was pointing some f-string lint error `C0209` in the job_list.py

Where strings where written `"{1}".format(var)` and were converted to `f"{var}"` allowing for a slimmer and more readable code


isimo00 (approved):
Could you add a more extensive description please?

PR #2284, opened 2025-04-08
Timestamps fixes

### LINKED ISSUE #2275: Wrong submit time [job_data.db->job_data table -> submit time]
Hello

There is an issue with the submit time being updated when it shouldn't be, which leads to storing an incorrect timestamp at some point.

Strangely, this doesn't happen in the integration test.

dbeltrankyl:
Instead of getting the job start and finish timestamp from the timestamp files, Autosubmit was always using the fallback values.

~This was happening only when job.fail_count == 0~ 

EDIT: For retrials, it was looking at tmp folder instead of tmp/Log_$expid

Closes #2275

### DIFF
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -1143,10 +1143,10 @@ class Job(object):
             int: Value in the index position, or 0 if the file or index does not exist.
         """
         if fail_count == -1:
-            logname = os.path.join(self._tmp_path, f"{self.stat_file}0")
+            logname = os.path.join(self._log_path, f"{self.stat_file}0")
         else:
             fail_count = str(fail_count)
-            logname = os.path.join(self._tmp_path, f"{self.stat_file}{fail_count}")
+            logname = os.path.join(self._log_path, f"{self.stat_file}{fail_count}")
         if os.path.exists(logname):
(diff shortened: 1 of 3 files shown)


dbeltrankyl:
About the coverage, the missing lines are being tested by the integration test.

dbeltrankyl:
I want to do a pre-release soon (preferably today ) basically because of this fix. Could you please review it? @VindeeR , @isimo00 ?

VindeeR:
Just the leave it here as e memo but when I tried to checkout to this branch I was having a problem due to the name of the branch :laughing:

dbeltrankyl:
> Just the leave it here as e memo but when I tried to checkout to this branch I was having a problem due to the name of the branch 😆

Ah, it was complaining about >? 

I admit that this branch doesn't have a good name ( I created it before the issue 😅 )

dbeltrankyl:
> pylint regarding protected member on the test/unit/test_job.py

Is it complaining about job._log_path? 

I don't have that complaint in Ruff, nor is it in the ide. 

But I think we should be able to call protected_members from a test file

dbeltrankyl:
Thanks for the feedback! merging, we can discuss the lint in another issue

PR #2293, opened 2025-04-10
Add operational experiments remote repository check before running

### LINKED ISSUE #2220: Forbid running operational experiments when there are changes in the project
As discussed yesterday, we would like to implement this feature. It has been introduced to the OC and so far there hasn't been any negative response.

The idea is to introduce a check in the `autosubmit run` for operational EXPIDs, similar to the one we have in `clean` or `archive` and make Autosubmit fail if there are changes in the project, that have not been pushed to Git or SVN.

Importantly, this should also check if there are changes in the submodules. I don't remember how strict the check was.

isimo00:
This PR introduces an additional check in `autosubmit run` to ensure that, if the experiment is operational, all local changes have been pushed to the remote repository before running

Closes #2220

### DIFF
--- a/autosubmit/git/autosubmit_git.py
+++ b/autosubmit/git/autosubmit_git.py
@@ -261,7 +261,14 @@ class AutosubmitGit:
         return True
 
     @staticmethod
-    def check_unpushed_changes(expid):
+    def check_unpushed_changes(expid: str) -> None:
+        """
+        Raises an AutosubmitCritical error if the experiment is operational, the platform is Git, and there are unpushed changes.
+
+        Args: expid (str): The experiment ID.
+
+        Returns: None
+        """
         if expid[0] == 'o':
             origin = Path(BasicConfig.expid_dir(expid).joinpath("conf/expdef_{}.yml".format(expid)))
             with open(origin, 'r') as f:
@@ -270,14 +277,13 @@ class AutosubmitGit:
                 project = data["PROJECT"]["PROJECT_TYPE"]
 
             version_controls = ["git", 
-                                "git",
-                                "svn"]
+                                "git submodule"]
             arguments = [["status", "--porcelain"],
-                        ["submodule", "foreach", "'git status --porcelain'"],
-                        "status"]
+                        ["foreach", "'git status --porcelain'"]]
             for version_control, args in zip(version_controls, arguments):
                 if project == version_control:
                     output = subprocess.check_output([version_control, args]).decode(locale.getlocale()[1])
                     if any(status.startswith(code) for code in ["M", "A", "D", "?"] for status in output.splitlines()):
+                        # M: Modified, A: Added, D: Deleted, ?: Untracked
                         raise AutosubmitCritical("Push local changes to remote repository before running", 7075)
 
--- a/test/unit/test_expid.py
+++ b/test/unit/test_expid.py
@@ -26,6 +26,7 @@ from textwrap import dedent
 from ruamel.yaml import YAML
 
 import pytest
+from pytest_mock import MockerFixture
 from mock import Mock, patch
 
 from autosubmit.autosubmit import Autosubmit
@@ -480,14 +481,13 @@ def test_perform_deletion(create_autosubmit_tmpdir, generate_new_experiment, set
         ('git', 'normal'),
         ('git', 'test'),
         ('git', 'evaluation'),
-        ('svn', 'operational'),
-        ('svn', 'normal'),
-        ('svn', 'test'),
-        ('svn', 'evaluation')
     ],
     indirect=["generate_new_experiment"]
 )
-def test_remote_repo_operational(generate_new_experiment, create_autosubmit_tmpdir, project_type, mocker):
+def test_remote_repo_operational(generate_new_experiment: str, create_autosubmit_tmpdir: str, project_type: str, mocker: MockerFixture) -> None:
+    '''
+    Tests the check_unpushed_changed function from AutosubmitGit, which ensures no operational test with unpushed changes in their Git repository is run.
+    '''
     expid = generate_new_experiment
     temp_path = Path(create_autosubmit_tmpdir) / expid / "conf" / f"expdef_{expid}.yml"
     


autosubmit/autosubmit.py
```diff
@@ -723,6 +723,20 @@ def run_command(args):
...
 
         if args.command == 'run':
+            if expid[0] == 'o':
+                origin = Path(BasicConfig.expid_dir(expid).joinpath("conf/expdef_{}.yml".format(expid))) 
+                with open(origin, 'r') as f:
+                    yaml = YAML(typ='rt')
+                    data = yaml.load(f)
+                    project = data["PROJECT"]["PROJECT_TYPE"]
+                if project == "git":
+                    output = subprocess.check_output(["git", "status"]).decode()
```
dbeltrankyl, line 733: Can you use .decode(locale.getlocale()[1])?

dbeltrankyl: Or this, ( this maybe should be a function. This code is in paramiko_platform only

```python
lang = locale.getlocale()[1]
        if lang is None:
            lang = locale.getdefaultlocale()[1]
            if lang is None:
                lang = 'UTF-8'
                
```

docs/source/troubleshooting/error-codes.rst
```diff
@@ -208,6 +208,9 @@ team in Git.
 +------+-----------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | 7074 | Profiling process failed                      | You can find more detailed information in the logs, as well as hints to solve the problem                                                                                        |
 +------+-----------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 7075 | Local changes have not been pushed to remote Git or SVN repo   | Commit and push locla changes before running the experiment
```
dbeltrankyl, line 211: locla -> local

test/unit/test_expid.py
```diff
@@ -15,13 +15,15 @@
...
 import pwd
 import sqlite3
 import tempfile
 from itertools import permutations, product
 from pathlib import Path
 from textwrap import dedent
+from ruamel.yaml import YAML
 
 import pytest
 from mock import Mock, patch
```
dbeltrankyl, line 29: Can you use the mocker fixture from pytest?

test/unit/test_expid.py
```diff
@@ -468,3 +470,33 @@ def test_perform_deletion(create_autosubmit_tmpdir, generate_new_experiment, set
     assert all(x in err_message for x in
                ["Cannot delete experiment entry", "Cannot delete directory", "Cannot delete structure",
                 "Cannot delete job_data"])
+
+
+@pytest.mark.parametrize("project_type", ['git', 'svn'])
```
dbeltrankyl, line 475: These mark.parametrize can be a single one

I think the syntax is this one:

```python
@pytest.mark.parametrize(
    "project_type, generate_new_experiment",
    [
        ('git', 'operational'),
        ('svn', 'operational')
    ],
    indirect=["generate_new_experiment"]
)
```

test/unit/test_expid.py
```diff
@@ -468,3 +470,33 @@ def test_perform_deletion(create_autosubmit_tmpdir, generate_new_experiment, set
...
+@pytest.mark.parametrize("generate_new_experiment", ['operational'], indirect=True)
+def test_remote_repo_operational(generate_new_experiment, create_autosubmit_tmpdir, project_type):
```
dbeltrankyl, line 477: I would also add test for the other types (testing, normal, evaluation) just to ensure that they work as they were before

test/unit/test_expid.py
```diff
@@ -468,3 +470,33 @@ def test_perform_deletion(create_autosubmit_tmpdir, generate_new_experiment, set
...
+            logconsole=True,
+            logfile="test.log",
+            expid= expid,
+            notransitive=False,
+            start_time=None,
+            start_after=None,
+            run_only_members = None,
+            profile=False
+        )
+        with pytest.raises(AutosubmitCritical): # add regex to match error message or code
```
dbeltrankyl, line 501: Maybe this is too general for a unit test; we could have false negatives here if we change the code related to the command, not necessarily to the code you've added here

dbeltrankyl (commented):
Hello @isimo00 

Thanks for the work! 

Some comments left. Also I want to comment somestuff:

> Importantly, this should also check if there are changes in the submodules. I don't remember how strict the check was.

Did you check this?

-------

I think that the test is too general for a unit test, as you're running the autosubmit run here. 

I would enclose the code that you've added into a function and test that function

autosubmit/git/autosubmit_git.py
```diff
@@ -257,3 +259,25 @@ def clone_repository(as_conf, force, hpcarch):
...
+        if expid[0] == 'o':
+            origin = Path(BasicConfig.expid_dir(expid).joinpath("conf/expdef_{}.yml".format(expid)))
+            with open(origin, 'r') as f:
+                yaml = YAML(typ='rt')
+                data = yaml.load(f)
+                project = data["PROJECT"]["PROJECT_TYPE"]
+
+            version_controls = ["git", 
+                                "git",
+                                "svn"]
```
dbeltrankyl, line 274: You can drop svn support ( deprecated)

dbeltrankyl: Also, git is repeated?

test/unit/test_expid.py
```diff
@@ -468,3 +471,37 @@ def test_perform_deletion(create_autosubmit_tmpdir, generate_new_experiment, set
...
+
+
+@pytest.mark.parametrize(
+    "project_type, generate_new_experiment",
+    [
+        ('git', 'operational'),
+        ('git', 'normal'),
+        ('git', 'test'),
+        ('git', 'evaluation'),
+        ('svn', 'operational'),
```
dbeltrankyl, line 483: same, you can drop svn support ( deprecated)

test/unit/test_expid.py
```diff
@@ -468,3 +471,37 @@ def test_perform_deletion(create_autosubmit_tmpdir, generate_new_experiment, set
...
+        ('svn', 'normal'),
+        ('svn', 'test'),
+        ('svn', 'evaluation')
+    ],
+    indirect=["generate_new_experiment"]
+)
+def test_remote_repo_operational(generate_new_experiment, create_autosubmit_tmpdir, project_type, mocker):
```
dbeltrankyl, line 490: Typing, 

Maybe also add a docstring, that explains what this is testing

dbeltrankyl (approved):
Thanks @isimo00 !

Some small comments about adding some comments/typing in the code or removed stuff that won't be used 

Afterwards it can be merged

autosubmit/git/autosubmit_git.py
dbeltrankyl: Ah, okay. Then don't remove it 😅

VindeeR:
@isimo00 I think I ended up reverting your commit (my reverts started right after you commit your change) Sorry for that but can you merge again?

autosubmit/git/autosubmit_git.py
```diff
@@ -257,3 +259,31 @@ def clone_repository(as_conf, force, hpcarch):
...
+    def check_unpushed_changes(expid: str) -> None:
+        """
+        Raises an AutosubmitCritical error if the experiment is operational, the platform is Git, and there are unpushed changes.
+
+        Args: expid (str): The experiment ID.
+
+        Returns: None
+        """
+        if expid[0] == 'o':
+            origin = Path(BasicConfig.expid_dir(expid).joinpath("conf/expdef_{}.yml".format(expid)))
```
kinow, line 273: This file may not exist, as not all AS4 experiments have `expdef_...` files. It would have been better, I think, if we used the parsed configuration, ignoring the file names, to retrieve whether we have a Git project or not. I think ClimateDT experiments may be using that file now, but for users it should be optional -- cc @TracyMcBean

Issue #2295, opened 2025-04-10
set all jobs at some chuncks as WAITING


youcheng-csc:
Hi Autosumbit Colleagues, 

I want to set all jobs (DN OPA APP , LRA, AQAU_ANALYSIS ....) at chunks 25-27 as WAITING (the simulation start from 19880101). I have tried
 ```autosubmit setstatus t05n -ftc "[ 19880101 [ fc0 [25-27] ] ], Any" -t WAITING -s -np```
but it echoes with error,
```[ERROR] Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='WARNING', command='setstatus', expid='t05n', noplot=True, save=True, status_final='WAITING', update_version=False, list=None, filter_chunks=None, filter_status=None, filter_type=None, filter_type_chunk='[ 19880101 [ fc0 [25] ] ], Any', filter_type_chunk_split=None, hide=False, group_by=None, expand=None, expand_status=None, notransitive=False, check_wrapper=False, detail=False)
[ERROR] This is the experiment: <t05n> which had an issue with the command: <setstatus> and it is currently using the Autosubmit Version: 4.1.13.
 [CRITICAL] An Error has occurred while setting some of the workflow jobs, no changes were made [eCode=7040]
```
Is there any wrong with the command? How to correct it?

Thank you in advance!

Cheng You

dbeltrankyl:
Hello @youcheng-csc,

The error message doesn't tell me much, but it is because the log of that part of the code is not working as it should. I've tested it, and it does not recognize the "Any" section there, as the -ftc was meant to be used for specific sections, not all.

I've replicated a case locally to be sure, and I found that you can use this one:

`autosubmit -lc setstatus t05n -fc "[ 19880101 [ fc0 [25-27] ] ]" -t WAITING -s -np`

Let me know if it works for you.

youcheng-csc:
Hi Daniel. Thank you for the reply. I have just tried and It does not work. It echoes folloiwng error,

```
error: argument -lc/--logconsole: invalid choice: 'setstatus' (choose from 'NO_LOG', 'INFO', 'WARNING', 'DEBUG')
usage: autosubmit [-h] [-v] [-lf {NO_LOG,INFO,WARNING,DEBUG}] [-lc {NO_LOG,INFO,WARNING,DEBUG}]
[...]

PR #2297, opened 2025-04-11
Fixes an issue with setstatus error message and Datelist when it is not an string from user config

dbeltrankyl:
fixes https://github.com/BSC-ES/autosubmit/issues/2295.

The error message was hidden due to a re-raise. I've removed the re-raise.

Also, I had the date list set with int instead of ". In other parts of Autosubmit, it doesn't matter, but it seems that for the setstatus it does


VindeeR (approved):
I just tested locally and the message is as expected.

Change looks good to me

dbeltrankyl:
Thanks!

PR #2306, opened 2025-04-14
Add type to run_experiment, respect when users define AUTOSUBMIT_CONFIGURATION

### LINKED ISSUE #2305: autosubmit run doesn't respect AUTOSUBMIT_CONFIGURATION
Found this while working on #2250. For the integration tests, I defined a different `autosubmitrc` file, but the command still looked for the file in my home directory.

kinow:
Closes #2305 

Will check tomorrow if I can write a test for this, although this will be covered by the integration tests in another PR.

I'm just separating changes that are not part of Postgres from the Postgres branches (cc @LuiggiTenorioK who suggested this good idea!).

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -80,7 +80,7 @@ from .history.experiment_history import ExperimentHistory
 import autosubmit.history.utils as HUtils
 import autosubmit.helpers.autosubmit_helper as AutosubmitHelper
 import autosubmit.statistics.utils as StatisticsUtils
-from autosubmit.helpers.utils import check_jobs_file_exists
+from autosubmit.helpers.utils import check_jobs_file_exists, get_rc_path
 
 from contextlib import suppress
 
@@ -2149,7 +2149,7 @@ class Autosubmit:
             p.work_event.set()
 
     @staticmethod
-    def run_experiment(expid, notransitive=False, start_time=None, start_after=None, run_only_members=None, profile=False):
+    def run_experiment(expid, notransitive=False, start_time=None, start_after=None, run_only_members=None, profile=False) -> int:
         """
         Runs and experiment (submitting all the jobs properly and repeating its execution in case of failure).
         :param expid: the experiment id
@@ -2158,7 +2158,7 @@ class Autosubmit:
         :param start_after: the expid after which the experiment should start
         :param run_only_members: the members to run
         :param profile: if True, the function will be profiled
-        :return: None
+        :return: exit status
 
         """
         # Start profiling if the flag has been used
@@ -3406,7 +3406,7 @@ class Autosubmit:
 
     @staticmethod
     def configure(advanced, database_path, database_filename, local_root_path, platforms_conf_path, jobs_conf_path,
-                  smtp_hostname, mail_from, machine, local):
+                  smtp_hostname, mail_from, machine: bool, local: bool):
         """
         Configure several paths for autosubmit: database, local root and others. Can be configured at system,
         user or local levels. Local level configuration precedes user level and user level precedes system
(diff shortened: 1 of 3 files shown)


autosubmit/autosubmit.py
```diff
@@ -3486,61 +3486,54 @@ def configure(advanced, database_path, database_filename, local_root_path, platf
...
-                parser.add_section("historicdb")
-                parser.set("historicdb", "path", str(historicdb_path))
-                parser.add_section("historiclog")
-                parser.set("historiclog", "path", str(historiclog_path))
-                parser.add_section("autosubmitapi")
-                parser.set("autosubmitapi", "url", autosubmitapi_url)
-                # parser.add_section("hosts")
-                # parser.set("hosts", "whitelist", " localhost # Add your machine names")
-                parser.write(config_file)
-                config_file.close()
```
kinow, line 3531: This file was closed only in the success path. Harmless to leave it open as the exceptions should crash the interpreter and close everything; but with NFS, GPFS, it doesn't hurt to make sure the file will be really closed. Thus, this change that's moving all the lines to the right...

autosubmit/helpers/utils.py
```diff
@@ -220,3 +221,31 @@ def strtobool(val):
...
+        return Path(os.environ['AUTOSUBMIT_CONFIGURATION'])
+
+    if machine:
+        rc_path = '/etc'
+    elif local:
+        rc_path = '.'
+    else:
+        rc_path = Path.home()
+
+    return Path(rc_path) / '.autosubmitrc'
```
kinow, line 251: Moved the logic from `autosubmit/autosubmit.py` to this smaller function to make it easier to test & reuse (e.g. in tests/fixtures).

kinow:
The lines not covered were not covered before either, and I'm afraid covering those if/elses is a bit hard for this PR. I can revert it if it's better, but I think it's worth reviewing it first, to see if it looks OK to be merged as it prevents file handles from leaking (even if that's quite impossible from happening :shrug: ).

LuiggiTenorioK (approved):
Seems pretty good! :100: 

It still feels weird that we have 2 boolean variables (`machine`, `local`) instead of one, but it is what it is :man_shrugging:

kinow:
> Seems pretty good! 💯
> 
> It still feels weird that we have 2 boolean variables (`machine`, `local`) instead of one, but it is what it is 🤷‍♂️

I thought exactly the same. For one split second I considered adding an `Enum`… but then I imagined how much of the code I'd have to update, and the possible issues, so I let that idea go :laughing: 

Thanks @LuiggiTenorioK !

autosubmit/helpers/utils.py
```diff
@@ -220,3 +221,31 @@ def strtobool(val):
...
         raise ValueError("invalid truth value %r" % (val,))
+
+
+def get_rc_path(machine: bool, local: bool) -> Path:
+    """Get the ``.autosubmit.rc`` path.
+
+    If the environment variable ``AUTOSUBMIT_CONFIGURATION`` is specified in the
+    system, this function will return a ``Path`` pointing to that value.
+
+    If ``machine`` is ``True``, it will use the file from ``/etc/.autosubmitrc``
```
dbeltrankyl, line 232: In /etc, the name is /etc/autosubmit

autosubmit/helpers/utils.py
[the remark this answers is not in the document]
kinow: Let me review the old code. I had it with etc/autosubmitrc, but the test kept creating /etc/.autosubmitrc.

I can still use /etc/autosubmitrc with the AUTOSUBMIT_CONFIGURATION, and I thought that's what we had in destine.

Thanks Dani!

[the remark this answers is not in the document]
kinow: Looking at the code in 4.1.13,

https://github.com/BSC-ES/autosubmit/blob/f5844f3eba3e49f445c6b7f7ea095d09ad463cc1/autosubmit/autosubmit.py#L3484-L3492

@dbeltrankyl I wrote a test for `/etc/autosubmitrc`, and it was failing. Then I realized it was using `/etc/.autosubmitrc`, and then I adjusted the code & test & comments to `/etc/.autosubmit`, which I thought was strange when writing it…

I translated the old code, and I think the code in 4.1.13 was using `/etc/.autosubmitrc` too. Let me look at older code.

PR #2250, opened 2025-03-26
Fixing Postgres tests

kinow:
Will try to continue tomorrow and finish before the end of the day on Friday.

- [x] Fix integration test run command test (failing intermittently locally)
- [x] Remove some mock from `autosubmit_config`, test with a real experiment
- [x] Move database tests to `database` package (in tests)
- [x] Move tests that use Postgres & SQlite to the integration folder
- [x] Remove GH Actions Postgres, start with testcontainers context manager, so we can run it automatically in our laptops (like the SSH integration test)

PR #2307, opened 2025-04-14
Separate integration and unit tests (for Postgres)

kinow:
Another PR like #2306 , related to #2250 . These changes were needed in the postgres branch as we are running DB integration tests there.

- [x] Separated unit and integration tests
- [x] Used pathlib
- [x] Use existing fixtures whenever possible (removing FakeBasicConfig and other local functions)

The changes in this pull request must be only in the tests, no code changed.

The coverage should be kept about the same or increased :+1:


kinow:
I will continue tomorrow working on the tests that failed here.

kinow:
After rebasing it today, I continued fixing the integration & unit tests after moving the tests as in the postgres branch. @LuiggiTenorioK it was a really good idea to move those changes out of the pg branch as I noticed a few things that were still missing (harder to see that when you have too many changes).

There are still more tests to fix, but once it gets merged the postgres branches - both - should get a lot simpler.

.github/workflows/ci.yaml
```diff
@@ -112,15 +112,6 @@ jobs:
...
 
-      # Run regression tests
-      - name: Regression tests
-        run: |
-          pytest \
-            --cov=autosubmit --cov-config=.coveragerc \
-            --cov-report=xml:test/coverage.xml --cov-append \
-            test/regression \
-            -m ''
-
```
kinow, line 123: This was run in integration and in regression build jobs. I've removed from integration, and now it only runs in the regression job.

test/integration/test_profiler.py
```diff
@@ -17,121 +17,47 @@
...
-    # This is set in _init_log which is not called
-    as_misc = Path(f"{run_tmpdir.strpath}/t000/conf/as_misc.yml")
-    with as_misc.open('w', encoding="utf-8") as f:
-        f.write("""
-            AS_MISC: True
-            ASMISC:
-                COMMAND: run
-            AS_COMMAND: run
-        """)
-    return log_dir
```
kinow, line 98: This code was removed as now `autosubmit_exp` creates a valid experiment, and runs `autosubmit create`, leaving it ready to be executed. The MISC file is missing, but there's an example in integration that calls the function that creates it (there's a TODO there to improve this).

kinow:
Coverage reported here:

image

Coverage reported in Codecov for master

image

So coverage-wise it should be OK. I didn't disable/delete any tests. Things were just moved.

kinow:
Maybe @dbeltrankyl or @LuiggiTenorioK would like to review it? Dani as he worked on the regression tests recently, or Luiggi as this change is part of the work for Postgres/EDITO.

LuiggiTenorioK (approved):
Seems pretty fine! :100: I liked that the fixtures are quite useful for many purposes, and prevent repeating ourselves a lot. Thanks @kinow!

kinow:
Thanks @LuiggiTenorioK !

PR #2310, opened 2025-04-15
first commit documenting the wrappers

VindeeR:
I'm planning on making 2 commits for this issue, one which includes the documentation of the wrapper, and one with the fix itself this way we can have the documentation in case something else pops up around the wrappers

### DIFF
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -23,7 +23,6 @@ Main module for Autosubmit. Only contains an interface class to all functionalit
 
 from collections import OrderedDict
 from pathlib import Path
-from xmlrpc.client import DateTime
 
 from autosubmit.job import job_utils
 import copy
@@ -3101,7 +3100,7 @@ class WrapperJob(Job):
         Parse a date from int to datetime.
         :param timestamp: time to be converted
         :type timestamp: int
-        :return: return time converted from int to datetime
+        :return: return time converted
         :rtype: datetime
         """
         value = datetime.datetime.fromtimestamp(timestamp)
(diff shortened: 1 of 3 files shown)


autosubmit/job/job.py
```diff
@@ -2980,7 +3035,12 @@ def update_failed_jobs(self, check_ready_jobs=False):
                 if job in real_running:
                     self.inner_jobs_running.append(job)
 
-    def cancel_failed_wrapper_job(self):
+    def cancel_failed_wrapper_job(self) -> None:
+        """
+        When a wrapper is cancelled or run into some problem all its jobs are cancelled,
+        if there are jobs on the list that are not Running, and is not Completed, or Failed set it as WAITING,
+        if not on these status and it is a vertical wrapper it will set the fail_count to the number of retrials.
```
dbeltrankyl, line 3042: It sets ( or should) the fail_count of the jobs to 0 only in case that the wrapper is FAILED

dbeltrankyl (approved):
Only one small comment, thanks!

autosubmit/job/job.py
VindeeR: > the fail_count of the jobs to 0 only

>            else:
>               if job.wrapper_type == "vertical":  # job is being retrieved internally by the wrapper
>                   job.fail_count = job.retrials

In the code it sets to job.retrials that I believe it is the amount of times a job had retried to be executed

VindeeR: > It sets ( or should) the fail_count of the jobs to 0 only in case that the wrapper is FAILED ( and it is a vertical wrapper)

for jobs that are Completed, or Failed it doesn't change the value int retrials, should we create an issue so that can be tackled?

PR #2319, opened 2025-04-22
Remove wheel from our build-system requires in pyproject.toml

### LINKED ISSUE #2318: Prevent issues with setuptools & wheels
I sent this link to myself over the long weekend to read it here at work: https://old.reddit.com/r/Python/comments/1jwbymm/psa_you_should_remove_wheel_from_your/

TL;DR, newer versions of `setuptools` include `wheel` code. But if you have a version of `wheel` installed, then it will use that version, which can cause unexpected build errors if the `setuptools` and `wheel` versions are not compatible.

kinow:
Closes #2318 

See also: https://old.reddit.com/r/Python/comments/1jwbymm/psa_you_should_remove_wheel_from_your/ & https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#declaring-the-build-backend

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools >= 77.0.3"]
 build-backend = "setuptools.build_meta"
 
 [project]


dbeltrankyl (approved):
Thanks!

kinow:
Rebasing, looks like `pip install` is failing but I think it's due to networking issues? Will kick the build again once it fails.

Issue #2324, opened 2025-04-23
Revert a bad change derived from maestro branch


dbeltrankyl:
Hello @isimo00 , @VindeeR 

@VindeeR  found an issue there:

https://github.com/BSC-ES/autosubmit/blame/32f127da5f15fc4ba08500a66b61b93a654e6039/autosubmit/platforms/slurmplatform.py#L552

This change was incorrect, as it now always calls the X11 code. 

(Although, I'm not sure why the tsuite is not complaining about this. Since it is also present 4.1.14)

I'm creating a PR

dbeltrankyl:
I'll cherry-pick the change once it is approved

kinow:
Feel free to merge that PR and leave the test for later, @dbeltrankyl

PR #2325, opened 2025-04-23
Revert a wrong change

dbeltrankyl:
Basically, the X11 method is always called when it shouldn't.

closes https://github.com/BSC-ES/autosubmit/issues/2324


VindeeR (approved):
Thanks for the quick action, and for the help with the issue

kinow (commented):
Looks good to me! I think this isn't too hard to cover with a test (just the `get_submit_cmd` function)?

dbeltrankyl:
> Looks good to me! I think this isn't too hard to cover with a test (just the get_submit_cmd function)?

I'll do it in another PR because this is necessary for other PR'S

Thanks to both!

kinow:
Thank **you** Dani!!

PR #2328, opened 2025-04-23
Add lsof +D check

### LINKED ISSUE #992: Autosubmit fails to delete an experiment [ Add lsof check ]
In GitLab by @ gbonet on Mar 23, 2023, 17:01

Hello @dbeltrankyl and @kinow,

### Autosubmit Version

v3.14.0

### Expid affected(If applicable)

**t0cg**

### Which task has issues? Where is the log(If applicable)

autosubmit delete. Log located on `/home/Earth/gbonet/logs/deletion.log`

### Summary

I am having problems deleting an experiment using autosubmit. I get an error message stating that some files weren not correctly deleted.

### Steps to reproduce

Trying to delete the experiment t0cg.

Whenever the delete command is executed on **t0cg** it fails and, since I am allowed to execute the command again, it seems that the experiment remains in the databse. The experiment folder is also still present in `/esarchive/autosubmit`

You would expect the delete command to fully delete the experiment from the filesystem and remove it from the database.

```
[0m[39m
[38;5;214m[ERROR] Trace: Some experiment files weren't correctly deleted
Please if the trace shows DATABASE IS LOCKED, report it to git
If there are I/O issues, wait until they're solved and then use this command again.
[0m[39m
[1m [31m[CRITICAL] Seems that something went wrong, please check the trace [eCode=7012][0m[39m
More info at https://autosubmit.readthedocs.io/en/v3.14.0/troubleshooting/error-codes.html

### PR DESCRIPTION
**Pull request by isimo00**
Adds a check to ensure no processes are running in an experiment directory before trying to delete experiment files.

Closes #992

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -81,6 +81,7 @@ import autosubmit.history.utils as HUtils
 import autosubmit.helpers.autosubmit_helper as AutosubmitHelper
 import autosubmit.statistics.utils as StatisticsUtils
 from autosubmit.helpers.utils import check_jobs_file_exists, get_rc_path
+from autosubmit.helpers.processes import process_id
 
 from contextlib import suppress
 
@@ -1469,7 +1470,8 @@ class Autosubmit:
 
         :raises AutosubmitCritical: If the experiment does not exist or if there are insufficient permissions.
         """
-        if not AutosubmitGit.check_directory_in_use(expid):
+        # if not AutosubmitGit.check_directory_in_use(expid):
+        if process_id(expid) is None:
             raise AutosubmitCritical("Ensure no processes are running in the experiment directory", 7076)
 
         experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1469,11 +1469,11 @@ class Autosubmit:
 
         :raises AutosubmitCritical: If the experiment does not exist or if there are insufficient permissions.
         """
-        experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
-        
         if not AutosubmitGit.check_directory_in_use(expid):
             raise AutosubmitCritical("Ensure no processes are running in the experiment directory", 7076)
 
+        experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
+        
         if experiment_path.exists():
             if force or Autosubmit._user_yes_no_query(f"Do you want to delete {expid} ?"):
                 Log.debug('Enter Autosubmit._delete_expid {0}', expid)
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1471,7 +1471,7 @@ class Autosubmit:
         """
         experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
         
-        if not AutosubmitGit.check_directory_in_use(experiment_path):
+        if not AutosubmitGit.check_directory_in_use(expid):
             raise AutosubmitCritical("Ensure no processes are running in the experiment directory", 7076)
 
         if experiment_path.exists():
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1470,6 +1470,9 @@ class Autosubmit:
         :raises AutosubmitCritical: If the experiment does not exist or if there are insufficient permissions.
         """
         experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
+        
+        if not AutosubmitGit.check_directory_in_use(experiment_path):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: autosubmit/git/autosubmit_git.py [hunk 1]
```diff
@@ -301,3 +301,16 @@ def check_unpushed_changes(expid: str) -> None:
                         # M: Modified, A: Added, D: Deleted, ?: Untracked
                         raise AutosubmitCritical("Push local changes to remote repository before running", 7075)
 
+    @staticmethod
+    def check_directory_in_use(dir_path: str) -> bool:
+        """
+        Lists all Open Files in a directory
+        """
+        result = subprocess.run(
+            ['lsof', '+D', dir_path],
```
kinow, line 310: Lsof may not be available (e.g. docker containers).

We might have to either add it to the container recipe and installation instructions, or maybe use a python lib (not sure if psutil has something like lsof).

kinow: `autosubmit stop` has code that checks if the experiment is running. Maybe that same function can be used to check if the expid is found in a process with `autosubmit run`, and if so warn the user.

kinow: https://github.com/BSC-ES/autosubmit/blob/438743344dac7c910a3ad89758bab27abcf33829/autosubmit/helpers/processes.py#L162

autosubmit/autosubmit.py
```diff
@@ -1470,6 +1470,9 @@ def delete(expid: str, force: bool) -> bool:
         :raises AutosubmitCritical: If the experiment does not exist or if there are insufficient permissions.
         """
         experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
```
kinow, line 1472: This can be moved after your check, to avoid declaring it altogether in case the `check_directory_in_use` fails. :grimacing:

autosubmit/git/autosubmit_git.py
```diff
@@ -301,3 +303,19 @@ def check_unpushed_changes(expid: str) -> None:
                         # M: Modified, A: Added, D: Deleted, ?: Untracked
                         raise AutosubmitCritical("Push local changes to remote repository before running", 7075)
 
+    @staticmethod
+    def check_directory_in_use(expid: str) -> bool:
+        """
+        Checks for open files and unfnished git-credential requests in a directory
```
kinow, line 309: s/unfnished/unfinished

autosubmit/git/autosubmit_git.py
```diff
@@ -301,3 +303,19 @@ def check_unpushed_changes(expid: str) -> None:
...
+        """
+        if process_id(expid) is not None:
+            return True
+        for proc in psutil.process_iter(['pid', 'name', 'cmdline']):
+            name = proc.info.get('name', '')
+            cmdline = proc.info.get('cmdline') or []
+            if 'git-credential' in name or any('git-credential' in arg for arg in cmdline):
+                return True
+        if bool(psutil.Process().open_files()):
+            return True
```
kinow, line 319: I think this `if bool()` means: If the current process has any open files, then abort.

But the current process is the `autosubmit delete` command, and we shouldn't abort if it has any files (e.g. libs?) open, I think? Only if something like `autosubmit run` is running?

autosubmit/git/autosubmit_git.py
```diff
@@ -301,3 +303,19 @@ def check_unpushed_changes(expid: str) -> None:
...
+        """
+        Checks for open files and unfnished git-credential requests in a directory
+        """
+        if process_id(expid) is not None:
+            return True
+        for proc in psutil.process_iter(['pid', 'name', 'cmdline']):
+            name = proc.info.get('name', '')
+            cmdline = proc.info.get('cmdline') or []
+            if 'git-credential' in name or any('git-credential' in arg for arg in cmdline):
+                return True
```
kinow, line 317: Do we need to check if `git-credential` is running? If we have 30 experiments with a lot of users connecting, like in the ClimateDT VM, couldn't we have multiple `git-credential` running, from different folders and for different users?

I thought maybe checking just if the `autosubmit run` was active would be enough?

kinow (commented):
Haven't looked at the test, just the code. Left a few comments, but I think it's going in the right direction! :raised_hands: Thanks

dbeltrankyl:
> Hi @dbeltrankyl do you know what should be modified from either check_directory_in_use in autosubmit/git/autosubmit_git.py or test_delete_experiment_not_owner in test/unit/test_expid.py so that the test doesn't fail? It looks like the test tries to delete an experiment without owner but the function I'm introducing raise an AutosubmitCritical because it detects an active process in the directory

The psutil is forced to raise a TypeError in that test.. I've tested this patch, which is just for mocking your new code addition for that test
 
[0001-patch-delete.patch](https://github.com/user-attachments/files/19960276/0001-patch-delete.patch)

dbeltrankyl (approved):
It looks good  to me but wait for @kinow to see your last comments about the section he reviewed

Thanks!

kinow (commented):
Some comments about finding the process and imports. Otherwise, looks neat!

Issue #1234, opened 2024-02-22
Error when running 'autosubmit create'


kinow:
In GitLab by @hrosegalb on Feb 22, 2024, 11:07

Hello @dbeltrankyl and @kinow,

### Autosubmit Version
4.0.101

### Summary
I am testing autosubmit locally with a `hello_world` script. When I tried to run `autosubmit create <expid>` to create the experiment, I get the following error:

```
 [CRITICAL] Traceback (most recent call last):
  File "/home/hgalbrai/.pyenv/versions/autosubmit_env/bin/autosubmit", line 37, in main
    Autosubmit.parse_args()
  File "/home/hgalbrai/.pyenv/versions/3.10.12/envs/autosubmit_env/lib/python3.10/site-packages/autosubmit/autosubmit.py", line 659, in parse_args
    Autosubmit._init_logs(args, args.logconsole, args.logfile, expid)
  File "/home/hgalbrai/.pyenv/versions/3.10.12/envs/autosubmit_env/lib/python3.10/site-packages/autosubmit/autosubmit.py", line 801, in _init_logs
    as_conf.reload(force_load=True)
  File "/home/hgalbrai/.pyenv/versions/3.10.12/envs/autosubmit_env/lib/python3.10/site-packages/autosubmitconfigparser/config/configcommon.py", line 1394, in reload
    self.experiment_data = self.normalize_variables(self.experiment_data)
  File "/home/hgalbrai/.pyenv/versions/3.10.12/envs/autosubmit_env/lib/python3.10/site-packages/autosubmitconfigparser/config/configcommon.py", line 508, in normalize_variables
    dependencies = job_data.get("DEPENDENCIES",7000)
AttributeError: 'NoneType' object has no attribute 'get'
'NoneType' object has no attribute 'get'
Unhandled error: If you see this message, please report it in Autosubmit's GitLab project
```

### Steps to reproduce

1. Create a new `expid` using the `Autosubmit-Template` repo. 

```
$ autosubmit expid -H "local" -d "Test in local"
```

2. Update the `expdef_a012.yml` file:

```
DEFAULT:
  # Job experiment ID.
  EXPID: "a012"
  # Default HPC platform name.
  HPCARCH: "local"
EXPERIMENT:
  # List of start dates
  DATELIST: '20000101'
  # List of members.
  MEMBERS: fc0
  # Unit of the chunk size. Can be hour, day, month, or year.
  CHUNKSIZEUNIT: month
[...]

PR #2299, opened 2025-04-11
4.1.14 release branch

dbeltrankyl:
This branch is to write the changelog and do the release after the t-suite rans sucesfully with this version

PR #2327, opened 2025-04-23
Wrapper status is not updated correctly

VindeeR:
Closes #2274.

File generation that AS will look for to try to update the status is created with a different name than what it searches for, this was creating a conflict on the validation and not updating it properly.

The file name is being reverted to how it used to be, and if needed the change it can be done by changing all the locations that it will impact

PR #2345, opened 2025-05-06
Fix database issues

dbeltrankyl:
**Check List**

I'll do it in another PR just to agilize this release. Manually tested

Already added from other fixes tries 

- [x] Documentation updated.
#2299

### DIFF
--- a/autosubmit/history/database_managers/experiment_history_db_manager.py
+++ b/autosubmit/history/database_managers/experiment_history_db_manager.py
@@ -403,10 +403,24 @@ class ExperimentHistoryDbManager(DatabaseManager):
         models = [Models.JobDataRow(*row) for row in job_data_rows][-1]
         return JobData.from_model(models)
 
-    def get_job_data_max_counter(self):
-        """ The max counter is the maximum count value for the count column in job_data. """
-        statement = "SELECT MAX(counter) as maxcounter FROM job_data"
-        counter_result = self.get_from_statement(self.historicaldb_file_path, statement)
+    def get_job_data_max_counter(self, job_name: str = None) -> int:
+        """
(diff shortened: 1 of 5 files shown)


dbeltrankyl:
Fixed: 

* counter now counts by job_name as it should
* Connect calls in case there are logs left to be recovered
* The #2327 introduced a bug in the stat file, as it was being searched in the wrong path ( tmp/LOG_$EXPID) instead of (tmp)
* In case that the stat_file was not found, for wrapper jobs it could be that the finish_time was None

I've manually tested it because we need it as soon as possible. I'll do the automatic tests in another PR

image

dbeltrankyl:
Can you have a look @kinow 🙏

kinow (commented):
@dbeltrankyl code-wise I can't find anything that looks odd. It's useful for me to review this too as I will have to port this to #2187 (so thanks for tagging me here!). The only thing to comment is the lack of tests. Is it urgent, or can we try to write some tests for this change?

kinow:
>I've manually tested it because we need it as soon as possible. I'll do the automatic tests in another PR

:laughing: you had already answered it! Hitting the approval button!

kinow (approved):

dbeltrankyl:
Thanks! I hope this is the last update for 4.1.14.. 🤞

kinow:
Fingers crossed! Good luck with the fix, @dbeltrankyl ! Let me know if you need any help (I can always use a break from postgres)

kinow:
BTW, great job on the docs and types!!! Even if we don't have time for the tests now, that's really useful & helpful (like when I need to port things to postgres and write tests :bow: )

Issue #2349, opened 2025-05-08
Delete the SGE platform


kinow:
From https://github.com/BSC-ES/autosubmit/pull/2337#discussion_r2079327625

The code is broken (functions that do not exist anymore), and hasn't been tested in a long time.

kinow:
Setting for next milestone as if we all agree, it should take just <1 minute to implement it 😬

kinow:
Already in the changelog for 4.1.15 👍 I only forgot to link this issue, I am sorry. Closing now.

PR #2362, opened 2025-05-19
Test Slurm platform on GitHub using a Slurm docker

VindeeR:
Partially addresses #977

This PR is aimed towards implementing the docker generation build to be used in the GH actions with that new complex and new tests can be executed and increase the coverage of the platforms and autosubmit

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,6 +7,8 @@ on:
     branches:
       - main
       - master
+permissions:
+  contents: read
 
 jobs:
   # lint: # Turn on when linting issues are resolved
@@ -118,7 +120,7 @@ jobs:
             --cov=autosubmit --cov-config=.coveragerc \
             --cov-report=xml:test/coverage.xml --cov-append \
             test/integration \
-            -m ''
+            -m 'not slurm'
 
       - name: Coverage report
         run: |
@@ -178,6 +180,81 @@ jobs:
           path: coverage.xml
           retention-days: 7
 
+  test-slurm:
+    runs-on: ubuntu-latest
+    services:
+      slurm:
+        image: autosubmit/slurm-openssh-container:25-05-0-1
+        ports:
+          - 2222:2222
+        options: --rm -it --cgroupns=host --privileged --volume /sys/fs/cgroup:/sys/fs/cgroup:rw --hostname slurmctld --name slurm-container
+
+    strategy:
+      matrix:
+        python-version: ["3.9"]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Copy SSH access key
+        run: |
+          docker cp slurm-container:/root/.ssh/container_root_pubkey /tmp/container_root_pubkey
+          chmod 600 /tmp/container_root_pubkey
+
+      - name: Configure SSH host
+        run: |
+          sudo bash -c 'mkdir /home/runner/.ssh/'
+          sudo bash -c 'echo -e $"Host localDocker\n  HostName localhost\n  User root\nStrictHostKeyChecking no\n UserKnownHostsFile /dev/null\n  IdentityFile /tmp/container_root_pubkey\n  Port 2222\n   ForwardX11 yes" >> /home/runner/.ssh/config'
+
+      - name: Run sinfo over SSH on SLURM container
+        run: |
+          ssh localDocker -p 2222 -i /tmp/container_root_pubkey sinfo
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install system dependencies
+        run: sudo apt-get install -y curl git graphviz rsync
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip packaging
+          pip install --upgrade -e .[all]
+
+      - name: Set up Git
+        run: |
+          # From: https://stackoverflow.com/questions/62960533/how-to-use-git-commands-during-a-github-action
+          # Set up dummy configuration for integration tests.
+          git --version
+          git config --global user.email "test@bsc.es"
+          git config --global user.name "GitHub Actions"
+
+      - name: Integration tests
+        run: |
+          pytest \
+            --cov=autosubmit --cov-config=.coveragerc \
+            --cov-report=xml:test/coverage.xml --cov-append \
+            test/integration \
+            -m 'slurm'
+
+      - name: Coverage report
+        run: |
+          coverage xml
+          coverage report
+
+      - name: Upload coverage artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage_slurm_py-${{ matrix.python-version }}
+          path: coverage.xml
+          retention-days: 7
+
   test-misc:
     runs-on: ubuntu-latest
     timeout-minutes: 5
@@ -206,7 +283,7 @@ jobs:
           cffconvert --validate -i CITATION.cff
 
   coverage:
-    needs: [test, test-integration, test-regression]
+    needs: [test, test-integration, test-regression, test-slurm]
     runs-on: ubuntu-latest
     timeout-minutes: 2
     steps:
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,46 @@
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+permissions:
+  contents: read
+
+jobs:
+  build-docker-images:
+    name: Build Docker images
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ vars.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: docker
+          # TODO: for a separate issue, try pushing it
+          push: false
+          load: true
+          tags: autosubmit/autosubmit:latest
+      
+      - name: Test the image
+        run: |
+          docker run --rm autosubmit/autosubmit:latest autosubmit --version
+          # https://github.com/BSC-ES/autosubmit/issues/2359
+          docker run --rm autosubmit/autosubmit:latest /bin/bash -c '[ -n $USER ] || exit 1'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+### 4.1.16: Unreleased
+
+- autosubmit/autosubmit container now includes the `$USER` environment variable
+  via its entrypoint #2359
+- Adding a Slurm Container to the CI/CD and creating tests to increase the
+  coverage of the Platforms #977
+
 ### 4.1.15: Unreleased
 
 The filter `-fp` of the command `autosubmit stats` changed in this release.
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1002,7 +1002,7 @@ class Autosubmit:
         eadmin = False
         owner = False
         current_user_id = os.getuid()
-        admin_user = "eadmin" # to be improved in #944
+        admin_user = "eadmin"# to be improved in #944
         try:
             eadmin = current_user_id == pwd.getpwnam(admin_user).pw_uid
         except Exception:
@@ -1011,7 +1011,8 @@ class Autosubmit:
         try:
             current_owner = pwd.getpwuid(current_owner_id).pw_name
         except (TypeError,KeyError):
-            Log.warning(f"Current owner of experiment {expid} could not be retrieved. The owner is no longer in the system database.")
+            Log.warning(f"Current owner of experiment {expid} could not be retrieved. The owner is no longer in the "
+                        f"system database.")
         if current_owner_id == current_user_id:
             owner = True
         elif raise_error:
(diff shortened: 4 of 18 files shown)


VindeeR:
@dbeltrankyl @kinow I believe the container are working now, at least in my local repo they are, but the access for the dockerhub needs to be created, and I was wondering which user/access I should use too

docker/README.md
```diff
@@ -117,9 +115,9 @@ $ docker run --rm \
   -ti \
   -p 2222:22 \
   -e DISPLAY=$DISPLAY \
-  -v $(pwd -P)/id_rsa:/home/autosubmit/.ssh/id_rsa \
-  -v $(pwd -P)/id_rsa.pub:/home/autosubmit/.ssh/id_rsa.pub \
-  -v $(pwd -P)/authorized_keys:/home/autosubmit/.ssh/authorized_keys \
+  -v ~/.ssh/id_rsa:/home/autosubmit/.ssh/id_rsa \
+  -v ~/.ssh/id_rsa.pub:/home/autosubmit/.ssh/id_rsa.pub \
+  -v ~/.ssh/authorized_keys:/home/autosubmit/.ssh/authorized_keys \
```
kinow, line 120: I think we shouldn't use our own keys.

Above, in this same docs, there's a line like this `docker run --rm -it --entrypoint /keygen.sh linuxserver/openssh-server`.

That line calls the linuxserver container utility tool to produce SSH keys. They are then stored locally (if I recall correctly).

These docs were showing the user how to produce random keys and use from the local directory (thus `$(pwd -P)`) for testing AS.

Not every user will want to map their keys inside container (I am in this group, for instance :grimacing:), so we either keep the previous version, or we change the docs where we show users how to create keys, and alert them about risks of binding their keys inside the container -- I find it easier to avoid a security risk like this whenever we can...

.github/workflows/ci.yaml
```diff
@@ -199,3 +199,63 @@ jobs:
...
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          registry: docker.io
+
+      - name: Build and push the first Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: ./docker/Dockerfile
+          push: false
```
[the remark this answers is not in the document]
kinow: And if we are not pushing, do we still need to login, and set up secrets/credentials for Docker here?

kinow (changes requested):

.github/workflows/ci.yaml
```diff
@@ -199,3 +199,63 @@ jobs:
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          push: true
+          tags: user/app:latest
```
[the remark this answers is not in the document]
VindeeR: No use for that section

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,38 @@
+name: DOCKER
```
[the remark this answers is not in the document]
VindeeR: never Nit-picking always looking out for quality :wink:

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,75 @@
...
+          git config --global user.name "GitHub Actions"
+
+      - name: Integration tests
+        run: |
+          pytest \
+            --cov=autosubmit --cov-config=.coveragerc \
+            --cov-report=xml:test/coverage.xml --cov-append \
+            test/unit/platforms/test_locplatform.py \
+            test/integration/platforms/test_slurmplatform.py \
+            -m ''
```
kinow, line 75: We may need a marker for Slum here.

autosubmit/autosubmit.py
```diff
@@ -3766,7 +3766,7 @@ def configure_dialog():
                                      height=20,
                                      width=80,
                                      form_height=10,
-                                     title='\Zb\Z1Mail notifications configuration:\Zn', colors='enable')
+                                     title='\\Zb\\Z1Mail notifications configuration:\\Zn', colors='enable')
```
kinow, line 3769: These were annoying me too. But I avoided changing them as I would have to confirm that it worked correctly in the terminal (visually inspect the output). Can you do that, for any of these lines modified? Just one, please :grimacing:

autosubmit/platforms/sgeplatform.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, expid, name, config):
         self.job_status['QUEUING'] = ['qw', 'hqw', 'hRwq', 'Rs', 'Rts', 'RS', 'RtS', 'RT', 'RtT']
         self.job_status['FAILED'] = ['Eqw', 'Ehqw', 'EhRqw', 's', 'ts', 'S', 'tS', 'T', 'tT', 'dr', 'dt', 'dRr', 'dRt',
                                      'ds', 'dS', 'dT', 'dRs', 'dRS', 'dRT']
-        self._pathdir = "\$HOME/LOG_" + self.expid
+        self._pathdir = "\\$HOME/LOG_" + self.expid
```
kinow, line 63: I think this platform file is being deleted in #2349 . If that happens before this one is merged, we need to pay attention to remove this file when fixing conflicts :+1:

autosubmit/platforms/slurmplatform.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, expid: str, name: str, config: dict, auth_password: str=None)
         self.job_status['QUEUING'] = ['PENDING', 'CONFIGURING', 'RESIZING']
         self.job_status['FAILED'] = ['FAILED', 'CANCELLED', 'CANCELLED+', 'NODE_FAIL',
                                      'PREEMPTED', 'SUSPENDED', 'TIMEOUT', 'OUT_OF_MEMORY', 'OUT_OF_ME+', 'OUT_OF_ME']
-        self._pathdir = "\$HOME/LOG_" + self.expid
+        self._pathdir = "\\$HOME/LOG_" + self.expid
```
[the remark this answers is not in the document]
VindeeR: it is since it gets an error otherwise, about not skipping properly, I'll add the comment :wink:

autosubmit/autosubmit.py
VindeeR: :saluting_face: yes, sir

autosubmit/platforms/slurmplatform.py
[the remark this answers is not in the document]
kinow: Is it an error or warning? Both of these worked for me,

```bash
$ python3 -c 'print("Hello \$HOME")'
Hello \$HOME
$ python3 -c 'print("Hello \\$HOME")'
Hello \$HOME
```

Or do we get an error elsewhere afterwards? In the IDE I thought what we had were warnings for this, I think.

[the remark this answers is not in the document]
VindeeR: I was getting this error in the GHActions on this [build](https://github.com/VindeeR/autosubmit/pull/3/commits/83fe01fe0001a0dd0501bdebcffbd1eefa1feba1)

image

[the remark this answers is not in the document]
kinow: I think the `SyntaxWarning` doesn't break the build, and in that case it was something else. But if you test that it doesn't break anything, the let's fix that warning too :+1: 
Thanks Erick!

kinow:
@VindeeR here's a way to replicate the test without waiting for slurm. We need to make this program to work in the container (i.e. we don't need to re-build all the time, maybe we can use `apt install` and change some `/etc` configs, etc., with things we find in slurm docs/blogs/etc.).

```c
#include <stdio.h>
#include <dbus/dbus.h>

int main(int argc, char *args) {
        DBusConnection *conn = NULL;
        DBusError err;

        dbus_error_init(&err);
        conn = dbus_bus_get(DBUS_BUS_SYSTEM, &err);

        if (dbus_error_is_set(&err)) {
                printf("%s: cannot connect to dbus system daemon: %s\n", __func__, err.message);
                dbus_error_free(&err);
        }

        if (!conn) {
                printf("Failed to create connection to dbus\n");
                return 1;
        }

        printf("OK!\n");
        return 0;
}
```

Inside the container, we can compile it with:

```bash
$ gcc test.c -o test `pkg-config --libs -cflags dbus-1`
```

Then run `./test`. You should get the error:

```bash
root@slurmctld:/# ./test 
main: cannot connect to dbus system daemon: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
Failed to create connection to dbus
```

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,78 @@
...
+          git config --global user.email "test@bsc.es"
+          git config --global user.name "GitHub Actions"
+
+      - name: Integration tests
+        run: |
+          pytest \
+            --cov=autosubmit --cov-config=.coveragerc \
+            --cov-report=xml:test/coverage.xml --cov-append \
+            test/integration/platforms/test_slurmplatform.py \
+            -m ''
```
kinow, line 78: We need to create a marker for Slurm. Then the integration tests should pass without any modification, and here you just set something like `-m 'slurm'`.

autosubmit/platforms/sgeplatform.py
kinow, line 1: Hmmm, we removed this one. I think you can rebase & delete this.

test/unit/platforms/test_locplatform.py
kinow, line 1: I had a look at the `master` branch, and I think this one exists there? Did you have to remove this in order to get Slurm to work?

kinow:
Had a quick look, the Docker/GH Actions part looks in order (will have another look with care/calm tomorrow). But left some comments about ptyest markers & coverage. The report on Codecov is showing things like `submit_Script` not covered, but I thought the pytest test we wrote was sending the job to the remote platform, and that that part of the code would now be covered. So that might still have some work pending, but otherwise it's nearly ready to be reviewed/merged. I think this will be included in 4.1.16 :clap: Excellent work, @VindeeR !

VindeeR:
> @VindeeR can I review the code now, or are you planning to submit anything new?
> 
> https://app.codecov.io/gh/BSC-ES/autosubmit/pull/2362/blob/autosubmit/platforms/slurmplatform.py
> 
> This is the main file that I'm more interested about for coverage & testing in this branch. And the one I will look more thoroughly while reviewing.

I was trying to introduce the mark you said `-m 'slurm'` but every time I add it, the build fail, but I don't plan on adding anything new, so you can go ahead with the review

autosubmit/platforms/sgeplatform.py
VindeeR: For some reason git wasn't properly tracking this file, which let to even when doing a rebase  the file wouldn't be deleted, and I didn't notice before thanks for checking it

kinow (commented):
@VindeeR can you add this to the `CONTRIBUTING.md` too, please?

````diff
(autosubmit4) kinow@ranma:~/Development/python/workspace/autosubmit$ git diff CONTRIBUTING.md
diff --git a/CONTRIBUTING.md b/CONTRIBUTING.md
index 07cf14ec..c8775fa8 100644
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,14 +88,25 @@ $ sudo setfacl --modify user:$USER:rw /var/run/docker.sock
 # $ sudo setfacl --remove user:$USER /var/run/docker.sock
 ```
 
+For the Slurm tests (pytest marker `slurm`) you will need the Slurm container
+built and running locally. We run it as a service in the CI pipeline, see the
+file `.github/workflows/ci.yaml` for details -- there is a Docker service initialized
+for the `test-slurm` job.
+
 Then you can run all the tests, with
 
 ```bash
 $ pytest -m ""
 ```
 
-or just the tests that require Docker:
+or just the tests that require Docker,
 
 ```bash
 $ pytest -m 'docker'
 ```
+
+or just the tests that require Slurm:
+
+```bash
+$ pytest -m 'slurm'
+```
````

File: test/integration/platforms/__init__.py [hunk 10]
```diff
@@ -0,0 +1,18 @@
...
+#
+# Autosubmit is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Integration tests for Autosubmit platforms."""
```
kinow, line 18: Missing newline at the end of the file.

kinow (commented):
Yeah, coverage reporting is missing some lines.

What I did on my laptop:

```bash
$ docker run --rm -it --cgroupns=host --privileged --volume /sys/fs/cgroup:/sys/fs/cgroup:rw --hostname slurmctld --name slurm-container -p 2222:2222 autosubmit/slurm-openssh-container:25-05-0-1

$ docker cp slurm-container:/root/.ssh/container_root_pubkey $HOME/.ssh/container_root_pubkey || echo "Failed to docker cp SSH key"

$ chmod 600 $HOME/.ssh/container_root_pubkey

$ cat <<- EOF >> $HOME/.ssh/config
Host localDocker
    HostName localhost
    User root
    StrictHostKeyChecking no
    UserKnownHostsFile /dev/null
    IdentityFile $HOME/.ssh/container_root_pubkey
    Port 2222
    ForwardX11 yes
EOF
```

Then, with the container `slurm-container` running and listening on `2222`, I ran the test with the debugger. But before that, I had set a breakpoint in each function [not covered](https://app.codecov.io/gh/BSC-ES/autosubmit/pull/2362/blob/autosubmit/platforms/slurmplatform.py) in the report.

image

It immediately triggered a breakpoint in `generate_submit_script`.

image

Will now look at the `coverage html` report.

kinow:
Did a `pip install -e .[all]`, deleted my existing `./htmlcov` directory and any existing `.coverage` file, and then issued this command.

```bash
$ pytest --verbose --strict-markers -n auto --max-worker-restart 0 --dist loadgroup --cov=autosubmit --cov-config=.coveragerc --cov-report=xml:test/coverage.xml --cov-append -m 'slurm' test/integration/platforms/test_slurmplatform.py
```

Then `coverage html`. @VindeeR locally generated coverage report actually shows `generate_submit_script` as covered.

image

So I think this means our codecov might be wrong? Since I copied the `pytest` lines from `pytest.ini` and from the GH Actions configuration for our test.

kinow (commented):
The test-slurm build was missing the coverage steps, so coverage reports were not being generated nor uploaded. @VindeeR is fixing that, and then I think we will be golden.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - autosubmit/autosubmit container now includes the `$USER` environment variable
   via its entrypoint #2359
+- Adding a Slurm-Docker-Container to the CI/CD and creating tests to increase the
+  coverage of the Slurm Platform and make sure the test are able to touch more code
```
kinow, line 6: Can you link the issue #977 at the end of the sentence, please? And I think we can shorten the sentence (increase the coverage == the test are able to touch more code?)... something like

> - Added a Docker container that launches a Slurm server in the CI/CD environment, and
>   added integration tests that use the container, improving the coverage of Slurm platform #977

kinow (changes requested):
I've marked all other files as reviewed. Just pending this one point about changelog...

kinow (approved):
Great work @VindeeR ! Approving assuming the GH Actions build passes and you think everything looks in order. Try to squash when merging the PR :+1: Thanks a ton!!!

PR #2365, opened 2025-05-19
Suppress unnecessary errors for nonexistent experiments

### LINKED ISSUE #1046: [enhancement] autosubmit delete shouldn't print a traceback when experiment does not exist
At the moment, running `autosubmit delete whatsmyname` will result in the following:

```bash
(autosubmit4) kinow@ranma:~/Development/python/workspace/autosubmit4-config-parser$ autosubmit delete whatsmyname
  File "/home/kinow/Development/python/workspace/autosubmit/bin/autosubmit", line 36, in main
    Autosubmit.parse_args()
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 648, in parse_args
    Autosubmit._init_logs(args, args.logconsole, args.logfile, expid)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 870, in _init_logs
    raise AutosubmitCritical("Experiment does not exist", 7012)
log.log.AutosubmitCritical:  
 [CRITICAL] Experiment does not exist [eCode=7012]
```

The traceback, the troubleshooting URL, the error code, the CRITICAL message, are all irrelevant to the user calling that. I believe the only important piece of information here is “Experiment does not exist”. It would be simpler to just return an exit code `1` and display that message.

Compare with other workflow manager tools:

1. Cylc

image

2. Nextflow

image

3. Autosubmit

image

isimo00:
This PR aims to declutter the console text displayed when users try to perform actions on experiments that do not exist, mailny Python's tracebacks.
Closes #1046 

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -907,13 +907,24 @@ class Autosubmit:
 
     @staticmethod
     def run_command(args: argparse.Namespace):
-        expid = "None"
-
         if hasattr(args, 'notransitive') and args.notransitive:
             warnings.warn('notransitive is deprecated and will be removed in a future major release!')
 
         if hasattr(args, 'expid'):
+            missing_expid = ""
             expid = args.expid
+            if expid :
+                expids = expid.replace(',', ' ').split(' ')
+                expids = [expid.lower() for expid in filter(lambda x: x, expids)]
+                expids = [x.strip() for x in expids]
+
+            for expid in expids:
+                exp_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(expid)
+                if not exp_path.exists():
+                    missing_expid +=  " "+expid
+            if missing_expid != "":
+                Log.error(f"Experiment{missing_expid} does not exist")
+                return 1
         if args.command != "configure" and args.command != "install":
             Autosubmit._init_logs(args, args.logconsole, args.logfile, expid)
         if args.command == 'run':
@@ -921,12 +932,8 @@ class Autosubmit:
                 raise AutosubmitCritical('Tracing is only available with profiling. Please add -p/--profile flag to run with tracing.', 7012)
             return Autosubmit.run_experiment(args.expid, args.start_time, args.start_after, args.run_only_members,
                                              args.profile, args.trace)
-        elif args.command == 'expid':
-            return Autosubmit.expid(args.description, args.HPC, args.copy, args.dummy, args.minimal_configuration,
-                                    args.git_repo, args.git_branch, args.git_as_conf, args.operational, args.testcase,
-                                    args.evaluation, args.use_local_minimal) != ''
         elif args.command == 'delete':
-            return delete_experiment(args.expid, args.force)
+            return delete_experiment(expids, args.force)
         elif args.command == 'monitor':
             return Autosubmit.monitor(args.expid, args.output, args.list, args.filter_chunks, args.filter_status,
                                       args.filter_type, args.hide, args.text, args.group_by, args.expand,
@@ -1111,7 +1118,14 @@ class Autosubmit:
                 expids = expid.split(" ")
             expids = [x.strip() for x in expids]
             for expid in expids:
-                exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, expid)
+                exp_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(expid)
+                as_conf = AutosubmitConfig(expid, BasicConfig, YAMLParserFactory())
+                as_conf.reload(force_load=True)
+
+                if len(as_conf.experiment_data) == 0:
+                    if args.command not in ["expid", "upgrade"]:
+                        raise AutosubmitCritical(f"Experiment {expid} has no yml data. Please, if you really wish to use "
+                                                 f"AS 4 prompt:\nautosubmit upgrade {expid}",7012)
                 tmp_path = os.path.join(exp_path, BasicConfig.LOCAL_TMP_DIR)
                 aslogs_path = os.path.join(tmp_path, BasicConfig.LOCAL_ASLOG_DIR)
                 if not os.path.exists(exp_path):
@@ -3120,9 +3134,9 @@ class Autosubmit:
             jobs_to_recover = job_list.get_job_list()
         else:
             jobs_to_recover = job_list.get_active()
-        
+
         selected_job_names = {job.name for job in jobs_to_recover}
-        
+
         # filters will be applied to all_jobs or only active_jobs, depending on the all_jobs flag
         if filter_section or filter_chunks or filter_status or filter_list:
             # Validate filters. Raises AutosubmitCritical if any filter is invalid, with a message specifying the issue.
@@ -5334,9 +5348,9 @@ class Autosubmit:
                 jobs_to_set_status = job_list.get_job_list()
                 selected_job_names = {job.name for job in jobs_to_set_status}
                 final_status = Autosubmit._get_status(final)
-                
+
                 Log.info("Filtering jobs...")
-                
+
                 selected_job_names = apply_job_filters(
                     job_list=job_list,
                     base_job_names=selected_job_names,
@@ -5348,7 +5362,7 @@ class Autosubmit:
                     filter_chunks_fn=Autosubmit._filter_jobs_by_chunks_splits,
                     status_from_str_fn=Autosubmit._get_status,
                 )
-                
+
                 # preserve job list ordering
                 final_list = [job for job in jobs_to_set_status if job.name in selected_job_names]
                 # Time to change status
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1129,6 +1129,7 @@ class Autosubmit:
                 aslogs_path = os.path.join(tmp_path, BasicConfig.LOCAL_ASLOG_DIR)
                 if not os.path.exists(exp_path):
                     if BasicConfig.DATABASE_BACKEND == 'sqlite':
+                        sys.tracebacklimit = 0
                         raise AutosubmitCritical("Experiment does not exist", 7012)
                     else:
                         # TODO: This needs #1352 issue, to have the workflow configuration fully
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -97,7 +97,7 @@ def new_experiment(description, version, test=False, operational=False, evaluati
                                  7011) from e
 
 
-def delete_experiment(expids: str, force: bool) -> bool:
+def delete_experiment(expids: list[str], force: bool) -> bool:
     """Deletes an experiment from the database,
     the experiment's folder database entry and all the related metadata files.
 
@@ -109,12 +109,10 @@ def delete_experiment(expids: str, force: bool) -> bool:
     # expid will come from argparse, which provides nix-style comma-separated values,
     # so here we parse the comma-separated values. ``.fromkeys`` keeps order and removes
     # duplicates.
-    expid_list = expids.replace(',', ' ').split(' ')
-    expid_list = [expid.lower() for expid in filter(lambda x: x, expid_list)]
 
     failed: list[str] = []
 
-    for expid in expid_list:
+    for expid in expids:
         try:
             _delete_experiment(expid, force)
         except Exception as e:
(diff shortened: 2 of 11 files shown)


VindeeR:
I believe we can close this PR @kinow @dbeltrankyl

kinow:
Has this been fixed, @VindeeR ? Last time I tried to delete a nonexisting exp I got an exception. But I haven't tried recently.

VindeeR:
I haven't tried it, I know it is suppressing the error in here, but I thought it was fixed for some reason.

I can test it, rebase, and merge if it was not

autosubmit/scripts/autosubmit.py
```diff
@@ -103,6 +103,8 @@ def main():
             return_value = Autosubmit.run_command(args)
         delete_lock_file()
     except BaseException as e:
+        if args.command == 'delete':
+            return 1
```
kinow: Don´t we have a better way of solving the issue without having to add this if here? Why was it needed?

autosubmit/autosubmit.py
```diff
@@ -1111,7 +1111,17 @@ def _init_logs(args, console_level='INFO', log_level='DEBUG', expid='None'):
                 expids = expid.split(" ")
             expids = [x.strip() for x in expids]
             for expid in expids:
-                exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, expid)
+                exp_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(expid)
+                if not exp_path.exists():
+                    Log.error(f"Experiment {expid} does not exist, therefore can't be deleted", 7012)
```
kinow: Maybe we can make the text a bit simpler, just "Experiment {expid} not found".

autosubmit/autosubmit.py
```diff
@@ -1111,7 +1111,17 @@ def _init_logs(args, console_level='INFO', log_level='DEBUG', expid='None'):
...
+                    quit()
```
kinow: This shouldn't be used, as far as I recall.

https://docs.python.org/3/library/constants.html#quit

It needs the site module installed/loaded, which is why normally in Python you see sys.exit.

We already have one point of exit in the code. It´d be better if we are able to keep that. Maybe we should modify the `delete_experiment` function to not raise an autosubmitcritical/error, and instead print the error and return False or 1.

kinow (changes requested):

autosubmit/scripts/autosubmit.py
VindeeR: if we don't stop there it will throw the same error as before again, the issue is that `BaseExpetion` will stop there at any error, but I think I just had an idea

kinow: Yeah, but I think we can change the `delete_experiment` function so it doesn't throw an error for experiments not found. That should remove the need for this change, I think/hope.

autosubmit/autosubmit.py
VindeeR: I was just over complicating things, I didn't realize there was a treatment later inside the `delete_experiment` so for some reason I believed that the code had to be stopped there

autosubmit/scripts/autosubmit.py
VindeeR: the error wasn't being throwing at the `delete_experiment` but at the `_init_logs` so all I needed to do is remove the error and return 1 in order to print them

test/integration/test_autosubmit.py
```diff
@@ -136,15 +137,16 @@ def test__init_logs_sqlite_exp_path_does_not_exist(autosubmit, autosubmit_exp, m
 
     monkeypatch.setattr(BasicConfig, 'DATABASE_BACKEND', 'sqlite')
 
-    with pytest.raises(AutosubmitCritical) as cm:
+    with pytest.raises(SystemExit):
```
kinow: We shouldn´t have SystemExit's happening. I think this is hiding some other issue.

autosubmit/autosubmit.py
```diff
@@ -1111,7 +1111,16 @@ def _init_logs(args, console_level='INFO', log_level='DEBUG', expid='None'):
...
+                exp_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(expid)
+                if not exp_path.exists():
+                    return 1
+                as_conf = AutosubmitConfig(expid, BasicConfig, YAMLParserFactory())
+                as_conf.reload(force_load=True)
+
+                if len(as_conf.experiment_data) == 0:
+                    if args.command not in ["expid", "upgrade"]:
+                        raise AutosubmitCritical(f"Experiment {expid} has no yml data. Please, if you really wish to use "
+                                                 f"AS 4 prompt:\nautosubmit upgrade {expid}",7012)
```
kinow: But I think this changes the behaviour a bit with deletion of multiple expids? (can't recall if thatś allowed).

Here if any expid doesn't exist we immediately stop the whole process. But maybe we should delete the ones we can, and say which ones failed.

If you do a `rm 1.txt does_not_exist.txt 3.txt`, Linux `rm` utility will give the user a message about `does_not_exist.txt` not existing, and will delete the remaining files, for example

test/integration/test_expid.py
VindeeR, line 1: Had to be changed due to removing the `raise AutosubmitCritical` at `autosubmit/experiment/experiment_common.py`

autosubmit/autosubmit.py
```diff
@@ -907,15 +907,12 @@ def parse_args() -> tuple[int, Optional[argparse.Namespace]]:
...
-
         if hasattr(args, 'notransitive') and args.notransitive:
             warnings.warn('notransitive is deprecated and will be removed in a future major release!')
 
-        if hasattr(args, 'expid'):
-            expid = args.expid
         if args.command != "configure" and args.command != "install":
-            Autosubmit._init_logs(args, args.logconsole, args.logfile, expid)
+            if Autosubmit._init_logs(args, args.logconsole, args.logfile) == 1:
+                return False
```
kinow: I don't know if this is safe. If Autosubmit cannot initialize the logs we probably have a serious problem and must raise an error...

kinow (commented):
I will review with more calm another day (sorry, tired brain & eyes 😅 ). Thanks for the os.path -> Path changes. I left a comment, but let me do a bit more thorough review another day before changing 👍 Thanks @VindeeR !

autosubmit/autosubmit.py
VindeeR: I believe in this scenario I changed it is fine, it'll only return this when the experiment is not created whenever the code runs smoothly it will never go to this return.

which means that as long as the experiment is created, the code inside the `_init_log`, ensures to always be setting the `set_file`. I believe it won't cause any issues regarding that from the changed code.

Although, I do believe there's no check for if it was initialized properly or not, I'll double check it.

autosubmit/experiment/experiment_common.py
```diff
@@ -134,7 +134,7 @@ def _delete_experiment(expid: str, force: bool) -> None:
 
     experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
     if not experiment_path.exists():
-        raise AutosubmitCritical("Experiment does not exist", 7012)
+        raise False
```
kinow: You cannot raise False, I believe. You must raise an exception. Is this currently being tested???

kinow (commented):
Found a `raise False` and stopped reviewing after that. Have a look at the code to confirm if that was intended, and if it's being tested. I think it's supposed to... give some strange runtime error? 😅

autosubmit/experiment/experiment_common.py
VindeeR: image

It was being tested.

I was going in between `return False` and `raise` and ended up getting them mixed up :sweat_smile: and because it raises an error inside the `try`, which was the idea, I didn't realize

Already changed just to `raise`

VindeeR: I just saw your other comment on the other [PR](https://github.com/BSC-ES/autosubmit/pull/3002#discussion_r3217665549)

And I believe this apply here too, the thing is if we print something here the log will keep showing unnecessary messages, the current flow allows for a cleaner logging, and better readability.

autosubmit/experiment/experiment_common.py
```diff
@@ -134,7 +134,7 @@ def _delete_experiment(expid: str, force: bool) -> None:
 
     experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
     if not experiment_path.exists():
-        raise AutosubmitCritical("Experiment does not exist", 7012)
+        raise
```
kinow: Is this covered in a test? I think `raise` ... would cause a runtime error? Don't we need to raise a specific exception?

kinow: Image

kinow (changes requested):

kinow:
Was going to review this one but looks like I already left some comments ~4 days ago (before the weekend). One more PR to go then.

autosubmit/experiment/experiment_common.py
VindeeR: @kinow which do you think should be the best approach?

autosubmit/autosubmit.py
VindeeR: I found a validation to raise an error inside the `log.py` when the log is not created properly, so this change is not necessary rolling it back

autosubmit/experiment/experiment_common.py
```diff
@@ -134,7 +134,7 @@ def _delete_experiment(expid: str, force: bool) -> None:
 
     experiment_path = Path(f"{BasicConfig.LOCAL_ROOT_DIR}/{expid}")
     if not experiment_path.exists():
-        raise AutosubmitCritical("Experiment does not exist", 7012)
+        raise Exception
```
kinow: Probably better to include a message, or it may show as simple `Exception` in the logs, and others will have to find the code used to figure out what happened. And I´d normally use a simpler exception like `ValueError` here. But the rest of the code in this function is raises `AutosubmitCritical`. Probably better to use that here too. @VindeeR there's one branch of the code that raises whatever exception was raised (a single `raise`). Maybe that too should be an `AutosubmitCritical`. Then we can update the pydoc of this function with something like

```
:raise AutosubmitCritical: If any error occurs while trying to delete the experiment from DB and disk.
```

autosubmit/experiment/experiment_common.py
kinow: I think an exception here, as the rest of the code fails with exceptions too.

autosubmit/experiment/experiment_common.py
kinow: ping this one @VindeeR

kinow (commented):
The result of deleting a nonexistent experiment after the change:

```
(venv) kinow@ranma:~/Development/python/workspace/autosubmit$ autosubmit delete 9rufsdoifds
[ERROR] Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='WARNING', command='delete', expid='9rufsdoifds', force=False, update_version=False)
[ERROR] This is the experiment: <9rufsdoifds> which had an issue with the command: <delete> and it is currently using the Autosubmit Version: <no version found>.
 [CRITICAL] Traceback (most recent call last):
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/scripts/autosubmit.py", line 104, in main
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 897, in run_command
    Autosubmit._init_logs(args, args.logconsole, args.logfile, expid)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 1132, in _init_logs
    raise AutosubmitCritical("Experiment does not exist", 7012)
autosubmit.log.log.AutosubmitCritical: Experiment does not exist
Error code: 7012

 [CRITICAL] Experiment does not exist [eCode=7012]
```

@VindeeR the main issue has some examples where the user is informed that the experiment does not exist without a Python traceback. https://github.com/BSC-ES/autosubmit/issues/1046

IMHO that makes more sense, as it's not really an exception. Here, we are sure the experiment does not exist and the best we can do is tell the user.

VindeeR:
As @kinow said, his new PR will be addressing this issue with a much larger structure change to improve readability and quality of the code, thus making this unneeded and deprecated. 

This PR will be close to address the issue at the #3171

Issue #1160, opened 2023-11-21
[enhancement][documentation] The examples in our docs must be self-contained


kinow:
We have some examples in our docs for frequency, splits, synchronization, wrappers, etc., that are really useful for users and also for developers.

I used some today to review @dbeltrankyl 's https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/387, but I had to copy the example, and then look at the graph, and guess what were the values under `EXPERIMENT`. Beginner users might struggle and give up using the examples.

kinow:
In GitLab by @dbeltrankyl on Dec 14, 2023, 14:53

Hello @kinow 

I'll expand this issue a bit to give more context and what to do aside of having self-contained experiments.

During !395 and !387 , we added a lot of logic into autosubmit workflow. 

Previously it was a very dummy code, in which autosubmit was just adding all the possible dependencies to all jobs, and then most of them were pruned with a transitive reduction. This process was painfully slow. 

The benefits of the dummy version were an "easy" integration of new features that affect the workflow as the transitive reduction was in charge of removing the redundant dependencies. 

From 4.1, this is no longer the case. 

So, to help the future developer who has to touch this new workflow logic ( per example for #1175  or #1177  ), I think we should generate, in the line of this issue, a database of experiments that contains all readthedocs examples ( and newer ones).

* [ ] This database should be run in the pipeline automatically.
* [ ] The database will consist of N amount of different experiments that match all the examples of the docs.
* [ ] Somehow.. retrieve this database experiments and autofill the docs? Not sure if doable * to match changes, if any
* [ ] The pipeline has to have somehow a checksum.
* [ ] The "checksum" of the plots can be done with the -d option ( this would require creating all the experiments once with 4.0, just to be safe ). Maybe there is a "image" diff viewer, but the simplest would be just using the -d option. This has already been manually tested and so far works for all but one experiment tested that the output has a different order

kinow:
In GitLab by @dbeltrankyl on Dec 14, 2023, 14:59

Also, I think that this database pipeline.. should be called only when the branch is ready to be merged not in each commit like we do usually, because it will take a bit and we already have resource issues with the runner

kinow:
Hi @dbeltrankyl ! Great idea on the database of experiments! We can possibly have these ones for the documentation, and a lot others to cover things that we know shall-not-fail.

[...]

PR #2366, opened 2025-05-20
[docs] Execute scripts to generate the documentation

VindeeR:
Partially addresses #1160

In this branch there wil be focus on 2 things:

- Automatize the documentation being generated on some of the functions executed by autosubmit
- Padronize the term EXPID throughout the documentation, so I'm removing the possibility of confusion
   - All the terms like a000|c000|cxxx|$expid|expid| became <EXPID>


.readthedocs.yaml
```diff
@@ -20,13 +20,18 @@ build:
...
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
    configuration: docs/source/conf.py
 
 # Optionally build your docs in additional formats such as PDF and ePub
-formats: all
+formats:
+  - htmlzip
+  - epub
+  - pdf
```
kinow, line 34: Interesting. The all didn't publish these other formats? Or was is required to change this in order to produce something outside of the scope of `all`?

docs/source/userguide/run/index.rst
```diff
@@ -412,49 +394,24 @@ Finally, you can launch Autosubmit *run* in background and with ``nohup`` (conti
...
-      -fa, --force_all      Stop all current running autosubmit processes
-      -c, --cancel          Orders to the schedulers to stop active jobs.
-      -fs FILTER_STATUS, --filter_status FILTER_STATUS
-                            Select the status (one or more) to filter the list of
-                            jobs. Default is SUBMITTED, QUEUING, RUNNING.
-			    
-      -t STATUS, --target STATUS
-                            Final status of killed jobs. Default is FAILED.
 
+.. runcmd:: autosubmit stop -h
```
kinow, line 406: :clap:

docs/source/userguide/create/index.rst
```diff
@@ -124,102 +94,6 @@ Example:
...
-         -H HPC, --HPC HPC     HPC to run experiment on it
-         -b BRANCH, --branch BRANCH
-                               branch of git to run (or revision from subversion)
-         -v, --update_version  Update experiment version
-
-Example:
-::
-
-    autosubmit test -c 1 -s 19801101 -m fc0 -H ithaca -b develop cxxx
-
```
kinow, line 222: GitHub UI is showing as if this command, but also all the text above about test cases was deleted. Is that really the case, and was it intentional?

kinow (commented):
Great work, @VindeeR ! To close #1160, I think what's missing is to create some examples (real workflows) used in docs, that use the same `runcmd` directive to run workflows and plot things, replacing the images (some which may be different now).

But we are going in the right direction here!

There's one part of the text that was deleted, but I am not sure if that was intentional. I left a comment about that, otherwise, looking great.

.readthedocs.yaml
VindeeR: For some reason, I couldn't figure exactly why, using the `all` was breaking the `..run` command, and it was breaking the build, I learned that removing the .pdf fixed a similar issue for someone else, but when I tried with all the format listed it worked so instead of removing one I let all of them as listed

docs/source/userguide/create/index.rst
kinow: Ah, got it. So let's remove that function too. Can you create a pull request for that?

kinow: Scratch that, I have some spare time now. Let me take care of that! :rocket:

kinow:
Using the RTD of this branch, https://autosubmit--2366.org.readthedocs.build/en/2366/, to test the Postgres branch now, #2187

docs/source/qstartguide/index.rst
```diff
@@ -45,32 +45,32 @@ Description of most used commands
...
-    * - **setstatus <expid>**
+    * - **setstatus <EXPID>**
       - Sets one or multiple jobs status to a given value.
 
 
 Create a new experiment
 =======================
 
-    ``autosubmit expid -dm -H "local" -d "Tutorial"``
+.. runcmd:: autosubmit expid -dm -H "local" -d "Tutorial"``
```
kinow, line 67: The rendered text shows the extra \`\` chars at the end... I think they can/have to be deleted?

docs/source/qstartguide/index.rst
```diff
@@ -90,20 +90,24 @@ The output of the command will show the expid of the experiment and generate the
...
+ To run an experiment, use ```autosubmit run <EXPID>```. Autosubmit runs experiments performing the following operations:
 
  - First, it **checks the experiment configuration**. If it is wrong, it won't proceed further.
  - Second, it **runs the experiment while retrieving all logs** from completed or failed tasks as they run.
  - Third, it manages all the **workflow steps by following the dependencies defined by the user** until all jobs are in COMPLETED or FAILED status. There can be jobs left in **WAITING** status if their dependencies are in **FAILED** status.
 
-While the experiment is running, it can be visualized via ``autosubmit monitor <expid>``.
+While the experiment is running, it can be visualized via ``autosubmit monitor <EXPID>``.
+
+.. runcmd:: find /home/docs/autosubmit/a000/plot/ -iname "*a000_*.png" -true -exec mv -- {} /home/docs/checkouts/readthedocs.org/user_builds/autosubmit/checkouts/2366/docs/source/qstartguide/dummy.pdf \;
```
kinow, line 108: Hmmm, maybe we can think in a one to make this simpler & portable. I cannot run this on my computer as it will fail.

kinow (commented):
Here's an example that's not self-contained, @VindeeR. Following the links from left/top to right/bottom in our docs, we start with the landing page, getting started, installation, and then the user guide.

The first part is to create an experiment (quite messy, we could simplify it, maybe in another PR), and then configure experiment, where we have instructions to add a heterogeneous job: https://autosubmit--2366.org.readthedocs.build/en/2366/userguide/configure/index.html#how-to-add-a-new-heterogeneous-job

(typo whcih in that page)

The section ends with

> This will create a new job named new_hetjob with two components that will be executed once.

Now we can ask ourselves:

* how can users know if that's working
* what's the output of create or inspect? does it matter/change?
* how can we, devs of AS, make sure that those docs are correct?

That's the main motivation for this issue of self-contained examples. To make sure these questions are answered with simple/clear instructions, and with automation so that we don't have to manually check that for each release.

kinow:
Most of the documentation in Configuring and Defining experiments in the user guide section are not self-contained :sweat: 

e.g. https://autosubmit--2366.org.readthedocs.build/en/2366/userguide/defining_workflows/index.html#running-jobs-once-per-startdate-member-or-chunk

There is a graph below the example YAML:

image

Simply pasting the JOBS in an existing experiment configuration will not render that same, unless they have the expected number of start dates, members, chunks.

But even writing what it's supposed to be configured, new users still need to go back to other sections to remember how to configure each. Easier to give the configuration and highlight what needs to be modified, so that the user is learning with the right context in mind.

kinow:
(and probably many examples broken...)

That's all! I'll not bother you here as I think I'm nearly done with my tests for Postgres. Thanks!

docs/source/qstartguide/index.rst
```diff
@@ -64,7 +64,7 @@ Description of most used commands
 Create a new experiment
 =======================
 
-.. runcmd:: autosubmit expid -dm -H "local" -d "Tutorial"``
+.. runcmd:: autosubmit EXPID -dm -H "local" -d "Tutorial"
```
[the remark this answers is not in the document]
VindeeR: yeap, it does, I fixed in my fork and forgot I didn't have replicated it here

kinow:
It'd be nice if we could

1. wrap it up so that it runs some commands with the runcmd
2. confirm the build time is not too bad (e.g. less than our unit tests job)
3. try to run it with coverage (something like this https://coverage.readthedocs.io/en/7.8.2/cmd.html)

manuel-g-castro:
real examples and workflows.

kinow (commented):
@VindeeR here's the idea I mentioned about coverage of Sphinx.

What we are doing is not very conventional... normally people don't care about the coverage of sphinx, as most people just have restructuredtext or markdown. But since we are running Python code to build the docs, with the directives like `runcmd`, I think it doesn't hurt to collect the coverage for that :grimacing: 

Here's a quick diff showing what I had in mind:

```diff
diff --git a/.github/workflows/ci.yaml b/.github/workflows/ci.yaml
index edb88b21..66fa1968 100644
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -178,6 +178,51 @@ jobs:
           path: coverage.xml
           retention-days: 7
 
+  # NOTE: The docs are already built on ReadTheDocs. We perform another quick build,
+  #       without checking or deploying the produced HTML, just to i) verify that
+  #       we can build it and ii) collect the coverage of the examples executed.
+  test-docs:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    strategy:
+      matrix:
+        python-version: [ "3.9" ]
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install system dependencies
+        run: sudo apt-get install -y curl graphviz rsync
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip packaging setuptools twine
+          pip install --upgrade -e .[all]
+
+      # Run regression tests
+      - name: Build docs
+        run: |
+          cd docs && make html_coverage
+
+      - name: Coverage report
+        run: |
+          coverage xml
+          coverage report
+
+      - name: Upload coverage artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage_docs-${{ matrix.python-version }}
+          path: coverage.xml
+          retention-days: 7
+
   test-misc:
     runs-on: ubuntu-latest
     timeout-minutes: 5
@@ -206,7 +251,7 @@ jobs:
           cffconvert --validate -i CITATION.cff
 
   coverage:
-    needs: [test, test-integration, test-regression]
+    needs: [test, test-integration, test-regression, test-docs]
     runs-on: ubuntu-latest
     timeout-minutes: 2
     steps:
diff --git a/docs/Makefile b/docs/Makefile
index 859b132c..a3f83a04 100644
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -196,6 +196,13 @@ coverage:
        @echo "Testing of coverage in the sources finished, look at the " \
              "results in $(BUILDDIR)/coverage/python.txt."
 
+html_coverage:
+       coverage run $$(which $(SPHINXBUILD)) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+       coverage combine
+       coverage html
+       @echo
+       @echo "Build finished. The HTML pages are in $(BUILDDIR)/html. Coverage in $(BUILDDIR)/htmlcov."
+
 xml:
        $(SPHINXBUILD) -b xml $(ALLSPHINXOPTS) $(BUILDDIR)/xml
        @echo
```

That adds a new `make`` target, `html_coverage` (leaving the default `coverage` untouched, who knows, we may use it some day? :man_shrugging:). It does build the HTML, just like Sphinx does. This shouldn't take long, probably less than one minute the whole job.

But, if you look at the command executed, instead of `sphinx-build` (the python executable of Sphinx used to build everything, identified by `SPHINXBUILD` env var), we call `coverage`. This is Ned's Coverage tool. The same used by `pytest-cov`. Some long time ago, most people called `coverage` directly, like what we have here, but nowadays it's integrated in test harness & plugins, like `pytest-cov`. But you can still use it directly.

So here we call that utility, passing the `sphinx-build` absolute path (fails with relative path). We also need to copy the `.coveragerc` of Autosubmit, but modify the paths,

```ini
# This is the Coverage.py configuration file. This is used by CI when running
# the tests and collecting coverage

[run]
branch = True
cover_pylib = False
concurrency = multiprocessing
data_file = .coverage
disable_warnings =
    trace-changed
    module-not-python
    module-not-imported
    no-data-collected
    module-not-measured
omit =
    build/*
parallel = True
source = ../autosubmit
timid = False

[report]
exclude_lines =
    pragma: no cover

    # Don't complain if tests don't hit defensive assertion code:
    raise NotImplementedError
    return NotImplemented

    # Ignore type checking code:
    if (typing\.)?TYPE_CHECKING:
    @overload( |$)

    # Don't complain about ellipsis (exception classes, typing overloads etc):
    \.\.\.

    # Ignore abstract methods
    @(abc\.)?abstractmethod

fail_under=0
ignore_errors = False
omit =
    tests/*
precision = 2
show_missing = False
skip_covered = False
sort = Name

[html]
directory = htmlcov

[xml]
output = coverage.xml
package_depth = 99
```

Place this file in `docs`, apply the diff, then `cd docs` and run `make html_coverage`. Then look at `htmlcov/index.html`. You should see something like this:

image

If all of this works, then I think you can try it in GH Actions and check if it uploads the report to Codecov. No idea whether it will improve our coverage or not, but we will see with time if we will be testing things only in the docs, perhaps with some of the future example workflows that we will have.

I hope it makes sense.

docs/source/ext/runcmdquiet.py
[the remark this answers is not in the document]
VindeeR: this makes more sense, I'll check on how to do it and try to implement later as an improvement

docs/source/conf.py
```diff
@@ -36,8 +36,9 @@
     'sphinx_rtd_theme',
     'sphinx_reredirects',
     'sphinx.ext.graphviz',
+    'sphinx.ext.coverage',
     'autosubmit_variables',
-    'helloworld',
+    'runcmdquiet',
```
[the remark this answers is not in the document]
VindeeR: I'm using this tag so I'll try to implement the silent before, otherwise it'll show commands that are not supposed to be shown

kinow:
I think we can exclude the docs from the code coverage report. Wdyt Erick?

VindeeR:
> I think we can exclude the docs from the code coverage report. Wdyt Erick?

I don't think it is necessarily a problem, or a bad thing.

Yes it'll decrease the coverage, but at the same time, it's showing a more accurate percentage, this could even help because there were a few `if-statement` that it started to show as well.

I think it's okay to leave it be even if it decreases IMHO but if you think it's better to remove we can easily remove too

image
image

.github/workflows/ci.yaml
```diff
@@ -283,7 +283,7 @@ jobs:
           cffconvert --validate -i CITATION.cff
 
   coverage:
-    needs: [test, test-integration, test-regression, test-slurm]
+    needs: [test, test-integration, test-regression]
```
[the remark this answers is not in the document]
VindeeR: Not too sure what happened there, but the rebase broke the ci.yaml, I think it should be fine now

docs/source/ext/runcmd.py
```diff
@@ -153,22 +153,21 @@ def run(self):
...
         # prompt = "prompt" in self.options
         # We patched this so that the prompt is displayed by default, similar
         # to how ``{code-block} console`` works.
-        prompt = True
         dedent_output = self.options.get("dedent-output", 0)
 
         # Dedent the output if required
         if dedent_output > 0:
-            output = "\n".join([x[dedent_output:] for x in output.split("\n")])
+            output = ""
```
kinow, line 166: Was this like this before? Just checking as I don't recall seeing this before.. but maybe I missed it. This doesn't seem to be dedenting the output anymore, instead it's removing the output now?

kinow: So now dedent silents the script?

kinow: Not really. It is like using a flag to indent text to actually delete text or silent it.

I think the correct approach would be to add a new tag to omit/silence/quiet the output. Leaving dedent to dedent it.

Otherwise, in some years another dev could need to use the dedent arg and be surprised when they find it actually does something else.

I will try to fix that later today or over next days.

VindeeR: I changed the name of the tag now too, so in this case that would be fixed, we wont have `dedent`, but we were already not using it

VindeeR: Sorry for poorly expressing myself before, but I just did a commit fixing it changing everything from `dedent` to `silent`

VindeeR: Okay, I think I just manage to add it, not sure what I did wrong before, most likely me missing something during previous tries on my fork, I'll make a couple of tests and commit it shortly @kinow

VindeeR:
It should all be good now, sorry for the mess

PR #2367, opened 2025-05-20
Reduce the total running time of test_expid (~10s)

### LINKED ISSUE #2355: [tests] Slowness in test_expid
The unit test `test_expid.py` stated taking 15+ seconds since a few days ago, when @VindeeR fixed a part of the code or tests. But I think the reason for that slowness might be not related to the PR that Erick was working on. Not entirely sure.

I noticed that as we now have `pytest --durations=5`, so it always shows the 5 slowest tests, and we used to have one taking 4 seconds and the other ones about 1 second or less.

Debugging it in the PostgreSQL branch (to make sure it wasn't due to the changes there), https://github.com/BSC-ES/autosubmit/pull/2187, I noticed that the test mocks a lot of calls, and then `Autosubmit.expid` gets called, but when it saves the experiment details, at the very end of the `Autosubmit.expid` function, it fails to locate the expid in the DB and the process reaches the 15 seconds timeout.

That's what's causing the test slowness now.

I think the solution will be to review the mocks in that test, or confirm if that's already covered by an integration test. If so, we might as well delete that unit test 👍

VindeeR:
closes #2355

Test has a problem in which due to an expected failure it is taking 10+ seconds to be finished looking for a way to mitigate this time to reasonable execution time

### DIFF
--- a/test/unit/test_expid.py
+++ b/test/unit/test_expid.py
@@ -23,6 +23,7 @@ from contextlib import contextmanager
 import pytest
 
 from autosubmit.autosubmit import Autosubmit
+from autosubmit.database import db_common
 from log.log import AutosubmitCritical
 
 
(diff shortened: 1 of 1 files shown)


VindeeR:
ooops @kinow I forgot to leave this issue as a draft thanks for the comment :smile:

kinow:
You left it as a draft alright, but I marked it as ready to review as it looked pretty much done, sorry :grimacing:

PR #2383, opened 2025-05-27
Logfile compression

### LINKED ISSUE #2266: Automatic compression of logfiles
This issue is to move to implement compression of log files in Autosubmit, at different levels:

- Compression of the Autosubmit log files
  - There could be a parameter to enable it and compress all the command-related logs excepting the current
- Compression of job .out and .err files
  - There should be another parameter to activate it
  - One option is to do the compression in the remote platform to save transfer time, but this could create a system dependency

On the API side, there should be an online decompression feature to show the logs. If it's done entirely in memory it could be dangerous. There could be a cache for large files.

Interestingly, vi and vim recent versions support reading compressed files, so we shouldn't create a lot of hassle

isimo00:
Add logfile compression using xz algorithm

Closes #2266

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -196,8 +196,8 @@ class Autosubmit:
                                 help="sets file's log level.")
             parser.add_argument('-lc', '--logconsole', choices=log_levels, default='WARNING', type=str,
                                 help="sets console's log level")
-            parser.add_argument('--compress_logs', action='store_true', default=False,  
-                                help='Compress log files after writing')
+            parser.add_argument('--compress_logs', type=int, default=0,  
+                                help='Number of compressed log files')
 
             subparsers = parser.add_subparsers(dest='command')
             # Run
@@ -836,6 +836,21 @@ class Autosubmit:
             if args.all or args.force_all:
                 expid_less.append("stop")
         global_log_command = ["delete", "archive", "upgrade"]
+
+        if args.compress_logs:
+            uncompressed_files = Log.find_uncompressed_files(BasicConfig.expid_aslog_dir(expid))
+            # Each log record has an .err and .out file: 2 files to compress
+            # per log record
+            num_files_to_compress = 2*args.compress_logs - len(uncompressed_files) 
+            while(num_files_to_compress != 0 and uncompressed_files):
+                # List resturns True if not empty
+                # Files are named with timestamps 
+                # so the most recent ones will be at the end of the list
+                Log.compress_logfile(uncompressed_files[-1])
+                uncompressed_files.pop()
+                num_files_to_compress -= 1
+
+
         if "offer" in args:
             if args.offer:
                 global_log_command.append("migrate")  # offer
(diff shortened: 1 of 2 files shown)


LuiggiTenorioK:
@isimo00 I saw that the changes update the writing functionality but not the reading one. For example, I think that the `autosubmit cat-log` needs to be updated as well.

@dbeltrankyl is there any other autosubmit command that reads the logs that I'm missing, or is it just `autosubmit cat-log`?

isimo00:
Will do and report how that works! Although I'm getting very behind with coding because of the presentation, so probably for next week

log/log.py
```diff
@@ -240,6 +241,8 @@ def set_file(file_path, type='out', level="WARNING"):
                     status_file_handler.setFormatter(LogFormatter(False))
                     status_file_handler.addFilter(custom_filter)
                     Log.log.addHandler(status_file_handler)
+                if compress:
+                    compress_logfile(file_path)
```
kinow, line 245: Is it compressing only after the log was written? It looks to me like this could be called shortly after the file is created the the Python logging handler is configured?

isimo00:
I could change the axis of the last graph to log scale, which would make differences more visual, but I'm afraid it can make file size less intuitive. I'll give it a shot

You pointed out something I forgot to mention, I did use Ulf's parameters for xz - dind't realize because I called the function I'm implementing for xz, which is tuned to do what Ulf does - maybe the comparison in size is not so fair then.

isimo00:
I think besides @kinow's comment, which I'll address tomorrow, this should be ready to go @LuiggiTenorioK. If you want to review it or check it isn't missing anything you'd need it to have please do so. I've stuck with Ulf's algorithm and flags.

autosubmit/autosubmit.py
```diff
@@ -891,35 +893,35 @@ def _init_logs(args, console_level='INFO', log_level='DEBUG', expid='None'):
             if args.command == "stop":
                 exp_id = "_".join(expids)
                 Log.set_file(os.path.join(BasicConfig.GLOBAL_LOG_DIR,
-                                          args.command + exp_id + '.log'), "out", log_level)
+                                          args.command + exp_id + '.log'), "out", args.compress_logs, log_level)
```
LuiggiTenorioK, line 896: Is it not the third argument the log level?

```suggestion
                                          args.command + exp_id + '.log'), "out", log_level, args.compress_logs)
```

The same happened with all the changes of the `Log.set_file` function in this file.

log/log.py
dbeltrankyl: Hello @isimo00,

Regarding the original request: https://github.com/BSC-ES/autosubmit/issues/2266#issue-2966975357.

I would introduce a call for the function `def _init_logs(args, console_level='INFO', log_level='DEBUG', expid='None')`, that checks if there is some uncompressed *.out, *err files.

Maybe it is worth considering if we can modify the parameter to be more than just TRUE or FALSE, allowing users to specify how many logs will remain visible. 

```python
aslogs_path = os.path.join(tmp_path, BasicConfig.LOCAL_ASLOG_DIR)
Log.set_file(os.path.join(aslogs_path, args.command + '.log'), "out", log_level)
Log.set_file(os.path.join(aslogs_path, args.command + '_err.log'), "err")
```

You could iterate over `aslogs_path` to find all uncompressed `.out` and `.err` files that do not match the current log files.

Based on the current code, correct me if I'm wrong, but it appears that the implementation would compress the current file. I believe we want to avoid compressing it until the end of the execution or the next command is run.

So, the easy solution is just to check it whenever the next command is launch. Another option, is to insert a call in the scripts/autosubmit.py so after everything is done, you compress the file ( but, I think is is more user friendly to have the last log uncompressed)


This is for the ASlogs.

Now we're missing the other part that I think is more important. 

The compression of the job logs ( $expid/tmp/LOG_$EXPID). 

Autosubmit only download them, in that case the idea is to compress them as soon as they're downloaded. ( Or compress in the remote and then download, but that creates a system dependency)

You could insert a call in def recover_platform_job_logs(self, as_conf) -> None: 

In this case, we only want to compress the out and err of the current job ( job.local_logs)

isimo00: Hi @dbeltrankyl I've implemented your suggestions to cast the `compressed_logs` argument as int rather than bool and added a function - called in `_init_logs` - that checks for uncompressed files and compresses them according to the user's input. I will reuse this new function for the job logs too.
Thank you!

autosubmit/autosubmit.py
```diff
@@ -196,6 +196,8 @@ def parse_args() -> Tuple[int, Optional[argparse.Namespace]]:
                                 help="sets file's log level.")
             parser.add_argument('-lc', '--logconsole', choices=log_levels, default='WARNING', type=str,
                                 help="sets console's log level")
+            parser.add_argument('--compress_logs', type=int, default=0,  
+                                help='Number of compressed log files')
```
LuiggiTenorioK, line 200: Could it be better to have it as a boolean as before?

It feels a little weird to have to guess how many uncompressed files I have to make the command work properly. Maybe it will be better to:

* Have the boolean `--compress_logs` to indicate to compress any new log generated by the command to run
* Another command like `autosubmit compress` to make operations to compress logs (list uncompressed logs, choose to compress all uncompressed, etc). Kinda like a "migration tool".

mcastril:
I am looking at the plots: https://github.com/BSC-ES/autosubmit/pull/2383#issuecomment-3023636693

and the comment:

> In terms of time, it is important to mention that xz does not have a well-maintained Python wrapping, so we call the subprocess Python module and run the compression via command line. 

I think it's a good idea to compare, but did you try what I had suggested in my [comment](https://github.com/BSC-ES/autosubmit/issues/2266#issuecomment-2893317170)? I thought LZMA in Python provided a native implementation for XZ.

https://docs.python.org/3/library/lzma.html

Who can take over this development?

dbeltrankyl:
> Who can take over this development?

@LuiggiTenorioK will (he offered voluntarily in the Slack autosubmit-dev channel)

kinow:
p.s. Some git conflicts :grimacing:

PR #2406, opened 2025-06-05
Add Docker option to run with postgres

LuiggiTenorioK:
Depends on https://github.com/BSC-ES/autosubmit/pull/2187 change to `autosubmit configure`

Running this with `docker build -e AS_PG_CONN_URL=postgres://...` set will configure the container to use Postgres automatically.

Since this depends on #2187 it will be good to test it with that branch using `docker build --build-arg=AUTOSUBMIT_GIT_REF=git+https://github.com/BSC-ES/autosubmit.git@as-postgres#egg=autosubmit[postgres] ...`


kinow (approved):
Looks good, and doesn't need a changelog entry IMO. I think this should be ready to be merged, @LuiggiTenorioK. Thanks!

LuiggiTenorioK:
> That's what I thought. I guess it should be OK as we will not add anything in our documentation, and if anyone tries that, they are probably doing something they are not supposed to do anyway, or they must know what they are doing with the container. But if you prefer to wait for that PR, I think we should make it a draft and link issues/PRs.

Yes, I think it is safe to merge as well. This is not the main Dockerfile after all.

PR #2425, opened 2025-06-17
Add $USER variable to our Docker container, and test it in GH Actions

### LINKED ISSUE #2359: [docker] AS container missing $USER env var
### Description

I think this was reported by @ainagaya in the workflow repository, but I missed the link to the issue.

To verify, start the container, and check variables that AS may use like `USER` and `HOME`. If these are missing some commands in AS may fail.

To fix it, add the variable in our `Dockerfile`, and also add a note in the documentation, in the installation docs about the need to have this variable propertly configured.

### Expected Behaviour

All commands working successfully.

kinow:
Closes #2359

Had some minutes today before going home, so wrote a quick fix for #2359, and then added some code to see if we can build with docker and run some basic docker tests.

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,46 @@
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+permissions:
+  contents: read
+
+jobs:
+  build-docker-images:
+    name: Build Docker images
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ vars.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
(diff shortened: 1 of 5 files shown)


docker/Dockerfile
```diff
@@ -190,13 +190,15 @@ RUN chown -R "${MAMBA_USER}:${MAMBA_USER}" "/home/${MAMBA_USER}/" && \
 
 COPY <<EOF /usr/local/bin/entrypoint.sh
 #!/bin/bash
+export USER=$(whoami)
```
kinow, line 193: This is the fix. `$USER` and other vars are set by bash, and not guaranteed in Linux to exist. That's why Docker doesn't set those, and we didn't get `$USER` as that's done normally in an interactive env with the default bash settings + profile/bashrc/etc.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,46 @@
...
+          # TODO: for a separate issue, try pushing it
+          push: false
+          load: true
+          tags: autosubmit/autosubmit:latest
+      
+      - name: Test the image
+        run: |
+          docker run --rm autosubmit/autosubmit:latest autosubmit --version
+          # https://github.com/BSC-ES/autosubmit/issues/2359
+          docker run --rm autosubmit/autosubmit:latest /bin/bash -c '[ -n $USER ] || exit 1'
```
kinow, line 46: This is the test :+1:

kinow:
Can you have a look at this, please, @VindeeR ? Thank you.

VindeeR (approved):
Cool detail with the permissions I gave it some read and it sounds like quite a few things can be done using them.

Docker builds normally too so all good to go, is this going to be released at the next build? v4.16?

Shouldn't it be better to wait after the release to merge?

Issue #2285, opened 2025-04-08
AS4.1.13 and 4.1.11 sbatch errors ( Revise the platform connectivity issues triggers for what is an AutosubmitCritical )


DMathas:
Hi @dbeltrankyl the RES used for the eMONARCH training data genration has been consumed, so in order to keep running I need to do some modifications in the configs. Could you help me how I should adjust my config and workflow?
In an exp still running with AS4.1.11 the following happens even if I set the MAXWAITINGJOBS and TOTALJOBS to `1`, causing the workflow to crash, for example in exp `a8zb`:
```
[WARNING] Script a8zb_20201216_002_231_POSTPROC_BUNDLED has some empty variables. An empty value has substituted these variables
[ERROR] TRACE:Bad Parameters.
set intel compilers as mpi wrappers backend
load impi/2021.10.0 (path, manpath, ld_library_path)
load mkl/2023.2.0 (ld_library_path)
load ucx/1.15.0 (path, ld_library_path, library_path, c_include_path,
cplus_include_path) 
load bsc/1.0 (path, manpath)
sbatch: error: maxsubmitjobsperaccount
sbatch: error: batch job submission failed: job violates accounting/qos policy (job submit limit, user's size and/or time limits)
 JOBS:['a8zb_20201216_002_231_POSTPROC_BUNDLED']
Command squeue -o %A,%.50j -n a8zb_20201216_002_231_POSTPROC_BUNDLED in glogin2.bsc.es warning: Set INTEL compilers as MPI wrappers backend
load impi/2021.10.0 (PATH, MANPATH, LD_LIBRARY_PATH)
load mkl/2023.2.0 (LD_LIBRARY_PATH)
load UCX/1.15.0 (PATH, LD_LIBRARY_PATH, LIBRARY_PATH, C_INCLUDE_PATH,
CPLUS_INCLUDE_PATH) 
load bsc/1.0 (PATH, MANPATH)

 [CRITICAL] Check job and queue specified in your JOBS definition in YAML. Sections that could be affected: POSTPROC_BUNDLED [eCode=7014]
```

In exps running with AS4.1.13 (for example `a900`) decreasing the MAXWAITINGJOBS and TOTALJOBS seems to help (only queing for now), but if set too high similar error:
```
[ERROR] TRACE:Bad Parameters.
set intel compilers as mpi wrappers backend
load impi/2021.10.0 (path, manpath, ld_library_path)
load mkl/2023.2.0 (ld_library_path)
[...]

PR #2428, opened 2025-06-18
WIP: Handle ConnectionError when executing commands over SSH

### LINKED ISSUE #1369: [enhancement] retry job submission on network errors
This happened in a testing suite for DestinE. The experiment was running fine, but apparently there was some network glitch and when Autosubmit tried to submit jobs, paramiko raised an exception, which caused a critical error.

Perhaps we can improve this, to have retrials for paramiko/network errors like this. This could help operators/devs in case the network becomes unstable due to traffic or maintenance.

---

From chat with @dbeltrankyl on the DestinE workflow repo (weekly 28):

@dbeltrankyl 

>> Is there a way to tell Autosubmit to try to send files N times with X interval? i.e. try to send it now and if it fails, sleep X seconds, then keep trying?

It should do something similar to that.

It raises an autosubmiterror or autosubmitcritical. If it is a critical it is just that is an unrecoverable error.

If it raises an error, like in the log, it should test all the connections from all platform and try again later. ( once all platforms are working again)

Can you show me the complete log in the submission?

---

@kinow 

>Can you show me the complete log in the submission?

I think it's this one you want, right:

```
Calculating possible ready jobs for LUMI-LOGIN
ESC[32mSection:DN can submit 1 jobs at this timeESC[0mESC[39m
ERROR:paramiko.transport:Socket exception: Connection reset by peer (104)
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/platforms/paramiko_platf
orm.py", line 1040, in send_command
    stdin, stdout, stderr = self.exec_command(command, x11=x11)
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/platforms/paramiko_platf
orm.py", line 969, in exec_command
    chan = self.transport.open_session()
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/paramiko-3.4.0-py3.9.egg/paramiko/transport.py", line 959, in 
open_session
    return self.open_channel(
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/paramiko-3.4.0-py3.9.egg/paramiko/transport.py", line 1090, in
 open_channel
    raise e
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/paramiko-3.4.0-py3.9.egg/paramiko/transport.py", line 2159, in
 run
    ptype, m = self.packetizer.read_message()
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/paramiko-3.4.0-py3.9.egg/paramiko/packet.py", line 463, in rea
d_message
    header = self.read_all(self.__block_size_in, check_rekey=True)
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/paramiko-3.4.0-py3.9.egg/paramiko/packet.py", line 308, in rea
d_all
    x = self.__socket.recv(n)
ConnectionResetError: [Errno 104] Connection reset by peer

  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/platforms/paramiko_platform.py", line 1417, in check_remote_log_dir
    if self.send_command(self.get_mkdir_cmd()):
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/platforms/paramiko_platform.py", line 1131, in send_command
    raise AutosubmitError(str(e),6016)
log.log.AutosubmitError:  

  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/job/job_packages.py", line 203, in submit
    self._send_files()
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/job/job_packages.py", line 240, in _send_files
    self.platform.send_file(self._job_scripts[job.name])
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/platforms/paramiko_platform.py", line 378, in send_file
    self.check_remote_log_dir()
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/platforms/paramiko_platform.py", line 1424, in check_remote_log_dir
    raise AutosubmitError("Couldn't send the file {0} to HPC {1}".format(
log.log.AutosubmitError:  

  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/platforms/platform.py", line 308, in submit_ready_jobs
    package.submit(as_conf, job_list.parameters, inspect, hold=hold)
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/job/job_packages.py", line 209, in submit
    raise AutosubmitCritical("Error while submitting jobs: {0}".format(e), 7013)
log.log.AutosubmitCritical:  

  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/autosubmit.py", line 2229, in run_experiment
    Autosubmit.submit_ready_jobs(as_conf, job_list, platforms_to_test, packages_persistence, hold=False)
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/autosubmit.py", line 2516, in submit_ready_jobs
    save_1, failed_packages, error_message, valid_packages_to_submit, any_job_submitted = platform.submit_ready_jobs(as_conf,
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/platforms/platform.py", line 346, in submit_ready_jobs
    raise AutosubmitCritical(e.message, e.code, e.trace)
log.log.AutosubmitCritical:  

  File "/appl/AS/4.1.9-final-beta/lib/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/EGG-INFO/scripts/autosubmit", line 59, in main
    return_value = Autosubmit.parse_args()
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/autosubmit.py", line 707, in parse_args
  File "/appl/AS/4.1.9-final-beta/lib64/python3.9/site-packages/autosubmit-4.1.9-py3.9.egg/autosubmit/autosubmit.py", line 2340, in run_experiment
    raise AutosubmitCritical(e.message, e.code, e.trace)
log.log.AutosubmitCritical:  

ESC[1m ESC[31m[CRITICAL] Error while submitting jobs:   [eCode=7013]ESC[0mESC[39m
(END)
```

This was logged right after other jobs succeeded, and I think the connection to submit new jobs was not successful, but it didn't try to submit again... I had a look at `job_packages` but I couldn't find a place with retrials for job submission if a network error occurs...

---

@dbeltrankyl 

> I think it's this one you want, right:

Yes, thanks

> This was logged right after other jobs succeeded, and I think the connection to submit new jobs was not successful, but it didn't try to submit again... I had a look at `job_packages` but I couldn't find a place with retrials for job submission if a network error occurs...

The problem is that this error is not handled as a "network error", I don't recall having this error before. 

When there is an ERROR, which is not the case, it will trigger the code of Autosubmit.py 2263 ( inside def run_experiment()

Amount other stuff, there is a `                                Autosubmit.restore_platforms(platforms_to_test, mail_notify=mail_notify,
                                                             as_conf=as_conf, expid=expid)`

Once the connection is restored, autosubmit run will return to the main loop and continue the experiment from where it crashed. 

However, if it is an Autosubmit critical, autosubmit will stop

So what is missing is to handle this error so the recovery routine can be triggered

The submission process is a little bit delicate tho, if there is a failure in check the status of the logs there is no issue, that command also has retries etc and when retrials are spent it will trigger the autosubmit error to make a new connection.

But the submission could "fail" and still something be submitted(which happened before), so we have to be careful with "ghost" jobs

kinow:
Closes #1369
Related to #2285 (not sure if this will fix that -- yet)

**Check List**

- [ ] Documentation updated.


autosubmit/job/job.py
```diff
@@ -2938,8 +2935,7 @@ def _check_inner_jobs_queue(self, prev_status :str) -> None:
             elif reason == '(JobHeldAdmin)':
                 Log.debug(
                     "Job {0} Failed to be HELD, canceling... ", self.name)
-                self._platform.send_command(
-                    self._platform.cancel_cmd + " {0}".format(self.id))
+                self._platform.send_command(self._platform.cancel_cmd + " {0}".format(self.id))
```
kinow, line 2938: Just tidying up code here, no real changes.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1018,103 +1037,91 @@ def x11_status_checker(self, session, session_fileno):
...
-                    except BaseException as e:
+                    except Exception as e:
                         raise AutosubmitCritical(f"Failed to execute command: {e}")
                     chan_fileno = chan.fileno()
                     self.poller.register(chan_fileno, select.POLLIN)
                     self.x11_status_checker(chan, chan_fileno)
                 else:
+                    chan = self.transport.open_session()
                     chan.exec_command(command)
+
```
kinow, line 1086: The changes above are mainly cosmetics. Worth mentioning: Catching `Exception` to avoid keyboard interrupt or system exit, etc. (recommended best practice, if we do have something from baseexception and not exception, there's probably not much we can do, and we can actually stall things by capturing it...)

And also the change in the logic above, before there were two `if x11/else`, now there's  single `if x11/else`.

kinow:
Already have a test case for success path in exec_command and send_command. I think I can modify that one to stop the SSH connection while exec_command is executed, simulating the SSHConnection, and then maybe shutdown the container to simulate the socket error.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -37,15 +56,59 @@ def wrapper(*args, **kwargs):
 
     return wrapper
 
-# noinspection PyMethodParameters
-class ParamikoPlatform(Platform):
```
kinow, line 41: The class is still here, just further down :point_down: , after these functions.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1018,103 +1052,93 @@ def x11_status_checker(self, session, session_fileno):
...
+                # FIXME: Huh, shouldn't it be the reverse here?
                 if str(e) in "SSH session not active":
                     self._ssh = None
                     self.restore_connection(None)
                 timeout = timeout + 60
                 retries = retries - 1
-        if retries <= 0:
-            return False , False, False
+
+        return False, False, False
```
kinow, line 1116: This will already be true, no need to check the `retries <= 0`

autosubmit/platforms/paramiko_platform.py
```diff
@@ -435,24 +473,22 @@ def send_file(self, filename, check=True):
             self.delete_file(filename)
         try:
             local_path = os.path.join(os.path.join(self.tmp_path, filename))
-            remote_path = os.path.join(
-                self.get_files_path(), os.path.basename(filename))
+            remote_path = os.path.join(self.get_files_path(), os.path.basename(filename))
```
kinow, line 476: No difference, just placed everything into one line as it fit the under ~120 limit

autosubmit/platforms/paramiko_platform.py
```diff
@@ -322,16 +374,22 @@ def connect(self, as_conf: 'AutosubmitConfig', reconnect: bool = False, log_reco
...
+                                              banner_timeout=60,
+                                              disabled_algorithms={'pubkeys': ['rsa-sha2-256', 'rsa-sha2-512']})
                     else:
                         try:
                             self._ssh.connect(self._host_config['hostname'], port, username=self.user,
-                                              key_filename=self._host_config_id, timeout=60 , banner_timeout=60)
+                                              key_filename=self._host_config_id, timeout=60, banner_timeout=60)
                         except Exception as e:
+                            Log.warning('SSH connect failed, will try again disabling RSA algorithms'
+                                        f'sha-256 and sha-512, error: {str(e)}')
```
kinow, line 389: Got this error in a test, and noticed it tried twice without telling the user. I guess it doesn't hurt to notify them in WARNING (or we can move to DEBUG if needed)

autosubmit/platforms/paramiko_platform.py
```diff
@@ -245,45 +317,18 @@ def interactive_auth_handler(self, title, instructions, prompt_list):
                     answers.append(twofactor_nonpush)
         return tuple(answers)
 
-    def map_user_config_file(self, as_conf) -> None:
```
kinow, line 248: Had to mock the SSH Client, and the SSH config, and the code was linked to this function, so I moved it up to a function (not using anything really from the object) and will check if we need some more tests.

kinow:
Rebased, tomorrow I will start squashing things and will continue working on that -- today was mainly meetings and reviewing some code with @VindeeR .

kinow:
Will fix the conflicts and rebase after 4.1.15 later -- busy working on deliverables today...

kinow:
Rebased earlier this morning, then adjusted the tests to re-use the same fixture. Going through the changes, removing cosmetics ones and leaving only the main changes.

test/unit/test_paramiko_platform.py
```diff
@@ -73,14 +74,23 @@ def ps_platform(tmpdir):
...
     assert platform.name == 'local'
     assert platform.expid == 'a000'
     assert platform.config["LOCAL_ROOT_DIR"] == platform.config["LOCAL_ROOT_DIR"]
     assert platform.header is None
     assert platform.wrapper is None
+    assert platform.header is None
     assert len(platform.job_status) == 4
+    # These calls are not implemented, but should not raise any error
+    platform.get_submit_script()
+    platform.generate_submit_script
```
kinow, line 93: Error here, fixing that...

kinow:
Rebased but tests still broken, pending to test if we can recover from missing `sacct` or other Slurm executables missing. And maybe more tests, and some docs about retries.

kinow:
Note: handle `socket.error`? Which is an alias for `OSError` now (any risk in handling `OSError`.... maybe it will include running out of memory?)

https://github.com/BSC-ES/autosubmit/issues/2591

kinow:
Thank you, @dbeltrankyl, will do ! I'm just writing tests and then will finish looking if we can retry if other errors happen or if that'll have to be left for a future PR.

docs/source/userguide/run/index.rst
```diff
@@ -446,3 +446,38 @@ To stop immediately experiment <EXPID>:
...
+executable does not exist on the remote platform, Autosubmit will log the error,
+and mark the job as ``FAILED``.
+
+However, if you have a networking issue between Autosubmit and your remote
+platform, then Autosubmit will log in ``INFO`` and ``WARNING`` and **will
+retry executing the command up to hard-coded ``3`` retries**.
+
+.. note::
+  We already have an issue created to make these retry settings configurable
+  by users and site admins.
```
kinow, line 483: @mcastril , @dbeltrankyl , I spent some time today reviewing how Autosubmit retries work for SSH commands (what's being fixed here), and I wrote it down here.

It's not covering what happens when you have wrappers, nor any other part of the code that may have retries. Only SSH connection and command execution.

docs/source/ext/autosubmitfigure.py
```diff
@@ -186,10 +186,11 @@ def _move_latest_image(experiment_plot_path: Path, target_path: Path, figure_nam
     :param target_path: Path where we will move the latest image produced with the ``runcmd`` directive to.
     :param figure_name: The name of the image used in our documentation.
     """
-    latest_image = max([png for png in experiment_plot_path.glob('*.png')], key=lambda f: f.stat().st_ctime)
-    if not latest_image:
+    images = [png for png in experiment_plot_path.glob('*.png')]
+    if not images:
```
kinow, line 190: Had this error after adding a new section :shrug: @VindeeR

kinow:
Changelog and docs done today, finished some code review to understand if we could do a bit more, and finished the code changes. Now it's really pending just the code coverage. So more tests over next days.

test/integration/conftest.py
```diff
@@ -126,6 +127,7 @@ def paramiko_platform() -> Iterator[ParamikoPlatform]:
     local_root_dir.cleanup()
 
 
+<<<<<<< HEAD
```
kinow, line 130: Grrrr

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1045,106 +1080,111 @@ def x11_status_checker(self, session, session_fileno):
...
                         if timeout_command == 0:
                             timeout_command = "infinity"
                         command = f'{command} ; sleep {timeout_command} 2>/dev/null'
-                    # command = f'export display {command}'
+                    # TODO: Why do we log the command here, and only in X11 mode?
                     Log.info(command)
-                    try:
-                        chan.exec_command(command)
-                    except BaseException as e:
-                        raise AutosubmitCritical(f"Failed to execute command: {e}")
```
kinow, line 1097: A chain might fail with `SSHException`, in this case we want it to be captured and retried up to N times, and only then raise the `AutosubmitCritical`.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1045,106 +1080,111 @@ def x11_status_checker(self, session, session_fileno):
...
                     except Exception as e:
-                        Log.warning(f"X11 display not found: {e}")
+                        Log.warning(f"X11 display not found: {str(e)}")
                         self.local_x11_display = None
-                    chan = self.transport.open_session()
                     chan.request_x11(single_connection=False, handler=self.x11_handler)
-                else:
-                    chan = self.transport.open_session()
-                if x11:
+
```
kinow, line 1117: Above, there's some simplification by using a `range` for retries, and separating/simplifying the logic for x11 (which doesn't appear to be used -- the code looks a bit broken :shrug: )

kinow:
Rebased, now will look at the coverage after lunch.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -70,19 +71,69 @@ def _create_ssh_client() -> paramiko.SSHClient:
...
-    def __init__(self, expid, name, config, auth_password = None):
-        """
+        if as_env_ssh_config_path:
+            return Path(as_env_ssh_config_path).expanduser()
 
-        :param config:
-        :param expid:
-        :param name:
-        """
+        return Path(f'~/.ssh/config_{as_env_current_user}').expanduser()
```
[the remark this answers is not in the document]
kinow: And this replaced the old code to load user SSH config, which is used by the testing suite. This will be validated as correct once it's used by the testing suite (even though there are tests here too).

autosubmit/platforms/paramiko_platform.py
```diff
@@ -179,46 +225,55 @@ def test_connection(self, as_conf: Optional['AutosubmitConfig']) -> Optional[str
...
+                                 f"will test another host: {str(e)}", 6002)
+
+        if not self.connected:
+            trace = (f'Can not create ssh or sftp connection to {self.host}: Connection could not be established'
+                     f' to platform {self.name}\n Please, check your expid on the PLATFORMS definition in YAML to'
+                     f' see if there are mistakes in the configuration\n Also Ensure that the login node listed'
+                     ' on HOST parameter is available(try to connect via ssh on a terminal)\n Also you can put'
+                     ' more than one host using a comma as separator')
+            error_message = 'Experiment cannot continue due to unexpected behaviour, Autosubmit will stop.'
+            raise AutosubmitCritical(error_message, 7050, trace)
```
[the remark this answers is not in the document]
dbeltrankyl: Hello @kinow 

We should be careful with this one, no? 

> t raises that when the error contains "Authentication failed."

I would say that, if it is the FIRST time it is trying to connect ( ex. during the prepare_run function) it should raise an AutosubmitCritical. 

And if it was at least, AUTH once, we can raise an AutosubmitError so it retries.

This is because in that way we ensure that the error is not coming from a missconfigured user/project... 

I'm wondering about what Error message paramiko raises if there is an 2FA error?

autosubmit/platforms/paramiko_platform.py
```diff
@@ -179,46 +225,55 @@ def test_connection(self, as_conf: Optional['AutosubmitConfig']) -> Optional[str
...
+                    #       about the subsequent ones? https://github.com/BSC-ES/autosubmit/issues/2595
+                    Log.printlog(f"Connection Failed to {self.host.split(',')[0]}, "
+                                 f"will test another host: {str(e)}", 6002)
+
+        if not self.connected:
+            trace = (f'Can not create ssh or sftp connection to {self.host}: Connection could not be established'
+                     f' to platform {self.name}\n Please, check your expid on the PLATFORMS definition in YAML to'
+                     f' see if there are mistakes in the configuration\n Also Ensure that the login node listed'
+                     ' on HOST parameter is available(try to connect via ssh on a terminal)\n Also you can put'
+                     ' more than one host using a comma as separator')
```
dbeltrankyl: I don't recall if we allow yaml list for this option? 

```
HOST:
  - login1.bsc.es
  - login2.bsc.es
```
  
  ...

autosubmit/platforms/paramiko_platform.py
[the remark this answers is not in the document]
dbeltrankyl: (bump this comment)

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1069,81 +1099,82 @@ def x11_status_checker(self, session, session_fileno):
...
         :return: the stdin, stdout, and stderr of the executing command
-        :raises SSHException: if the server fails to execute the command
         """
-        while retries > 0:
+        for retry in range(0, retries):
+            Log.debug(f'Executing command {command}, retry #{retry + 1} out of {retries}')
             try:
+                chan: Channel = self.transport.open_session()
+
+                # TODO: x11 appears to never be called anywhere in the code base?
```
dbeltrankyl: Not sure what do you mean with this?

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1069,81 +1099,82 @@ def x11_status_checker(self, session, session_fileno):
...
+                Log.warning(f'A networking error occurred while executing command [{command}]: {str(e)}')
+                if not self.connected or not self.transport or not self.transport.active:
                     self.restore_connection(None)
-                timeout = timeout + 60
-                retries = retries - 1
-        if retries <= 0:
-            return False, False, False
+                else:
+                    Log.info('The SSH transport is still active, will not try to reconnect')
+                # TODO: We need to understand why we are increasing in increments of 60 seconds, then document it.
```
dbeltrankyl: Is to not spam reconnects when the platform is down. There has to be a timeout so we don't risk to have a ip ban

The AutosubmitError is a loop that can last some days ( a number that I think it is modifiable?). When this is triggered, we're already sure that is not a miss configuration issue ( bad parameters , and the permission checker function raises an AutosubmitCritical)

dbeltrankyl (approved):
Some comments but good changes overall. Thanks!

dbeltrankyl:
Hello ! 

@VindeeR , @kinow 

I reviewed this some time ago, so my idea is:

- to rebase this once the rest of the 4.1.16 is merged ( maybe today? ) 
- Add a new module for DestinE based on this branch
- Tell them that the 4.1.16 is ready to be tested
 
 I think the code should work already, and what is missing are minor things, and the tests which can be done when @kinow is back

dbeltrankyl:
Issues that likely won't be included in 4.1.16 then:

- PBS (we can do a 4.1.17 release with only that, or a dev module for Kai and Leo once it is stable )
- And https://github.com/BSC-ES/autosubmit/pull/2474

dbeltrankyl:
I've rebased this one 

def test_run_command_raises_autosubmit(autosubmit_exp: AutosubmitExperimentFixture, autosubmit: Autosubmit, mocker, command: str):

These tests are failing now in this branch in the recovery and run, because they no longer raise an Exception related to a platform change

autosubmit/platforms/paramiko_platform.py
kinow: Ha, I couldn't get it to be called no matter how hard I tried in tests. Then, after writing the Docker test I realized it was being called really only if you had an actual X server. Mocking it... well, it could work, perhaps... but you would have to spend days mocking things to get everything working 😆 (that's the direction I was going, when I left this comment and gave up).

It's now outdated as we added a test in another PR. Let me remove this comment. Thanks Dani!

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1069,81 +1099,82 @@ def x11_status_checker(self, session, session_fileno):
...
-                    chan = self.transport.open_session()
-                if x11:
+                    chan.request_x11(single_connection=False, handler=self.x11_handler)
+
                     if "timeout" in command:
                         timeout_command = command.split("timeout ")[1].split(" ")[0]
                         if timeout_command == 0:
                             timeout_command = "infinity"
                         command = f'{command} ; sleep {timeout_command} 2>/dev/null'
+                    # TODO: Why do we log the command here, and only in X11 mode?
```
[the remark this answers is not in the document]
kinow: Could be, I've removed this `Log.info(command)`. Thanks Dani!

autosubmit/platforms/paramiko_platform.py
kinow: @dbeltrankyl this is where I tried to use a list for hosts. But maybe I didn't configure it correctly.

docs/source/userguide/wrappers/index.rst
```diff
@@ -187,7 +187,8 @@ The extend_wallclock parameter allow the users to provide extra headroom for the
 Retrials
```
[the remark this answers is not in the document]
kinow: Ah, I left this one intentionally, @dbeltrankyl .

It's because this page has a series of parameters, like Queue, Extend_wallclock, and Retrials.

So even though we want to use retries now, the parameter name is still retrials.

image

autosubmit/platforms/paramiko_platform.py
kinow: Got it. I think this TODO will remain and next time one of us work on that code we can see if we figure out how to fix it, document it, or if we will need a separate issue for this later. (Might be just to keep the x11 open, as you said). Thanks

autosubmit/platforms/paramiko_platform.py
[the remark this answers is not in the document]
kinow: > We should be careful with this one, no?

:+1: 

>>    t raises that when the error contains "Authentication failed."
>
>I would say that, if it is the FIRST time it is trying to connect ( ex. during the prepare_run function) it should raise an AutosubmitCritical.
> And if it was at least, AUTH once, we can raise an AutosubmitError so it retries.
>
>This is because in that way we ensure that the error is not coming from a missconfigured user/project...

Hmmm, I see. I think while that could work, it would be harder to implement without bugs, and harder to test and verify when there are errors (i.e. harder to maintain).

Maybe we could never raise AutosubmitCritical, and retry the 2 times?

My understanding is that we are verifying if the error contains "Authentication failed", which may indicate a failure to authenticate, or anything else. And that we are tying our code to either Paramiko or OpenSSH code, not to the SSH protocol.

If we look at the SSH protocol, https://datatracker.ietf.org/doc/html/rfc4253, nowhere it says that an implementation must return the text "Authentication failed". It only says "If no algorithm satisfying all these conditions can be found, the connection fails, and both sides MUST disconnect."

image

We are, however, checking if we get an exception with the error message that either OpenSSH server returns, https://github.com/search?q=repo%3Aopenssh%2Fopenssh-portable%20authentication%20failed&type=code:

image

or Paramiko returns (I think we are reading the text from Paramiko, but I don't know if Paramiko never returns the server message).

https://github.com/paramiko/paramiko/blob/8697158113a3eab25ed1d1d541ebe2cad10169a7/paramiko/auth_handler.py#L246

A site could choose to use [DropBear](https://github.com/mkj/dropbear), [LSH](https://git.lysator.liu.se/lsh/lsh), [Apache MINA](https://mina.apache.org/), [Teleport](https://goteleport.com/), etc.... or OpenSSH or Paramiko could change their text message.

So we could either document that the authentication will stop the simulation if we have any error that returns "Authentication failure" in the message, and that we assume that this message is returned by a server like OpenSSH, or we can simply say that we will try 2 times, and stop the simulation if that doesn´t happen.

We could modify the logic to catch Paramiko's `AuthenticationException`... but I don´t know if that covers all possible authentication failures. But we could say to users that if we get an authentication error from Paramiko, we will not retry on the first connection, and only on an actual reconnect.

I think I am more inclined toward the simpler solution of retrying blindly, and when there are no more retries, raise AutosubmitCritical. WDYT?

> I'm wondering about what Error message paramiko raises if there is an 2FA error?

That's easy to test now with the Docker tests added in https://github.com/BSC-ES/autosubmit/pull/2604

With this diff:

```diff
diff --git a/test/integration/test_paramiko_platform.py b/test/integration/test_paramiko_platform.py
index 566ed4c7..f101269d 100644
--- a/test/integration/test_paramiko_platform.py
+++ b/test/integration/test_paramiko_platform.py
@@ -234,12 +234,12 @@ def test_send_file_errors(exp_platform_server: ExperimentPlatformServer):
 @pytest.mark.parametrize(
     'cmd,error,x11_enabled,mfa_enabled',
     [
-        ('whoami', None, True, False),
-        ('parangaricutirimicuaro', AutosubmitError, True, False),
-        ('whoami', None, False, False),
-        ('parangaricutirimicuaro', AutosubmitError, False, False),
-        ('whoami', None, True, True),
-        ('parangaricutirimicuaro', AutosubmitError, True, True),
+        # ('whoami', None, True, False),
+        # ('parangaricutirimicuaro', AutosubmitError, True, False),
+        # ('whoami', None, False, False),
+        # ('parangaricutirimicuaro', AutosubmitError, False, False),
+        # ('whoami', None, True, True),
+        # ('parangaricutirimicuaro', AutosubmitError, True, True),
         ('whoami', None, False, True),
         ('parangaricutirimicuaro', AutosubmitError, False, True)
     ]
@@ -260,7 +260,7 @@ def test_send_command(cmd: str, error: Optional[Exception], x11_enabled: bool, m
         exp_platform_server.platform.two_factor_method = 'token'
         exp_platform_server.platform.pw = 'password'
         # 55192054 comes from the Docker setup for 2FA, see docker/ssh/linuxserverio-ssh-with-2fa-x11/README.md
-        mocker.patch('autosubmit.platforms.paramiko_platform.input', return_value='55192054')
+        mocker.patch('autosubmit.platforms.paramiko_platform.input', return_value='123')
 
     exp_platform_server.platform.connect(None, reconnect=False, log_recovery_process=False)
```

We are passing the wrong token to Paramiko. The exception we get:

```bash
Job list created successfully
Using /home/bdepaula/.ssh/config as SSH configuration file
 [CRITICAL] 2FA authentication failed: Authentication failed.[eCode=7000]

test/integration/test_paramiko_platform.py:233 (test_send_command[parangaricutirimicuaro-AutosubmitError-False-True])
cmd = 'parangaricutirimicuaro'
error = <class 'autosubmit.log.log.AutosubmitError'>, x11_enabled = False
mfa_enabled = True
request = <FixtureRequest for <Function test_send_command[parangaricutirimicuaro-AutosubmitError-False-True]>>
mocker = <pytest_mock.plugin.MockerFixture object at 0x7cd78c485190>

    @pytest.mark.parametrize(
        'cmd,error,x11_enabled,mfa_enabled',
        [
            # ('whoami', None, True, False),
            # ('parangaricutirimicuaro', AutosubmitError, True, False),
            # ('whoami', None, False, False),
            # ('parangaricutirimicuaro', AutosubmitError, False, False),
            # ('whoami', None, True, True),
            # ('parangaricutirimicuaro', AutosubmitError, True, True),
            ('whoami', None, False, True),
            ('parangaricutirimicuaro', AutosubmitError, False, True)
        ]
    )
    @pytest.mark.docker
    def test_send_command(cmd: str, error: Optional[Exception], x11_enabled: bool, mfa_enabled: bool,
                          request: pytest.FixtureRequest, mocker):
        """This test opens an SSH connection (via sftp) and sends a command."""
        if x11_enabled:
            request.applymarker('x11')
        if mfa_enabled:
            request.applymarker('mfa')
    
        exp_platform_server: ExperimentPlatformServer = request.getfixturevalue('exp_platform_server')
    
        if mfa_enabled:
            exp_platform_server.platform.two_factor_auth = mfa_enabled
            exp_platform_server.platform.two_factor_method = 'token'
            exp_platform_server.platform.pw = 'password'
            # 55192054 comes from the Docker setup for 2FA, see docker/ssh/linuxserverio-ssh-with-2fa-x11/README.md
            mocker.patch('autosubmit.platforms.paramiko_platform.input', return_value='123')
    
>       exp_platform_server.platform.connect(None, reconnect=False, log_recovery_process=False)

integration/test_paramiko_platform.py:265: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../autosubmit/platforms/paramiko_platform.py:439: in connect
    self.transport.auth_interactive(self.user, self.interactive_auth_handler)
/home/bdepaula/micromamba/envs/autosubmit/lib/python3.9/site-packages/paramiko/transport.py:1756: in auth_interactive
    return self.auth_handler.wait_for_response(my_event)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <paramiko.auth_handler.AuthHandler object at 0x7cd78c535580>
event = <threading.Event object at 0x7cd7b20eaeb0>

    def wait_for_response(self, event):
        max_ts = None
        if self.transport.auth_timeout is not None:
            max_ts = time.time() + self.transport.auth_timeout
        while True:
            event.wait(0.1)
            if not self.transport.is_active():
                e = self.transport.get_exception()
                if (e is None) or issubclass(e.__class__, EOFError):
                    e = AuthenticationException(
                        "Authentication failed: transport shut down or saw EOF"
                    )
                raise e
            if event.is_set():
                break
            if max_ts is not None and max_ts <= time.time():
                raise AuthenticationException("Authentication timeout.")
    
        if not self.is_authenticated():
            e = self.transport.get_exception()
            if e is None:
                e = AuthenticationException("Authentication failed.")
            # this is horrible.  Python Exception isn't yet descended from
            # object, so type(e) won't work. :(
            # TODO 4.0: lol. just lmao.
            if issubclass(e.__class__, PartialAuthentication):
                return e.allowed_types
>           raise e
E           paramiko.ssh_exception.AuthenticationException: Authentication failed.

/home/bdepaula/micromamba/envs/autosubmit/lib/python3.9/site-packages/paramiko/auth_handler.py:263: AuthenticationException
```

So we get a similar "Authentication failed." from Paramiko.

autosubmit/platforms/paramiko_platform.py
dbeltrankyl: Ok, so we can open an issue for It's not that necessary right now. 

Because HOST: mn1, mn2, mn3 works, no? 

or 

HOST: mn1 mn2 mn3

kinow:
I think this means this can be reviewed, @dbeltrankyl , while I continue searching what's causing this intermittent failure in #2745

autosubmit/platforms/paramiko_platform.py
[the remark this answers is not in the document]
dbeltrankyl: > My understanding is that we are verifying if the error contains "Authentication failed

Yes, but you're right, the message could change at any time

> I think I am more inclined toward the simpler solution of retrying blindly, and when there are no more retries, raise AutosubmitCritical. WDYT?

What we want to achieve is a way of differentiating when we're connecting for the first time, to see that our Auth is right ( user, project, passwordless)

The problem I see with this approach is that if it occurs during the "recovery loop" due to a non-related Auth issue, Autosubmit will end, and we don't want that. (For example, it happens on the weekend; the platform is down for hours, but eventually it comes back...) Autosubmit should be able to resume the run.


However, looking at the code... I believe it is safe to raise an **AutosubmitError** instead of **AutosubmitCritical**, as the first connection is made outside the try: except: that triggers the Autosubmit recovery mode. 


The first connection is made in the prepare_run function, which is outside the main loop, so if an AutosubmitError is raised at that point it will act as an AutosubmitCritical

kinow:
No problem at all, Dani. Here's the patch applied.

```diff
diff --git a/autosubmit/platforms/paramiko_platform.py b/autosubmit/platforms/paramiko_platform.py
index 23e9e0cf..d03e50ae 100644
--- a/autosubmit/platforms/paramiko_platform.py
+++ b/autosubmit/platforms/paramiko_platform.py
@@ -255,6 +255,7 @@ class ParamikoPlatform(Platform):
 
         :param as_conf: Autosubmit configuration.
         :param log_recovery_process: Indicates that the call is made from the log retrieval process.
+        :raises AutosubmitError: If the connection was not established even with multiple connection retries.
         """
         Log.info('Restoring SSH connection...')
         with suppress(Exception):
@@ -285,7 +286,7 @@ class ParamikoPlatform(Platform):
                      ' on HOST parameter is available(try to connect via ssh on a terminal)\n Also you can put'
                      ' more than one host using a comma as separator')
             error_message = 'Experiment cannot continue due to unexpected behaviour, Autosubmit will stop.'
-            raise AutosubmitCritical(error_message, 7050, trace)
+            raise AutosubmitError(error_message, 7050, trace)
 
     def agent_auth(self, port: int) -> bool:
         """Attempt to authenticate to the given SSH server using the most common authentication methods available.
diff --git a/test/integration/conftest.py b/test/integration/conftest.py
index b7a85f25..e14c85ef 100644
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -192,7 +192,7 @@ def autosubmit_exp(
         conf_dir = exp_path / "conf"
         global_logs = Path(BasicConfig.GLOBAL_LOG_DIR)
         global_logs.mkdir(parents=True, exist_ok=True)
-        #TODO: Autosubmit.install issue https://github.com/BSC-ES/autosubmit/issues/1328
+        # TODO: Autosubmit.install issue https://github.com/BSC-ES/autosubmit/issues/1328
         Path(BasicConfig.STRUCTURES_DIR).mkdir(parents=True, exist_ok=True)
         exp_tmp_dir = exp_path / BasicConfig.LOCAL_TMP_DIR
         aslogs_dir = exp_tmp_dir / BasicConfig.LOCAL_ASLOG_DIR
diff --git a/test/integration/test_paramiko_platform.py b/test/integration/test_paramiko_platform.py
index 566ed4c7..5ed88b3f 100644
--- a/test/integration/test_paramiko_platform.py
+++ b/test/integration/test_paramiko_platform.py
@@ -24,7 +24,6 @@ from getpass import getuser
 from pathlib import Path
 from typing import cast, Generator, Optional, Protocol, Union, TYPE_CHECKING
 
-
 import paramiko
 import pytest
 from paramiko import ChannelFile  # type: ignore[import]
@@ -35,6 +34,7 @@ from autosubmit.log.log import AutosubmitCritical
 from autosubmit.log.log import AutosubmitError
 from autosubmit.platforms.headers.slurm_header import SlurmHeader
 from autosubmit.platforms.paramiko_submitter import ParamikoSubmitter
+from autosubmit.platforms.slurmplatform import SlurmPlatform
 
 if TYPE_CHECKING:
     # noinspection PyProtectedMember
@@ -820,3 +820,17 @@ def test_test_connection(exp_platform_server: 'ExperimentPlatformServer', ssh_se
     # TODO: This function is odd, if it reconnects, it will return ``"OK"``, but when it's all good
     #       then it will return ``None``.
     assert None is exp_platform_server.platform.test_connection(None)
+
+
+@pytest.mark.docker
+def test_failed_connection_raises_as_error(autosubmit_exp, tmp_path):
+    """Test that failing to restore a connection, even with retries, results in ``AutosubmitError``."""
+    exp = autosubmit_exp(_EXPID, {})
+    platform_config = {
+        "LOCAL_ROOT_DIR": exp.as_conf.basic_config.LOCAL_ROOT_DIR,
+        "LOCAL_TMP_DIR": str(tmp_path),
+        "LOCAL_ASLOG_DIR": str(tmp_path)
+    }
+    slurm_platform = SlurmPlatform(exp.expid, 'slurm_platform', platform_config, None)
+    with pytest.raises(AutosubmitError):
+        slurm_platform.restore_connection(None, log_recovery_process=False)
```

What error code do you think we should use, @dbeltrankyl ? The same?

dbeltrankyl:
Hm, maybe 6003?  Failed connection, wrong configuration.

7000+ are meant to be for critical errors

kinow:
6003 it is! Changed, squash/pushed :+1:

PR #2449, opened 2025-07-04
Getting rid of unused code, and doing some cleaning

### LINKED ISSUE #2136: Function process_times has an impossible "else" in job_utils.py
As I was creating a coverage test for the job_utils.py I found the following piece of code

                        
                   filtered[0].transit = 0
                        for i in range(len(filtered) - 1, 0, -1):
                            filtered[i].transit = max(filtered[i].queue -
                                                      filtered[i - 1].total, 0)

                        # Positive or zero transit time
                        positive = len(
                            [job for job in filtered if job.transit >= 0])

                        if positive > 1:
                            for i in range(0, len(filtered)):
                                if filtered[i].transit >= 0:
                                    temp_index = i
                                    if i > 0:
                                        # Only consider after the first job
                                        filtered[i].queue = max(filtered[i].queue -
                                                                filtered[i - 1].total, 0)
                                        fixes_applied[filtered[i].name] = filtered[i - 1].total
                                else:
                                    filtered[i].queue = max(filtered[i].queue -
                                                            filtered[temp_index].total, 0)
                                    fixes_applied[filtered[i].name] = filtered[temp_index].total
                                # it is starting of level

as per how the code works the filtered[i].transit will never be smaller than 0 which makes impossible for the execution of the else statement.

Should this piece of code be deleted or modified?
job/job_util.py::line 417

### LINKED ISSUE #2441: [Code] var save is always false
As I was creating the test to increase the coverage of the `Slurm Platform` I found the following code

Image

I believe this can be either removed as `save` would always be `False` never allowing the function `save()` to be executed or maybe fixed so it can change values accordingly if this was something that passed unnoticed.

[Code to be checked](https://github.com/BSC-ES/autosubmit/blob/0a5a35380e0d913d634aaa4ba7e2c9ed5dc6d6f7/autosubmit/autosubmit.py#L2283C25-L2287C124)

VindeeR:
Closes #2441
Closes #2136 

As I was creating tests and investigating the Slurm platform I run into some unused or bad code.

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2282,10 +2282,6 @@ class Autosubmit:
                             raise AutosubmitError("Config files seems to not be accessible", 6040, str(e))
                         total_jobs, safetysleeptime, default_retrials, check_wrapper_jobs_sleeptime = Autosubmit.get_iteration_info(as_conf,job_list)
 
-                        save = False
-                        # End Check Current jobs
-                        if save:  # previous iteration
-                            job_list.backup_save()
                         # This function name is totally misleading, yes it check the status of the wrappers, but also orders jobs the jobs that  are not wrapped by platform.
                         jobs_to_check, job_changes_tracker = Autosubmit.check_wrappers(as_conf, job_list, platforms_to_test, expid)
                         # Jobs to check are grouped by platform.
@@ -2336,7 +2332,6 @@ class Autosubmit:
                             except Exception as e:
                                 Log.warning(
                                     "Couldn't recover the Historical database, AS will continue without it, GUI may be affected")
-                        job_changes_tracker = {}
                         if Autosubmit.exit:
                             Autosubmit.check_logs_status(job_list, as_conf, new_run=False)
                             job_list.save()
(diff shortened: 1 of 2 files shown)


VindeeR:
@isimo00 if you have some time could you please review the changes here, I think it should be simple enough

autosubmit/autosubmit.py
```diff
@@ -5555,15 +5550,15 @@ def _get_submitter(as_conf):
...
-            return ParamikoSubmitter()
-        else:
+        # try:
+        as_conf.get_communications_library()
+        # except Exception as e:
+        #     communications_library = 'paramiko'
+        # if communications_library == 'paramiko':
+        #     return ParamikoSubmitter()
+        # else:
             # only paramiko is available right now.
```
isimo00, line 5560: This section now doesn't support exceptions no?

VindeeR: This function will try to get the value `Paramiko` of the API, whatever it exists or not it'll will return `Paramiko`, and inside the function there's a treatment for exception raising more specific warnings.

So I thought it wouldn't make sense have this exception since it'll always return `Paramiko` whether there's or there's not the value in the API, and that exception would override the more specific one.

VindeeR:
I forgot to remove the assign to the variable `communications_library` I'll adjust it immediately

isimo00 (approved):

PR #2450, opened 2025-07-07
Add file attachment for FAILED status update

### LINKED ISSUE #2050: Add configuration options for mail notification
Related to #1997 

We need configuration options for users to:

- Disable attachments
- Restrict the max size of attachments

This must be implemented in a way so that users (think of sysadmins) are able to control that themselves, without having to ask AS developers to modify code :+1:

### LINKED ISSUE #2055: Email notification on finished chunk or/and experiment
Hi @isimo00 ,

How can I get an email notification when job X is finished in chunk Y?

Alternatively, is there a way to get the notification when the experiment finishes?

To put it in context, my experiment has 20 chunks. I want to get notified when the experiment finishes. Currently, I can set the job to notify_on COMPLETE, but that generates 20 emails, when I only want one for the last job to finish. Also, for future cases, setting the notification for chunk Y would be nice.

Thank you

isimo00:
Add mail notifications with logfile attachments for job status updates to `FAILED`

Closes #2055 
cc @cguzman95


autosubmit/notifications/mail_notifier.py
```diff
@@ -212,6 +213,11 @@ def notify_status_change(
             ('Autosubmit', self.config.MAIL_FROM))
         message['Subject'] = f'[Autosubmit] The job {job_name} status has changed to {status}'
         message['Date'] = email.utils.formatdate(localtime=True)
+        
+        if status == "FAILED":
+            message.attach(MIMEText(message_text))
+            self._collect_logfiles(message, exp_id)
```
kinow, line 219: I think there are no tests for this yet. Could you add one, please, @isimo00 ? The rest of the code looks OK. Would be good to document it somewhere, and add a changelog entry too under 4.1.16 :+1: Thanks!

isimo00: I have to say, however, that tests related to the mail notifications should be replaced at some point (hopefully before I leave) by new ones using the SMTP container

isimo00: Sure thing!🤞🏻

isimo00:
Actually I started integrating the fake mail server container for testing in #2135 so I'd rather finish it there - I requested a review form you and I guess I forgot to check out your comment - @kinow. Back to it now

autosubmit/notifications/mail_notifier.py
isimo00: We're really coordinated today... 😅

kinow:
I believe this was a request from a user, to include the logs in the notification emails. I think Irene finished it, but it never got merged as we didn't have time to review it. Maybe we can include this one in 4.1.16? WDYT, @dbeltrankyl ? Assuming the review is positive and there's nothing missing?

VindeeR:
@kinow @dbeltrankyl I'm rebasing this as I believe it was finished but never reviewed, based on the comments, maybe we can add this to master if all tests pass

kinow:
Thanks @VindeeR !

autosubmit/notifications/mail_notifier.py
```diff
@@ -208,13 +210,23 @@ def notify_status_change(
...
             ('Autosubmit', self.config.MAIL_FROM))
         message['Subject'] = f'[Autosubmit] The job {job_name} status has changed to {status}'
         message['Date'] = email.utils.formatdate(localtime=True)
+        if status == "FAILED":
+            message.attach(MIMEText(message_text))
+            self._collect_logfiles(message, exp_id)
 
-        self._send_message(mail_to, self.config.MAIL_FROM, message)
+        for mail in mail_to:  # expects a list
+            message['To'] = email.utils.formataddr((mail, mail))
```
dbeltrankyl: Is it the key.. 'To' or 'TO'? in the other place is written 'TO'

dbeltrankyl (changes requested):
Hello @VindeeR 

Thanks for retaking the work from Irene, 

Just one comment is relevant to this approval: I believe we're sending the wrong log. Can you check that? 

Thanks!

dbeltrankyl (changes requested):
Thanks @VindeeR ! 

There is some stuff to adress

dbeltrankyl (commented):
Some minor stuff commented, thanks for the work!

autosubmit/notifications/mail_notifier.py
```diff
@@ -155,6 +155,24 @@ class MailNotifier:
     def __init__(self, basic_config):
         self.config = basic_config
 
+    def _collect_logfiles(self, message: 'MIMEMultipart', exp_id: str):
+        """Generate a compressed file with all the logs from the LOG_<EXPID> folder"""
+        run_log_files_err = [f for f in self.config.expid_log_dir(exp_id).glob('*.err') if Path(f).is_file()]
+        run_log_files_out = [f for f in self.config.expid_log_dir(exp_id).glob('*.out') if Path(f).is_file()]
+        if run_log_files_err and run_log_files_out:
```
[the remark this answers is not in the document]
dbeltrankyl: I think you should move all this to the log_recovery process, in that way the log will be there for sure and you don't need to do extra communication

autosubmit/notifications/mail_notifier.py
```diff
@@ -46,15 +46,16 @@ def _compress_file(
...
-    :type file_path: Path
+    :type file_path: list[Path]
     :return: The Path object of the compressed file.
     :rtype: str
     :raises AutosubmitError: The file cannot be compressed.
     """
     try:
-        zip_file_name = Path(temporary_directory.name, f'{file_path.name}.zip')
+        zip_file_name = Path(temporary_directory, 'log.zip')
         with zipfile.ZipFile(zip_file_name, 'w', zipfile.ZIP_DEFLATED) as zip_file:
```
dbeltrankyl: Now the log compression is toggleable; by default, it is on. In these cases, when the code reaches this part, the logs will already be compressed. 

I think it is weird to have compressed files inside a .zip file, and, for example, if someone wants to download the log from the phone, which is a possible scenario for operators who aren't on the work pc when the notification appears, it would be counterintuitive. Maybe detect whether the log is compressed beforehand to avoid an extra compression? 

Also, it doesn't hurt to attach the err and out in two different files, but that choice is up to you

dbeltrankyl (commented):
Hello! thanks for the update some comments

autosubmit/notifications/mail_notifier.py
dbeltrankyl: Sorry for the delay 

They're compressed but they maintain the same filename ( not tar, zip is added) if I recall correctly cc: @LuiggiTenorioK

autosubmit/notifications/mail_notifier.py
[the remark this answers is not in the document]
dbeltrankyl: I agree that it is better involved in the mailnotifier, but the log info is not reachable from there as it is called from the main process which doesn't have the log variables updated

autosubmit/notifications/mail_notifier.py
LuiggiTenorioK: In the log recovery, the `.out`/`.err` files are compressed in the remote platform first before transferring them to local. When the file is compressed, the saved filename is changed to the compressed filename, so `.gz` or `.xz` is added at the end of that filename.

VindeeR:
@dbeltrankyl tagging you here just so you can find it later for review

VindeeR:
@dbeltrankyl Just rebased this PR

kinow (changes requested):
This one needs to be rebased and have more tests added, at least for the code handling happy-path and errors (as this is not tested by the testing suite, so we would have users telling us whether it's working or not).

VindeeR:
> This one needs to be rebased and have more tests added, at least for the code handling happy-path and errors (as this is not tested by the testing suite, so we would have users telling us whether it's working or not).

Tests were added, we too have had some other tests related to sending email added at another PR, that makes use of the server, and here we have a test that simulates an email server, and handling it's submission of email failure.

I added a couple other simpler tests, just to check the output errors generated and fields.

Let me know if there's anything else I might be missing.

Thanks for the review @kinow

kinow:
Thanks @VindeeR ! I'm just waiting for the GH Actions build to finish.

kinow:
I've kicked the jobs. An integration test failed..

kinow (commented):
The line not tested in autosubmit.py looks OK. There's one missing in basicconfig.py, but it's exactly the new line for the attachment setting. I think that'd make sense to be tested here, as we are adding here.

Mail notifier missing/partially covered are OK. Configcommon has a partial covered, but not important for this change. The helpers/utils.py remains with many parts not yet tested. Are these being tested in an integration test here, @VindeeR , and not appearing in the coverage, or are these still missing some tests?

VindeeR:
@kinow those lines were not tested before, not new lines, but the lines it is complaining is within an exception, I just removed the suppression on it because it was already inside a try...exception, so it makes no sense to have it in there.

Everything else should be fine, I'll rebase it and should be ready for review

autosubmit/autosubmit.py
```diff
@@ -2358,7 +2359,7 @@ def run_experiment(expid: str, start_time: Optional[str] = None, start_after: Op
                                 else:
                                     mail_notify = False
                                 times = times + 1
-                                Autosubmit.restore_platforms(platforms_to_test, mail_notify=mail_notify,
+                                restore_platforms(platforms_to_test, mail_notify=mail_notify,
```
kinow, line 2295: > @kinow those lines were not tested before, not new lines, but the lines it is complaining is within an exception, I just removed the suppression on it because it was already inside a try...exception, so it makes no sense to have it in there.

Are you talking about this part, @VindeeR ?

Image

I agree it was not tested, but since this PR is removing the function, it'd be good to write a test too. This way we confirm that the imported function works (good to confirm it works both because it was moved, and someone reading this part alone could wonder if there were any modifications in the `restore_platforms` -- in case of an operational bug restoring platforms, for instance).

autosubmit/helpers/utils.py
```diff
@@ -108,82 +108,98 @@ def check_experiment_ownership(
...
+                Log.RESULT,
+            )
         else:
-            if platform.connected:
-                platform.connected = False
+            if platform_to_test.connected:
+                platform_to_test.connected = False
                 Log.printlog(
-                    f"[{platform.name}] Connection successful to host {platform.host}, "
+                    f"[{platform_to_test.name}] Connection successful to host {platform_to_test.host}, "
```
kinow, line 178: We should be able to test this one? We changed the variable name, and also changed in the f-string. Normally it'd be safe, but with Python the only part to be sure this works is with a test or in runtime when users use this code. I think the code was not well-tested before... but now it's on you to write, I'm afraid. I had to write a bunch of tests for code I didn't write, but were part of #2778 🤷 😥

autosubmit/helpers/utils.py
```diff
@@ -108,82 +108,98 @@ def check_experiment_ownership(
...
                     email = as_conf.get_mails_to()
                     if "@" in email[0]:
-                        Notifier.notify_experiment_status(MailNotifier(BasicConfig), expid, email, platform)
-            platform_issues += f"\n[{platform.name}] Connection Unsuccessful to host {platform.host} "
+                        Notifier.notify_experiment_status(
+                            MailNotifier(BasicConfig), expid, email, platform_to_test
+                        )
+            except Exception as e2:
+                Log.debug(f"Unexpected exception sending email notification: {str(e2)}")
+            platform_issues += f"\n[{platform_to_test.name}] Connection Unsuccessful to host {platform_to_test.host} "
```
kinow, line 154: I think this part is also not covered, but it's less risky. If you find it's too hard to write a test for this, I'd be OK with that I think. If an operational workflow fails to send emails, that would be OK. If it tails to restore a platform (reconnect/read config/etc), then someone would have to stop and work with operators to solve it (normally Dani or I, or maybe someone from the workflow/A7 team...).

autosubmit/autosubmit.py
VindeeR: > > @kinow those lines were not tested before, not new lines, but the lines it is complaining is within an exception, I just removed the suppression on it because it was already inside a try...exception, so it makes no sense to have it in there.
> 
> Are you talking about this part, @VindeeR ?
> Image
> 
> I agree it was not tested, but since this PR is removing the function, it'd be good to write a test too. This way we confirm that the imported function works (good to confirm it works both because it was moved, and someone reading this part alone could wonder if there were any modifications in the `restore_platforms` -- in case of an operational bug restoring platforms, for instance).

Sorry I was referring to the `helpers/utils.py` as the rest I thought it was okay, but I will address what you said :+1:

autosubmit/helpers/utils.py
VindeeR: I did a simple test, the issue is that the failure is being mocked I`ll try to add a more robust test, failing to generate the email, I believe there should be a test somewhere related to it

autosubmit/config/basicconfig.py
```diff
@@ -131,6 +132,8 @@ def __read_file_config(file_path):
             BasicConfig.SMTP_SERVER = parser.get('mail', 'smtp_server')
         if parser.has_option('mail', 'mail_from'):
             BasicConfig.MAIL_FROM = parser.get('mail', 'mail_from')
+        if parser.has_option('mail', 'attachment'):
+            BasicConfig.MAIL_FROM = parser.get('mail', 'attachment')
```
kinow: `MAIL_FROM` is set to `mail.attachment`?

test/integration/platforms/test_platform_restore.py
```diff
@@ -0,0 +1,195 @@
...
+    with pytest.raises(AutosubmitCritical) as cm:
+        restore_platforms([exp.platform], mail_notify=True, as_conf=None, expid=None)
+
+    assert expected_cm in str(cm.value.message)
+    assert expected_log in mock_log_printlog.call_args.args[0]
+
+
+def test_restore_platforms_raises(get_next_expid, autosubmit_exp, mocker) -> None:
+    """When a platform yields a problematic message, an AutosubmitCritical is raised."""
+    #needs to be finished
```
kinow: leftover comment? I think this can be removed?

test/integration/platforms/test_platform_restore.py
```diff
@@ -0,0 +1,195 @@
...
+            "Issues while checking the connectivity of platforms.",
+            False,
+        ),
+    ],
+)
+def test_restore_platforms_connection_exception_on_exception(
+    get_next_expid, autosubmit_exp, mocker, exception, expected_log, expected_cm: str, connected: bool,
+) -> None:
+    """When a platform yields a problematic message, an AutosubmitCritical is raised."""
+    exp = autosubmit_exp(get_next_expid(), experiment_data={})
```
kinow: Before `get_next_expid` was necessary for a short time. I think the test was written at the time. Now, it's optional. Fine either way here 👍

test/unit/test_mail.py
```diff
@@ -15,7 +15,10 @@
 # You should have received a copy of the GNU General Public License
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
```
kinow: We can probably avoid future annotations.

autosubmit/autosubmit.py
kinow, line 1: File deleted. Rebasing we need to remove it here too.

autosubmit/config/configcommon.py
```diff
@@ -1325,6 +1325,7 @@ def check_autosubmit_conf(self, no_log=False) -> bool:
                 else:
                     mails = mails.split(' ')
                 self.experiment_data["MAIL"]["TO"] = mails
+                self.experiment_data["MAIL"]["ATTACHMENT"] = parser_data["MAIL"].get("ATTACHMENT", False)
```
kinow: Maybe ATTACHMENT could have a different name? I **think** this is a flag that enables/disables adding attachments for failures. Maybe `FAILURE_INCLUDES_ATTACHMENTS`, or `INCLUDE_ATTACHMENTS`, etc.? (I'm terrible choosing names.)

test/integration/platforms/test_platform_restore.py
VindeeR: if it is not needed anymore it is better to make it a standard not to use

PR #2451, opened 2025-07-07
Fix git check for operational experiments

### LINKED ISSUE #2446: Do not rely on expdef_**** files (e.g. in autosubmit_git for oper experiments)
### Description

`expdef_***.yml` files were always present in AS3 experiments. In AS4 experiments users may use whatever they want, and in the end we parse & combine them, producing the `conf/metadata/experiment_data.yml` file, that shows what AS has in memory.

See this comment for an example added in 4.1.14: https://github.com/BSC-ES/autosubmit/pull/2293/files#r2182051227

* Autosubmit version: 4.1.14, 4.1.15
* Machine/VM/environment name (if applicable): ClimateDT
* Experiment ID (if applicable):o00b

### Reproducible Example

Create an `o---` experiment, using Git, and minimal. Try running it, and you should get an exception similar to:

```bash
2025-07-03 09:49:17,676 Autosubmit is running with 4.1.15
[WARNING] 2025-07-03 09:49:17,781 Yaml file /appl/AS/AUTOSUBMIT_DATA/o00b/proj/git_project/conf/additional_jobs/transfer-False.yml not found
[ERROR] 2025-07-03 09:49:18,276 Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='WARNING', command='run', expid='o00b', notransitive=False, update_version=False, start_time=None, start_after=None, run_only_members=None, profile=False)
[ERROR] 2025-07-03 09:49:18,277 This is the experiment: <o00b> which had an issue with the command: <run> and it is currently using the Autosubmit Version: 4.1.15.
2025-07-03 09:49:18,277 Traceback (most recent call last):
  File "/appl/AS/4.1.15/lib64/python3.9/site-packages/autosubmit/scripts/autosubmit.py", line 105, in main
  File "/appl/AS/4.1.15/lib64/python3.9/site-packages/autosubmit/autosubmit.py", line 724, in run_command
    AutosubmitGit.check_unpushed_changes(expid)
  File "/appl/AS/4.1.15/lib64/python3.9/site-packages/autosubmit/git/autosubmit_git.py", line 290, in check_unpushed_changes
    with open(origin, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/appl/AS/AUTOSUBMIT_DATA/o00b/conf/expdef_o00b.yml'

[CRITICAL] 2025-07-03 09:49:18,277 Unexpected error: [Errno 2] No such file or directory: '/appl/AS/AUTOSUBMIT_DATA/o00b/conf/expdef_o00b.yml'.
2025-07-03 09:49:18,277 More info at https://autosubmit.readthedocs.io/en/master/troubleshooting/error-codes.html
```

### Expected Behaviour

No exception raised, and instead a nice message displayed to user if appropriate.

kinow:
Use cwd for subprocess calls, pass an array to subprocess calls, use as_conf instead of expdef_*, simplify logic by checking only for git changes (it includes submodules).

Closes #2446 

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,3 +141,12 @@ or just the tests that require Slurm:
 ```bash
 $ pytest -m 'slurm'
 ```
+
+## Random ports
+
+Some tests require random ports. This table below can be useful for troubleshooting
+if we ever start running out of ports.
+
+- The integration test `test_paramiko_platform.py` uses the range `2500` to `3000`
+- The integration test `test_mail.py` uses the range `3500` to `4000`
+- The integration test `test_autosubmit_git.py` uses the range `4000` to `4500`
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -50,7 +50,6 @@ from pyparsing import nestedExpr
 from ruamel.yaml import YAML
 
 import autosubmit.helpers.autosubmit_helper as AutosubmitHelper
-import autosubmit.history.utils as HUtils
 import autosubmit.statistics.utils as StatisticsUtils
 from autosubmit.database.db_common import create_db
 from autosubmit.database.db_common import delete_experiment, get_experiment_descrip
@@ -60,7 +59,7 @@ from autosubmit.database.db_structure import get_structure
 from autosubmit.experiment.detail_updater import ExperimentDetails
 from autosubmit.experiment.experiment_common import copy_experiment
 from autosubmit.experiment.experiment_common import new_experiment
-from autosubmit.git.autosubmit_git import AutosubmitGit
+from autosubmit.git.autosubmit_git import AutosubmitGit, check_unpushed_changes, clean_git
 from autosubmit.helpers.processes import process_id
 from autosubmit.helpers.utils import check_jobs_file_exists, get_rc_path
 from autosubmit.helpers.utils import strtobool
@@ -719,9 +718,7 @@ class Autosubmit:
             expid = args.expid
         if args.command != "configure" and args.command != "install":
             Autosubmit._init_logs(args, args.logconsole, args.logfile, expid)
-
         if args.command == 'run':
-            AutosubmitGit.check_unpushed_changes(expid)
             return Autosubmit.run_experiment(args.expid, args.notransitive,args.start_time,args.start_after, args.run_only_members, args.profile)
         elif args.command == 'expid':
             return Autosubmit.expid(args.description,args.HPC,args.copy, args.dummy,args.minimal_configuration,args.git_repo,args.git_branch,args.git_as_conf,args.operational,args.testcase,args.evaluation,args.use_local_minimal) != ''
@@ -2240,6 +2237,16 @@ class Autosubmit:
                     raise
                 except Exception as e:
                     raise AutosubmitCritical("Error in run initialization", 7014, str(e))  # Changing default to 7014
+
+                as_conf_config = as_conf.experiment_data.get('CONFIG', {})
+                git_operational_check_enabled = as_conf_config.get('GIT_OPERATIONAL_CHECK_ENABLED', True)
+
+                if git_operational_check_enabled:
+                    Log.debug('Checking for dirty local Git repository')
+                    check_unpushed_changes(expid, as_conf)
+                else:
+                    Log.warning('Git operational check disabled by user')
+
                 Log.debug("Running main running loop")
                 did_run = False
                 #########################
@@ -2952,9 +2959,8 @@ class Autosubmit:
                 if project_type == "git":
                     Log.info("Registering commit SHA...")
                     autosubmit_config.set_git_project_commit(autosubmit_config)
-                    autosubmit_git = AutosubmitGit(expid)
                     Log.info("Cleaning GIT directory...")
-                    if not autosubmit_git.clean_git(autosubmit_config):
+                    if not clean_git(autosubmit_config):
                         return False
                     Log.result("Git project cleaned!\n")
                 else:
--- a/autosubmit/git/autosubmit_git.py
+++ b/autosubmit/git/autosubmit_git.py
@@ -1,38 +1,33 @@
-#!/usr/bin/env python3
-
-# Copyright 2015-2020 Earth Sciences Department, BSC-CNS
-
+# Copyright 2015-2025 Earth Sciences Department, BSC-CNS
+#
 # This file is part of Autosubmit.
-
+#
 # Autosubmit is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
-
+#
 # Autosubmit is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
-
-from os import path
-
+#
 # You should have received a copy of the GNU General Public License
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
 import locale
 import os
-import psutil
 import re
 import shutil
 import subprocess
 from pathlib import Path
-from ruamel.yaml import YAML
 from shutil import rmtree
 from time import time
-from typing import List, Union
+from typing import List, Optional, Union
 
-# from autosubmit import Autosubmit
-from autosubmit.helpers.processes import process_id
 from autosubmitconfigparser.config.basicconfig import BasicConfig
+from autosubmitconfigparser.config.configcommon import AutosubmitConfig
+
 from log.log import Log, AutosubmitCritical
 
 Log.get_logger("Autosubmit")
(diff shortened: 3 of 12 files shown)


autosubmit/git/autosubmit_git.py
```diff
@@ -45,6 +42,45 @@
...
+        proj_dir = as_conf.get_project_dir()
+
+        for git_cmd, args in zip(git_cmds, arguments):
+            output = subprocess.check_output([*git_cmd, *args], cwd=proj_dir)
+            output = output.decode(locale.getlocale()[1])
+            for line in output.splitlines():
+                tokens = line.strip().split(' ')
+                if tokens[0] in _GIT_STATUS_DIRTY_FLAG:
+                    raise AutosubmitCritical("Push Git local changes to remote repository before running", 7075)
+
```
kinow, line 82: The function was already static, and it had no state-dependency to the class. So it made no difference the function being here in the module, or in an object. The advantage for this one is that we don't need to create an object, avoid risks of state, and it could be faster (I think it's faster to load from a module directly?).

But also simpler to test/call.

test/conftest.py
```diff
@@ -139,7 +139,7 @@ def autosubmit_exp(
     def _create_autosubmit_exp(
             expid: Optional[str] = None,
             experiment_data: Optional[Dict] = None,
-            wrapper = False,
+            wrapper=False,
```
kinow, line 142: Fix linter warning...

autosubmit/git/autosubmit_git.py
```diff
@@ -45,6 +42,45 @@
...
+        ]
+        arguments = [
+            ["status", "--porcelain"],
+            ["foreach", "'git status --porcelain'"]
+        ]
+
+        proj_dir = as_conf.get_project_dir()
+
+        for git_cmd, args in zip(git_cmds, arguments):
+            output = subprocess.check_output([*git_cmd, *args], cwd=proj_dir)
```
kinow, line 76: Deconstructing the arrays here (this `*`), from what I recall, you can pass a string or an array, but not mixing them. So now everything should be just an array of strings. This was one of the bugs we had.

autosubmit/git/autosubmit_git.py
```diff
@@ -45,6 +42,45 @@
...
+            output = output.decode(locale.getlocale()[1])
+            for line in output.splitlines():
+                tokens = line.strip().split(' ')
```
kinow, line 79: Added an `strip()` to the call. Before, we were testing if the line started with the `M`, `?`, etc., flags. However, if you look closely at the terminal output, there is a space before the first letter.

I only noticed this when I removed the `mocker` and created a real small Git repository to test this. This was one of the bugs we had... :bug:

kinow:
Rebased, and I think the code is ready now, including tests and changelog. I think we don't have anything about this new feature in our docs, so I will write one/two paragraphs.

Will also run some tests locally, and if everything works, I'll mark it as ready for review and ask Erick or Irene to review it.

kinow:
I think now I have the code to check for code that was not committed, as well as code that was not pushed to remote. Now I just need to add more tests for both a Git repository, and for Git submodules.

kinow:
Turns out we can remove the code that checks for changes in submodules. I created an integration test with Docker + `git_http_backend` (comes with Git, you need an Apache httpd/nginx in front, found a container with that).

It serves a real Git repository, and I noticed that the coverage never reached the code for submodules. That's because I was first checking for changes not committed or pushed in the project. And any modification in a local submodule reflects in the check for changes in the repository that contains that submodule.

i.e. a single check and we can check for changes in the project and submodules. Coverage should improve now from the previous ~65%

autosubmit/git/autosubmit_git.py
```diff
@@ -275,45 +322,3 @@ def is_git_repo(git_repo: str) -> bool:
...
-                        ["foreach", "'git status --porcelain'"]]
-            for version_control, args in zip(version_controls, arguments):
-                if project == version_control:
-                    output = subprocess.check_output([version_control, args]).decode(locale.getlocale()[1])
-                    if any(status.startswith(code) for code in ["M", "A", "D", "?"] for status in output.splitlines()):
-                        # M: Modified, A: Added, D: Deleted, ?: Untracked
-                        raise AutosubmitCritical("Push local changes to remote repository before running", 7075)
-
-    @staticmethod
-    def check_directory_in_use(expid: str) -> bool:
```
kinow, line 307: @isimo00 you wrote this function, and I think it was quite recently. I checked in my IDE, and it looks like it's never used? Do you recall if this was supposed to be called somewhere, and if we had tests for this?

I noticed, first, because the coverage report warned for low coverage here as I added the return variable. Thanks!

CONTRIBUTING.md
```diff
@@ -141,3 +141,12 @@ or just the tests that require Slurm:
...
 ```
+
+## Random ports
+
+Some tests require random ports. This table below can be useful for troubleshooting
+if we ever start running out of ports.
+
+- The integration test `test_paramiko_platform.py` uses the range `2500` to `3000`
+- The integration test `test_mail.py` uses the range `3500` to `4000`
+- The integration test `test_autosubmit_git.py` uses the range `4000` to `4500`
```
kinow, line 152: @isimo00, @dbeltrankyl , @VindeeR I've added the prots we are using for tests here, so that in future PRs we can reference what's using what, and for choosing the next range if needed :+1:

test/integration/test_utils/git.py
```diff
@@ -0,0 +1,122 @@
...
+        f'git -c protocol.file.allow=always submodule add {url} {name}',
+        'git add .',
+        'git commit -am "Add submodule"'
+    ]
+
+    if push:
+        commands.append(f'git push {remote} {branch}')
+
+    command = ';'.join(commands)
+    check_output(command, cwd=str(path), shell=True)
```
kinow, line 122: These are handy functions for writing tests that manipulate contents of Git repositories. In the future, we can use these to test the remaining `git` package -- which has only 45.64% of coverage right now, but shouldn't be too hard to test the rest.

image

kinow:
Whew, tired but this is now done :sleepy: . I will finish the documentation tomorrow morning, have a final review of the code, and then it should be ready for review.

CONTRIBUTING.md
isimo00: Thanks!

kinow:
Started writing a few more tests (not planning to spend much time on this, but since I'm changing this file...), and then I noticed that our `clean_git` function did the same implemented here, but instead of `git status --porcelain` it used `git diff-index HEAD`.

I replaced the old code by calls to the new code, keeping log messages and exceptions raised. We already had tests for the git clean, so will use those to check if I broke anything :)

autosubmit/git/autosubmit_git.py
isimo00: You're right, it's not used. We ended up chagning how we checked this, and it's with `process_id` now - https://github.com/BSC-ES/autosubmit/blob/8fcf0dfd44643e1e65172966c0f5c680317e51ed/autosubmit/autosubmit.py#L1495 -. It's safe to remove it. Thanks for raising this!

kinow:
Tests & new docs written. PR ready for review.

kinow:
Hi @dbeltrankyl ! I was going to ask Irene or Erick to review this one, but maybe you could review this one when you have time. This is the fix for an issue that prevented the operators from using AS with 4.1.15, as an oper experiment wouldn't run due to a bug with how we checked for git changes for `o` experiments.

test/integration/test_utils/git.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if not path.exists():
+        path.mkdir(parents=True)
+
+    git_init_args = []
+    if bare:
+        git_init_args.append('--bare --shared=true')
+
+    commands = [
+        f'git -c init.defaultBranch=master init . {" ".join(git_init_args)}'
```
dbeltrankyl, line 46: Shouldn't be the default main? ( just to follow the """new""" convention )

kinow: Ah, good catch. Soon we will rename the branch in Autosubmit, and then this test would be the odd one, as everything else would be following this """new""" convention (gave a chuckle :laughing: ). Thanks

test/integration/test_utils/git.py
```diff
@@ -0,0 +1,128 @@
...
+
+            with open(clone_repo / 'README.md', 'w') as f:
+                f.write('This is a test repository of Autosubmit.')
+
+            git_commit_all_in_dir(clone_repo, branch=branch, push=True)
+
+    # git-http-backend (which comes with Git) may have issues with directory
+    # permissions depending on the settings and file system permissions.
+    # Thus, we chmod everything to 0x777 here.
+    system(f'chmod -R 0777 {str(path)}')
```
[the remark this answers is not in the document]
kinow: Hmmm, yeah, should work on macOS, but not on Windows. Just confirmed that Python doesn't have a way to recursively chmod, but it might have at some point: https://discuss.python.org/t/adding-recursive-support-in-os-chmod-os-chown-shutil-chown/27816/2

For now I think AS will build only on Linux and macOS (there are other system calls that I think are Linux/bsd only too). Thanks for the reminder, I'll keep that in mind and try to use Python code more than system calls if possible (and replace this one at some point, maybe I just find a way to run the container with the UID/GID/username from the host machine, and then we wouldn't have to worry about chmod at all...)

kinow:
Added a flag, `CONFIG.GIT_OPERATIONAL_CHECK_ENABLED` to toggle this feature on/off. Will write a test for that, probably tomorrow (going to look at the Sphinx issue with Erick now)

test/integration/test_utils/git.py
kinow: Added a `# TODO` in the code so we know we can use `main` when using Git 2.28+. My old BSC laptop, with Ubuntu 22.04 doesn't have this new version, which is required to modify the default branch when creating a bare local repository (which we use for testing here).

autosubmit/autosubmit.py
```diff
@@ -2240,6 +2237,16 @@ def run_experiment(expid, notransitive=False, start_time=None, start_after=None,
...
                     raise AutosubmitCritical("Error in run initialization", 7014, str(e))  # Changing default to 7014
+
+                as_conf_config = as_conf.experiment_data.get('CONFIG', {})
+                git_operational_check_enabled = as_conf_config.get('GIT_OPERATIONAL_CHECK_ENABLED', False)
+
+                if git_operational_check_enabled:
+                    Log.debug('Checking for dirty local Git repository')
+                    check_unpushed_changes(expid, as_conf)
+                else:
+                    Log.warning('Git operational check disabled by user')
```
kinow, line 2248: Now we use a flag here to control where we check for changes not committed/pushed, or not.

dbeltrankyl: Thanks Bruno. Let me know if you want to me to revise it again

kinow:
Tests appear to be stable again after #2472. Coverage looks OK (hard to go into the missing lines... and those Log.bla functions are definitely being called, no idea why Codecov is reporting them wrongly). Ready for a new review, please, @dbeltrankyl . Ta

autosubmit/git/autosubmit_git.py
```diff
@@ -45,57 +40,114 @@
...
+
+    :param expid: The experiment ID.
+    :param as_conf: Autosubmit configuration object.
+    """
+    project_type = as_conf.get_project_type()
+    if expid[0] == 'o' and project_type == 'git':
+        proj_dir = Path(as_conf.get_project_dir())
+
+        if not _check_uncommitted_code(proj_dir):
+            raise AutosubmitCritical("You must commit and push your code to the remote Git repository in an "
```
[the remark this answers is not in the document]
kinow: Will give it a try, might be helpful to users.

test/integration/test_utils/git.py
```diff
@@ -0,0 +1,137 @@
...
+
+"""Helper functions for working with Git in integration tests."""
+
+from os import system
+from pathlib import Path
+from subprocess import check_output
+from tempfile import TemporaryDirectory
+
+# TODO: Rename branch to main to match the other repositories; note that this
+#       needs to wait until our laptops & CICD env are using Git 2.28+, in order
```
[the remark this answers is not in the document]
kinow: Hmmm, we could add it, but we would still have the `if` and the `master` if < 2.28, and `main` otherwise... as this is a test, I think we can keep `master` here for simplicity and at some point in the future remove this TODO adding a note somewhere like `CONTRIBUTING` maybe, explaining to build it you must have Git 2.28+ as otherwise this test would fail.

docs/source/userguide/run/index.rst
```diff
@@ -49,6 +43,13 @@ More info on password-less ssh can be found at: http://www.linuxproblem.org/art_
 
 .. caution:: After launching Autosubmit, one must be aware of login expiry limit and policy (if applicable for any HPC) and renew the login access accordingly (by using token/key etc) before expiry.
 
+When running operational experiments (i.e. an experiment whose EXPID starts with ``'o'``, e.g. ``o001``),
```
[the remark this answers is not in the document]
kinow: Ah, I think I wrote the docs before adding the flag. Good catch.

kinow:
Modified the functions  to, instead of "check", to "get_uncommitted", and get_code_not_pushed. These functions are `Optional[str]`.

Tested first by creating an operational experiment with Git. Then added a file without committing it.

image

Then I committed it but didn't push to GitHub.

image

Then git reset it hard, removing any files or commits, and the workflow ran fine.

dbeltrankyl (approved):
Hello @kinow

Looks good! thanks

PR #2457, opened 2025-07-07
[Documentation] NUMTHREAD wrong use

### LINKED ISSUE #2123: Review documentation where it mentions NUMTHREADS
Today @orioltinto had a question about some job variables not being correctly translated as he expected them to be. I had a look, and checked our docs first. If a user searches by `--cpus-per-task`, they find two hits, one being [this one](https://autosubmit.readthedocs.io/en/master/userguide/variables.html#custom-directives).

Image

I tried the `NUMTHREADS` but it didn't work as expected. A quick debugging in the IDE, and found that I needed to use `THREADS` instead. It's not clear (or I don't remember) why we have threads/numthreads, but if one is deprecated or not used, then I guess it'd be better to remove it from the docs and warn the users to update their configs.

VindeeR:
Closes #2123 

Numthreads was being referenced wrongly in the documentation, this is a fix to indicate to the user which one would be the correct directive to use

**Check List**

- [x] Documentation updated.

### DIFF
--- a/docs/source/userguide/variables.rst
+++ b/docs/source/userguide/variables.rst
@@ -64,8 +64,8 @@ Custom directives
 -----------------
 
 There are job variables that Autosubmit automatically converts into
-directives for your batch server. For example, ``NUMTHREADS`` will
-be set in a Slurm platform as ``--SBATCH --cpus-per-task=$NUMTHREADS``.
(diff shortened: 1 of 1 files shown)


kinow (approved):
:+1:

PR #2458, opened 2025-07-07
[Refactoring] Changing how path is generated

VindeeR:
Updating path handling and updating how to locate files in the Job List

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,15 @@ How to cite Autosubmit:
     2016 International Conference on High Performance Computing & Simulation (HPCS), 
     Innsbruck, 2016, pp. 895-900. doi: 10.1109/HPCSim.2016.7568429
 
+AUTOSUBMIT IS MAINTAINED ON THE FOLLOWING MACHINES
+==================================================
+
+- bscesautosubmit{01/02}.bsc.es -- Autosubmit 3, BSC virtual machine
+- bsceshub{02/03/04}.bsc.es -- Autosubmit 3 and Autosubmit 4, BSC virtual machine
+- climate-dt -- Autosubmit 4, CSC virtual machine
+- EDITO-infra -- Autosubmit 4, MOi and VLIZ cloud infrastructure (Docker, Kubernetes)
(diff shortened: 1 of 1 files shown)


autosubmit/job/job_list.py
```diff
@@ -189,10 +189,10 @@ def generate(self, as_conf, date_list, member_list, num_chunks, chunk_ini, param
...
         if force:
             Log.debug("Resetting the workflow graph to a zero state")
-            if os.path.exists(os.path.join(self._persistence_path, self._persistence_file + ".pkl")):
-                os.remove(os.path.join(self._persistence_path, self._persistence_file + ".pkl"))
-            if os.path.exists(os.path.join(self._persistence_path, self._persistence_file + "_backup.pkl")):
-                os.remove(os.path.join(self._persistence_path, self._persistence_file + "_backup.pkl"))
+            if Path(self._persistence_path +"/"+ self._persistence_file + ".pkl").exists():
+                Path(self._persistence_path +"/"+ self._persistence_file + ".pkl").rmdir()
+            if Path(self._persistence_path +"/"+ self._persistence_file + "_backup.pkl").exists():
+                Path(self._persistence_path +"/"+ self._persistence_file + "_backup.pkl").rmdir()
```
kinow, line 195: I think we don't need the `"/"`. Comma can be used (I believe they designed it as a drop-in replacement of `os.path`.

Also, since we'll reuse the values, then maybe first declare a variable for each path, then do a

```py
if persistence_pkl_path.exists():
  persistence_pkl.path.rmdir()
```

But... I think it shouldn't be `rmdir`. Probably it needs to be `unlink()`?

If this works, and doesn't break CICD, @VindeeR , then we need a test for this change (i.e. you modified test with something that should break and it didn't, so either we don't have tests, or they are not good enough).

kinow (changes requested):

autosubmit/job/job_list.py
VindeeR: This was an really old commit I saw pending with.

I'm going to add the tests and fix what you pointed out.

Thanks for such a quick review :raised_hands:

PR #2459, opened 2025-07-08
[Refactoring] Redundant functions

### LINKED ISSUE #2364: Remove autosubmit.history.utils function create_path_if_not_exists
Besides being a mouthful to say its name, that function is redundant with pathlib. Its source code.

```python
def create_path_if_not_exists(path):
  # type : (str) -> bool
  if not os.path.exists(path):
    os.makedirs(path)
    return True
  return False
```

Which is the same as 

```py
Path(path).mkdirs(exists_ok=True)
```

In the code, currently, nothing checks the returned boolean. So not raising an error should be feature-similar, without breaking anything, but replaces all those lines by a single one from stdlib.

Found while testing/reviewing #2187

VindeeR:
Closes #2364 

Two redundant functions were removed from the utils class
- create_path_if_not_exists
- create_path_if_not_exists_group_permission
Both function had as purpose to create the folder and add the proper permission, which could be done using Path()

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1435,7 +1435,7 @@ class Autosubmit:
             exp_folder.mkdir(mode=dir_mode)
             required_dirs = ["conf", "pkl", "tmp", "tmp/ASLOGS", f"tmp/LOG_{exp_id}", "plot", "status"]
             for required_dir in required_dirs:
-                Path(exp_folder / required_dir).mkdir(mode=dir_mode)
+                Path(exp_folder / required_dir).mkdir(parents=True, mode=dir_mode)
             Log.info(f"Experiment folder: {exp_folder}")
         except OSError as e:
             try:
@@ -3505,7 +3505,7 @@ class Autosubmit:
                     Log.error("Not a valid path. You must include '~/' at the beginning.")
             database_path = Path(database_path).expanduser().resolve()
             # if not os.path.exists(database_path):
-            HUtils.create_path_if_not_exists(database_path)
+            Path(database_path).mkdir(parents=True, exist_ok=True)
             # Log.error("Database path does not exist.")
             # return False
             while database_filename is None:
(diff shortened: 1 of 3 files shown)


autosubmit/autosubmit.py
```diff
@@ -3505,7 +3505,7 @@ def configure(advanced, database_path, database_filename, local_root_path, platf
                     Log.error("Not a valid path. You must include '~/' at the beginning.")
             database_path = Path(database_path).expanduser().resolve()
             # if not os.path.exists(database_path):
-            HUtils.create_path_if_not_exists(database_path)
+            Path(database_path).mkdir(exist_ok=True)
```
kinow, line 3508: Also needs parents? :grimacing: Like old code...

autosubmit/monitor/monitor.py
```diff
@@ -625,7 +630,7 @@ def generate_output_stats(self, expid: str, joblist: list[Job], output_format="p
         output_complete_path_stats = os.path.join(BasicConfig.DEFAULT_OUTPUT_DIR, output_filename)
         is_default_path = True
         if is_owner or is_eadmin:
-            create_path_if_not_exists_group_permission(os.path.join(BasicConfig.LOCAL_ROOT_DIR, expid, "stats"))
+            Path(Path(BasicConfig.LOCAL_ROOT_DIR, expid, "stats")).mkdir(exist_ok=True, mode=0o774)
```
kinow, line 633: s/mode=0o774/mode=_DEFAULT_MKDIR_GROUP_PERMISSION (and for my OCD the order could match the one below :rofl: )

autosubmit/monitor/monitor.py
```diff
@@ -638,7 +643,8 @@ def generate_output_stats(self, expid: str, joblist: list[Job], output_format="p
                 is_default_path = False
         if is_default_path:
             Log.info("You don't have enough permissions to the experiment's ({}) folder. The output file will be created in the default location: {}".format(expid, BasicConfig.DEFAULT_OUTPUT_DIR))
-            create_path_if_not_exists_group_permission(BasicConfig.DEFAULT_OUTPUT_DIR)
+
+            Path(BasicConfig.DEFAULT_OUTPUT_DIR).mkdir(mode=_DEFAULT_MKDIR_GROUP_PERMISSION, exist_ok=True)
```
kinow, line 647: Ditto here :grimacing:

kinow (approved):
Looks great now! 🤟

PR #2474, opened 2025-07-21
CSCDESTINCLIMADT-794: Fix 'list' object has no attribute 'status'

kinow:
**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -2264,10 +2264,9 @@ class JobList(object):
                     jobs.append(job)
         return jobs
 
-    def get_in_queue_grouped_id(self, platform):
-        # type: (object) -> Dict[int, List[Job]]
-        jobs = self.get_in_queue(platform)
-        jobs_by_id = dict()
+    def get_in_queue_grouped_id(self, platform) -> dict[str, list[Job]]:
+        jobs: list[Job] = self.get_in_queue(platform)
+        jobs_by_id: dict[str, list[Job]] = dict()
         for job in jobs:
             if job.id not in jobs_by_id:
                 jobs_by_id[job.id] = list()
--- a/test/unit/test_job_list.py
+++ b/test/unit/test_job_list.py
@@ -817,3 +817,20 @@ def test_retrieve_times(job_list, jobs_as_dict, tmp_path, make_exception, second
                                                     job_times=None, seconds=seconds, job_data_collection=None)
             assert retrieve_data.name == job.name
             assert retrieve_data.status == Status.VALUE_TO_KEY[job.status]
+
+
+def test_get_in_queue_grouped_id_same_job_id(job_list, mocker):
+    """Test that in a list of Jobs where they all have the same ID, they
+    are returned grouped correctly.
+
+    TODO: maybe we should understand what triggers that (wrappers, recover a wrapper?).
+    """
+    platform = mocker.MagicMock()
+    platform.name = 'dummy_platform'
+    platform.serial_platform = platform
+
+    for job in job_list._job_list:
+        job.platform = platform
+
+    jobs_by_id = job_list.get_in_queue_grouped_id(platform)
+    assert all([isinstance(job, list) for job in jobs_by_id.values()])


autosubmit/job/job_list.py
```diff
@@ -2264,10 +2264,9 @@ def get_jobs_by_section(self, section_list: list, banned_jobs: list = None,
                     jobs.append(job)
         return jobs
 
-    def get_in_queue_grouped_id(self, platform):
-        # type: (object) -> Dict[int, List[Job]]
-        jobs = self.get_in_queue(platform)
-        jobs_by_id = dict()
+    def get_in_queue_grouped_id(self, platform) -> dict[str, list[Job]]:
+        jobs: list[Job] = self.get_in_queue(platform)
+        jobs_by_id: dict[str, list[Job]] = dict()
```
kinow, line 2269: Adding types, as they show in the IDE where the error is. No other changes here yet.

kinow:
When you `create` the workflow without wrappers, with just two jobs, you get two jobs in job_list created in `update_genealogy`, both with the same ID.

image

image

When you run it, it again calls the same `update_genealogy`, getting two jobs in the job_list, with the same ID, but only one job each time is executed (depend on each other).

---

When the jobs are wrapped, they have different IDs in `update_genealogy`. When you `run` it, job_list will have two jobs, different IDs too.

However, in `get_in_queue_grouped_id`, when it calls `self.get_in_queue(platform)`, it will return two jobs, with the exact same ID.

image

At that point, when the function goes to `check_wrappers`, we still don't get/reproduce the bug from @youcheng-csc because the code enters the first part of the `if` statement, and not the second where the bug happened.

image

That's why we normally don't see this bug happening.

---

When you run the same experiment after it failed, without creating it first, in `update_genealogy` now instead of two different IDs, you get two identical IDs :dizzy: My guess is that it's loading from disk, and since the last run had the same ID, that's what was saved.

Here, the code enters the second part of the `if` statement in `check_wrappers`, but now it has a single job, as the first failed, and the second succeeded.

---

In the existing experiment, I have two jobs. One fails, then I recover (similar to what @youcheng-csc did), and re-launch it. One completed, one failed. But having a single failed job, doesn't trigger the part of the code we want with a `list` (noticed this as I had spent some time reviewing the code and writing that unit test).

I increased `NUMCHUNKS` to `3`,  and modified the `SIM` `SCRIPT` to

```shell
# Fails on the second and third chunks
CHUNK="%CHUNK%"
if [ "$CHUNK" -eq 1 ]
then
  echo "OK!"
else
  echo "Uh oh"
  crashit!
fi
```

image

Here, I setstatus so I had one job failed, and two waiting in the wrapper.

image

Still not able to reproduce Cheng's issue. Getting closer, but can't figure out what makes the code to enter that branch of the `if` with a `list`. (At least we now know what happened.)

kinow:
Just before I cleaned everything, I spent some more minutes looking at the variables in different frames in the debugger, and taking note of the DB and its data.

Now I'm really grateful, as the bug repeated every time I launched the debugger -- this indicates that this is not a random error, but based on data/state.

Before, I was chasing how `job_list.job_package_map` could be empty. But I noticed in the debugger, that at one of the frames, the function/self had no job_package, it was empty (I think it was supposed to be a list or dict).

I had looked at the table in SQLite, and it was indeed empty.

Now I cannot reproduce the bug anymore, but my `job_package` is populated.

image

So now we just need to understand what caused my `job_package` to be empty. And I suspect @youcheng-csc 's DB had the same when the issue happened.

kinow:
Using the `LOCALDOCKER` Slurm container as platform, and crashing Autosubmit on `save_wrappers`, then re-running `autosubmit run o002` reproduces the bug with the Docker container too.

So the issue is not isolated to BSC MareNostrum5, but really something with the AS code, and probably with some state that's persisted or not upon errors. So the proper fix is to identify what state is being persisted or not, where/why, write a test to reproduce and fail. Then fix the error, and finally fix the code that's failing `mypy` (the easy fix that we have been postponing).

kinow:
Now that I can use the local Docker container it's a lot simpler to run several tests.

Cleaned everything (i.e. deleted logs, DB, tmp files, etc., job data, structure, etc.), then create + run crashing the `save_wrapper`. Now, I ran `autosubmit run` and reproduced the error.

Then saved the pickle and job packages. Job data is empty (0 bytes).

I had repeated the steps without crashing `save_wrapper` and that works. But what's more interesting, is that if instead of running `autosubmit run` after crashing, I run `autosubmit recovery`, then the issue cannot be reproduced.

That means `autosubmit recovery` fixes whatever state is "bad" in the pickle or DBs. So I recovery'd the experiment, and got the following diff'ing objects.

The sqlite DBs job package look identical, no job package or job package wrapper.

The pickle, however, has the following differences (`pip install deepdiff`, remove Include Sources from PyCharm launcher):

```python
{}
{'values_changed': {"root['_status']": {'new_value': 1, 'old_value': 2},
                    "root['ready_date']": {'new_value': 20250812134102,
                                           'old_value': 20250812134037}}}
{'values_changed': {"root['_status']": {'new_value': 0, 'old_value': 2}}}
done
```

The script used to produce that diff:

```py
import pickle
from pprint import pprint

from deepdiff import DeepDiff

with open('/home/kinow/autosubmit/o002/pkl/job_list_o002.pkl', 'rb') as fd:
    job_list = pickle.load(fd)

with open('/home/kinow/autosubmit/o002/pkl/job_list_o002_broken.pkl', 'rb') as fd:
    job_list_broken = pickle.load(fd)

assert len(job_list) == len(job_list_broken)

for i, pair in enumerate(zip(job_list.values(), job_list_broken.values())):
    differences = DeepDiff(pair[1], pair[2])
    pprint(differences)

print('done')
```

This output shows that the `_status` was changed from `2` (SUBMITTED` ) to `1` (`READY`). And the ready date was also modified, but I don't think that's important.

kinow:
Fixing the pull request to look at coverage, and to prepare the fix for the test.

test/unit/provenance/test_rocrate.py
```diff
@@ -504,114 +503,6 @@ def test_rocrate_main_fail_missing_rocrate(
...
-        job2.processors = '1'
-
-        mocked_job_list.get_job_list.return_value = [job1, job2]
-        mocked_job_list.get_ready.return_value = []  # Mock due the new addition in the job_list.load()
-        mocked_job_list.get_waiting.return_value = []  # Mocked due the new addition in the job_list.load()
-        autosubmit = Autosubmit()
-        r = autosubmit.rocrate(_EXPID, path=Path(temp_dir))
-        assert r
-
-
```
kinow, line 614: Move to an integration test, so we could use `autosubmit_exp` and remove all the mock calls (this test failed as the code and the mocks were not matching -- the danger of mocking too much).

autosubmit/platforms/wrappers/wrapper_factory.py
kinow, line 1: Cosmetic changes here. Nothing important.

autosubmit/job/job_packages.py
kinow, line 1: Cosmetic changes here too. Shouldn't matter to the issue in question, but reduce the warnings/code clutter.

autosubmit/autosubmit.py
```diff
@@ -136,13 +129,12 @@ def __call__(self, parser, namespace, values, option_string=None):  # noqa: F841
...
 
+
 class Autosubmit:
-    """
-    Interface class for autosubmit.
-    """
+    """Interface class for autosubmit."""
 
     def __init__(self):
-        self.command = None
```
kinow, line 145: Never used...

autosubmit/autosubmit.py
```diff
@@ -719,35 +716,38 @@ def run_command(args):
...
                                       args.expand_status, args.hide_groups, args.notransitive, args.check_wrapper,
-                                      args.txt_logfiles, args.profile, detail=False)
+                                      args.txt_logfiles, args.profile)
         elif args.command == 'stats':
             return Autosubmit.statistics(args.expid, args.filter_type, args.filter_period, args.output,
                                          args.section_summary, args.jobs_summary, args.hide, args.notransitive)
         elif args.command == 'clean':
             return Autosubmit.clean(args.expid, args.project, args.plot, args.stats)
         elif args.command == 'recovery':
+            # FIXME: remove transitive
```
kinow, line 735: Created issue for this, but added a few FIXME's (whoever fixes that should see and remove it).

autosubmit/autosubmit.py
```diff
@@ -5559,31 +5568,16 @@ def _get_chunks(out):
...
-        """
-        Returns the submitter corresponding to the communication defined on autosubmit's config file
+    def _get_submitter(as_conf) -> ParamikoSubmitter:
+        """Returns the submitter corresponding to the communication defined on autosubmit's config file.
 
         :return: submitter
-        :rtype: Submitter
         """
+        # FIXME: Why are we loading the communication library if we only return one type of submitter?
+        #        Quite sure Erick/vindeer found this same thing confusing...
```
kinow, line 5577: @VindeeR saw this, I believe. Adding a note here so we don't forget it.

kinow:
@dbeltrankyl , I removed some changes, but now the remaining ones are mainly docs, types, linter. I can undo some, but at least types and the TODO/FIXME added I think would be good to add. Can you review it whenever you have time, and let me know what you think? If there are parts you prefer that we do not touch, just mention and I'll revert :+1: 

I also noticed that some of the linter issues I'm fixing here, @VindeeR is also fixing in his #2550 PR. Whoever merges first, would save the other from some work (and other devs too, in future PRs :grimacing: ).

Cheers

autosubmit/job/job_packager.py
```diff
@@ -660,40 +705,59 @@ def _divide_list_by_section(self, jobs_list):
         sections_split = dict()
         jobs_by_section = dict()
 
-        for wrapper_name,jobs_in_wrapper in self.jobs_in_wrapper.items():
+        # FIXME: At some point Erick had a problem in the documentation where an example
```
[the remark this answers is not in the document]
kinow: Excellent! This is what I was trying to remember. Let me create an issue for this: https://github.com/BSC-ES/autosubmit/issues/2568

test/unit/test_job_list.py
```diff
@@ -45,8 +46,12 @@ def as_conf(autosubmit_config):
     })
 
 
+# TODO: this file has ``setup_job_list`` and ``job_list`` that create job lists.
```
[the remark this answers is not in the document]
kinow: Thanks Dani!

autosubmit/autosubmit.py
```diff
@@ -3100,66 +3104,70 @@ def recovery(expid, noplot, save, all_jobs, hide, group_by=None, expand=list(),
...
+        except Exception as e:
             raise AutosubmitCritical("Couldn't restore the experiment workflow", 7040, str(e))
 
         try:
             packages = JobPackagePersistence(expid).load()
 
-            groups_dict = dict()
+            groups_dict: Optional[dict[str, Any]] = None
+
+            # TODO: Wouldn't it make sense to move/combine `if group_by` and `if not plot`? They seem to share
```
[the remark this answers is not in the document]
kinow: Thanks, @dbeltrankyl . Issue created to combine these two, and then I will also move the other code that doesn't need to be executed when `-np` is passed. https://github.com/BSC-ES/autosubmit/issues/2570

autosubmit/job/job_list.py
```diff
@@ -3428,3 +3406,48 @@ def ts_to_datetime(timestamp):
...
+    run_only_members = as_conf.get_member_list(run_only=True)
+
+    job_list = JobList(expid, as_conf, YAMLParserFactory(), get_job_list_persistence(expid, as_conf))
+    job_list.generate(
+        as_conf, date_list, as_conf.get_member_list(), as_conf.get_num_chunks(), as_conf.get_chunk_ini(),
+        as_conf.experiment_data, date_format, as_conf.get_retrials(), as_conf.get_default_job_type(),
+        wrapper_jobs, new=new, run_only_members=run_only_members
+    )
+
+    # TODO: If we load a job list with submitted statuses, should we reset then here? Or alert the user
```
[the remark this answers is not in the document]
kinow: That was part of the issue here, Dani. Somehow, I got an empty job_package_map/or dict (can't recall the name of the variable). Normally it's filled with something, but somehow, that ended up empty and recovery + run triggered the same issue Cheng had.

Later, I tried to reproduce that, but without luck, and the only way I got the exception Kai & Cheng had was by doing that forced crash before `save_wrappers`. Strange.

I'll leave this TODO in case it helps whoever works on a similar issue. That could be removed in the postgres branch, or later if we refactor/improve that area of the code.

autosubmit/autosubmit.py
```diff
@@ -4170,10 +4175,11 @@ def database_fix(expid):
...
                 os.popen(bash_command).read()
                 exp_history.initialize_database()
-
-            except Exception:
-                Log.warning("It was not possible to restore the jobs_data.db file... , a new blank db will be created")
-                result = os.popen(f"rm {database_path}").read()
+            except Exception as e:
+                Log.warning("It was not possible to restore the jobs_data.db file... , "
+                            f"a new blank db will be created: {str(e)}")
+                # FIXME: replace os.popen by Path#unlink?
```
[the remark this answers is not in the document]
kinow: With postgres we should be able to avoid that with transactions. They will guarantee things are persisted correctly, or it fails otherwise (whereas with NFS & normal Linux I/O it's hard, unless you wait for the `flush` call in Linux, I think... in NFS it varies by vendor, and there are always bugs... :bug: :lady_beetle: )

autosubmit/autosubmit.py
```diff
@@ -1877,33 +1850,34 @@ def check_wrappers(
...
-                        if platform.name in jobs_to_check:
-                            jobs_to_check[platform.name].append([job, job_prev_status])
-                        else:
-                            jobs_to_check[platform.name] = [[job, job_prev_status]]
-        return jobs_to_check,job_changes_tracker
+                else:
+                    # TODO: ``JobList.get_in_queue_grouped_id`` appears to return only jobs that are submitted,
+                    #       running, queuing, unknown, and held. Is it possible to have failed jobs too?
+                    # Adds to a list all running jobs to be checked. Failed jobs are ignored.
+                    for job in [j for j in jobs if j.status != Status.FAILED]:
```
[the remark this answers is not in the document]
kinow: > I did this in my branch, and what happened is that "wrapped" jobs were wrongly being considered as "simple" jobs. Will this raise issues further down? Do we double-check somewhere that the job is simple and not a wrapped one? Maybe it is worth doing, even if it is a Log.warning for us to debug if something weird happen

What was it that you did in your branch, Dani? Make `get_in_queue_grouped_id` return a list of jobs, and iterate like here?

The rest of the code should be doing the same as before, ignoring `FAILED` job. I think even though this function check wrappers, it checks and reorders all jobs. I think you had left a comment just before this function is called

>  # This function name is totally misleading, yes it check the status of the wrappers,
>  # but also orders jobs the jobs that  are not wrapped by platform.

I don't know where we could add this check. It'd be something like `if job.wrapper_type in [...]`? And then check if it contains simple, or if it's in the valid states?

> Also, in my branch, I encountered another bug (but I'm 99.9% sure it's due to the memory changes which should be solved) where the job_list object ID differed from the packages_map job object ID.

This would trigger the bug of Cheng & Kai too, and could be the/or part of the root cause of this issue, I think. Interesting.

kinow:
Thanks for the review, @dbeltrankyl . I think I addressed all your comments agreeing with them, deleting TODOs that you answered already, or creating issues.

Just this one, https://github.com/BSC-ES/autosubmit/pull/2474/files#r2292870824, I couldn't answer very well, I think.

Could you have another look, please, @dbeltrankyl ?

I rebased, and pushed the changes that are simply removing the TODO/FIXME you answered. If an intermittent test fails (test_send_data or one in test_slurm) then I'll kick the build and then it should pass.

Thanks

autosubmit/autosubmit.py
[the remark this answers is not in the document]
dbeltrankyl: So what I asked there was if that else is executed, what will happen futher down in the execution? because the job will stop being considered wrapped

autosubmit/job/job_list.py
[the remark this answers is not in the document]
dbeltrankyl: > recovery + run triggered the same issue Cheng had.

Hm, I see that can happen if the wrapper was running before using recovery, because some jobs inside the wrapper could be completed and maybe the create erases the current wrappers ? or maybe the load of the previous run is not saving everything

> hat could be removed in the postgres branch, or later if we refactor/improve that area of the code.

Well, the saving / loading is enterily different on that branch, so yes

If possible or we have time, I'll add a test to emulate a run recovery and wrappers in the #2187 ( or leave it for a different pr)

kinow:
Rebased, still pending to look at one comment from Dani's feedback.

kinow:
Rebased, but will hold until #2577 is merged, then rebase this one again and then this PR should have a lot less changes.

dbeltrankyl:
This should be done for 4.1.16. It was already working but blocked by #2577

dbeltrankyl:
> This should be done for 4.1.16. It was already working but blocked by https://github.com/BSC-ES/autosubmit/pull/2577

What I'm not sure is when the #2577 will be ready? Because maybe we can merge this one without the other as it is a DestinE bug.   cc @kinow

kinow:
The #2551 should be ready to be reviewed, once @VindeeR has time to finish reviewing #2604 (which contains tests for some of his changes). Then I will rebase this one and we should be ready to review/merge this one.

kinow:
Rebased, but there are more changes to be moved to another branch. Will start doing that tomorrow.

dbeltrankyl:
ah I had an old version locally... let me check again

kinow:
Ah, thanks for spotting the typo. I will see why my PyCharm is not complaining about it. Tomorrow I will nuke my venv and Conda envs. Maybe my containers are using some broken version.

dbeltrankyl:
Did you performed 

$ sudo setfacl --modify user:$USER:rw /var/run/docker.sock

?

For a while, I don't know the PR that caused it, but I commented in one of my PR( but it was not related to any change done by me), that the DB postgres tests weren't being launched correctly when using -m "" 

Now I see that with -m "" and without that enabled, all the tests are failing due to a permission error in Docker (even the ones that are not running with Docker 🤔), not sure if it is related to last week's changes or the Pr that caused the first issue

Found it! I commented on it there https://github.com/BSC-ES/autosubmit/pull/2700#issuecomment-3511397145 

So I suspect you didn't set the Docker flag.

kinow:
This is helpful, thanks @VindeeR ! I am nuking all my env and downloading half of the whole Internet with Docker and Conda here :) thanks!!

VindeeR:
Just a bit more of a complement, running the tests that failed individually immediately after works

kinow:
Looks like the infinite loop is stuck in the SSH transport.

image

I also noticed I had a lot of Docker Slurm containers still alive on my laptop. TestContainers is supposed to manage the containers and kill them with a sidekick container (when you launch a test with slurm docker, you should notice always at least 2 containers, SSH Slurm, and another one from testcontainers using testcontainers/ryuk).

kinow:
First problem I found is that now the autosubmit_exp ignores JOBS passed via EXPERIMENT_DATA by default. That sounds counterintuitive. If I give it settings, by default I think it must use them, unless I disable. Right now it's the opposite. Any `JOBS` passed is ignored unless you tell the fixture to use that.

Second problem appears to be in the joblist. Undoing that change got the test to fail at least. It's failing due to the other change in autosubmit.py. But looks like I can debug this problem with GH actions that way, as I couldn't get the test to work on PyCharm on my home or BSC laptops.

kinow:
I see your point, @dbeltrankyl . But right now if any developer passes this:

```yaml
autosubmit_exp('a000',
EXPERIMENT: {...},
PLATFORMS:
  PLATFORM: {...}
JOBS: {
  A: {
    FILE: bla
  }
}
)
```

Somebody reading this code might expect `JOBS` to be used (why not? they are valid AS config). But the current logic ignores it.

I think what you are saying is that if I pass

```yaml
autosubmit_exp('a000',
EXPERIMENT: {...},
PLATFORMS:
  PLATFORM: {...}
# NO JOBS
)
```

then Autosubmit includes the default jobs which can be risky or misleading. I agree with that. But I think we didn't need to change it to ignore the JOBS. Instead, we could change inside the fixture to do something like

```
def autosubmit_exp():
   def inner_fn(experiment_data: dict):
       if 'JOBS' not in experiment_data:
           raise ValueError('The fixture needs the test JOBS')  # or remove the default jobs?
```

And if a test needs an experiment that's job-less, then we just add a flag like you did, but with the default False.

That way, the old style of the tests would work. And if you needed that for a new test, then this new test would be the exception and avoid loading the jobs.

The changes in joblist are in this PR. The infinite loop is gone and it fails, so now I can keep poking the code to see what part is causing the issue.

kinow:
Hmmm, I think we were talking about different things, @dbeltrankyl . I thought that flag was to delete the jobs, because this is what I had in my jobs in `/tmp/pytest-of-kinow/....`:

image

I looked at my last tests and they have jobs. It was after I changed the flag to not delete the jobs, but I also changed more things (re-created again conda env, re-installed Docker, rebooted, fiddled with firewall, and changed branches a few times). So maybe there's something happening for me.

Will continue looking into that on Monday. Thanks!

autosubmit/autosubmit.py
```diff
@@ -1953,35 +1953,35 @@ def check_wrappers(
         :param expid: a string with the experiment id
         :return: non-wrapped jobs to check and a dictionary with the changes in the jobs status
         """
-        jobs_to_check: dict[str, list[list[Job]]] = dict()
+        jobs_to_check: dict[str, list[list[Job]]] = defaultdict(list)
```
kinow: We had the pattern

```py
if v not in dic:
  dic[v] = []
else:
  dic.append(v)
```

so replaced that by the defaultdict(list) here so we don't need that if any longer.

test/integration/platforms/test_slurmplatform.py
```diff
@@ -1146,3 +1151,94 @@ def test_check_if_packages_are_ready_to_build(autosubmit_exp):
...
+                'TEMP_DIR': '',
+                'TYPE': 'slurm',
+                'USER': 'root',
+                'MAX_PROCESSORS': 1,
+                'PROCESSORS_PER_NODE': 1,
+            }
+        }
+    })
+
+    mocked_job_list_save_wrappers = mocker.patch.object(JobList, 'save_wrappers', side_effect=ValueError)
```
dbeltrankyl: Just a question and curiosity 

since it is an integration test, why mock here instead of actually saving it? 

It is because I'm trying always to avoid mocking stuff in integration tests so far

test/integration/platforms/test_slurmplatform.py
```diff
@@ -1146,3 +1151,94 @@ def test_check_if_packages_are_ready_to_build(autosubmit_exp):
...
+
+    with pytest.raises(ValueError):
+        exp.autosubmit.run_experiment(expid=_EXPID)
+
+    mocked_job_list_save_wrappers.reset_mock()
+
+    # NOTE: On ``master`` before the fix (commit d635461f4ecac42985ba584e2cbfa65223ea2151),
+    #       this fails instead of raising ``ValueError``, showing the same exception
+    #       reported by users, ``AttributeError: 'list' object has no attribute 'status'``.
```
dbeltrankyl: So, if we're checking if the AttributeError is raised, shouldn't we avoid mocking the save and see if no exceptions are raised? 

Surely the fix fixes the AttributeError, but if we do further changes on the code for this function, it may contain a bug, and it could happen that we will see that this function is covered and passing as it is raising the expected valueError due to the mock

test/unit/test_job_list.py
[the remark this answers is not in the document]
dbeltrankyl: > or do it here directly..

In a sub-branch of the 4.2 one I mean

test/integration/platforms/test_slurmplatform.py
```diff
@@ -1146,3 +1151,94 @@ def test_check_if_packages_are_ready_to_build(autosubmit_exp):
...
+    ``status`` of a Python ``list`` object has also been fixed in that pull
+    request -- the fix was postponed to see if we could locate the root
+    cause, even though this may not be the root cause.
+
+    Ref:
+
+    * https://github.com/BSC-ES/autosubmit/issues/2463
+    * https://jira.eduuni.fi/browse/CSCDESTINCLIMADT-794
+    """
+    exp = autosubmit_exp(_EXPID, experiment_data={
```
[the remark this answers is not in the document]
kinow: Oh, d'oh. At first I didn't understand where the exp.expid could be reused. Now I see it. Updated it in two places. Thanks

test/integration/platforms/test_slurmplatform.py
```diff
@@ -1146,3 +1150,93 @@ def test_check_if_packages_are_ready_to_build(autosubmit_exp):
...
+    exp.autosubmit._check_ownership_and_set_last_command(
+        exp.as_conf,
+        exp.expid,
+        'run')
+
+    # Here we just confirm that the exception is raised for whatever mysterious
+    # reason (I/O error, out of memory, etc.).
+    with pytest.raises(ValueError):
+        exp.autosubmit.run_experiment(expid=exp.expid)
+
```
dbeltrankyl: Hello @kinow , @VindeeR 

Maybe the infinite loop test issue is there? 

I rebased the 4.2.0 branch, and there this test was failing.

I notice that the save_wrapper is never unmocked. 

Shouldn't this line be added? after the first run?

    JobList.save_wrappers = real_save_wrappers  # restore original

kinow: Ha! Had the same issue as @VindeeR, @dbeltrankyl . I'm testing this fix there now. :crossed_fingers:

Issue #2462, opened 2025-07-10
Implement check of performance metrics after each chunk


MissterP:
Hi!

### Problem

I would like to implement a feature that checks if some performance metrics (e.g.  SYPD) are within a threshold (from the DestinE wiki) after a simulated chunk of an experiment finishes. Moreover, if some of the performance metrics are not within the threshold, then an email will be sent to an operator//group of operators (their email would be at the config files of an experiment). The objective is to only implement the check for SYPD , but I would like to take into account for the implementation that could be more performance checked at the future.

### Proposed Solution

I would like to implement this feature by re-using the access of information after a chunk finishes and the email system. Before starting the implementation, I would like to request information about the code that manages the information of a chunk of an experiment once it finishes. 

Thank you!

dbeltrankyl:
1) 

This can be integrated into our actual subprocess and would be the optimal solution, as it would be beneficial to have it for all experiments, not just DestinE-related. Maybe, if needed, you can hardcode some stuff to make it compatible with only DestinE, and we (AS developers) can extend it at some point to make it more general and add more metrics.

The proper place would be:

- autosubmit->platform->platform.py -> recover_platform_job_logs_wrapper() which calls to:
- autosubmit->platform->platform.py -> recover_platform_job_logs()

It won't delay Autosubmit main loop and will be triggered as soon as a log is available to recover. 

You could add a conditional there to check the current job.section, chunk, split etc.

The optimal would be to have an AS Yaml setting that allows you to only calculate stuff for some sections etc, but it can be done later or we can help to do it during the PR (cc @kinow )


You could recover the NUMCHUNKS from the autosubmit config

2 and 3) 

Between these, I'm not sure what is best design-wise as it depends on the workflow, so this is a question for the workflow developers more than us. Also, I think that adding one extra job per chunk would be a bit (potentially) much overhead due the queuing times, maybe you could just do it every X chunks using the frequency parameter of Autosubmit


> The main problem is how to send the e-mail without increasing the complexity of the implementation and affecting to the maintenance. We would like to ask if there is a way that from the script we could use the e-mail system from Autosubmit (Reusability), or if there is a way to replicate the functionality of the e-mail system to the script (Duplication?). Maybe @isimo00 can provide information related to how the e-mail system could be replicated.


The best way to do this is to extend the Autosubmit e-mail functionality so you can add/extend a custom message. The problem would be finding the file that contains the information the mail needs, and we still would need to develop something in Autosubmit itself

[...]

PR #2481, opened 2025-07-23
Implement check of performance metrics after each chunk

paparici:
This pull request introduces a new **performance monitoring** for **Autosubmit**, enabling the **computation** and **notification** (using **e-mail**) of **performance** **metrics** for **jobs** if they aren't within a threshold ([issue](https://github.com/BSC-ES/autosubmit/issues/2462)). The changes include the addition of a base performance class,  specific implementations for job types and projects (e.g. DestinE). It's important to indicate that only is implemented the metric `SYPD`. 

An example of the **new configuration** for the performance:  
```
PERFORMANCE:
- PROJECT: 'DESTINE'
- SECTION: 
  - SIM
  - POST
- NOTIFY_ON: 
 - COMPLETED 
 - FAILED 
- NOTIFY_TO: 
  - e-mail1 
  - e-mail2
```

The configuration lets the user to select the jobs to compute and notify performance metrics. Moreover, if a project has specific performance metrics, they will be computed.  In the implementation, I did a dummy `DestinESIMPerformance` to show how it could be implemented (e.g. changes the threshold for SYPD, new metrics...).

**Question**: I would like to ask how to add integration tests (the unit tests are completed) for this implementation and where should I include it (tests for platforms.py?).

**P.S**: I'm looking for the threshold of **SYPD** for **DESTINE**. So, I would like to ask to have reviewed the code at the meantime, therefore I will need an additional commit to update the threshold.


MissterP:
@VindeeR told me to check for pylint errors. So, I will check that.

autosubmit/performance/base_performance.py
```diff
@@ -35,17 +36,20 @@ class PerformanceMetricInfo(NamedTuple):
...
-       
+
 
 class BasePerformance(ABC):
     """Base class for performance metrics calculation"""
 
-    _mail_notifier = MailNotifier(BasicConfig()) # Default MailNotifier with BasicConfig 
+    _mail_notifier = MailNotifier(
+        BasicConfig()
+    )  # Default MailNotifier with BasicConfig
```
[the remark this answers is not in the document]
MissterP: Hi! I tried to use **black** to reduce pylint warnings. I just saw that in .pylintrc states that max_size is 120 so, I will do the changes. Thank you for the information! 

I would like to ask if is there a standard tool for solving pylint warnings in Autosubmit?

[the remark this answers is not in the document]
kinow: Not yet. We have an issue to add that, but we haven't had time yet. For now, the best option is to try to lint new code, leaving code not related to your change unchanged (easier to review :+1: ) (unless you are adding docs, mypy, or fixing code that's hard to read)

MissterP:
Hi! It's not necessary to run the tests. It's still a work in progress. I just want to update my branch with the changes of master.

kinow:
> Hi! Still working on some issues detected from the integration tests. I would like to ask why the start_timestamp of a job could be greater than the finish_timestamp (executing tests in local) even when in the configuration of the job, the scripts states sleep 5. Moreover, I'm running the tests with the parameter -n 0.

Hmmm, weird. Could be, a) `sleep` is mocked, b) bug in our logic, c) something is taking too long or running too fast in the tests. We will try to replicate it later, thanks for the detailed info, Pablo :+1:

test/unit/performance/utils.py
```diff
@@ -0,0 +1,66 @@
...
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
+
+class Utils:
+    """
+    Utility class for common operations.
+    """
+
+    @staticmethod
+    def get_SYPD(start_timestamp: int, finish_timestamp: int, chunk_size: str, chunk_size_unit: str) -> float:
```
kinow, line 25: As this class contains only static functions, it probably make more sense to make these functions, without a class, and then import then when/if needed.

The naming also needs to be lowercase (uppercase in Python is rarely used, reserved for constants [which are also normally not constants.... but that's just Python :shrug: ]).

test/unit/performance/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    def get_under_performance(value: float, threshold: float) -> float:
+        """
+        Calculate the percentage under performance based on the value and threshold.
+        
+        :param value: The actual value to compare against the threshold.
+        :type value: float 
+
+        :param threshold: The threshold value to compare against.
+        :type threshold: float
+
```
kinow, line 60: In Python docstrings, parameters and types, etc, do not need spaces between them.

test/unit/performance/utils.py
```diff
@@ -0,0 +1,66 @@
...
+        elif chunk_size_unit == "month":
+            return duration_seconds / (30 * 24 * 3600) * float(chunk_size)
+        elif chunk_size_unit == "day":
+            return duration_seconds / (24 * 3600) * float(chunk_size)
+        raise ValueError(f"Unsupported chunk size unit: {chunk_size_unit}")
+
+    @staticmethod
+    def get_under_performance(value: float, threshold: float) -> float:
+        """
+        Calculate the percentage under performance based on the value and threshold.
```
kinow, line 53: You can place this comment right after the `"""`, https://peps.python.org/pep-0008/#documentation-strings

That's like the header of the docs. It's normally short, very brief, just stating what it does in a few chars. Anything else in the lines below, with one blank line separating paragraphs.

autosubmit/performance/type_job/SIM/SIM_performance.py
kinow, line 1: Python files normally do not have upper case, folders too are kept in lower case (unless we have a good/strong reason for doing so).

kinow (commented):
> Hi @kinow!
>
>I did a refactor to use the Injection pattern so that I can get the MailNotifier instance from a utils.py from my performance folder. However, I have the same issue were the patch isn't applying because it's being applied after the imports.

Do you mean in `mock_base_performance_notifier`? Could you point which part you need help with? I'm a bit lost as there are quite a few files in this PR :grimacing:

test/unit/performance/utils.py
MissterP: On it! Thanks for the info!

test/unit/performance/utils.py
MissterP: On it! Thanks for the info!

autosubmit/performance/type_job/SIM/SIM_performance.py
MissterP: On it! Thanks for the info!

MissterP: > Python files normally do not have upper case, folders too are kept in lower case (unless we have a good/strong reason for doing so).

Before starting to apply lint, PEP8, etc. I wanted to check that the integration tests are successful. However, I really appreciate your comments.

MissterP:
> Python files normally do not have upper case, folders too are kept in lower case (unless we have a good/strong reason for doing so).

Before starting to apply lint, PEP8, etc. I wanted to check that the integration tests are successful. However, I really appreciate your comments!

PR #2521, opened 2025-07-30
[config] Fix #2484 and #2485, add mypy type hints, fix typos, and imports

### LINKED ISSUE #2484: [config] If `experiment_data.yml` produces an error saving, it deletes the file quietly, without logging the exception
### Description

I am fixing the last unit test of #2482, to close #2052. The test is failing as there is an error saving the YAML (the experiment data has a function, that cannot be serialized, I believe).

The code catches the exception, and deletes the file, but never alerts the user or developer. It'd be useful to have something logged, at least at DEBUG level.

* Autosubmit version: master

### Reproducible Example

Modify the code or mock something to raise an error when YAML dumps data onto a file.

Image

### Expected Behaviour

If there is an error saving any YAML to the disk, we log something useful so that users and/or developers are aware and can take some action.

### LINKED ISSUE #2485: [config] AutosubmitConfig.check_autosubmit_conf always returns True
### Description

Found while moving config-parser code back to this repository and reading the code.

Image

In the code above, it will return True or else True. The return False is unreachable. Not sure if it is breaking anything, but logging this issue so we don't forget it.

kinow:
Closes #2484 
Closes #2485 

Fixes typos, adds mypy type hints, fixes several IDE warnings, and removes unused functions.

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/config/configcommon.py
+++ b/autosubmit/config/configcommon.py
@@ -29,16 +29,16 @@ from collections import defaultdict
 from contextlib import suppress
 from datetime import datetime, timedelta
 from pathlib import Path
-from typing import List, Union, Any, Tuple, Dict
+from typing import Any, Optional, Union
 
 from bscearth.utils.date import parse_date
 from configobj import ConfigObj
 from pyparsing import nestedExpr
 from ruamel.yaml import YAML
 
+from autosubmit.config.basicconfig import BasicConfig
+from autosubmit.config.yamlparser import YAMLParserFactory
 from log.log import Log, AutosubmitCritical, AutosubmitError
-from .basicconfig import BasicConfig
-from .yamlparser import YAMLParserFactory
 
 
 class AutosubmitConfig(object):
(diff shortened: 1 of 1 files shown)


autosubmit/config/configcommon.py
```diff
@@ -203,10 +204,9 @@ def get_section(self, section, d_value="", must_exists=False):
                 "{0} must exists. Check that subsection {1} exists.".format(section_str, str(current_level)), 7014)
         if current_level is None or (not isinstance(current_level, numbers.Number) and len(current_level) == 0):
             return d_value
-        else:
-            return current_level
+        return current_level
 
-    def get_wchunkinc(self, section):
+    def get_wchunkinc(self, section: str) -> str:
```
kinow, line 209: There are a few functions where I think we would have to change the code in order to being able to add types. For those functions, I left the code untouched as we will probably need good tests first. For other functions like this one I simply added the types as in pydocs or after reading the code.

kinow:
Had some minutes before lunch, and decided to tackle the two issues I found while migrating autosubmitconfigparser to autosubmit (will migrate the remaining pull requests and archive that repo). @VindeeR can you have a look at this PR, please? Cheers

PR #2527, opened 2025-07-30
[config] fix runtime error in current_variables when reference variable does n…

kinow:
From https://github.com/BSC-ES/autosubmit-config-parser/pull/79

@dbeltrankyl 's original comment:

This fixes an issue when the key to substitute is an int, float. This only happens on running time. 

I'm not entirely sure why. 

However, it occurs when you have a current variable that autosubmit generates during runtime, which refers to a variable that does not exist.

Example:

```yaml
PLATFORMS:
      BLA:
        type: ps
        BLA_PROCESSORS: 3
   
JOBS:
  JOB:
    PROCESSORS: "%CURRENT_BLA_PROCESSORS%"
    PLATFORM: BLA
```

This works.

But this doesn't

```yaml
PLATFORMS:
      BLA:
         type: ps
         
JOBS:
  JOB:
    PROCESSORS: "%CURRENT_BLA_PROCESSORS%"
    PLATFORM: BLA
```

Here, Autosubmit understands that Processors is 1 ( default value when processors is None )

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/config/configcommon.py
+++ b/autosubmit/config/configcommon.py
@@ -1025,6 +1025,8 @@ class AutosubmitConfig(object):
             keys = parameters.get(str(dynamic_var[0][start_long:-1]), None)
             if not keys:
                 keys = parameters.get(str(dynamic_var[0]), None)
+                if type(keys) in [int, float]:
+                    keys = None
         else:
             keys = dynamic_var[1]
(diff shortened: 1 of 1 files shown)


dbeltrankyl:
I believe we applied a different fix for this one

I'll close the pr but keep the branch for a while just in case

PR #2530, opened 2025-07-31
Define SSH connection banner timeout in tests

kinow:
**Check List**

The `banner_timeout` argument in the `connect` function for Paramiko is not set by default. Setting it is the recommended solution for an issue we had in our integration tests these past days:

image

This PR sets the timeouts in tests for connection, auth, and banner, all defined with the same value of 90 seconds (which should be enough, hopefully.

- [ ] Documentation updated.

### DIFF
--- a/test/integration/test_paramiko_platform.py
+++ b/test/integration/test_paramiko_platform.py
@@ -35,6 +35,8 @@ Note that tests will start and destroy an SSH server. For unit tests, see ``para
 in the ``test/unit`` directory."""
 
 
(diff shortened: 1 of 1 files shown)


VindeeR (approved):
I had problems with permission on the tests:
integration/test_paramiko_platform.py::test_send_file[filename]
integration/test_paramiko_platform.py::test_send_file[filename_long_path]

But I believe your changes are unrelated, I too did a few tests in order to try and see if the issue would happen and it didn't, so I believe it is working :crossed_fingers:

PR #2537, opened 2025-08-05
Safe delete

dbeltrankyl:
Derived from #2187, I found that the delete was not safe and could result in the deletion of anything if it was directly called (without the command) 

This fixes it

**Check List**

- [x] Documentation updated.


kinow (approved):
:pray: thanks Dani!!! Approving assuming GH Actions pass.

PR #2545, opened 2025-08-12
Add log compression command

### LINKED ISSUE #2266: Automatic compression of logfiles

LuiggiTenorioK:
Closes #2266
Supersedes: #2383 

This PR implements a command to compress the logs of a given experiment using the lzma algorithm. The idea is to separate concerns to simplify the solution. Given:

> * Compression of the Autosubmit log files
>    * There could be a parameter to enable it and compress all the command-related logs excepting the current


LuiggiTenorioK:
@dbeltrankyl @kinow this PR supersedes what Irene was working on (#2383)

Now, I have created a new command to compress all the command logs (ASLOGS): `autosubmit compresslogs $EXPID`

The idea is to have a command to do this operation without depending on another command at first. Then, we can trigger the log compression in a modular way at the end of the execution by adding `--compresslogs` to a command that uses an expid (e.g., `autosubmit --compresslogs create $EXPID`, `autosubmit --compresslogs run $EXPID`, etc).

@dbeltrankyl I think we can extend it later to not compress the logs used in the current command or skip N logs as you mentioned. For now, I think this fulfills the main need.

I'll be looking on compressing .out and .err logs before or after transfer.

LuiggiTenorioK:
@kinow btw do you know why integration tests sometimes fail? I have been re-running them, thinking it was something temporary, but now I'm starting to worry :sweat_smile:

kinow (commented):
Not reviewing yet, just organizing list of PRs to review today/over next week. @LuiggiTenorioK your work has very little from Irene's work, you seem to have used some of her old code like `is_xz_file`, and improved others like `find_uncompressed_files`, used the same algorithm she was using, but changed how things are done and improved (pathlib/types/etc.). It's missing her script for comparing the algorithms (I think @mcastril had also suggested adding one to that script, to check the performance of... lzma perhaps?). Did you consider keeping Irene's commit, and either including your work in the same commit as co-author, or as a second commit? If that's OK with you, when reviewing it I can bring her commit (I'd just start a blank branch locally, bring her commit, then apply your changes manually adding a second commit and push-force and ask you to take a look).

mcastril:
Is there any con for using the cmd cli? I tend to think that errors with invoked programs are harder to debug since we don't control the verbosity. Additionally, I guess this will create an extra dependency. `xz` is widespread in Linux distributions but it could not be present in minimal or very light ones. That's mainly what I mentioned the Python implementation. 

Unless we have the ability to run the cmd only if it exist, and run the Python implementation otherwise, without producing an error.

The flag option also looks good. I had thought about an option in the config, but since the commands are logged to, I guess there's no problem about it

kinow:
Running the cmd cli (i.e. forking the linux process from Python) has a performance penalty that was visible in Irene's matplotlib plots. It has a security issue as there are attacks that require code to fork a process. And, as you said, it also creates a dependency to a binary on a platform (that may or may not work, may have extra dependencies to `module`'s which may get broken if IT/Ops change anything, etc.).

I think we could implement different compression algorithms, some that depend on tools being available like `xz`, but some that use pure python. That way, we could tell users (e.g. Andrea/KIT, EDITO, etc.) to use one that uses al algo that comes with Python, and others that need a higher compression ratio to install and use their binaries, like `xz` on DestinE (as we know `xz` is available there as that's what Ulf is using).

I think that's a bit more of work for us, but it'd make AS more portable and it'd pay off in the long run.

LuiggiTenorioK:
I made another script to include the comparison between Python LZMA and the xz CLI (threaded and multi-threaded), and I can confirm the penalty that Bruno mentioned. Also, there is no visible improvement using multi-threads in the xz cli.

compression_comparison

Additionally, there is a small improvement in compression ratio using zstd compression, but not sure if it is enough to consider replacing lzma since zstd is not as widespread as lzma yet.

kinow:
zstandard is looking quite good to save disk space, better than xz (I think Fedora was recommending to use that over xz after the security issue). Is zstandard forking a process to compress, and calling a system utility too?

As for compression time, zipfile looks a lot a lot faster. Maybe the space saved with others is not worth the complexity and time to compress.

Thanks Luiggi!

LuiggiTenorioK:
> zstandard is looking quite good to save disk space, better than xz (I think Fedora was recommending to use that over xz after the security issue). Is zstandard forking a process to compress, and calling a system utility too?

I used the same python lib that Irene was using (`zstandard`), and it seems to be using C bindings (CFFI) to run a compiled C library, which should be faster than forking the process.

> As for compression time, zipfile looks a lot a lot faster. Maybe the space saved with others is not worth the complexity and time to compress.

That's true, and is also even more widespread, even for non-technical users. However, that space saved in the others might be quite significant for big files.

dbeltrankyl:
> That's true, and is also even more widespread, even for non-technical users. However, that space saved in the others might be quite significant for big files.

If this is for compression in the remote, I think we don't need to prioritize the speed as it will be done in the side process which doesn't affect the overall perfomance

dbeltrankyl:
Also, there wouldn't be much to compress if we go job by job

kinow:
Remote zipfile, pure Python. I think it'll be simpler to start with this implementation then, and later, if needed, we go xz, multiple options to do on remote, local, etc. (i.e. start simpler, then complicated if needed).

LuiggiTenorioK:
> Remote zipfile, pure Python. I think it'll be simpler to start with this implementation then, and later, if needed, we go xz, multiple options to do on remote, local, etc. (i.e. start simpler, then complicated if needed).

And how about this PR? These are only local logs (`ASLOGS`), so standalone compatibility is not a factor if there are python libraries available. I think we can stay with lzma as it is or upgrade to zstandard. However, it will be inconsistent with what we will do in remote in another PR.

LuiggiTenorioK:
> Maybe we can re-work this PR to use zipfile or leave as-is if that works on multiple platforms? First we can wait for @mcastril to catch up on the previous 3-4 comments, and see if he agrees.
> 
> WDYT, Luiggi?

Maybe we can stay with `lzma` until Miguel reads the comments.

However, I was just thinking that `gzip` might be better than `zipfile` since we are just compressing single files. Here is the plot with `gzip` included:

compression_comparison

There is no significant improvement between zip and gzip for my test cases, but the community always mentions that gzip is better :sweat_smile:

kinow:
Kicking the slurm build...

kinow:
Ah, I commented in the other PR about documentation, but just saw that it's linking to this one. So I think the work is not done yet. Ignore my other message, @LuiggiTenorioK , @dbeltrankyl :+1: :bow:

Issue #1157, opened 2023-11-20
[bug] Script templates with exit prevent tailer from being created


kinow:
Hello @dbeltrankyl and @mcastril 

### Autosubmit Version

Latest from `master`, but I recall seeing this issue some time ago. I remembered about this after code-reviewing a merge request by @ainagaya to DE workflow.

### Expid affected(If applicable)

NA

### Which task has issues? Where is the log(If applicable)

NA

### Summary

A template script with `exit 0`, `exit 1`, etc., will cause the final job script to contain an `exit N` before the tailer. Thus, the tailer is never reached and the experiment hangs.

### Steps to reproduce

1. Create a local project with a single job and a script similar to this one:

```bash
fn() {
echo "OI"
}

exit 0

```

### What is the current bug behavior?

The final script looks like this:

```bash
#!/bin/bash

###############################################################################
#                   LOCAL_SETUP a005 EXPERIMENT
###############################################################################
###################
# Autosubmit header
###################
locale_to_set=$(locale -a | grep ^C.)
if [ -z "$locale_to_set" ] ; then
    # locale installed...
    export LC_ALL=$locale_to_set
else
    # locale not installed...
    locale_to_set=$(locale -a | grep ^en_GB.utf8)
    if [ -z "$locale_to_set" ] ; then
        export LC_ALL=$locale_to_set
    else
        export LC_ALL=C
    fi 
fi

set -xuve
job_name_ptrn='/home/kinow/autosubmit/a005/tmp/LOG_a005/a005_LOCAL_SETUP'
echo $(date +%s) > ${job_name_ptrn}_STAT

################### 
# AS CHECKPOINT FUNCTION
###################
# Creates a new checkpoint file upon call based on the current numbers of calls to the function

AS_CHECKPOINT_CALLS=0
function as_checkpoint {
    AS_CHECKPOINT_CALLS=$((AS_CHECKPOINT_CALLS+1))
    touch ${job_name_ptrn}_CHECKPOINT_${AS_CHECKPOINT_CALLS}
}

###################
# Autosubmit job
###################

fn() {
echo "OI"
}

exit 0


###################
# Autosubmit tailer
###################
set -xuve
echo $(date +%s) >> ${job_name_ptrn}_STAT
touch ${job_name_ptrn}_COMPLETED
exit 0
```

[...]

PR #2550, opened 2025-08-13
[Enhanced] Delete wrapper code

### LINKED ISSUE #2522: Delete code for wrappers with local platform
This was implemented for regression tests, but now with Docker + slurm|ssh, we don't need that anymore. Furthermore, there are a few issues when using the local platform with wrappers, as the code is not the same that handles Slurm wrappers.

VindeeR:
Closes #2522

This change was focused on cleaning old and unnecessary code and adjust the tests to make use of the container

**Check List**

- [ ] Documentation updated.


kinow (commented):
I think I saw more code about local platform & wrappers in our code base... maybe in config parser... and somewhere else? We might have to search for `'local'` and `"local"`...

But this is looking fantastic. Thanks a lot for updating the tests! Bet that was fun :)

kinow:
Check if we need this or not: https://github.com/BSC-ES/autosubmit/blob/936290f9bcc9693fb723e2220067e6a341ff62e7/autosubmit/config/configcommon.py#L1589-L1590

kinow (commented):
Huh, lots of new errors in integration tests.

VindeeR:
> Huh, lots of new errors in integration tests.

I believe it's related to the signature of the functioni changed, I'm still trying to understand which exactly could be

autosubmit/job/job_packages.py
```diff
@@ -401,7 +405,8 @@ class JobPackageThread(JobPackageBase):
     """
     FILE_PREFIX = 'ASThread'
 
-    def __init__(self, jobs, dependency=None, jobs_resources=dict(),method='ASThread',configuration=None,wrapper_section="WRAPPERS", wrapper_info= {}):
+    def __init__(self, jobs, dependency=None, jobs_resources: dict = None, method='ASThread',
```
dbeltrankyl, line 408: Hello @VindeeR 

About the bug you commented through slack, I think there is the issue. 

jobs_resources should be initialized to as  `jobs_resources = dict()` not None

kinow:
I got these same errors locally, quite consistently.

```
 ERROR:paramiko.transport:Exception (client): Error reading SSH protocol banner
ERROR:paramiko.transport:Traceback (most recent call last):
ERROR:paramiko.transport:  File "/opt/hostedtoolcache/Python/3.9.23/x64/lib/python3.9/site-packages/paramiko/transport.py", line 2363, in _check_banner
ERROR:paramiko.transport:    buf = self.packetizer.readline(timeout)
ERROR:paramiko.transport:  File "/opt/hostedtoolcache/Python/3.9.23/x64/lib/python3.9/site-packages/paramiko/packet.py", line 395, in readline
ERROR:paramiko.transport:    buf += self._read_timeout(timeout)
ERROR:paramiko.transport:  File "/opt/hostedtoolcache/Python/3.9.23/x64/lib/python3.9/site-packages/paramiko/packet.py", line 665, in _read_timeout
ERROR:paramiko.transport:    raise EOFError()
ERROR:paramiko.transport:EOFError
ERROR:paramiko.transport:
ERROR:paramiko.transport:During handling of the above exception, another exception occurred:
ERROR:paramiko.transport:
ERROR:paramiko.transport:Traceback (most recent call last):
ERROR:paramiko.transport:  File "/opt/hostedtoolcache/Python/3.9.23/x64/lib/python3.9/site-packages/paramiko/transport.py", line 2179, in run
ERROR:paramiko.transport:    self._check_banner()
ERROR:paramiko.transport:  File "/opt/hostedtoolcache/Python/3.9.23/x64/lib/python3.9/site-packages/paramiko/transport.py", line 2367, in _check_banner
ERROR:paramiko.transport:    raise SSHException(
ERROR:paramiko.transport:paramiko.ssh_exception.SSHException: Error reading SSH protocol banner
```

You have to kick the build, and we need to investigate it later.

test/integration/platforms/test_run_slurm_check_db.py
```diff
@@ -252,7 +262,7 @@ def _check_files_recovered(as_conf, log_dir, expected_files) -> dict:
     for f in log_dir.glob('*'):
         files_check_list[f.name] = not any(
             str(f).endswith(f".{i}.err") or str(f).endswith(f".{i}.out") for i in range(retrials + 1))
-    stat_files = [str(f).split("_")[-1] for f in log_dir.glob('*') if "STAT" in str(f)]
+    stat_files = [str(f).split("_")[-1] for f in log_dir.parent.glob('*') if "STAT" in str(f)]
```
kinow, line 265: Took me a long time to find this :laughing: 

The previous test was made for the results of running wrappers with the local platform. So it was working as the STAT_0 file existed in the log_dir. However, with a slurm platform, the log_dir in the remote contains the file, and then the log recovery will download it to tmp, not to the tmp/log_<EXPID> :face_exhaling:

kinow:
Test integration broken. There are a few tests failing, many complaining that the submit & finish time do not agree (one is greater than the other :dizzy: ). Test Slurm broken on GH, but passing locally. Will have some work adjusting it, as now it's running with testcontainers, instead of using a GH container service. This should make the tests more reliable, and fix the issues we had with intermittent failures :crossed_fingers:

kinow:
Rebased.

kinow:
About the tests that are failing because the latest finish is greater than the previous submit, from what I could understand, it's happening because the job `_STAT_` files are not being updated with the finish time.

There's some code called by the process recovery that it writes total stats, and it tries to find the finish time in that `_STAT_` file, but when it's missing it gets the system current time.

So, what happens is that we have the correct submit and start time. And the jobs have already finished at some point in the past.

Then, more jobs are submitted, and again get correct submit and start (these jobs are retries).

In the meantime, at some point, the log recovery process will call that function, notice the missing `finish` time, and use the system time.

When that occurs, the real finish time is ignored, and the actual finish time for the first job will be greater than the submit time of the retry of the same job (because the log retrieval uses a different sleep/interval).

I believe the solution is to figure out what's causing the job stat file not to have the finish time. If this bug is confirmed, I think this is something that could be affecting other random bugs in AS, or bugs in the API & GUI. Will confirm once I figure out how to fix it.

kinow:
Ah, looks like it has been like this for a while, and by design.

image

This is an excerpt from the job cmd, directly from the Slurm container (@VindeeR , this container was a real deal, changed how we write tests :clap: ).

The last part in the tailer is responsible for writing the `_COMPLETED` file, and also for updating the `_STAT_${RETRY}` file with the finish time.

This means that any time a script finishes successfully, we get the correct `finish` time in the DB. And that every time a job fails to run we get the time that the log retrieval process ran and processed the stats file of the job (which can be micro/milli/seconds later, or even minutes or hours). Causing disparity between submit/finish in retries.

I think I can have a go at fixing it in Shell, and address #1157, but not sure if it won't break anything by fixing this :grimacing:

test/unit/test_job.py
```diff
@@ -2088,21 +2088,3 @@ def test_wrapper_job_cancel_failed_wrapper_job_error(autosubmit_config, mocker):
...
-    as_conf = autosubmit_config(_EXPID, {})
-    platform = mocker.MagicMock()
-    platform.get_pscall.return_value = pid_found
-    wrapper_job = WrapperJob(_EXPID, 1, 'WAITING', 0, [], '00:30', platform, as_conf, False)
-    wrapper_job.platform_name = 'local'
-    wrapper_job.processors = 2  # not serial
-
-    wrapper_job.cancel_failed_wrapper_job()
-
-    assert platform.send_command.called == pid_found
```
kinow, line 2108: Not needed anymore, as wrappers do not run on local anymore.

autosubmit/job/job_common.py
```diff
@@ -77,296 +76,6 @@ class bcolors:
...
-        else:
-            number = string_number
-        try:
-            number = float(number) * multiplier
-        except Exception as exp:
-            number = 0.0
-            pass
-    return number
-
-
```
kinow, line 369: Isolated the template snippets code to `a.job.template`.

autosubmit/job/job.py
```diff
@@ -2523,48 +2484,40 @@ def create_wrapped_script(self, as_conf, wrapper_tag='wrapped'):
...
-        :rtype: bool
+        :param as_conf: Autosubmit configuration.
+        :param show_logs: Whether to display logs or not.
+        :return: Returns ``True`` if the script is well-formed, otherwise returns ``False``.
         """
-
-        out = False
         parameters = self.update_parameters(as_conf, set_attributes=False)
         template_content, additional_templates = self.update_content(as_conf, parameters)
-        if template_content is not False:
```
kinow, line 2543: `template_content` is a `str`, never `False` (it may be falsey, maybe it should be ìf template_content`?)

PR #2569, opened 2025-08-22
Force yes in command prompts

LuiggiTenorioK:
Small PR to bypass confirmation prompts in `stop` and `pklfix` commands. Perfect to make it easier to integrate into users' automation scripts.

Based on what was missing for https://github.com/BSC-ES/autosubmit-api/pull/215

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -694,7 +694,7 @@ class Autosubmit:
             subparser.add_argument('-f', '--force', default=False, action='store_true',
                                    help='Forces to stop autosubmit process, equivalent to kill -9')
             group.add_argument('-a', '--all', default=False, action='store_true',
-                               help='Stop all current running autosubmit processes, will ask for confirmation')
+                               help='Stop all current running autosubmit processes, will ask for confirmation unless -y is used')
             group.add_argument('-fa', '--force_all', default=False, action='store_true',
                                help='Stop all current running autosubmit processes')
             subparser.add_argument(
@@ -4077,6 +4077,8 @@ class Autosubmit:
 
         :param expid: experiment identifier
         :type expid: str
+        :param force: force the operation without confirmation
+        :type force: bool
         :return:
         :rtype: 
         """
@@ -5874,6 +5876,8 @@ class Autosubmit:
         :type current_status: str
         :param status: status to change the active jobs to
         :type status: str
+        :param force_yes: force yes answer to prompts
+        :type force_yes: bool
         """
         from autosubmit.helpers.processes import process_id, retrieve_expids
         from autosubmit.job.job_utils import cancel_jobs
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -591,6 +591,12 @@ class Autosubmit:
             subparser = subparsers.add_parser(
                 'pklfix', description='restore the backup of your pkl')
             subparser.add_argument('expid', help='experiment identifier')
+            subparser.add_argument(
+                "-f",
+                "--force",
+                action="store_true",
+                help="force restore without confirmation",
+            )
 
             # Update Description
             subparser = subparsers.add_parser(
@@ -691,6 +697,13 @@ class Autosubmit:
                                help='Stop all current running autosubmit processes, will ask for confirmation')
             group.add_argument('-fa', '--force_all', default=False, action='store_true',
                                help='Stop all current running autosubmit processes')
+            subparser.add_argument(
+                "-y",
+                "--yes",
+                default=False,
+                action="store_true",
+                help="Automatically answer yes to prompts",
+            )
             subparser.add_argument('-c', '--cancel', action=CancelAction, default=False, nargs=0,
                                 help='Orders to the schedulers to stop active jobs.')
             subparser.add_argument('-fs', '--filter_status', type=str, default="SUBMITTED, QUEUING, RUNNING",
@@ -797,13 +810,13 @@ class Autosubmit:
         elif args.command == 'dbfix':
             return Autosubmit.database_fix(args.expid)
         elif args.command == 'pklfix':
-            return Autosubmit.pkl_fix(args.expid)
+            return Autosubmit.pkl_fix(args.expid, args.force)
         elif args.command == 'updatedescrip':
             return Autosubmit.update_description(args.expid, args.description)
         elif args.command == 'cat-log':
             return Autosubmit.cat_log(args.ID, args.file, args.mode, args.inspect)
         elif args.command == 'stop':
-            return Autosubmit.stop(args.expid, args.force, args.all, args.force_all, args.cancel, args.filter_status, args.target)
+            return Autosubmit.stop(args.expid, args.force, args.all, args.force_all, args.cancel, args.filter_status, args.target, args.yes)
 
     @staticmethod
     def _init_logs(args, console_level='INFO', log_level='DEBUG', expid='None'):
@@ -4058,7 +4071,7 @@ class Autosubmit:
         update_experiment_descrip_version(expid, version=Autosubmit.autosubmit_version)
 
     @staticmethod
-    def pkl_fix(expid):
+    def pkl_fix(expid, force: bool = False):
         """
         Tries to find a backup of the pkl file and restores it. Verifies that autosubmit is not running on this experiment.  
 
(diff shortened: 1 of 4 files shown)


test/integration/scripts/test_stop.py
```diff
@@ -0,0 +1,41 @@
...
+_EXPID = "t111"
+
+
+def test_autosubmit_stop_command_invocation(autosubmit_exp, mocker: MockerFixture):
+    autosubmit_exp(_EXPID, experiment_data={})
+
+    mocker.patch("sys.argv", ["autosubmit", "stop", "-y", _EXPID])
+
+    with patch("autosubmit.autosubmit.Autosubmit.stop") as mock_stop:
+        mock_stop.return_value = 0
```
kinow, line 35: This is a bug in the code. It should return bool or int (and in the future we must have a single return type for the script functions called from `main`). But this is mocking to return 0, and below asserting 0 is returned (which will never fail). We might as well just call it, I think, without asserting anything as we are sure of its behaviour (as we have the bug in our code :sweat: )

test/integration/scripts/test_stop.py
kinow, line 1: I think this test is not testing the `-y` flag. At least, running with coverage in pycharm does not highlight the line where you placed the `force_yes`. Coverage goes from script to autosubmit.py where it calls the stop. And if we change the signature of stop, adding parameters at the beginning, we would not only have to add a test for that, but also fix this test as it asserts an argument exists in a specific position.

Maybe we could test that `input` is called or not, based on the `-y` value?

```py
def test_autosubmit_stop_command_invocation(autosubmit_exp, mocker: MockerFixture):
    """Test that passing ``-y`` to ``autosubmit stop`` prevents the user prompt."""
    autosubmit_exp(_EXPID, experiment_data={})

    mocked_input = mocker.patch('autosubmit.autosubmit.input', return_value='yes')

    mocker.patch("sys.argv", ["autosubmit", "stop", _EXPID])
    main()
    # ``input`` called;
    mocked_input.assert_called_once()
    mocked_input.reset_mock()

    mocker.patch("sys.argv", ["autosubmit", "stop", "-y", _EXPID])
    main()
    # ``input`` not called.
    mocked_input.assert_not_called()
```

Not sure if we need to parametrize, as there are only two variations. But with this it should verify that the code works at least up to the call to `input` (or no call, when `-y` is given).

test/integration/scripts/test_pklfix.py
```diff
@@ -0,0 +1,80 @@
...
+    mocker.patch("sys.argv", ["autosubmit", "pklfix", "-f", _EXPID])
+
+    with patch("autosubmit.autosubmit.Autosubmit.pkl_fix") as mock_pklfix:
+        mock_pklfix.return_value = 0
+        assert 0 == main()
+
+        mock_pklfix.assert_called_once()
+
+        passed_args = mock_pklfix.call_args[0]
+        assert passed_args[1] is True
```
kinow, line 43: I think here we also could check if the user was prompted for input or not.

test/integration/scripts/test_pklfix.py
```diff
@@ -0,0 +1,80 @@
...
+):
+    exp = autosubmit_exp(_EXPID, experiment_data={})
+
+    as_conf = exp.as_conf
+
+    # Create empty pkl files
+    exp_path = os.path.join(as_conf.basic_config.LOCAL_ROOT_DIR, _EXPID)
+    pkl_folder_path = os.path.join(exp_path, "pkl")
+    current_pkl_path = os.path.join(pkl_folder_path, f"job_list_{_EXPID}.pkl")
+    backup_pkl_path = os.path.join(pkl_folder_path, f"job_list_{_EXPID}_backup.pkl")
```
kinow, line 58: We are slowly replacing `os.path` calls by `pathlib` in the code.

test/integration/scripts/test_pklfix.py
```diff
@@ -0,0 +1,80 @@
...
+
+    os.makedirs(pkl_folder_path, exist_ok=True)
```
kinow, line 60: This shouldn't be needed. `autosubmit_exp` calls `create` by default, unless you specify `create=False` after @VindeeR 's PR is merged (one of his PR's is adding it).

autosubmit/autosubmit.py
```diff
@@ -691,6 +697,13 @@ def parse_args() -> Tuple[int, Optional[argparse.Namespace]]:
                                help='Stop all current running autosubmit processes, will ask for confirmation')
             group.add_argument('-fa', '--force_all', default=False, action='store_true',
                                help='Stop all current running autosubmit processes')
+            subparser.add_argument(
+                "-y",
+                "--yes",
+                default=False,
+                action="store_true",
+                help="Automatically answer yes to prompts",
```
kinow, line 705: Hmmm, some help statements start in upper case, while others in lower case. Noted, will add an issue for that later.

autosubmit/autosubmit.py
```diff
@@ -4058,7 +4071,7 @@ def get_files(root_dir_, extensions,files=""):
         update_experiment_descrip_version(expid, version=Autosubmit.autosubmit_version)
 
     @staticmethod
-    def pkl_fix(expid):
+    def pkl_fix(expid, force: bool = False):
```
kinow, line 4074: Missing docstring (new IDE warning)

autosubmit/autosubmit.py
```diff
@@ -5886,7 +5900,7 @@ def stop(expids: str, force=False, all_expids=False, force_all=False, cancel=Fal
             expids = [
                 expid
                 for expid in expids
-                if input(f"Confirm stopping the experiment: {expid} (y/n)[enter=y]? ").lower() in truthy_values
+                if force_yes or input(f"Confirm stopping the experiment: {expid} (y/n)[enter=y]? ").lower() in truthy_values
```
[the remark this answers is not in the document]
LuiggiTenorioK: Actually, you cannot use `--all` + `--force-all` because they are declared in the same exclusive group. Maybe this is an issue that should be addressed in a different PR.

test/integration/scripts/test_pklfix.py
LuiggiTenorioK: `test_pklfix_bypass_prompt_confirmation` indirectly tests that when `force=False`

test/integration/scripts/test_stop.py
```diff
@@ -0,0 +1,68 @@
...
+
+    mocker.patch("sys.argv", ["autosubmit", "stop", "-y", _EXPID])
+
+    with patch("autosubmit.autosubmit.Autosubmit.stop") as mock_stop:
+        main()
+
+        mock_stop.assert_called_once()
+
+        passed_args = mock_stop.call_args[0]
+        assert passed_args[7] is True
```
kinow, line 44: This is still brittle as if we add an arg before we will have to adjust this test.

From what I understood, this test only checks that the `-y` is parsed by argparse and results in the stop functon receiving the force yes. That's also covered by the test below, so I think we can remove this test entirely. Coverage showing it's not touching the changes in `stop()`

image

Or is it testing another part of the code not covered by `test_stop_bypass_prompt_confirmation`

test/integration/scripts/test_pklfix.py
```diff
@@ -0,0 +1,84 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test for the autosubmit pklfix command"""
+
+from pathlib import Path
+from unittest.mock import patch
```
kinow, line 21: Can you replace this by the pytest fixture. Most of the code is already using that one, so it makes it more uniform and avoids any issues in case `pytest` changes anything related to mockers/mocking.

test/integration/scripts/test_pklfix.py
kinow: `test_autosubmit_pklfix_command_invocation` appears to be testing just that the arg in position `1` is `True`. That's covered indirectly by `test_pklfix_bypass_prompt_confirmation` which is also less brittle, so I think this test can be deleted.

test/integration/scripts/test_stop.py
```diff
@@ -0,0 +1,68 @@
...
+
+    with patch("builtins.input") as mock_input:
+        mock_input.return_value = "no"
+
+        assert main() is None
+
+        if force_yes:
+            mock_input.assert_not_called()
+        else:
+            mock_input.assert_called_once()
```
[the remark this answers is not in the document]
kinow: Realized you are checking called once, hadn't noticed that, thought it was called, sorry.

test/integration/scripts/test_stop.py
```diff
@@ -0,0 +1,46 @@
...
+    # Mock command line arguments
+    passed_args = ["autosubmit", "stop"] + (["-y"] if force_yes else []) + [_EXPID]
+    mocker.patch("sys.argv", passed_args)
+
+    mock_input = mocker.patch("builtins.input")
+    mock_input.return_value = "no"
+
+    assert main() is None
+
+    assert mock_input.call_count == (0 if force_yes else 1)
```
kinow, line 46: :muscle: :clap: hadn't thought about doing it this way!

kinow (approved):
Thank you, @LuiggiTenorioK !!!

kinow:
Ah! @LuiggiTenorioK we need a changelog entry for this too, please :bow:

LuiggiTenorioK:
> Ah! @LuiggiTenorioK we need a changelog entry for this too, please 🙇

I forgot :sweat_smile:  Added!

PR #2575, opened 2025-08-26
Re-write test_send_file without mocks

### LINKED ISSUE #2470: [test] send_file integration test is flaky
### Description

From today's meeting, @VindeeR kicked the build of #2451 as send_data had failed. I remember seeing that test fail somewhere too. It could be due to conflict in the port allocated, or something else.

* Autosubmit version: master, 4.1.15+

### Reproducible Example

NA

### Expected Behaviour

Test behavior is deterministic.

kinow:
Closes #2470 

**Check List**

- [ ] Documentation updated.


autosubmit/platforms/paramiko_platform.py
```diff
@@ -213,10 +210,8 @@ def restore_connection(self, as_conf: 'AutosubmitConfig', log_recovery_process:
                     raise AutosubmitCritical(f"First connection to {self.host} is failed, check host configuration"
                                              f" or try another login node ", 7050, str(e))
             while self.connected is False and retry < retries:
-                try:
+                with suppress(Exception):
```
kinow, line 213: This is changing and being tested on #2428

autosubmit/platforms/paramiko_platform.py
```diff
@@ -480,10 +469,8 @@ def send_file(self, filename, check=True):
         except IOError as e:
             raise AutosubmitError(f'Can not send file {os.path.join(self.tmp_path, filename)} to '
                                   f'{os.path.join(self.get_files_path(), filename)}', 6004, str(e))
-        except BaseException as e:
-            raise AutosubmitError(
-                'Send file failed. Connection seems to no be active', 6004)
-
+        except Exception as e:
+            raise AutosubmitError(f'Failed to send file, the SSH connection may be inactive: {str(e)}', 6004)
```
kinow, line 473: Just moved to same line, rewrote the message.

autosubmit/platforms/psplatform.py
```diff
@@ -57,16 +42,23 @@ def __init__(self, expid, name, config):
...
 
+    def get_checkAlljobs_cmd(self, jobs_id):
+        pass
+
+    def parse_Alljobs_output(self, output, job_id):
+        pass
+
+    def parse_queue_reason(self, output, job_id):
+        pass
+
```
kinow, line 53: Moved functions to after the constructor -- can cover it with some tests, but looks silly (and we might not need these functions in the future depending on how we re-design it -- with abc/use parent function/etc.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1497,20 +1482,35 @@ def closeConnection(self):
...
                 self._ftpChannel.rmdir(path)
                 self._ftpChannel.mkdir(path)
                 self._ftpChannel.rmdir(path)
             return True
         except Exception as e:
-            return False
-    
+            Log.warning(f'Failed checking remote permissions (2): {str(e)}')
+        return False
+
```
kinow, line 1513: Added logging (doesn't hurt, may be useful to us)

kinow:
The code changed is actually really silly. Logging, moving things to same line, or changing the text of exceptions. Some coverage warnings are in lines moved to another location in the same file (after constructor). So I think the coverage report can be ignored for those changes. The only more serious change, which is replacing some code with `suppress` is going to happen in #2428 too.

kinow:
@VindeeR I was going to spend more time on this, but I think we can actually review it now. I'll trigger the build one more time, and if it's green, then can you review it, please?

It's fixing (hopefully for good) the `test_send_file` that was failing intermittently. See comments above about what changed and coverage.

Thanks!

autosubmit/platforms/paramiko_submitter.py
```diff
@@ -71,8 +70,8 @@ def load_platforms(self, asconf, retries=5, auth_password = None, local_auth_pas
         :rtype: dict
         """
         exp_data = asconf.experiment_data
-        raise_message=""
-        platforms_used = list()
+        raise_message = ""
+        platforms_used = []
```
VindeeR, line 74: is there any difference between the two ways? is there one considered standard? from the explanation I got online functionality wise both works the same way but the first invokes the object to create the list and the second create the list immediately after.

Do you think it would be better to standardize everything like this from now on?

kinow: I use the `list` call when I need to make something a list (like some view or iterator, like those returned by `itertools` functions). And I use `[]` when I need an empty list as it's shorter (a lot in Python is about making it shorter, especially with indentation in the language).

Following the zen of python, https://pep20.org/#obvious, you get `[]` that does one thing, and it's easy/fast for others to grasp what you intend. Whereas `list()` is normally used to convert something to a list... or create a list. If you separate how you use these, then it's easier for others/you to read it -- in theory :shrug:

kinow:
Thanks @VindeeR !

@LuiggiTenorioK looks like the test that failed is the one we added this morning:

```bash
FAILED test/integration/scripts/test_stop.py::test_stop_bypass_prompt_confirmation[True] - assert 7000 is None
 +  where 7000 = main()
===== 1 failed, 136 passed, 20 skipped, 637 warnings in 158.29s (0:02:38) ======
```

kinow:
I've rebased, and it should fix the build :crossed_fingers: but we might have to look at this fail later on.

PR #2580, opened 2025-08-28
Squashed Base Postgres with SQLAlchemy and database_backend

dbeltrankyl:
**Check List**

- [x] Documentation updated.


dbeltrankyl:
This is the base implementation of Postgres without the job list #2187 

As commented in the meeting the idea is to merge this into master

dbeltrankyl:
It is the first time I've tried to revert to a commit and create a new branch 

Since the tests aren't passing, I guess it is a side effect of the rebases done in the #2187, so the branch has to be revised to be compatible with the master

dbeltrankyl:
@kinow So now the question is whether it's worth spending time making this compatible with master, which might cause conflicts in #2187. Or should I just handle the conflicts as I was doing? Maybe the latter is better...

kinow:
> @kinow So now the question is whether it's worth spending time making this compatible with master, which might cause conflicts in https://github.com/BSC-ES/autosubmit/pull/2187. Or should I just handle the conflicts as I was doing? Maybe the latter is better...

I think we could spend next week testing the branch with sqlalchemy, and perhaps go over some 4.1.16 issues while tests run? Then try to have it tested with some simple experiments in DestinE, and if nothing strange happens, we merge it to master, and then rebase/fix conflicts in the PR with joblist?

dbeltrankyl:
> I think we could spend next week testing the branch with sqlalchemy, and perhaps go over some 4.1.16 issues while tests run? Then try to have it tested with some simple experiments in DestinE, and if nothing strange happens, we merge it to master, and then rebase/fix conflicts in the PR with joblist?

Sorry @kinow, forgot to answer this one, citing also your other comment

>Great @dbeltrankyl ! Just to confirm, the plan is that this is the PR with postgres, to be reviewed, tested and merged for 4.1.16. Then another PR will have the joblist changes right?

I have not taken any more action on this yet. 

> and then rebase/fix conflicts in the PR with joblist?

In the job list one, we don't have issues right now ( it is updated)

My point is:

1) This has a lot of issues because the rebases messed up this single commit
2) The rebase was done with all the commits together, so the conflicts were resolved, having the other commits in mind.
3) I think it will take more time to resolve conflicts on this commit than possible conflicts in the other branch.

4) Joblist one is not too far from being finished
5) Luiggi is already using the other branch for the API, so I want to finish it ASAP and don't use time on testing this commit alone


So, taking into account that this was originally intended to alleviate the conflicts on the other branch, I think it doesn't make much sense to use time to resolve stuff that was already solved in the other branch

kinow:
Hi @dbeltrankyl I'm not following what's happening in the other PR then. Does that PR #2187 contain the joblist pkl -> db, or just sqlalchemy + postgres support?

If the former, I believe we'll have to wait for 4.2.0 before merging that, even if it's being tested now. If the latter, I think we could review it and merge it to master to include in 4.1.16 as it would be just sqlalchemy using sqlite + pickle file... at least that's what I thought we were planning to do to reduce the number of rebases & conflicts.

Sorry if I'm not following it correctly.

dbeltrankyl:
Hello @kinow 

> Does that PR https://github.com/BSC-ES/autosubmit/pull/2187 contain the joblist pkl -> db,

Yes

> f the former, I believe we'll have to wait for 4.2.0 before merging that,
Agree

> at least that's what I thought we were planning to do to reduce the number of rebases & conflicts.

Yes, that was the intention 

However, when I tried (as you can see in this branch), I had all the tests failing. 

I checked the cause of it,  I noticed that all the previous rebases done were also affecting this commit, not only the HEA,D which dificults the task of merging in 4.1.16. Since I did some modifications to the code in the HEAD of that branch to resolve conflicts which I would have to do again but differently


So my conclusion is https://github.com/BSC-ES/autosubmit/pull/2580#issuecomment-3248009641, which I think is better to merge everything for 4.2. 

I can deal with the conflicts as I was doing before

dbeltrankyl:
yes

PR #2581, opened 2025-08-29
WIP, not working

### LINKED ISSUE #2574: KeyError on HPCOFFLINE_DN_PLATFORM
### Description

Hi, I'm trying to run the Testing Suite for v5.3.0 of the workflow which includes a new Autosubmit variable HPCOFFLINE_DN_PLATFORM.

* Autosubmit version: 4.1.15 but also occurs with 4.1.14
* Machine/VM/environment name (if applicable): LUMI and MN5
* Experiment ID (if applicable): t061, t063, and t06a
  * /appl/AS/AUTOSUBMIT_DATA/t063/tmp/ASLOGS/20250826_120439_run_err.log

### Reproducible Example

Look at any of the experiments listed above (t061, t063, t06a).

### Expected Behaviour

We expected this variable to be substituted with the correct value.

kinow:
Closes #2574 

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -40,9 +40,6 @@ from pathlib import Path
 from time import sleep
 from typing import Dict, Set, Tuple, Union, Any, List, Optional
 
-from autosubmit.config.basicconfig import BasicConfig
-from autosubmit.config.configcommon import AutosubmitConfig
-from autosubmit.config.yamlparser import YAMLParserFactory
 from bscearth.utils.date import date2str
 from portalocker import Lock
 from portalocker.exceptions import BaseLockException
@@ -51,6 +48,9 @@ from ruamel.yaml import YAML
 
 import autosubmit.helpers.autosubmit_helper as AutosubmitHelper
 import autosubmit.statistics.utils as StatisticsUtils
+from autosubmit.config.basicconfig import BasicConfig
+from autosubmit.config.configcommon import AutosubmitConfig
+from autosubmit.config.yamlparser import YAMLParserFactory
 from autosubmit.database.db_common import create_db
 from autosubmit.database.db_common import delete_experiment, get_experiment_descrip
 from autosubmit.database.db_common import get_autosubmit_version, check_experiment_exists
@@ -74,13 +74,12 @@ from autosubmit.job.job_list_persistence import JobListPersistencePkl
 from autosubmit.job.job_package_persistence import JobPackagePersistence
 from autosubmit.job.job_packager import JobPackager
 from autosubmit.job.job_utils import SubJob, SubJobManager
+from autosubmit.log.log import Log, AutosubmitError, AutosubmitCritical
 from autosubmit.migrate.migrate import Migrate
 from autosubmit.notifications.mail_notifier import MailNotifier
 from autosubmit.notifications.notifier import Notifier
 from autosubmit.platforms.paramiko_submitter import ParamikoSubmitter
 from autosubmit.platforms.platform import Platform
-from autosubmit.platforms.submitter import Submitter
-from autosubmit.log.log import Log, AutosubmitError, AutosubmitCritical
 
 dialog = None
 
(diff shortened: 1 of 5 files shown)


kinow:
(To test it, I was `autosubmit create -cw -np a00g && autosubmit inspect a00g`, then verifying `experiment_data.yml` and one Shell script.)

autosubmit/job/job.py
```diff
@@ -1625,12 +1614,8 @@ def update_current_parameters(self, as_conf: AutosubmitConfig, parameters: dict)
 
         return parameters
 
-    def process_scheduler_parameters(self, job_platform, chunk):
```
kinow: Not used.

autosubmit/job/job.py
```diff
@@ -1512,13 +1508,6 @@ def update_status(self, as_conf: AutosubmitConfig, failed_file: bool = False) ->
 
         return self.status
 
-    @staticmethod
-    def _get_submitter(as_conf):
```
kinow: Not used. It was used in a test, but I simply created a `ParamikoSubmitter` directly instead.

autosubmit/config/configcommon.py
```diff
@@ -2889,22 +2889,6 @@ def get_parser(parser_factory, file_path):
                     "{}\n This file and the correctness of its content are necessary.".format(str(exp)))
         return parser
 
-    @staticmethod
-    def parse_placeholders(content, parameters) -> str:
```
kinow: Never used.

kinow:
@dbeltrankyl fixed, but now a regression test is failing. I debugged it a bit more, and looks like the regression data had a placeholder, but now with this change it needs to be updated (as AS will update it too).

image

This was supposed to fix the test, but actually it also loads my user `~/platforms.yml` file because of the configuration copied from the ClimateDT workflow :sweat_smile: 

image

I will update the regression test to use the expanded/substituted value so the test passes, and remove the user home  platforms.yml file so the test doesn't load it.

autosubmit/autosubmit.py
dbeltrankyl: Does autosubmit run calls this function? I thought this was only called by the -cw flag or inspect

dbeltrankyl (commented):
Hello @kinow 

Thanks for the work! 

I have two concerns. Can you look at them before I click approve? thanks!

kinow:
Thanks for reviewing this, @dbeltrankyl . I will start the CES report & timesheet, so I will probably come back to this issue next week. Cheers

autosubmit/autosubmit.py
kinow: Ah, you are right. It's because the integration test is calling wrappers too. Let me add another test without it.

autosubmit/autosubmit.py
```diff
@@ -1849,8 +1848,11 @@ def generate_scripts_andor_wrappers(as_conf, job_list, jobs_filtered, packages_p
                 continue
             wrapper_jobs[wrapper_section] = as_conf.get_wrapper_jobs(wrapper_data)
         Log.warning("Aux Job_list was generated successfully")
+
```
kinow: @dbeltrankyl code completely removed here. It was not necessary, just a left-over from my first try at fixing it. Thanks for spotting that!

dbeltrankyl (approved):
Everything seems fine! Thanks! One comment about the changelog and that's it you can merge after changing it :)

For the future, regarding the review procedure, I have one suggestion: 

I think that once another developer performs a first review, the commits shouldn't be rebased with the already-reviewed commits. 

In that way, after the owner of the PR addresses the feedback, the reviewer will only have to look at the following commits and not the whole code again 

What do you think? 

Edit: Then, after the approval, we can rebase everything into a single commit

So it would look like

rebased # 1 commit
review -> feedback
rebased 2 # 2 commits
review -> feedback
rebase 3 ... # 3 commits
approved
merge # 1 commit

PR #2600, opened 2025-09-16
[Documentation] Generate images automatically in docs build, including them in our Sphinx docs

### LINKED ISSUE #1160: [enhancement][documentation] The examples in our docs must be self-contained

### LINKED ISSUE #2069: Mistake in documentation generated
https://autosubmit.readthedocs.io/en/master/userguide/modifying_workflow/index.html

As I was talking to @kinow we found a problem at the documentation

"How to restart the experiment" > "How to recover an experiment" 
Can be found the Sphinx syntax "::" at the page

VindeeR:
Closes #2069
Closes #1160 

Improving the quality of the documentation generating it in an automated way ensuring images to always be up to date with instruction from provided

It uses our `runcmd` to execute `autosubmit` commands (create, run, etc.), placing images in a folder. Then, it moves them to the `_static`, and uses the `figure` directive to produce the `<figure>` used in the final docs. But, instead of injecting the `figure`, we get its HTML output, and send a `raw` directive.

This way, we control when the image will be used. Otherwise, `figure` would fail, saying that the image does not exist, as the images exist in the build, and not in the source directory, failing the `sphinx-build` execution.

An example of another directive that does something similar, used as reference: https://github.com/matplotlib/matplotlib/blob/799f26066b5b110fe7eee841ad979aee90238234/lib/matplotlib/sphinxext/plot_directive.py (although they do something more complicated, rendering a Jinja2 template, and then injecting values into the Sphinx state machine…

**Check List**

- [x] Documentation updated.


kinow (approved):
:+1: thanks!!!

PR #2618, opened 2025-09-30
Quickfix Log.set_file infinite loop

### LINKED ISSUE #2513: Infinite loop in Log.set_file
### Description

I'm writing tests for our logging module in #2475 . Doing so, I noticed one part where the tests were a bit strange. Apparently, we have an infinite loop if we fail to `set_file`, as the `retries` variables is never incremented, so the check in the `while` will keep forever the file it's searching for is not created, independent of the number of retries.

* Autosubmit version: 4.1.15+, master

### Reproducible Example

Image

### Expected Behaviour

I think it's supposed to try until the file exists, or until the number of retries runs out.

LuiggiTenorioK:
Closes #2513

I did see this error before when I did not have enough permissions to create a file. The solution is quite trivial.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ the filter the jobs. Not using any value for `-fp` still returns all jobs.
 - Fix PS platform to work with the local machine #2374
 - Fixed a `ZeroDivisionError` when using RO-Crate or `stats`, and also an issue
   where the message said `None` could not be iterable. #2389
+- Fixed a bug that produces an infinite loop when it is not possible to create log files #2618
 
 **Enhancements:**
 


autosubmit/log/log.py
```diff
@@ -201,7 +201,13 @@ def get_logger(name="Autosubmit") -> None:
         logging.getLogger(name)
 
     @staticmethod
-    def set_file(file_path: str, type='out', level="WARNING") -> None:
+    def set_file(
+        file_path: str,
+        type="out",
+        level="WARNING",
+        max_retries: int = 3,
+        timeout: int = 5,
```
kinow, line 209: missing docstring 😬

autosubmit/log/log.py
```diff
@@ -271,7 +275,12 @@ def set_file(file_path: str, type='out', level="WARNING") -> None:
                     status_file_handler.addFilter(custom_filter)
                     Log.log.addHandler(status_file_handler)
                 os.chmod(file_path, 509)
-            except Exception as e:  # retry again
+            except Exception:  # retry again
+                retries += 1
+                if retries >= max_retries:
+                    raise AutosubmitCritical(
+                        message=f"Could not create log file {file_path} after {max_retries} attempts"
```
kinow, line 282: Maybe we could add the exception to the message here at least, `message=f"Could not.... attempts: {str(e)}"?

test/unit/test_log.py
```diff
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
+import pytest
+import pytest_mock
```
kinow, line 19:
```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from pytest_mock import MockFixture
```

Then in the test signature `mocker : 'MockFixture'`. This way we only keep imports of code used in the code (avoids actually importing the code just for types which may slow down code (not important in a test file to be honest) and avoids issues with circular dependencies.

kinow (commented):
Looking neat! A few comments, but I think after those it should be pretty much ready.

kinow (approved):
Looks good to em! @LuiggiTenorioK can you add a changelog and merge, please? Thanks!

PR #2654, opened 2025-10-15
[Enhancement] Mypy - annotation fix

VindeeR:
Mypy fix that tries to avoid making use of annotations

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -249,7 +249,7 @@ class Job(object):
         self.check = 'true'
         self.check_warnings = False
         self.packed = False
-        self.hold = False  # type: bool
+        self.hold: bool = False
         self.distance_weight = 0
         self.level = 0
         self._export = "none"
@@ -310,7 +310,7 @@ class Job(object):
                 self.status
         self.validate_template = False
 
-    def clean_attributes(self):
+    def clean_attributes(self) -> None:
         if self.status == Status.FAILED and self.fail_count >= self.retrials:
             return None
         self.rerun_only = False
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -248,7 +248,7 @@ class Job(object):
         self.check = 'true'
         self.check_warnings = False
         self.packed = False
-        self.hold = False  # type: bool
+        self.hold: bool = False
         self.distance_weight = 0
         self.level = 0
         self._export = "none"
@@ -284,9 +284,9 @@ class Job(object):
         self._notify_on = None
         # The three variables under this message are related to the #PR2918 that is a development
         # focused on adding the key information for computing the simulated years for the CPMIPS metrics.
-        self._cpmip_thresholds = {} 
-        self._chunk_size = None 
-        self._chunk_size_unit = None 
+        self._cpmip_thresholds = {}
+        self._chunk_size = None
+        self._chunk_size_unit = None
         self._processors_per_node = None
         self.ec_queue = None
         self.platform_name = None
@@ -309,7 +309,7 @@ class Job(object):
                 self.status
         self.validate_template = False
 
-    def clean_attributes(self):
+    def clean_attributes(self) -> None:
         if self.status == Status.FAILED and self.fail_count >= self.retrials:
             return None
         self.rerun_only = False
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -682,7 +682,7 @@ class JobList(object):
             filters = [{}]
         return filters
 
-    def _check_dates(self, relationships: Dict, current_job: Job) -> {}:
+    def _check_dates(self, relationships: Dict, current_job: Job) -> dict:
         """Check if the current_job_value is included in the filter_from and retrieve filter_to value
 
         :param relationships: Remaining filters to apply.
@@ -741,7 +741,7 @@ class JobList(object):
         filters_to_apply = self._unify_to_filters(filters_to_apply)
         return filters_to_apply
 
-    def _check_chunks(self, relationships: Dict, current_job: Job) -> {}:
+    def _check_chunks(self, relationships: Dict, current_job: Job) -> dict:
         """Check if the current_job_value is included in the filter_from and retrieve filter_to value
 
         :param relationships: Remaining filters to apply.
@@ -774,7 +774,7 @@ class JobList(object):
         filters_to_apply = self._unify_to_filters(filters_to_apply, current_job.splits)
         return filters_to_apply
 
-    def _unify_to_filter(self, unified_filter, filter_to, filter_type, splits=None) -> {}:
+    def _unify_to_filter(self, unified_filter, filter_to, filter_type, splits=None) -> dict:
         """Unify filter_to filters into a single dictionary
 
         :param unified_filter: Single dictionary with all filters_to
@@ -2592,7 +2592,7 @@ class JobList(object):
             job.platform.add_job_to_log_recover(job)
         return log_recovered
 
-    def check_if_log_is_recovered(self, job: Job) -> Path:
+    def check_if_log_is_recovered(self, job: Job) -> Optional[Path]:
         """Check if the log is recovered.
 
         Conditions:
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -683,7 +683,7 @@ class JobList(object):
             filters = [{}]
         return filters
 
-    def _check_dates(self, relationships: Dict, current_job: Job) -> {}:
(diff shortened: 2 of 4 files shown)


autosubmit/job/job.py
```diff
@@ -2698,6 +2644,16 @@ def recover_last_ready_date(self) -> None:
...
+decorator_list: list = []
+list_decorators = ['x11', 'x11_options', 'tasktype', 'jobname', 'script', 'fail_count', 'retrials', 'checkpoint',
+                    'sdate', 'member', 'chunk', 'split', 'delay', 'wallclock', 'hyperthreading', 'nodes', 'numthreads',
+                    'threads', 'cpus_per_task', 'numtask', 'tasks', 'tasks_per_node', 'scratch_free_space', 'memory',
+                    'memory_per_task', 'frequency', 'synchronize', 'dependencies', 'delay_retrials', 'packed', 'export',
+                    'custom_directives', 'splits', 'notify_on', 'status', 'log_recovered', "current_queue", 'numproc',
+                    'processors', 'processors_per_node']
+for decorator_to_import in list_decorators:
+    decorator = Job(name=decorator_to_import)
+    decorator_list.append(decorator.__doc__)
```
kinow: Note to self: check why this was added.

autosubmit/job/job_list.py
```diff
@@ -744,7 +744,7 @@ def _check_splits(self, relationships, current_job):
         filters_to_apply = self._unify_to_filters(filters_to_apply, current_job.splits)
         return filters_to_apply
 
-    def _unify_to_filter(self, unified_filter, filter_to, filter_type, splits=None) -> {}:
+    def _unify_to_filter(self, unified_filter, filter_to, filter_type, splits=None) -> dict:
```
kinow: 👏 👌

autosubmit/job/job_list.py
```diff
@@ -2732,8 +2727,6 @@ def update_list(self, as_conf: AutosubmitConfig, store_change: bool = True,
                                 Log.debug(f"Setting job: {job.name} status to: READY "
                                           "(conditional jobs are completed/failed)...")
                                 break
-                            if as_conf.get_remote_dependencies() == "true":
-                                all_parents_completed.append(job.name)
```
kinow: Hmmm, it's looking like we never have this get_remote_dependencies...?

autosubmit/job/job_list.py
```diff
@@ -3049,7 +2977,7 @@ def print_with_status(self, status_change: Optional[dict[Any, Any]] = None, noco
         for job in all_jobs:
             if len(job.parents) == 0:
                 roots.append(job)
-        visited = list()
+        visited: list = []
```
kinow: when doing it like this, it can be useful to specify the list element (e.g. `list[str]`) when we know it. Will check what's the type here later.

kinow (commented):
There are a few things I got curious about in this change, so I added a few notes for myself, and I will review them in the morning with calm and ☕ 

But it's really good to have PRs like this adding types :) 🙏 thank you a lot Erick!!!

autosubmit/job/job.py
```diff
@@ -130,38 +130,29 @@ class Job(object):
...
+        'stat_file', '_status', 'prev_status', 'new_status', 'priority', '_parents', '_children', '_fail_count',
+        'expid', 'parameters', '_tmp_path', '_log_path', 'check', 'check_warnings', '_packed', 'hold',
+        'distance_weight', 'level', '_export', '_dependencies', 'running', 'start_time', 'ext_header_path',
+        'ext_tailer_path', 'edge_info', 'total_jobs', 'max_waiting_jobs', 'exclusive', '_retrials',
+        'current_checkpoint_step', 'max_checkpoint_step', 'reservation', 'delete_when_edgeless', 'het', 'updated_log',
+        'submit_time_timestamp', 'start_time_timestamp', 'finish_time_timestamp', '_script', '_log_recovery_retries',
+        'ready_date', 'wrapper_name', 'is_wrapper', '_wallclock_in_seconds', '_notify_on', '_processors_per_node',
+        'ec_queue', 'platform_name', '_serial_platform', 'submitter', '_shape', '_x11', '_x11_options',
+        '_hyperthreading', '_scratch_free_space', '_delay_retrials', '_custom_directives', '_log_recovered',
+        'packed_during_building', 'workflow_commit'
```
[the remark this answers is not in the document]
VindeeR: I just formatted to occupy the max space in a line and save lines, no real changes here

kinow:
Just reviewing an AS API PR for Luiggi, then I will come back to review this one.

autosubmit/job/job.py
```diff
@@ -222,74 +213,73 @@ def __init__(self, name=None, job_id=None, status=None, priority=None, loaded_da
...
-        self.current_checkpoint_step = 0
-        self.max_checkpoint_step = 0
-        self.reservation = ""
+        self.current_checkpoint_step: int = 0
+        self.max_checkpoint_step: Optional[int] = 0
+        self.reservation: Optional[str] = ""
         self.delete_when_edgeless = False
         # hetjobs
-        self.het = None
+        self.het: dict[Any, Any]
```
kinow: I think while harmless, here we are setting the variable as `None` but declaring as `dict`? Maybe it should be initialized/

autosubmit/job/job.py
```diff
@@ -1429,13 +1383,12 @@ def update_status(self, as_conf: AutosubmitConfig, failed_file: bool = False) ->
...
         """
-        previous_status = self.status
-        self.prev_status = previous_status
+        self.prev_status = self.status
         new_status = self.new_status
         if new_status == Status.COMPLETED:
             Log.debug(
                 "{0} job seems to have completed: checking...".format(self.name))
-            self._platform.get_completed_files(self.name, wrapper_failed=self.packed)
+            self._platform.get_completed_files(self.name, wrapper_failed=self.packed) # type: ignore
```
kinow: Two spaces before a `#`, or it adds a lint warning.

autosubmit/job/job.py
```diff
@@ -3043,8 +2999,8 @@ def _check_running_jobs(self) -> None:
             wait = 2
             retries = 5
             over_wallclock = False
-            content = ''
-            while content == '' and retries > 0:
+            content: list[str] = list()
```
kinow: `[]` instead of `list()` 😬

autosubmit/job/job.py
kinow: And the `self._platform` shouldn't be `None`. Maybe define the `self._platform` as `Optional['Platform']` and then set the ignore here to the error complained by mypy

In the output of mypy, it shows the category of the type error, which using `Optional` results in `union-attr` (the union type `None` doesn't have the function `.get_completed_files`, but we know it's not `None` at this point).

So we can reduce the scope of the ignore: `  # ignore[union-attr]`. That should raise other issues if we introduce any type bugs, but will silence mypy for this line.

autosubmit/job/job.py
```diff
@@ -2339,7 +2286,7 @@ def get_wrapped_content(self, as_conf: AutosubmitConfig, parameters: dict):
     def _get_paramiko_template(self, snippet: 'TemplateSnippet', template, parameters) -> str:
         current_platform = self._platform
         return ''.join([
-            snippet.as_header(current_platform.get_header(self, parameters), self.executable),
+            snippet.as_header(current_platform.get_header(self, parameters), self.executable), # type: ignore
```
kinow: Two spaces before `#`.

autosubmit/job/job.py
```diff
@@ -2808,7 +2764,7 @@ def __init__(
         # divide jobs in dictionary by state?
         self.wallclock = total_wallclock  # Now it is reloaded after a run -> stop -> run
         self.running_jobs_start: OrderedDict = OrderedDict()
-        self._platform: 'ParamikoPlatform' = platform
+        self._platform: 'ParamikoPlatform' = platform  #type: ignore
```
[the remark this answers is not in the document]
kinow: And this can be removed if in the parent `Job` we set the type as `Optional['Platform']`.

autosubmit/job/job.py
kinow: Hmm, but where is it used after you create the `decorator_list`? I think the replacement for the decorator is more or less following this idea. But you still have the `autosubmit_parameters`. That can be removed too, I think.

This part I think still needs some further work, but maybe we can work together to finish it for 4.1.16, or bump it to 4.1.17 (I think you have nearly 80% of the work done here, which is why I think there might still be time for 4.1.16).

VindeeR:
I just realized that the changes thought to be committed on this PR were already implemented so I'll be closing it

Issue #2583, opened 2025-08-29
AttributeError: 'NoneType' object has no attribute 'set'


kinow:
### Description

This issue happened a long time ago for me while working on `conftest.py` and some integration tests.

Some weeks ago, I noticed an experiment from @youcheng-csc had the same in its logs (I'll chase that and update this issue later). Then, I had the same problem with the following test (it's missing to set the AS command, see other integration tests):

```py
def test_run_bug_save_wrapper_crashes(as_exp):
    exit_code = as_exp.autosubmit.run_experiment(expid=_EXPID)
    _assert_exit_code(final_status, exit_code)
```

Then, @VindeeR reported seeing this same problem in GH Actions when he added a commit changing `test_run_command_integration`.

This issue is not happening a lot, but it's important we understand why it's happening, and prevent before it stops an operational experiment (the experiment of Cheng was suffering from #2474 , so maybe that's why we didn't hear about it yet).

* Autosubmit version: master, 4.1.15
* Machine/VM/environment name (if applicable): DestinE, GH Actions
* Experiment ID (if applicable):
* Experiment tasks & log path if applicable):
  * e.g. `<EXPID>_19900101_fc0_1_<TASK_NAME>` 
  * e.g. `/appl/AS/AUTOSUBMIT_DATA/<EXPID>/tmp/ASLOGS`
  * (And any other information you may have, the more, the merrier.)

### Reproducible Example

See comment above.

### Expected Behaviour

No exception raised during runtime.

VindeeR:
I spent the whole day trying to figure out how does this issue come to happen, and I haven't make much progress, what I found out was the following.

1. Is an issue related to the work_event being set a `None` instead of `set`
2. The error happens at the `p.work_event.set()`
3. When the `submitter` is generated it's values is already as None
4. I could only replicate this issue at the branch `969b6e0f`
5. The `Plarform Class` uses the function `prepare_process()` to set the `ctx.Event` into the `self.platform` but not the platform that return
6. the function `load_process_info()` is supposed to copy the data from the self platform to the platform that return but this data is not passed

<img width="1075" height="317" alt="Image" src="https://github.com/user-attachments/assets/237fd58f-40c2-4983-b4b3-6935f8ec450c" />

<img width="893" height="145" alt="Image" src="https://github.com/user-attachments/assets/867b53a7-80b3-4d28-8c8e-2ce392669fc0" />

[...]

PR #2663, opened 2025-10-21
[Bug] - 'NoneType' object has no attribute 'set'

### LINKED ISSUE #2611: Experiment not starting: [CRITICAL] Unexpected error: 'NoneType' object has no attribute 'set'.
### Description
Half of the time I am trying to run my exp (`a9sg`), I need to re-create the exp multiple times to get rid of an error that seems to be randomly caused, or I don't fully understand. See info below, thanks in advance!
* Autosubmit version: v4.1.14
* Machine/VM/environment name (if applicable): hub5
* Experiment ID (if applicable): a9sg
* Experiment tasks & log path if applicable): /gpfs/scratch/bsc32/bsc954883/a9sg/LOG_a9sg

### Reproducible Example
You can clone autosubmit exp `a9sg` and try to create and run it to reproduce (very likely) the bug.

### Expected Behaviour
It will 50% of the time after doing `create exp` or running `scripts/prepare_autosubmit.sh` create this error:
```
[ERROR] Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='WARNING', command='run', expid='a9sg', notransitive=False, update_version=False, start_time=None, start_after=None, run_only_members=None, profile=False)
[ERROR] This is the experiment: <a9sg> which had an issue with the command: <run> and it is currently using the Autosubmit Version: 4.1.14.
 [CRITICAL] Unexpected error: 'NoneType' object has no attribute 'set'.
```

VindeeR:
Closes #2611 #2583

There was a flow that would always end up giving an Error 'NoneType' object has no attribute 'set', due to set being outside the validation

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -1859,17 +1859,15 @@ class JobList(object):
             return [job for job in completed_jobs if job.packed is False]
         return completed_jobs
 
-    def get_completed_failed_without_logs(self, platform: Any = None) -> List[Any]:
-        """Returns a list of completed or failed jobs without updated logs.
+    def get_completed_failed_without_logs(self) -> list[Any]:
+        """
+        Returns a list of completed or failed jobs without updated logs.
 
-        :param platform: Job platform, defaults to None.
-        :type platform: Platform
         :return: List of completed and failed jobs without updated logs.
         :rtype: List[Job]
         """
 
         completed_failed_jobs = [job for job in self._job_list if
-                                 (platform is None or job.platform.name == platform.name) and
                                  (job.status == Status.COMPLETED or job.status == Status.FAILED) and
                                  job.updated_log is False]
 
--- a/autosubmit/platforms/platform.py
+++ b/autosubmit/platforms/platform.py
@@ -17,14 +17,13 @@
 
 import atexit
 import multiprocessing
-from multiprocessing.context import SpawnContext
 import os
 import queue  # only for the exception
 import time
 import traceback
 from contextlib import suppress
-from multiprocessing.synchronize import Event
 from multiprocessing.queues import Queue
+from multiprocessing.synchronize import Event
 # noinspection PyProtectedMember
 from os import _exit  # type: ignore
 from pathlib import Path
@@ -42,6 +41,7 @@ if TYPE_CHECKING:
     from autosubmit.job.job import Job
     from autosubmit.job.job_list import JobList
     from autosubmit.job.job_package_persistence import JobPackagePersistence
+    from multiprocessing.process import BaseProcess
 
 
 def _init_logs_log_process(as_conf: 'AutosubmitConfig', platform_name: str) -> None:
@@ -113,7 +113,7 @@ class CopyQueue(Queue):
         self.timeout = timeout
         super().__init__(maxsize, ctx=ctx)
 
-    def put(self, job: Any, block: bool = True, timeout: float = None) -> None:
+    def put(self, job: Any, block: bool = True, timeout: Optional[float] = None) -> None:
         """Puts a job into the queue if it is not a duplicate.
 
         :param job: The job to be added to the queue.
@@ -131,7 +131,7 @@ class Platform:
     Class to manage the connections to the different platforms.
     """
     # This is a list of the keep_alive events, used to send the signal outside the main loop of Autosubmit
-    worker_events = list()
+    worker_events: list[Event] = []
     # Shared lock between the main process and a retrieval log process
     lock = multiprocessing.Lock()
 
@@ -148,9 +148,10 @@ class Platform:
         :param auth_password: Optional password for two-factor authentication.
         :type auth_password: str or list, optional
         """
+        self.ctx = self.get_mp_context()
         self.connected = False
-        self.expid = expid  # type: str
-        self._name = name  # type: str
+        self.expid: str = expid
+        self._name: str = name
         self.config = config
         self.tmp_path = os.path.join(
             self.config.get("LOCAL_ROOT_DIR", ""), self.expid, self.config.get("LOCAL_TMP_DIR", ""))
@@ -209,12 +210,12 @@ class Platform:
             self.pw = None
         self.max_waiting_jobs = 20
         self.recovery_queue: Optional[Queue] = None
-        self.work_event = None
-        self.cleanup_event = None
-        self.log_retrieval_process_active = False
-        self.log_recovery_process = None
+        self.work_event: Optional[Event] = None
+        self.cleanup_event: Optional[Event] = None
+        self.log_retrieval_process_active: bool = False
+        self.log_recovery_process: Optional['BaseProcess'] = None
         self.keep_alive_timeout = 60 * 5  # Useful in case of kill -9
-        self.processed_wrapper_logs = set()
+        self.processed_wrapper_logs: set[str] = set()
         self.compress_remote_logs = False
         self.remote_logs_compress_type = "gzip"
         self.compression_level = 9
@@ -231,7 +232,7 @@ class Platform:
             self.compress_remote_logs = platform_config.get("COMPRESS_REMOTE_LOGS", False)
             self.remote_logs_compress_type = platform_config.get("REMOTE_LOGS_COMPRESS_TYPE", "gzip")
             self.compression_level = platform_config.get("COMPRESSION_LEVEL", 9)
-            
+
         self.log_queue_size = log_queue_size
         self.remote_log_dir = None
 
@@ -857,18 +858,18 @@ class Platform:
         This method sets the cleanup event to signal the log recovery process to finish,
         waits for the process to join with a timeout, and then resets all related variables.
         """
-        if self.cleanup_event:
+        if self.cleanup_event is not None:
             self.cleanup_event.set()  # Indicates to old child ( if reachable ) to finish.
-        if self.log_recovery_process:
+        if self.log_recovery_process is not None:
             # Waits for old child ( if reachable ) to finish. Timeout in case of it being blocked.
             self.log_recovery_process.join(timeout=60)
         # Resets everything related to the log recovery process.
-        self.recovery_queue = None
+        self.recovery_queue = self.ctx.Queue()
         self.log_retrieval_process_active = False
         self.remove_workers(self.work_event)
-        self.work_event = None
-        self.cleanup_event = None
-        self.log_recovery_process = None
+        self.work_event = self.ctx.Event()
+        self.cleanup_event = self.ctx.Event()
+        self.log_recovery_process = self.ctx.Process()
         self.processed_wrapper_logs = set()
 
     def load_process_info(self, platform):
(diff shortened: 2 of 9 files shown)


kinow:
From today's meeting:

We can test it with @dbeltrankyl 's suggestion:

- Set a breakpoint in the code somewhere (see next points when deciding where)
- Find a way to stop the log recovery (kill it? :hocho: )
- Before it's relaunched, while the breakpoint is waiting, check that `conf/as_misc.yml` contains `AS_COMMAND: run` and then  run another command, like `autosubmit monitor`
- Now, look at the `conf/as_misc.yml` again, and confirm the `AS_COMMAND` shows `monitor`.
- Finally, release the breakpoint, and see what happens and report it here :grimacing: 

If I understood correctly (Dani can correct me here) the log recovery may not be initialized, and the event or another object may be `None`, causing this issue (or a very similar one).

autosubmit/platforms/platform.py
```diff
@@ -97,7 +97,7 @@ class CopyQueue(Queue):
     A queue that copies the object gathered.
     """
 
-    def __init__(self, maxsize: int = -1, block: bool = True, timeout: float = None, ctx: Any = None) -> None:
+    def __init__(self, maxsize: int = -1, block: bool = True, timeout: float = 0.0, ctx: Any = None) -> None:
```
kinow: Interesting. We declare a `self.timeout` on `CopyQueue`, but I cannot find where it's used.

Looking inside `Queue` (parent class, from Python stdlib), I also cannot see where `self.timeout` gets used for the `Queue` object -- it only receives timeouts via function args.

You can see that the `CopyQueue.put` function, for instance, receives a timeout. That may be completely different than this `self.timeout` you set the type for here. My guess is that this `self.timeout` can be removed, but probably for a 4.1.17 or later version. 🫠

autosubmit/platforms/platform.py
```diff
@@ -1135,7 +1136,7 @@ def get_file_size(self, src: str) -> Union[int, None]:
         """
         raise NotImplementedError  # pragma: no cover
 
-    def read_file(self, src: str, max_size: int = None) -> Union[bytes, None]:
+    def read_file(self, src: str, max_size: int = 0) -> Union[bytes, None]:
```
kinow: The implementations are still using `None`. `locplatform` and `paramiko_platform`. In both cases, I think passing `max_size` as `None` may be OK? I think `read` calls the C function `read`, https://man7.org/linux/man-pages/man2/read.2.html, which accepts `0`. I think Python might be translating `None` to `0` (if you really want to be sure, you can check cpython's code -- or revert this one as the implementations are not following this 🤷‍♂️ ?)

autosubmit/autosubmit.py
```diff
@@ -2261,7 +2261,10 @@ def refresh_log_recovery_process(platforms: list[Platform], as_conf: AutosubmitC
             if not p.log_recovery_process or not p.log_recovery_process.is_alive():
                 p.clean_log_recovery_process()
                 p.spawn_log_retrieval_process(as_conf)
-            p.work_event.set()
+            try:
+                p.work_event.set()
+            except Exception as e:
+                Log.debug(f"Error while loading work_event: {str(e)}")
```
kinow: What exception are we expecting here?

If it's the error of `None`, then `AttributeError` could be suppressed or logged.

`Event.set()` doesn't appear to raise any exceptions, 

- https://docs.python.org/3/library/multiprocessing.html#multiprocessing.managers.SyncManager.set
- https://github.com/python/cpython/blob/02202c117b5702f3325e62b07ccdeaa125fc8722/Lib/multiprocessing/synchronize.py#L343-L347
  - https://github.com/python/cpython/blob/02202c117b5702f3325e62b07ccdeaa125fc8722/Lib/threading.py#L475-L520
  - https://github.com/python/cpython/blob/02202c117b5702f3325e62b07ccdeaa125fc8722/Lib/multiprocessing/synchronize.py#L279-L306

So it might be better to check if is `None`, or maybe leave as it was?

autosubmit/platforms/platform.py
VindeeR: Yeah, I was fixing a different PR that was changing the Paramiko and I changed there to not be a mandatory parameter, at the end of the day it works the same, I'll be removing it here as well

VindeeR: I think it might be better to tackle each file as we work on them no? to avoid too many MyPy changes all together

autosubmit/autosubmit.py
VindeeR: It raises `AttributeError: NoneType no set attribute set()`

I put it in the test, and forgot to change in the code sorry, I'll make the changes

autosubmit/platforms/platform.py
kinow: Yes, but in this case you are declaring `max_size` here to be one type, leaving the other files with different types. In this case a developer looking at this could be confusing which direction to follow based on types. So for cases where a type spreads across files/implementations, then it might be safer/less confusing to fix everything together, to avoid bugs, IMHO.

autosubmit/platforms/platform.py
```diff
@@ -661,7 +662,7 @@ def delete_file(self, filename: str):
         raise NotImplementedError  # pragma: no cover
 
     # Executed when calling from Job
-    def get_logs_files(self, exp_id: str, remote_logs: tuple[str, str]) -> None:
+    def get_logs_files(self, exp_id: str, remote_logs: tuple[str, str]):
```
kinow: I think you remived the return type? Any reason for that?

kinow (commented):
Will continue reviewing it later. Need more coffee.

autosubmit/platforms/platform.py
VindeeR: this functions has a 

```py
return NotImplementedError
```

as return, to which Mypy was complaining

autosubmit/platforms/platform.py
```diff
@@ -133,7 +133,7 @@ class Platform(object):
     Class to manage the connections to the different platforms.
     """
     # This is a list of the keep_alive events, used to send the signal outside the main loop of Autosubmit
-    worker_events = list()
+    worker_events: list = []
```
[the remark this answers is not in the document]
kinow: No, I mean, `list[Event]`? Or is it another type? :laughing:

autosubmit/platforms/platform.py
kinow: In that case, the fix would be to keep the right type in the signature and `raise NotImplementedError` (raise, not return). We had this in many other parts of the code, returning instead of raising. I thought there were no other occurrences of that :disappointed_relieved:

autosubmit/platforms/platform.py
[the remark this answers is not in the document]
VindeeR: Oh okay, yeah it makes sense :rofl: my bad, thanks for noticing :raised_hands:

autosubmit/platforms/platform.py
```diff
@@ -211,13 +212,13 @@ def __init__(self, expid: str, name: str, config: dict, auth_password: Optional[
         else:
             self.pw = None
         self.max_waiting_jobs = 20
-        self.recovery_queue = None
-        self.work_event = None
-        self.cleanup_event = None
-        self.log_retrieval_process_active = False
-        self.log_recovery_process = None
+        self.recovery_queue: Queue = self.ctx.Queue()
```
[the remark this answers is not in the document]
kinow: Maybe these values could be made `Optional` @VindeeR ?

I think the code expects `spawn_log_retrieval_process` and/or `prepare_process` to be called before these objects are used. `spawn_...` will call `prepare_process` where these objects are initialized.

Setting to a default value here in the constructor might be useful to make sure these objects are never used when `None`, but I think this could hide bugs? In the sense that you could have another event or queue used, then those function are called, re-initialize these objects, discarding the ones initially used, and then there would be some inconsistency in the system that could be hard to track.

test/integration/scripts/test_stats.py
```diff
@@ -20,10 +20,12 @@
 from pathlib import Path
 
 from autosubmit.scripts.autosubmit import main
+from test.integration.conftest import sleep
 
 _EXPID = 't000'
 
 
+@sleep(1)
```
kinow: Ah, I thought about a different way to fix this test, that I think could be even more deterministic.

```
$ python3
Python 3.10.12 (main, Aug 15 2025, 14:32:43) [GCC 11.4.0] on linux
>>> import multiprocessing
>>> multiprocessing.active_children()
[]
>>> 
```

Or maybe something with psutil. When you run the `autosubmit run` we need to ensure that it finished and the log recovery process, or any other child process (jobs? 🤷 ) have completed. So instead of simply sleeping, you can implement a timeout of 10 or 30 seconds that waits until all processes have finished or fail with a nice message saying "Waited 30 seconds but processes [151561, 323232, etc] are still alive!". That way we would know why the test failed.

test/integration/conftest.py
```diff
@@ -655,3 +655,21 @@ def as_db(request: 'FixtureRequest', autosubmit: Autosubmit, tmp_path: 'LocalPat
     autosubmit_exp('____')
 
     return backend
+
+
+def sleep(timeout, retry=3):
```
kinow: Maybe rename to timeout? Or wait_something? But you got the gist of it very quickly! 💪 👍

kinow (commented):
8/9 files reviewed! Only have some questions about `platform.py`, and comments in other places. Great job so far! Thanks Erick.

autosubmit/job/job_list.py
```diff
@@ -1869,18 +1869,15 @@ def get_completed(self, platform=None, wrapper=False):
             return [job for job in completed_jobs if job.packed is False]
         return completed_jobs
 
-    def get_completed_failed_without_logs(self, platform: Any = None) -> List[Any]:
+    def get_completed_failed_without_logs(self) -> List[Any]:
```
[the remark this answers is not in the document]
VindeeR: oops, that flew under the radar :laughing: thanks for noticing

test/unit/platforms/test_locplatform.py
```diff
@@ -82,3 +84,14 @@ def test_get_stat_file(count: int, stats_file_exists: bool, job_fail_count: int,
...
+def test_refresh_log_recovery_process(autosubmit, autosubmit_config):
+    as_conf = autosubmit_config('t000', {}, reload=False, create=False)
+    as_conf.misc_data["AS_COMMAND"] = 'run'
+
+    local = LocalPlatform(expid='t000', name='local', config=as_conf.experiment_data)
+    local.clean_log_recovery_process()
+    local.log_recovery_process = BaseProcess()
+
+    with patch('multiprocessing.process.BaseProcess.is_alive', return_value=True):
+        autosubmit.refresh_log_recovery_process(platforms=[local], as_conf=as_conf)
```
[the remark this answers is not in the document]
VindeeR: hmmm... I guess this is a bit tricky, this tests if `[platform].work_event` would crash, because before, since it wasn't initialized, it was crashing the run, which was failing the test by default, but if it isn't crashing it is working.

In theory, I could assert if `[platform].work_event` is not `None` but now with the initialization they'll never be so it makes no sense to assert that.

What do you think?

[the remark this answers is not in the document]
kinow: Ah yes. Even if we remove the default assignment in the constructor (another thread here), I think this wouldn't make much sense.

Maybe we should move it to an integration test and drop the mock.

- a small experiment with one task, `SCRIPT: sleep 60`
- `p = Process(target=some function that calls autosubmit.run)`, `p.start()`
- then we can `assert event is not None`, call `refresh_log_recovery_process`, and now check if the instances changed, or if they are in some other state we expected they to be
- check if `p.is_alive()` & if so terminate+join (the terminate must kill it, I think).

test/integration/test_log_recovery.py
```diff
@@ -313,3 +309,20 @@ def test_create_a_new_copy(local, pjm, slurm, ps_platform, ecaccess):
...
+    local.clean_log_recovery_process()
+
+    assert local.work_event is not None
+
+    p = local.work_event
+
+    with mocker.patch('multiprocessing.process.BaseProcess.is_alive', return_value=True):
+        autosubmit.refresh_log_recovery_process(platforms=[local], as_conf=as_conf)
+        assert p == local.work_event
+        assert local.work_event.is_set()
```
kinow: 👏

test/integration/scripts/test_stats.py
```diff
@@ -20,10 +20,12 @@
 from pathlib import Path
 
 from autosubmit.scripts.autosubmit import main
+from test.integration.conftest import wait_child
 
 _EXPID = 't000'
 
 
+@wait_child(1)
```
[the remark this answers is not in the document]
VindeeR: I read your message but got confused, I was going to ask you more about it later once you had some free time, thanks for the help <3

PR #2668, opened 2025-10-22
[Implementation] Implementing PBS for Miyabi

### LINKED ISSUE #2478: FileNotFoundError: [Errno 2] No such file or directory: '/home/kinow/autosubmit/o002/tmp/LOG_o002'
### Description

Found while troubleshooting o007/climatedt, for #2474 . At some point in my experiment I decided to delete older files, and start over, so I deleted pkl contents, conf/metadata, job_data in the AS metadata folder, and everything inside `tmp`.

When I tried running again, I got the following exception.

```bash
[ERROR] Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='DEBUG', command='run', expid='o002', notransitive=False, update_version=False, start_time=None, start_after=None, run_only_members=None, profile=False)
[ERROR] This is the experiment: <o002> which had an issue with the command: <run> and it is currently using the Autosubmit Version: 4.1.15.
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/scripts/autosubmit.py", line 105, in main
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 724, in run_command
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 2285, in run_experiment
    jobs_to_check, job_changes_tracker = Autosubmit.check_wrappers(as_conf, job_list, platforms_to_test, expid)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 1875, in check_wrappers
    wrapper_job, save = Autosubmit.manage_wrapper_job(as_conf, job_list, platform,
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 1825, in manage_wrapper_job
    wrapper_job.check_status(
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/job/job.py", line 2906, in check_status
    self.cancel_failed_wrapper_job()
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/job/job.py", line 3162, in cancel_failed_wrapper_job
    self._check_running_jobs()
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/job/job.py", line 3044, in _check_running_jobs
    if not os.stat(log_dir):
FileNotFoundError: [Errno 2] No such file or directory: '/home/kinow/autosubmit/o002/tmp/LOG_o002'
```

Reading the code, 

```py
            if not os.stat(log_dir):
                os.mkdir(log_dir)
                os.chmod(log_dir, 0o770)
```

I believe the intention was to create the directory if it did not exist. However, `os.stat` raises `FileNotFoundError`. I think the fix here is to simply call `Path(bla).exists()` instead.

* Autosubmit version: master, 4.1.15+

### Reproducible Example

- Create an experiment
- Run a few times
- Delete tmp contents
- Try running again

### Expected Behaviour

AS creates the files/directories as expected, or fails complaining files/directories do not exist, no exception raised.

### LINKED ISSUE #2659: Add PBS platform back (for Miyabi)
NOTE: this is for Miyabi, which ends on March 2026; they need to run a 100 member ensemble of 20-30 years control on Miyabi-C (not G). So the sooner we have a stable version ready for them to use, the merrier. Maybe 4.1.16 would be safer.

### Problem

We removed the PBS platform some releases before as the code that was ported from AS3 didn't work with AS4.

We now need to add it back for Miyabi / HANAMI.

Miyabi appears to have 2FA, so we will have to think how we'll handle that, or if we'll need to request a longer expiration date, or an exclusion of the 2FA token.

### Proposed Solution

1. Use Miyabi to validate the platform
2. Use the version from the last Git commit with the platform as reference
3. Re-write it using the Slurm platform as that's the best working remote platform we have
4. Test that we can submit jobs, retrieve logs, and transfer files
5. Test directives, especially related to resources like CPU, GPU, mem, account, wallclock
6. Test on Miyabi-C, submitting very small jobs, maybe one that runs a bit longer
7. Test on Miyabi-G and run something like pytorch+list devices, and/or nvidia-smi, making sure we have a job with a GPU available

Some refs for Miyabi

- https://www.cc.u-tokyo.ac.jp/events/seminar/files/20250116-5.pdf
- https://miyabi-www.jcahpc.jp/login

VindeeR:
Closes #2659 
Closes #2478
Related #2900 

Re-implementation of the PBS platform focused on integration with the Miyabi project 

**Check List**

- [ ] Documentation updated.


autosubmit/platforms/headers/pbs_header.py
```diff
@@ -0,0 +1,374 @@
...
+import textwrap
+
+
+class PbsHeader(object):
+    """Class to handle the Lindgren headers of a job"""
+
+    # noinspection PyMethodMayBeStatic,PyUnusedLocal
+    def get_queue_directive(self, job, parameters, het=-1):
+        """
+        Returns queue directive for the specified job
```
kinow: Would be good if we aligned on how pydoc comments are structured. This line you used is the third line of the documentation.

```
""" < --- first lline
 < --- empty, where sphinx/other tools would just add a break line
Returns queue... < -- your text
```

Some tools are able to handle this. But it's easier and more consistent if we follow something like https://peps.python.org/pep-0257/#multi-line-docstrings

We have both formats in our code. Old libs normally follow PEP-257. I noticed IDE/AI sometimes generate one or the other.

kinow:
Conflict added by my PR fixing the integration tests, @VindeeR :grimacing: sorry

autosubmit/config/configcommon.py
```diff
@@ -2503,7 +2502,7 @@ def get_notifications(self) -> str:
 
     # based on https://github.com/cbirajdar/properties-to-yaml-converter/blob/master/properties_to_yaml.py
     @staticmethod
-    def ini_to_yaml(root_dir: Path, ini_file: str) -> None:
+    def ini_to_yaml(root_dir: Path, ini_file: Path) -> None:
```
dbeltrankyl: Are we sure that in every call to this we send a Path, not a string?

autosubmit/platforms/headers/pbs_header.py
```diff
@@ -0,0 +1,240 @@
...
+    def get_account_directive(self, job, parameters, het=-1):
+        """Returns account directive for the specified job
+
+        :param job: job to create account directive for
+        :type job: Job
+        :return: account directive
+        :rtype: str
+        """
+        if parameters['CURRENT_PROJ'] != '':
+            return f"qsub -q {parameters['CURRENT_QUEUE']} -l {parameters['CURRENT_SELECT']} -W group_list={parameters['CURRENT_PROJ']} ./a00b/LOG_a00b/a00b_LOCAL_SETUP.cmd"
```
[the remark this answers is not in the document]
dbeltrankyl: >             return f"qsub -q {parameters['CURRENT_QUEUE']} -l {parameters['CURRENT_SELECT']} -W group_list={parameters['CURRENT_PROJ']} ./a00b/LOG_a00b/a00b_LOCAL_SETUP.cmd"

Also this is hardcoded to a specific experiment 😅

autosubmit/platforms/headers/pbs_header.py
```diff
@@ -0,0 +1,240 @@
...
+    def get_nodes_directive(self, job, parameters, het=-1):
+        """Returns nodes directive for the specified job
+
+        :param job: job to create nodes directive for
+        :type job: Job
+        :return: nodes directive
+        :rtype: str
+        """
+        if parameters['NODES'] != '':
+            return f"qsub -l nodes={parameters['NODES']}"
```
dbeltrankyl: Can you check if this is translated as 

`#PBS qsub {directive}`

instead of

`#PBS {directive}`?

autosubmit/platforms/headers/pbs_header.py
```diff
@@ -0,0 +1,240 @@
...
+    def get_memory_per_task_directive(self, job, parameters, het=-1):
+        """Returns memory per task directive for the specified job
+
+        :param job: job to create memory per task directive for
+        :type job: Job
+        :return: memory per task directive
+        :rtype: str
+        """
+        if parameters['MEMORY_PER_TASK'] != '':
+            return f"qsub -l vmem={parameters['MEMORY_PER_TASK']}"
```
dbeltrankyl: Can you check if this is translated as 

`#PBS qsub {directive}`

instead of

`#PBS {directive}`?

autosubmit/platforms/headers/pbs_header.py
```diff
@@ -0,0 +1,240 @@
...
+    def get_threads_per_task(self, job, parameters, het=-1):
+        """Returns threads per task directive for the specified job
+
+        :param job: job to create threads per task directive for
+        :type job: Job
+        :return: threads per task directive
+        :rtype: str
+        """
+        if parameters['NUMTHREADS'] != '':
+            return f"qsub -l ppn={parameters['NUMTHREADS']}"
```
dbeltrankyl: Can you check if this is translated as 

`#PBS qsub {directive}`

instead of

`#PBS {directive}`?

autosubmit/platforms/headers/pbs_header.py
```diff
@@ -0,0 +1,240 @@
...
+        :param job: job to create tasks per node directive for
+        :type job: Job
+        :return: tasks per node directive
+        :rtype: str
+        """
+        if int(parameters['TASKS']) > 1:
+            return f"qsub -l nodes={parameters['TASKS']}:ppn={parameters['TASKS']}"
+        return ""
+
+    def get_select_directive(self, job, parameters, het=-1):
```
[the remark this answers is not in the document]
dbeltrankyl: or 

#%QUEUE_DIRECTIVE%
#%SELECT_DIRECTIVE%
#%WALLCLOCK%:00
#%JOBNAME%

and add the PBS in the get's of before (instead of qsub)

autosubmit/platforms/pbsplatform.py
```diff
@@ -0,0 +1,657 @@
...
+                            for id_ in jobid:
+                                self.send_command(self.cancel_job(id_))
+                    jobs_id = None
+                    self.connected = False
+                    if e.trace is not None:
+                        has_trace_bad_parameters = str(e.trace).lower().find("bad parameters") != -1
+                    else:
+                        has_trace_bad_parameters = False
+                    if (has_trace_bad_parameters or e.message.lower().find("invalid partition") != -1 or
+                            e.message.lower().find(" invalid qos") != -1 or
```
dbeltrankyl: I think.. the message will be different on this platform. Can you check it? 

Submit a wrong configured script, for example, with 999 node,s and see what output you receive

autosubmit/platforms/pbsplatform.py
```diff
@@ -0,0 +1,657 @@
...
+
+        :rtype: None
+        """
+        self.root_dir = os.path.join(
+            self.scratch, self.project_dir, self.user, self.expid)
+        self.remote_log_dir = os.path.join(self.root_dir, "LOG_" + self.expid)
+        self.cancel_cmd = "qdel"
+        self._submit_cmd = (f'qsub -q {self.config.get("PLATFORMS", {}).get(self.name.upper(), {}).get("QUEUE", "debug-g")} '
+                            # f'-e {self.remote_log_dir}.err.0 -o {self.remote_log_dir}.out.0 '
+                            f'-l {self.config.get("PLATFORMS", {}).get(self.name.upper(), {}).get("SELECT", "select=1:mpiprocs=2")} -W group_list={self.project_dir} {self.remote_log_dir}/')
```
dbeltrankyl: Same here. 

The header is the one that should set everything

In the qsub you only want to pass the working dir and the script itself

dbeltrankyl (commented):
Hello @VindeeR ! 

Good progress! but there are few things to address. I commented upon them :). Also commented other stuff related to mypi fixes 

I did not check the tests

autosubmit/platforms/headers/pbs_header.py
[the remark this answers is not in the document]
VindeeR: > (just a question) What is select? Maybe doc what the qsub flags do

`SELECT` is a new directive I implemented that deals with the `-l`, which is a mandatory argument.

It can be used to specify both the `nodes` and the `mpiprocs` inside of it either `select=2` or `select=2:mpiprocs=4`, for example.

I know we have the directive for `nodes` but I don't think we do for `mpi` and since the command is supposed to be just 1 I thought it would be better to create a new one where they can be explicitly at the users discretion, instead of creating more complexity to the code.

Both ways are possible and I can change if you think it'll be better.

autosubmit/platforms/headers/pbs_header.py
```diff
@@ -0,0 +1,240 @@
...
+    def get_exclusive_directive(self, job, parameters, het=-1):
+        """Returns account directive for the specified job
+
+        :param job: job to create account directive for
+        :type job: Job
+        :return: account directive
+        :rtype: str
+        """
+        if str(parameters['EXCLUSIVE']).lower() == 'true':
+            return "qsub -l naccesspolicy=singlejob"
```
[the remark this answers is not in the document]
VindeeR: I was following the same pattern seen in `SLURM` I believe I can delete this code since they're not currently been used and later on implemented if needed, but I need to double check to make sure it will all be working just fine if I remove them or at least removing it's implementation for now

autosubmit/platforms/headers/pbs_header.py
VindeeR: ditto for last comment

autosubmit/platforms/headers/pbs_header.py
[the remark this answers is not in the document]
dbeltrankyl: Also, we already support MPI, for that reason, we have these variables:

```
tasks 
threads
```
processors:
--ntasks=# : Number of "tasks" (use with distributed parallelism).
tasks (aka, MPI)
--ntasks-per-node=# : Number of "tasks" per node (use with distributed parallelism).
Threads: (openmp, per example)
--cpus-per-task=# : Number of CPUs allocated to each task (use with shared memory parallelism).

autosubmit/platforms/pbsplatform.py
```diff
@@ -0,0 +1,657 @@
...
+        Generates a command to get data from a job
+        JobId, State, NCPUS, NNodes, Submit, Start, End, ConsumedEnergy, MaxRSS, AveRSS%25.
+
+        :param job_id: ID of a job.
+        :param job_id: str
+
+        :return: Command to get job energy.
+        :rtype: str
+        """
+        return (f'sacct -n --jobs {job_id} -o JobId%25,State,NCPUS,NNodes,Submit,'
```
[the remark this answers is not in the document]
VindeeR: I didn't delete this command because I didn't know if we were going to use it for `Miyabi`.

Especially because I believe this function is only used by the 'SLURM' platform, do you know by any chance if it is supposed to be used by `Miyabi`?

`if job_data_dc and type(self.platform) is not str and self.platform.type == "slurm":`

autosubmit/platforms/pbsplatform.py
```diff
@@ -0,0 +1,758 @@
...
+                                self.send_command(self.cancel_job(id_))
+                    jobs_id = None
+                    self.connected = False
+                    if e.trace is not None:
+                        has_trace_bad_parameters = str(e.trace).lower().find("violates resource limits") != -1
+                    else:
+                        has_trace_bad_parameters = False
+                    if (has_trace_bad_parameters or e.message.lower().find("violates resource limits") != -1 or
+                            e.message.lower().find("Unknown Resource") != -1 or
+                            e.message.lower().find("Job violates queue") != -1 or
```
dbeltrankyl: same

autosubmit/platforms/pbsplatform.py
```diff
@@ -0,0 +1,758 @@
...
+        :param output_lines: Output of the ssh command.
+        :type output_lines: str
+        :param x11: Enable x11 forwarding, to enable graphical jobs.
+        :type x11: bool
+
+        :return: List of job ids that got submitted and had an output.
+        :rtype: Union[list[int], int]
+        """
+        try:
+            if output_lines.find("failed") != -1:
```
dbeltrankyl: Same here, maybe convert the string to upper or lower so it is less prone to errors... 

Also, does PBS output have the words "failed" if the job fails to be submitted?

autosubmit/platforms/headers/pbs_header.py
```diff
@@ -0,0 +1,240 @@
...
+            #
+            #%RESERVATION_DIRECTIVE%
+            #PBS -q %QUEUE_DIRECTIVE%
+            #PBS -l %SELECT_DIRECTIVE%
+            #PBS -l walltime=%WALLCLOCK%:00
+            #PBS -N %JOBNAME%
+            #PBS -o %CURRENT_SCRATCH_DIR%/%CURRENT_PROJ_DIR%/%CURRENT_USER%/%DEFAULT.EXPID%/LOG_%DEFAULT.EXPID%/%OUT_LOG_DIRECTIVE%
+            #PBS -e %CURRENT_SCRATCH_DIR%/%CURRENT_PROJ_DIR%/%CURRENT_USER%/%DEFAULT.EXPID%/LOG_%DEFAULT.EXPID%/%ERR_LOG_DIRECTIVE%
+            #PBS -W group_list=%CURRENT_PROJ_DIR%
+            #PBS -j oe
```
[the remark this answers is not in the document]
dbeltrankyl: But that changes from project to project. 

This would be the equivalent to the -A in other platforms. 

So:

```yaml
PLATFORMS:
 miyabi: 
   USER: test
   PROJECT: oe
   scratch_dir: bla
```

should set

```
#PBS -j oe
remote_proj_dir = /bla/oe/test
```

```yaml
PLATFORMS:
 miyabi: 
   USER: test
   PROJECT: oe
   SCRATCH_PROJECT_DIR: somethingelse
   scratch_dir: bla
```

```
#PBS -j oe
remote_proj_dir = /bla/somethingelse/test
```

dbeltrankyl (commented):
Hello Erick! 

Good work! 

Left some small comments,

I didn't look at the tests part tho.


After the changes, 

it would be good if @Lerriola tested it before merging to master, in case we need to make additional changes due to incorrect behaviour with real templates or something along those lines. 

So after you address the feedback about the code, I can deploy a dev ( or tell you how to ) module in the hubs so he can test it ( we don't need to have all the tests )

Or maybe better, he can just do a git clone of this branch and test it, which would be much faster and easier to do further changes as he would only need to do a git pull

dbeltrankyl:
> @dbeltrankyl I can't seem to reply to your #PBS -j oe comment so I can't reference it here, but here's the explanation to it.

> As per Miyabi user guide I was sent the -j oe parameter is used to Merge output of standard error to standard output. and in all the examples in the manual they use it.

> Hence my decision of building the header this way, but it can be easily removed.

So this is not related to the project at all, no?

> Merge output of standard error to standard output.

Also.. This would explain the need of having to create manually the ERR file:

`            touch %CURRENT_SCRATCH_DIR%/%CURRENT_PROJ_DIR%/%CURRENT_USER%/%DEFAULT.EXPID%/LOG_%DEFAULT.EXPID%/%ERR_LOG_DIRECTIVE%`


What happens if you don't add the `-j oe` and comment out the touch? Will the file be created?


In any case, in the rest of the platforms, we don't merge the .out and .err so I would remove it to mantain a coherence

Thanks for looking at it !

dbeltrankyl:
@Lerriola 

I recommend that you, instead of a module, just generate a venv for you 

I FYI the slack @kinow  message of last week:

> Also, if you ever need it, you can clone autosubmit in your home dir, python -m venv venv; cd $autosubmit-clone; pip install . and then activate that venv and try using it directly. It should be able to use the global DB and normal filesystem, and then you'd be using that version of AS.
> Useful if we are not available to install a new module, or you want to test a certain commit or branch, etc.



To clone this: 

`
git clone https://github.com/BSC-ES/autosubmit -b miyabi_pbs_platform
`

manuel-g-castro:
Hi, @VindeeR. Maybe [this](https://unix.stackexchange.com/questions/50508/reusing-ssh-session-for-repeated-rsync-commands) work?

Lerriola:
Thanks @dbeltrankyl :+1: The current local-setup and synch are basically cloning model sources if needed and then transferring them to the remote setup. So I could implement an option that depending on a key in main (so it is not platform dependant) I can avoid the additional ssh (rsync) connection.

I think this is probably the easiest option. I imagine that mounting with sshfs would still require using the OTP codes, right? I'm also not sure when the ssh connection is triggered, if only at the initial mounting or any time the remote mounted filesystem is used. In any case, at some point the OTP code will be required.

Lerriola:
Hi @VindeeR and @dbeltrankyl,

The experiment I'm using is **a81i** (on the bsc-hubs), and it uses a branch that includes the workflow changes for running on Miyabi.  Here's the MR with all the changes https://gitlab.earth.bsc.es/ces/hanami/wp4p1_workflow/-/merge_requests/2

Thanks!

Lerriola:
Thanks for the answer @dbeltrankyl. I was checking, and I don't have a `~/.bashrc` file. At least in home.

dbeltrankyl:
Hello @VindeeR 

I was checking @Lerriola's logs, and I think the main problem is that, in interactive mode, the module  init script is not being called, so the qsub command is not found. 

So it should be a matter of generating a .bashrc file that loads that script 

On a side note, he showed me the logs from you, and I noticed that the qsub command is not returning the job_id, and instead it gets [], The jobid is recovered later by the fallback mechanism. Is that intended?

Lerriola:
Okay! So I manually added the user .bashrc file similar to the one that @VindeeR has, and now the workflow is able to run.

Lerriola:
Hi @VindeeR! Thanks for all the help : ) I've been running the workflow and I've managed to get NICOCO compiled and to run part of a chunk of the simulation :tada:  The issue is that I think the script is exiting prematurely. I'm not sure if this is because of Miyabi or the newer version of autosubmit. 

The cmd runs properly, and the mpiexec runs as should, but it finishes prematurely here:

```bash
 + echo 'Start of the NICOCO and regridding mpiexec runs - Fri Feb  6 10:58:39 PM JST 2026'
mpiexec -n "${NICAM_PROC}" "${NICAM_DRIVER}" : -n "${COCO_PROC}" "${COCO_DRIVER}" 2>&1
+ mpiexec -n 160 /work/jh250055/m44004/aahd/nicoco/build/NICAM/NICAM/bin/nhm_driver : -n 32 /work/jh250055/m44004/aahd/nicoco/build/NICAM/COCO/bin/MYB/cgcmO.t42l40hmed52full-a8o1x32.mpmd
r=$?
+ r=0
```

The same job on Marenostrum5 does the following:

```bash
# Run the components passing the different variables for mpiexec
echo "Start of the NICOCO and regridding mpiexec runs - $(date)"
++ date
+ echo 'Start of the NICOCO and regridding mpiexec runs - Tue Dec  2 15:02:21 CET 2025'
mpiexec -n "${NICAM_PROC}" "${NICAM_DRIVER}" : -n "${COCO_PROC}" "${COCO_DRIVER}" 2>&1
+ mpiexec -n 40 /gpfs/scratch/bsc32/bsc032060/aa7c/nicoco/build/NICAM/NICAM/bin/nhm_driver : -n 32 /gpfs/scratch/bsc32/bsc032060/aa7c/nicoco/build/NICAM/COCO/bin/MN5/cgcmO.t42l40hmed52full-a8o1x32.mpmd
mpiexec -n "${ICO2LL_PROC}" "${ICO2LL_DRIVER}" 2>&1
+ mpiexec -n 40 /gpfs/scratch/bsc32/bsc032060/aa7c/nicoco/build/NICAM/NICAM/bin/fio_netcdf_ico2ll_mpi
echo "End of the NICOCO and regridding mpiexec runs - $(date)"
++ date
+ echo 'End of the NICOCO and regridding mpiexec runs - Tue Dec  2 15:09:01 CET 2025'
```
Any ideas what could be happening or what could be done? CC: @dbeltrankyl

VindeeR:
@Lerriola I just finished the merge you can test with the fix implemented by @dbeltrankyl

dbeltrankyl:
Great! In fact this was a development made by @kinow I merely pushed it to 4.1.16

VindeeR:
I believe that, since it is working on simpler workflows, we can merge this PR, and in the eventual need of fix or change, we create new specific issues from now on @dbeltrankyl @kinow do you agree?

dbeltrankyl:
> @dbeltrankyl I noticed you had merged a 4.1.16 branch into master can I merge this branch too?

Ah, my bad, I thought I told you in this issue 😅 . Yes you can

dbeltrankyl:
But I think @kinow wanted to do a review

VindeeR:
> But I think @kinow wanted to do a review

No problem just checking to see if it can be progressed, maybe should have said reviewed not merged :rofl: 
Thanks :raised_hands: @kinow Could you review please?

VindeeR:
@dbeltrankyl I removed the wrapper from this branch and added it to a new branch #2926 
I'm doing tests on the hub trying to replicate the behavior

LuiggiTenorioK:
From Copilot :copilot: :

## PR #2668 Review: Semaphore Leakage & Process/Queue Termination Analysis

This PR implements the PBS platform for Miyabi and includes several fixes to the multiprocessing infrastructure in `platform.py` and `autosubmit.py`. Below is a comprehensive analysis of all semaphore/resource leakage risks and improper process/queue termination patterns.

---

### ✅ WHAT THE PR FIXES WELL

1. **`clean_log_recovery_process()` now calls `terminate()` before `join()`** (platform.py L846-848) — Previously, the code only called `join(timeout=60)` without ever sending `terminate()`, meaning a stuck child process could hang the parent for 60 seconds and then still remain alive as a zombie.

2. **Queue cleanup added** (platform.py L851-853) — `recovery_queue.close()` and `recovery_queue.join_thread()` are now called before recreating the queue. Previously, the old `multiprocessing.Queue` (which has an internal feeder thread and OS pipe file descriptors) was simply replaced, leaking the feeder thread and FDs each time `clean_log_recovery_process()` ran.

3. **Null checks for `work_event`** (autosubmit.py L2115, platform.py L857) — Prevents `AttributeError` when `work_event` is `None` (e.g. when `DISABLE_RECOVERY_THREADS=true` or before `spawn_log_retrieval_process` has been called).

4. **`is_alive()` guard before `join()`** (autosubmit.py L2369) — At experiment end, the exit path now checks `is_alive()` before calling `join()`, which is correct.

5. **Process reset to `None` instead of `self.ctx.Process()`** (platform.py L861) — Creating an empty `Process()` object was confusing and could mislead `is_alive()` checks.

---

### 🔴 CRITICAL ISSUES: Semaphore/Resource Leakage

#### Issue 1: `send_cleanup_signal()` will crash on `NoneType` (platform.py L941-948)

```python
def send_cleanup_signal(self) -> None:
    if self.log_recovery_process.is_alive():  # ← CRASHES if None!
        self.work_event.clear()               # ← CRASHES if None!
        self.cleanup_event.set()              # ← CRASHES if None!
        self.log_recovery_process.join(timeout=60)
```

This atexit handler is registered at line 904 during `prepare_process()`. After the PR's change at line 861 (`self.log_recovery_process = None`), if `clean_log_recovery_process()` runs before exit, the atexit handler will raise `AttributeError: 'NoneType' object has no attribute 'is_alive'`. Since `atexit` handlers swallow exceptions silently in CPython, this means **the cleanup signal is never sent and the child process may linger**.

**Additionally**, atexit handlers accumulate — every call to `prepare_process()` (i.e., every process respawn) registers a *new* pair of atexit handlers. There is no deduplication. After N respawns, there are N stale atexit handlers referencing old (already-cleaned) state.
#### Issue 11: `_exit(0)` in child process bypasses cleanup (platform.py L92, L1078)

The child process calls `os._exit(0)` which is intentional (to avoid running atexit handlers in the child), but this also means any queued data in the `recovery_queue`'s internal buffer that hasn't been flushed to the pipe will be lost. If the child put items in the queue but the feeder thread hasn't flushed them yet, those items are silently dropped.

---

### 🟢 AREAS THAT ARE CORRECT

- The `daemon = True` flag (L913) ensures child processes don't prevent parent exit
- The `spawn` context (L918) is correct for avoiding fork-related issues
- The `CopyQueue.put()` serialization via `__getstate__()` avoids sharing live objects
- The `wait_for_work()` / `wait_mandatory_time()` event loop pattern is reasonable

---

### SUMMARY: Risk Map

| Location | Issue | Severity | Type |
|----------|-------|----------|------|
| `platform.py:945-948` | `send_cleanup_signal()` crashes on `None` | 🔴 Critical | Process leak (orphaned child) |
| `platform.py:248` | `err.message` → `AttributeError` prevents event removal | 🔴 Critical | Event leak in `worker_events` |
| `platform.py:855-860` | `self.ctx` may be `None` on early cleanup | 🔴 Critical | Crash / unhandled exception |
| `platform.py:848` | `join(timeout=60)` with no kill fallback | 🔴 Critical | Zombie process leak |
| `platform.py:851-853` | Queue closed after `terminate()` | 🟡 Moderate | Potential `join_thread()` hang |
| `platform.py:134,239,249` | `worker_events` list unsynchronized | 🟡 Moderate | Race condition / stale events |
| `platform.py:904-905` | Unbounded atexit handler accumulation | 🟡 Moderate | Handler leak + stale references |
| `platform.py:900` | `del self.recovery_queue` without `close()` | 🟡 Moderate | FD / feeder thread leak |
| `autosubmit.py:2371` | `join()` with no timeout | 🟡 Moderate | Main process hang |
| `platform.py:856,862` | Duplicate flag assignment | 🟢 Minor | Code quality |

VindeeR:
> From Copilot :copilot: :
> ## PR #2668 Review: Semaphore Leakage & Process/Queue Termination Analysis
> 
> This PR implements the PBS platform for Miyabi and includes several fixes to the multiprocessing infrastructure in `platform.py` and `autosubmit.py`. Below is a comprehensive analysis of all semaphore/resource leakage risks and improper process/queue termination patterns.
> ### ✅ WHAT THE PR FIXES WELL
> 
>     1. **`clean_log_recovery_process()` now calls `terminate()` before `join()`** (platform.py L846-848) — Previously, the code only called `join(timeout=60)` without ever sending `terminate()`, meaning a stuck child process could hang the parent for 60 seconds and then still remain alive as a zombie.
> 
>     2. **Queue cleanup added** (platform.py L851-853) — `recovery_queue.close()` and `recovery_queue.join_thread()` are now called before recreating the queue. Previously, the old `multiprocessing.Queue` (which has an internal feeder thread and OS pipe file descriptors) was simply replaced, leaking the feeder thread and FDs each time `clean_log_recovery_process()` ran.
> 
>     3. **Null checks for `work_event`** (autosubmit.py L2115, platform.py L857) — Prevents `AttributeError` when `work_event` is `None` (e.g. when `DISABLE_RECOVERY_THREADS=true` or before `spawn_log_retrieval_process` has been called).
> 
>     4. **`is_alive()` guard before `join()`** (autosubmit.py L2369) — At experiment end, the exit path now checks `is_alive()` before calling `join()`, which is correct.
> 
>     5. **Process reset to `None` instead of `self.ctx.Process()`** (platform.py L861) — Creating an empty `Process()` object was confusing and could mislead `is_alive()` checks.
> 
> 
> ### 🔴 CRITICAL ISSUES: Semaphore/Resource Leakage
> #### Issue 1: `send_cleanup_signal()` will crash on `NoneType` (platform.py L941-948)
> 
> ```python
> def send_cleanup_signal(self) -> None:
>     if self.log_recovery_process.is_alive():  # ← CRASHES if None!
>         self.work_event.clear()               # ← CRASHES if None!
>         self.cleanup_event.set()              # ← CRASHES if None!
>         self.log_recovery_process.join(timeout=60)
> ```
> 
> This atexit handler is registered at line 904 during `prepare_process()`. After the PR's change at line 861 (`self.log_recovery_process = None`), if `clean_log_recovery_process()` runs before exit, the atexit handler will raise `AttributeError: 'NoneType' object has no attribute 'is_alive'`. Since `atexit` handlers swallow exceptions silently in CPython, this means **the cleanup signal is never sent and the child process may linger**.
> 
> **Additionally**, atexit handlers accumulate — every call to `prepare_process()` (i.e., every process respawn) registers a _new_ pair of atexit handlers. There is no deduplication. After N respawns, there are N stale atexit handlers referencing old (already-cleaned) state.
> #### Issue 11: `_exit(0)` in child process bypasses cleanup (platform.py L92, L1078)
> 
> The child process calls `os._exit(0)` which is intentional (to avoid running atexit handlers in the child), but this also means any queued data in the `recovery_queue`'s internal buffer that hasn't been flushed to the pipe will be lost. If the child put items in the queue but the feeder thread hasn't flushed them yet, those items are silently dropped.
> ### 🟢 AREAS THAT ARE CORRECT
> 
>     * The `daemon = True` flag (L913) ensures child processes don't prevent parent exit
> 
>     * The `spawn` context (L918) is correct for avoiding fork-related issues
> 
>     * The `CopyQueue.put()` serialization via `__getstate__()` avoids sharing live objects
> 
>     * The `wait_for_work()` / `wait_mandatory_time()` event loop pattern is reasonable
> 
> 
> ### SUMMARY: Risk Map
> Location 	Issue 	Severity 	Type
> `platform.py:945-948` 	`send_cleanup_signal()` crashes on `None` 	🔴 Critical 	Process leak (orphaned child)
> `platform.py:248` 	`err.message` → `AttributeError` prevents event removal 	🔴 Critical 	Event leak in `worker_events`
> `platform.py:855-860` 	`self.ctx` may be `None` on early cleanup 	🔴 Critical 	Crash / unhandled exception
> `platform.py:848` 	`join(timeout=60)` with no kill fallback 	🔴 Critical 	Zombie process leak
> `platform.py:851-853` 	Queue closed after `terminate()` 	🟡 Moderate 	Potential `join_thread()` hang
> `platform.py:134,239,249` 	`worker_events` list unsynchronized 	🟡 Moderate 	Race condition / stale events
> `platform.py:904-905` 	Unbounded atexit handler accumulation 	🟡 Moderate 	Handler leak + stale references
> `platform.py:900` 	`del self.recovery_queue` without `close()` 	🟡 Moderate 	FD / feeder thread leak
> `autosubmit.py:2371` 	`join()` with no timeout 	🟡 Moderate 	Main process hang
> `platform.py:856,862` 	Duplicate flag assignment 	🟢 Minor 	Code quality

Thanks for the help with that @LuiggiTenorioK :heart: I'll open a PR to treat the pointed issues in a different branch avoiding this branch's delay to merge

VindeeR:
@dbeltrankyl I think this PR can be reviewed now, I'll open a different PR to address the things addressed by @LuiggiTenorioK

dbeltrankyl:
Hello @VindeeR 

Do you mind if we first look this one? #2939 

I can help with a patch to adapt this pr after the other one if needed

dbeltrankyl:
I'll rebase it and adapt the changes for tghe new master

dbeltrankyl:
Hello @VindeeR 

I think it should be working now with the new flow https://github.com/BSC-ES/autosubmit/pull/2668/changes/4b9a16bd9fd68248946b4b80b839f20fa2464b1f#diff-7f963dd467fb190257e1feb6aab085ab18887cfa58a36b215c7fc68906cf090cR283 use this view to see what the commit changed

Also fixed some typos etc

Issue #2432, opened 2025-06-25
HPC* variables randomly missing


kinow:
### Description

From https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/issues/1125 (link protected, but will summarize here to non-BSC'ers).

@ainagaya gathered three cases where variables were not replaced by AS. In all of these cases, eventually the variables were replaced, without anything obvious done by users that could have caused this.

* Autosubmit version: 4.1.14 (probably 4.1.15)
* Machine/VM/environment name (if applicable): destine
* Experiment ID (if applicable): t062, a2j1, t05s

### Reproducible Example

This is the tricky part.

### Expected Behaviour

The behavior is consistent (either works, or doesn't)

kinow:
Thanks Dani!

kinow:
@dmcclainbsc reported having the same problem in the testing suite 2 weeks ago, on August 26. Internal links:

- https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/merge_requests/789#note_388622
- https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/issues/1220#note_388592

Experiment t063, with 4.0.14. Thanks for letting us know, Damien!

kinow:
And I think this issue may have bitten @mandresm 4 days ago, in DestinE experiment `t05t`. Internal link: https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/issues/1160#note_391984

mandresm:
> And I think this issue may have bitten [@mandresm](https://github.com/mandresm) 4 days ago, in DestinE experiment `t05t`. Internal link: https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/issues/1160#note_391984

I have just cleaned the experiment with the testing_suite `--clean` command (sorry), so there is not much left to check. I am running it again, I'll report if the issue persists

kinow:
> Hello! I believe this is another example:
> 
> /appl/AS/AUTOSUBMIT_DATA/t096/tmp/LOG_t096/t096_19900101_fc0_2_SIM.20250918121211.err
> 
> HPCROOTDIR was not substituted which led to failure.

@dmcclainbsc I had a look at your experiment, and cloned it into a new experiment in my local laptop. A quick create + inspect, showed me that the `HPCROOTDIR`appears to be created in that template file as

```bash
set -xuve

# HEADER

[...]

PR #2669, opened 2025-10-22
Fix wrapper issue and maybe another one

dbeltrankyl:
**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/config/configcommon.py
+++ b/autosubmit/config/configcommon.py
@@ -2840,7 +2840,7 @@ class AutosubmitConfig(object):
             content = content.replace(match, parameters.get(match[1:-1], ""))
         return content
 
-    def get_current_wrapper(self, section):
+    def get_current_wrapper(self, section: str) -> dict:
         """Returns the wrapper configuration for a given job section.
 
         :param section: job section
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -2290,7 +2290,6 @@ class Job(object):
         for event in self.platform.worker_events:  # keep alive log retrieval workers.
             if not event.is_set():
                 event.set()
-        parameters.update(as_conf.default_parameters)
         return parameters
 
     def init_platform(self, as_conf: AutosubmitConfig) -> None:
--- a/test/unit/test_job.py
+++ b/test/unit/test_job.py
@@ -1718,7 +1718,7 @@ def test_update_dict_parameters_invalid_script_language(platform_name: Optional[
     [None, "", "  ", "some-string", "a", "123", "True"],
     ids=["None", "empty", "spaces", "some-string", "a", "123", "True"]
 )
-def test_job_parameters(reservation: Optional[str], tmp_path: Path) -> None:
+def test_job_parameters(reservation: Optional[str], tmp_path: Path, autosubmit_config) -> None:
     """
     Parametrized test for job reservation propagation.
 
@@ -1768,8 +1768,7 @@ def test_job_parameters(reservation: Optional[str], tmp_path: Path) -> None:
     basic_config = FakeBasicConfig()
     basic_config.read()
     basic_config.LOCAL_ROOT_DIR = str(tmp_path)
-
-    config = AutosubmitConfig(expid, basic_config=basic_config, parser_factory=YAMLParserFactory())
+    config = autosubmit_config(expid, basic_config=basic_config)
     config.reload(True)
     parameters = config.load_parameters()
 
@@ -1963,6 +1962,7 @@ def test_job_parameters_resolves_all_placeholders(autosubmit_config, monkeypatch
         },
     }
     as_conf.experiment_data = as_conf.experiment_data | additional_experiment_data
+    # Needed to monkeypatch reload to avoid overwriting experiment_data ( the files doesn't exist in a unit-test)
     monkeypatch.setattr(as_conf, 'reload', lambda: None)
     job = Job(_EXPID, '1', Status.WAITING, 0)
     job.section = 'TEST_JOB_2'


autosubmit/job/job.py
```diff
@@ -1564,48 +1564,47 @@ def get_metric_folder(self, as_conf: AutosubmitConfig) -> str:
...
+        parameters['CURRENT_TYPE'] = parameters.get('CURRENT_TYPE', self.platform.type)
+        parameters['CURRENT_SCRATCH_DIR'] = parameters.get('CURRENT_SCRATCH_DIR', self.platform.scratch)
+        parameters['CURRENT_PROJ_DIR'] = parameters.get('CURRENT_PROJ_DIR', self.platform.project_dir)
+        parameters['CURRENT_ROOTDIR'] = parameters.get('CURRENT_ROOTDIR', self.platform.root_dir)
+        parameters['CURRENT_LOGDIR'] = parameters.get('CURRENT_LOGDIR', self.platform.get_files_path())
+
         for key, value in as_conf.jobs_data[self.section].items():
             parameters[f"CURRENT_{key.upper()}"] = value
 
+        for key, value in as_conf.get_current_wrapper(self.section).items():
```
dbeltrankyl: Added current wrappers values ( has the highest prio)

autosubmit/job/job.py
```diff
@@ -1564,48 +1564,47 @@ def get_metric_folder(self, as_conf: AutosubmitConfig) -> str:
 
     def update_current_parameters(self, as_conf: AutosubmitConfig, parameters: dict) -> dict:
```
dbeltrankyl: Should we update HPC* ones there? Currently, it is done only in the as_conf.reload() #2432 

CC: @kinow it would be a "cannon" way of solving the #2432 issue if you don't find a way of properly explaining why it happens ( better this than nothing...)

test/unit/test_job.py
```diff
@@ -698,109 +595,6 @@ def test_total_processors(self):
             self.job.nodes = test['nodes']
             assert self.job.total_processors == test['expected']
 
-    def test_job_script_checking_contains_the_right_variables(self):
```
dbeltrankyl: This test was testing.. nothing at all? too many mocks I've added a new one

test/unit/test_job.py
```diff
@@ -1918,3 +1712,274 @@ def test_update_dict_parameters_invalid_script_language(platform_name: Optional[
...
+        template_content, additional_templates = job.update_content(config, parameters)
+        assert not additional_templates
+        assert f'#SBATCH --reservation={reservation}' in template_content
+
+
+def test_job_parameters_resolves_all_placeholders(autosubmit_config, monkeypatch):
+
+    as_conf = autosubmit_config('t000', {})
+
+    additional_experiment_data = {
```
dbeltrankyl: Extracted and simplified version of a real case

test/unit/test_job.py
```diff
@@ -1918,3 +1712,274 @@ def test_update_dict_parameters_invalid_script_language(platform_name: Optional[
...
+            "WRAPPER_0": {
+                "JOBS_IN_WRAPPER": "TEST_JOB_2",
+                "MAX_WRAPPED": 2,
+                "TYPE": "vertical",
+                "WRAPPER_HAS_PRIO": "whatever_from_wrapper",
+            }
+        },
+    }
+    as_conf.experiment_data = as_conf.experiment_data | additional_experiment_data
+    monkeypatch.setattr(as_conf, 'reload', lambda: None)
```
dbeltrankyl: I'm not writting files because it is a unit test. This is the only mock so the reload doesn't overwrite the experiment_data set

dbeltrankyl:
At the end, the change was quite complicated because it was breaking working stuff. 

The problem was with the order of loading the custom_directives and also the dynamic variables dict was never emptied, even if no futher variables could be replaced. 

Originally I thought that the problem was the placeholders in the wrapper cmd, but the real issue there is that it mustn't have any placeholder starting with ( they should be already resolved). Probably we will need to remove the find placeholders call of that part of the code but we can do it in another PR since doing it was breaking tests.

Ready to review cc: @kinow , @VindeeR

test/integration/conftest.py
```diff
@@ -470,7 +469,8 @@ def slurm_server(session_mocker, tmp_path_factory):
 
     with docker_container \
             .with_env('TZ', 'Etc/UTC') \
-            .with_bind_ports(2222, ssh_port) as container:
+            .with_bind_ports(2222, ssh_port) \
+            .with_name(container_name) as container:
```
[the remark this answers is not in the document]
kinow: Also not sure why that would happen, but as there is a `with_name` function, and the other approach passes to `kwargs` which passes to other functions and at some point uses it (or not), I think your change to `.with_name` is simpler if that works (i.e. if build is green).

Thanks Dani!

autosubmit/config/configcommon.py
```diff
@@ -2838,3 +2839,16 @@ def parse_placeholders(content, parameters) -> str:
             # replace all '%(?<!%%)\w+%(?!%%)' with parameters value
             content = content.replace(match, parameters.get(match[1:-1], ""))
         return content
+
+    def get_current_wrapper(self, section):
```
kinow:
```suggestion
    def get_current_wrapper(self, section: str) -> dict:
```

test/unit/test_job.py
```diff
@@ -1918,3 +1712,274 @@ def test_update_dict_parameters_invalid_script_language(platform_name: Optional[
...
+            TEMP_DIR: ''
+    '''))
+
+    # bootstrap config and generate jobs
+    basic_config = FakeBasicConfig()
+    basic_config.read()
+    basic_config.LOCAL_ROOT_DIR = str(tmp_path)
+
+    config = AutosubmitConfig(expid, basic_config=basic_config, parser_factory=YAMLParserFactory())
+    config.reload(True)
```
kinow: It didn't work using `autosubmit_config` here? ☝️

dbeltrankyl:
Thanks @kinow !

I've applied the modifications, if it passes I'll squash & merge it

kinow (approved):

PR #2673, opened 2025-10-24
Dependencies fixes: (multi-day-apps) but, it could fix/break more at once 

dbeltrankyl:
**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -424,18 +424,21 @@ class JobList(object):
 
         :return:
         """
+        from itertools import combinations
+        from networkx import NetworkXError
+
         delete_relations = set()
         for section, jobs in problematic_jobs.items():
             for child_name, parents in jobs.items():
-                for parent_name in parents:
-                    for another_parent_name in list(parents)[1:]:
-                        if self.graph.has_successor(parent_name, another_parent_name):
-                            delete_relations.add((parent_name, child_name))
-                        elif self.graph.has_successor(another_parent_name, parent_name):
-                            delete_relations.add((another_parent_name, child_name))
+                parents_list = list(parents)
+                for parent_name, another_parent_name in combinations(parents_list, 2):
+                    if self.graph.has_successor(parent_name, another_parent_name):
+                        delete_relations.add((parent_name, child_name))
+                    elif self.graph.has_successor(another_parent_name, parent_name):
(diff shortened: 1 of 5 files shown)


dbeltrankyl:
Ok, found the bug. 

There is a part of the code that revises each of the "sections" of the current dependency.

Then it checks the dependencies of the [1,N] Dependency.

The objective is to add dependencies without redundancy: if the dependency you set depends on another section you've already set in your dependencies that you're referring to, Autosubmit would skip it since the previous job already has that dependency set. 

But, this only takes into account a section without "distances". So it is skipping dependencies that shouldn't be skipped. 

This was added because some workflows have the dependency defined without the need of having it defined.. but it doesn't make sense to mitigate it in the code. 

So, from the user's perspective, it may have been added for a reason or by mistake, but it should be the user who sets it as he/she want. 

So I'll just remove that part of the code.. and when I do that. Only one test broke ( auto-monarch with crossdates). I'm looking at it

autosubmit/job/job_list.py
```diff
@@ -1417,20 +1436,33 @@ def _manage_job_dependencies(self, dic_jobs, job, date_list, member_list, chunk_
...
+            chunk: Optional[int],
+            chunk_list: List[int],
+            member: Optional[str],
+            member_list: List[str],
+            date: Optional[str],
+            date_list: List[str],
+            dependency: Any
+    ) -> Tuple[bool, Tuple[Optional[int], Optional[str], Optional[str]]]:
+        """
+        Compute the target chunk/member/date for a dependency with a +/- distance.
```
kinow: Would be good if we started documenting the code following PEP 257, where the first comment is a very short note about the code. Followed by a blank line, then the body of the comment.

It's actually parsed that way, with a title for the comment, followed by the body. And then args/returns/etc. using Google/napoleon'etc. styles.

But this

```
"""
Bla.
"""
```

is not the same as 

```
"""Bla."""
```

It's similar to the example in this part, https://peps.python.org/pep-0257/#handling-docstring-indentation

> The docstring in this example contains two newline characters and is therefore 3 lines long. The first and last lines are blank

It's harmless, but eventually we might want to have it the way of that PEP for better integration with tools like Sphinx, and also for consistency, I think.

dbeltrankyl:
Thanks! 

Rebased and squashed, will be merged when pipeline success

Issue #2666, opened 2025-10-22
Dependencies bugs tracking


dbeltrankyl:
Hello 

This is mainly for me, that is the one that will work on the dependencies issues. 

I'll tag all of them there because touching that code means that something we fix we could broke for other workflow  ( or maybe the bug is the same for all the issues) 

As like other times I fixed something about the dependencies, I'll add the workflow to the regression tests:

Destine:
* [ ] https://github.com/BSC-ES/autosubmit/issues/2631
* [ ] https://github.com/BSC-ES/autosubmit/issues/2639 

Bsc:
* [ ] https://github.com/BSC-ES/autosubmit/issues/2565
* [ ] https://github.com/BSC-ES/autosubmit/issues/2518
* [ ] https://github.com/BSC-ES/autosubmit/issues/2466
* [ ] https://github.com/BSC-ES/autosubmit/issues/2397

dbeltrankyl:
All issues are now resolved and merged into master, closing this

PR #2680, opened 2025-10-27
All filter acting wrong. #2565

### LINKED ISSUE #2565: DEPENDENCIES configuration
* Autosubmit version: 4.1.15-conda
* Machine: hub03
* Experiment ID: a9r1

Hi AS team!

I am trying to configure a job that, for each chunk index `i`, `i_JOB2` depends on `i_JOB1` for all members. In other words, I want something similar to:

A job `j_i_JOB2` depends on `x_i_JOB1` for every x in MEMBERS, where `i` is a CHUNK and `j` is a MEMBER.

Here are the configurations I have tried with 2 MEMBERS and 2 or 3 CHUNKS. In both cases, using CHUNKS_FROM.ALL or CHUNKS_FROM._NUMBER_, the result is the same workflow.  It seems like the configuration I want is only applied to the last chunk:

Image

I've also tested this with 3 MEMBERS and 3 CHUNKS, and once again, what I want only happens in the last chunk:

Image

Could you please help me with this?

Feel free to ask if something is not clear, it's a bit tricky to explain :leaves:

Thanks :cherry_blossom:

dbeltrankyl:
closes https://github.com/BSC-ES/autosubmit/issues/2565.

Part of https://github.com/BSC-ES/autosubmit/issues/2666

closes #2565.
**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -363,9 +363,9 @@ class JobList(object):
 
     def _add_dependencies(
             self,
-            date_list: List[Any],
-            member_list: List[Any],
-            chunk_list: List[int],
+            date_list: list[Any],
+            member_list: list[Any],
+            chunk_list: list[int],
             dic_jobs: DicJobs,
             option: str = "DEPENDENCIES",
     ) -> None:
@@ -374,7 +374,7 @@ class JobList(object):
         Iterate experiment `JOBS` sections to:
         - build deep dependency maps (with and without distance metadata),
         - compute and attach edges for each job,
-        - add dependencies that couldn't (or not solved yet) be safetly added for later pruning,
+        - add dependencies that couldn't (or not solved yet) be safely added for later pruning,
         - add per-edge metadata to job objects.
 
         :param date_list: List of dates used by the experiment.


dbeltrankyl:
Ready to review, just to mention that I did the less invasive changes.

autosubmit/job/job_list.py
```diff
@@ -353,23 +353,52 @@ def _deep_map_dependencies(self, section, jobs_data, option, dependency_list=set
...
+        for sep in ("+", "-"):
+            if sep in dep:
+                return dep.split(sep, 1)[0]
+        return dep
+
+    def _add_dependencies(
+            self,
+            date_list: List[Any],
+            member_list: List[Any],
+            chunk_list: List[int],
```
kinow: s/List/list 😬

kinow (commented):
@dbeltrankyl I was looking inside `_deep_map_dependencies`, and noticed that they also strip keys there, so I thought we could re-use the new function you wrote for that, reducing duplicated code.

But, there's something that I don't understand in that function.

It takes a `strip_keys: bool`, defaulting to `True`. Which is fine. I take it it defines whether you want to strip keys or not.

It ignores and returns immediately if section already in list, which makes sense (recursivity), or adds the section to the list.

The next step is then to check, if `strip_keys` is **`False`**, then it checks if the section contains `+` or `-`, and then does the `section.split` (i.e. what your `_strip_key` function is doing, I think).

Then, for each dependency of the section it checks if `strip_keys` is **`True`**, hen it calls the `dependency.split`, which I thought would be doing the same thing as your function.

Do you know why first `strip_keys` being `False` triggers the code to split by `+`, `-`, and then in the same code if `strip_keys` is `True` is what triggers that again? 😵

dbeltrankyl:
@kinow

> Do you know why first strip_keys being False triggers the code to split by +, -, and then in the same code if strip_keys is True is what triggers that again? 😵

Yeah, it's confusing, and there should be a better way to do it, for sure. 

Context:

The outcome of the function should be this: 

What this is supposed to do is navigate through 

JOBS.JOB.DEPENDENCY

These keys can have distances ( sim-1, sim+1) 

There are two maps, one with distances and another one without them

The one without distances is straightforward; strip all the keys always

ex:

current_section:
map["section"] = ("other_section")
Then we want to navigate to other_section
map["other_section"] = ("other_other_section")

and so on 

The one with distances, 

current_section:
map_d["section"] = ("other_section-1")
Then we want to navigate to other_section, but we need to strip the other_section for the key name
map_d["other_section"] = ("other_other_section",)

---------------------------------

Debug:

The first call (i=0), always has the section stripped as it comes from the jobs_data

So, this can be ignored for i=0

```python
        if not strip_keys:
            if "+" in section:
                section = section.split("+")[0]
            elif "-" in section:
                section = section.split("-")[0]
```

This one is for the dependencies, here we can have +/-

```python
        for dependency in dependencies_keys:
            if strip_keys:
                if "+" in dependency:
                    dependency = dependency.split("+")[0]
                elif "-" in dependency:
                    dependency = dependency.split("-")[0]
```

map: i=1: new_section= dependency_stripped 
map_distances: i=1: the new_section=dependency can have +/-

So, this in map_distances case i=1

```python
        if not strip_keys:
            if "+" in section:
                section = section.split("+")[0]
            elif "-" in section:
                section = section.split("-")[0]
```

will strip the key from the section, so we can find it in the jobs_data dict ( jobs.job)

kinow (approved):
Added a comment about a typo, and types that can use `list` instead of `List`. I don't know if your new small function to strip keys can be used in the other function, Dani, but that can be done later as well. Feel free to merge as is or fix anything of these small issues if you have time. Thanks!

kinow:
(Moved the closes #issuenumber to main comment as I **think** it only works there?)

PR #2619, opened 2025-10-01
4.2.0 branch -> Joblist on db, tests, main run logic changes, Status: reworked, recovery/setstatus

dbeltrankyl:
Continues https://github.com/BSC-ES/autosubmit/pull/2187 because https://github.com/BSC-ES/autosubmit/pull/2187 head will be set to the first commit. 

cc @kinow

PR #2694, opened 2025-11-05
Avoid Autosubmit's bash script interpreting task's script and provoking all variables to be unbounded

### LINKED ISSUE #2718: Bash template causes unbound variables in 4.1.16
### Description
When testing rocrate feature with autosubmit/4.1.16 in the climateDT dev VM found a weird behavior causing an unbound variable error in all the jobs.
We tried submitting the scripts manually and they failed the same way.
The main difference I see with the previous versions is the following header which it was not appearing before when running the workflow and i suspect is the root of the problem:
```
r=0
+ r=0
set +e
+ set +e
bash -e <<__AS_CMD__
set -xuve
```
What we see in the err file is that the whole script is printed and that the job fails because of unbound variables.
What usually happens is that you get  the execution printed line by line (set -xuve) but in this case we don't see anything like this. Just the whole printed script and the unbound variable error. 
To double check I also tried running the autoHPCW workflow and got exactly the same error.

Additionally, we tried submitting remote jobs and it failed the same way.
* Autosubmit version: autosubmit/4.1.16-dev-e54728b
* Machine/VM/environment name (if applicable): climateDT-wf-dev1
* Experiment ID (if applicable): a001 a002
  /appl/AS/AUTOSUBMIT_DATA/a001/tmp/LOG_a001/a001_LOCAL_SETUP.20251124103818.err
  /appl/AS/AUTOSUBMIT_DATA/a001/tmp/LOG_a001/a001_19900101_REMOTE_SETUP.20251125153214.err

### Reproducible Example

You can copy a001 and run it. It fails in the local_setup.

cc @ainagaya @dbeltrankyl @kinow @VindeeR

manuel-g-castro:
fixes #2718 

Hi, @kinow @dbeltrankyl @VindeeR 

This merge request solves a critical issue that breaks every single workflow we have. 

The problem is that now the task's bash script is passed as a [heredoc](https://www.gnu.org/software/bash/manual/bash.html#Here-Documents), but it is not guarded by quotes. therefore the bash script executing the Autosubmit code (header, tailer, etc) makes the parameter and variable substitution.

As far as I could tell, this bug was introduced by 

https://github.com/BSC-ES/autosubmit/blob/761c2084db40213bba1a79492af68ed938d336b9/autosubmit/job/template/bash.py#L95-L98

I am marking this PR as draft because I would like for someone (@VindeeR ?) to write a test to check if the variables are being substituted correctly, so that the CI/CD is able to detect similar errors. 

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ use it without a failure, but there will be a warning displayed asking you to
 update your command line. This argument will be completely removed on a later
 release.
 
+**Known issues:**
+
+- Due to a new heredoc blocks to handle failures in the templates, `script.cmd.err` line numbers are offset by ~5 lines (e.g. reported line 10 → actual line 15). #2694 #2718
+- Please adjust accordingly when debugging.
+
 **Bug fixes:**
 
 - Fixed issue with the verification of dirty Git local repositories in operational experiments #2446
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2335,6 +2335,8 @@ class Autosubmit:
                     Log.warning('Git operational check disabled by user')
 
                 Log.debug("Running main running loop")
+                Log.warning("Known issue: Due to recent changes in Autosubmit's script generation, error line numbers in "
+                            "`script.cmd.err` files may be offset by ~5 lines. Please adjust accordingly when debugging.")
                 did_run = False
                 #########################
                 # AUTOSUBMIT - MAIN LOOP


dbeltrankyl:
Hello @manuel-g-castro 

I had the same bug, but I didn't notice it was applying to more stuff that the checkpoint function. 

These are the changes I have on 4.2.0

https://github.com/BSC-ES/autosubmit/pull/2619/files#diff-0d5f69819455cf3b8eed9082a76824477f4c4e91fadd27403d6e9561aa8d13b7

https://github.com/BSC-ES/autosubmit/blob/4450cd912589d7522b7056463c14c2490b47b27b/autosubmit/job/template/bash.py

manuel-g-castro:
Hi, @kinow ! Thanks for the comment.

> Now, unfortunately, users must look at the lines in the final script, not in their original scripts.

I do not think this was ever the case. For me, at least, every time that I need to debug a script of mine, I need to check the `.cmd` script generated by Autosubmit. Otherwise the lines will be mismatched with the ones from my template script. 

I think I am not being clear on my point of view. So let me explain an example.

This is my script, and because of the `set -xuve`, it fails because of unbounded variables:

```
#!/bin/bash

set -xuve
echo ${UNBOUNDED_VARIABLE}
```

If i execute the workflow with this script, i get the following error on the .err:

```
/app/autosubmit/experiments/a000/tmp/LOG_a000/a000_LOCAL_SETUP.cmd: line 45: UNBOUNDED_VARIABLE: unbound variable
```

which correctly indicates the faulty line at the 45ft line of the .cmd. Below you have the snipped of the script (with numbering).

```
    38	###################
    39	# Autosubmit job
    40	###################
    41	
    42	#!/bin/bash
    43	
    44	set -xuve
    45	echo ${UNBOUNDED_VARIABLE}
    46	
    47	###################
    48	# Autosubmit tailer
    49	###################
```

Now, if I do the same with the changes in master (`c80654`), the error points to the line that executes bash -- 46 --, but not the line where the error actually happens in the script  -- 51 --:

```
/app/autosubmit/experiments/a000/tmp/LOG_a000/a000_LOCAL_SETUP.cmd: line 46: UNBOUNDED_VARIABLE: unbound variable
```

Here is the snipped of the final .cmd script (with line numbers).

```
    46	bash -e <<__AS_CMD__
    47	set -xuve
    48	#!/bin/bash
    49	
    50	set -xuve
    51	echo ${UNBOUNDED_VARIABLE}
    52	
    53	__AS_CMD__
```

kinow:
Thanks @dbeltrankyl , @manuel-g-castro ! I can review these changes and add more tests if needed. Thanks a lot!!! :bow:

dbeltrankyl:
Hello @manuel-g-castro 

I told our DestinE co-workers to test this branch, and it is working fine for them. 

We can merge this as it is (I'll add a warning somewhere, tho, about the line discrepancy... not sure where? Maybe at the start of the run? And for 4.1.17 or 4.2.1, try to fix the line discrepancy. 

Mainly because without this, I think not all scripts will work well 

What do you think?

manuel-g-castro:
Okay! No problems. I can open an issue for the line discrepancy so we do not forget about it.

PR #2705, opened 2025-11-11
Fixes template issue #2426

### LINKED ISSUE #2426: Variable not correctly substituting in additional file
### Description

Hi. I am trying to enable multiple start dates within one experiment for DestinE. However, AQUA FDB files are silently failing to generate in Remote Setup because Autosubmit is not recognizing the variable (either %CHUNK_START_DATE% or %SDATE%) in the additional file aqua-True. This later causes LRA Generator jobs to fail when they cannot find the files.

* Autosubmit version: 4.1.14
* Machine/VM/environment name (if applicable): MN5
* Experiment ID (if applicable): t06y
  * /appl/AS/AUTOSUBMIT_DATA/t06y/tmp/LOG_t06y/t06y_19900101_REMOTE_SETUP.20250616090605.err
  * I tried many different versions of this variable (see [here](https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/merge_requests/859#note_374930)) but none of them worked.

### Reproducible Example

See experiment t06y.

### Expected Behaviour

CHUNK_START_DATE or SDATE should be pulled from the current run and be used in the AQUA file generation in Remote Setup. For example, if the experiment datelist is '199010101 20200101', and the current sim is 20200101, the AQUA files should generate with only 20200101 as the date (eg config_catalog_20200101_REMOTE_SETUP).

dbeltrankyl:
Closes #2426

**Check List**

- [x] Documentation updated.

### DIFF
--- a/test/unit/test_job.py
+++ b/test/unit/test_job.py
@@ -2278,9 +2278,9 @@ def test_job_parameters_resolves_all_placeholders(autosubmit_config, monkeypatch
     assert parameters["TESTDATES.START_DATE"] == "20200101"
     assert parameters["TESTDATES.START_DATE_WITH_SPECIAL"] == "20200101"
     assert parameters["EXPERIMENT.DATELIST"] == 20200101
-    # TODO: This should be a list, but it isen't. Also adding more that one element in the list is not working either
-    # TODO: This issue doesn't come from this PR and here it was added to test other part of the code
-    # TODO: Needs to be fixed in another PR (as_conf.substitute_dynamic_variables). There is an issue already for that
+    # TODO: This should be a list, but it isn't. Also, adding more than one element to the list is not working either.
+    # TODO: This issue isn't caused by this PR, and it was added here to test another part of the code.
+    # TODO: Needs to be fixed in another PR (as_conf.substitute_dynamic_variables). There is already an issue for that.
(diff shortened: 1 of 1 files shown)


dbeltrankyl:
> There's just one thing I notice on your test, is that it only test for the case of the STARTDATES being a list and your code does and if that is not checked.

Hm, I think that parameter was already in the config of the test, (not sure why tho) 

But in this PR I'm testing 

CHUNK_START_DATE ( which can't be a list)
TESTDATES.START_DATE
TESTDATES.START_DATE_WITH_SPECIAL
EXPERIMENT.DATELIST ( which can't be a list right now just a string separated by " " although eventually it should be possible )

VindeeR:
> > There's just one thing I notice on your test, is that it only test for the case of the STARTDATES being a list and your code does and if that is not checked.
> 
> Hm, I think that parameter was already in the config of the test, (not sure why tho)
> 
> But in this PR I'm testing
> 
> CHUNK_START_DATE ( which can't be a list) TESTDATES.START_DATE TESTDATES.START_DATE_WITH_SPECIAL EXPERIMENT.DATELIST ( which can't be a list right now just a string separated by " " although eventually it should be possible )

Sorry, I think I had a brain fart moment when I wrote that, I meant to say that your test only checks if they're a `string` but it doesn't go to the `else` where it checks if it is a `list`

image

Issue #2697, opened 2025-11-06
Wrapper engine integration


pablogoitia:
Hello, Autosubmit team 👋🏻

I am opening this issue in order to address possible questions about the wrapper engine integration process. The tools that we have chosen are [Flux](https://flux-framework.org/) and [Parsl](https://parsl-project.org/), being Flux the first one to be integrated.

By now, what I am doing is to debug Autosubmit's `create` and `run` commands to find and understand the components involved in the wrapper creation and its later submission. The idea is, once I have taken my notes, to review how Flux's submission batch scripts work and start making some small adaptations to cover elemental cases (for example, running a vertical wrapper).

If possible, and if it does not consume much time, I would like to temporally implement Flux as another [wrapping method](https://autosubmit.readthedocs.io/en/latest/userguide/wrappers/index.html#method), as AS_Thread or Srun, to facilitate debugging, integration, and testing for myself.

dbeltrankyl:
> Determine the cause why AS is sending wrong job lists to the wrapper builder. Is it AS' fault or is it mine? Have I declared correctly the dependencies in the configuration file? Is the multi-section case natively supported for vertical-horizontal wrappers?

It is an AS fault. 

> To implement horizontal-vertical wrappers.

I think you will face similar issues 

In https://github.com/BSC-ES/autosubmit/pull/2619 it should be fixed ( not 100% sure)

Maybe you can just focus on implementing it, and once it's in master or as a test in a different branch, I can rebase for 4.2 and launch your tests there

pablogoitia:
> It is an AS fault.

Thanks for checking it, @dbeltrankyl!

> I think you will face similar issues

No problem because I think for now I can just check whether the jobs are running in the same order that appears in the lists.

> Maybe you can just focus on implementing it, and once it's in master or as a test in a different branch, I can rebase for 4.2 and launch your tests there

Sounds good to me, thanks Dani

pablogoitia:
This could be future work: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_26.html

pablogoitia:
[...]

PR #2708, opened 2025-11-13
Implementation of the new delegated wrappers and Flux integration

### LINKED ISSUE #2968: A new "adaptative" and delegated wrapper type allowing dependencies between sections
### Motivation

Now that Flux has been integrated as an _in situ_ engine for task aggregation, it has been proposed to extend the current implementation to support "adaptive" or "delegated" hybrid wrappers that can handle sections with cross-dependencies, which are not supported by the current hybrid wrapper types (vertical-horizontal and horizontal-vertical).

The idea behind this new type of wrapper is that it adapts to the structure of the tasks it contains, abstracting the workflow developer from the choice of the wrapper type that best suits the execution needs, and making the wrapper creation more flexible.

### Conceptualization

For this purpose, a new type of wrapper will be created, compatible with current wrapping policies and operating exclusively under the `flux` method. The communication between Autosubmit and Flux will be done through "standardized" data structures that facilitate the integration of new tools for _in situ_ wrapper management if necessary in the future.

The work will be based on the existing implementation in PR #2708 (issue #2697).

pablogoitia:
Closes #2968 

Old issue: #2697

**Check List**

- [x] Documentation updated.


autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -121,6 +122,316 @@ def _indent(self, text, amount, ch=' '):
...
+
+            # Check if the job completed successfully
+            if [ -f "${{job_name}}_COMPLETED" ]; then
+                echo "The job $job_name has been COMPLETED"
+            else
+                echo "The job $job_name has FAILED"
+                touch "${{job_name}}_FAILED"
+                wrapper_failed=1
+            fi
+        done
```
manuel-g-castro: Should you not also add the STAT file? 

Although I am not sure if these is one of the files that will stop being generated. Can you clarify, @kinow?

autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -121,6 +122,316 @@ def _indent(self, text, amount, ch=' '):
...
+
+                # Check if the job completed successfully
+                if [ -f "${{job_name}}_COMPLETED" ]; then
+                    echo "The job $job_name has been COMPLETED"
+                else
+                    echo "The job $job_name has FAILED"
+                    touch "${{job_name}}_FAILED"
+                    touch "WRAPPER_FAILED"
+                    return 1
+                fi
```
manuel-g-castro: Same comment as before. Should you not also generate the STAT files?

autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -121,6 +122,316 @@ def _indent(self, text, amount, ch=' '):
...
+            return 0
+        }}
+
+        # Execute vertical wrappers in parallel
+        for ((i = 0; i < ${{#scripts[@]}}; i++)); do
+            execute_vertical_wrapper "${{scripts[$i]}}" &
+        done
+
+        # Wait for all vertical wrappers (subprocesses) to finish
+        wait
```
manuel-g-castro: Woah! Did not knew that there was this handy barrier in bash to wait for processes in the background!!

Today I learned!

autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -121,6 +122,316 @@ def _indent(self, text, amount, ch=' '):
...
+
+        # Debug commands
+        # flux resource list
+        """).format(self.processors, ' '.join(str(s) for s in self.job_scripts), '\n'.ljust(13))
+
+class FluxVerticalHorizontalWrapperBuilder(FluxWrapperBuilder):
+    # TODO: [ENGINES] Check error handling behavior
+    # TODO: [ENGINES] Delete the "flux resource list" after testing
+    # TODO: [ENGINES] Observe what happens if a single job arrives. Is it inside another list?
+    # TODO: [ENGINES] Should multi-section wrappers be supported?
```
manuel-g-castro: What do you mean with multi-section?

manuel-g-castro (commented):
Overall I am really happy with the implementation. 

I find it very elegantly done, given the high degree of difficulty of the code base. 

I made some minor comments here and there, but my major two comments regard the creation of the STAT files and a file to check if the wrapper failed in the horizontal-vertical case. 

The first point could turn to be unnecessary if @kinow 's PR stops the creation of this file. 

The second point is a suggestion just to avoid yet-another-file in the parallel file system. 

Whenever I have time, I plan to install this branch and try the wrappers.

Thanks!

autosubmit/platforms/wrappers/wrapper_builder.py
pablogoitia: Multi-section wrappers (as I call them) are aggregations that comprise more than one section. For example, "SIM" and "DA" in this [advanced example](https://autosubmit.readthedocs.io/en/latest/userguide/wrappers/index.html#advanced-example-set-up-an-crossdate-wrapper).

I had this question because, when assigning inline "--nslots" in `flux batch`, I had to add some more logic to deduce the amount of "slots" for each section, and I didn't know if I should add it to all the types of wrapper. Now the "--nslots" are indicated in the job's Flux header, so the TODO makes no sense anymore.

autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -121,6 +122,316 @@ def _indent(self, text, amount, ch=' '):
...
+        fi
+
+        # Debug commands
+        # flux resource list
+        """).format(self.processors, ' '.join(str(s) for s in self.job_scripts), '\n'.ljust(13))
+
+class FluxVerticalHorizontalWrapperBuilder(FluxWrapperBuilder):
+    # TODO: [ENGINES] Check error handling behavior
+    # TODO: [ENGINES] Delete the "flux resource list" after testing
+    # TODO: [ENGINES] Observe what happens if a single job arrives. Is it inside another list?
```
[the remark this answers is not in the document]
dbeltrankyl: There is a code that checks whether a wrapper is balanced (the same number of jobs in all the lists). But it won't stop autosubmit execution. It only shows a warning. 

So, yes, one single job might be in some wrapper_list[i][j]. 

By default, and from memory, the min/max wrapped is set to 2, but the user could put it to 1. Using:

```yaml
Wrappers:
 WRAPPER:
    MIN_WRAPPED_H: 1
    MAX_WRAPPED_H: 1
    MIN_WRAPPED_V: 1 
    MAX_WRAPPED_V: 1
```

autosubmit/platforms/wrappers/wrapper_builder.py
manuel-g-castro: So I understand that a section is a workflow task? And, with more tasks, you need to get more information on the cpus that each task uses. Correct?

pablogoitia:
Hi,  @manuel-g-castro. Commit https://github.com/BSC-ES/autosubmit/pull/2708/changes/7a0aec4789a7dbdcfea1d555325e26dd48efd22c should have solved the issue that you reported in this comment https://github.com/BSC-ES/autosubmit/pull/2720#issuecomment-3616579515, about the specific case where tasks per node are specified without the node count. Now, the expected behavior is to not specify a node count if it is not explicitly given by the job configuration, still complying with [RFC25](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.html) standards.

Thanks again for reporting it and for leaving such a detailed comment!

pablogoitia:
> @VindeeR
> 
> This can be merged in master now ( after rebasing). I did a new branch for 4.1.16 based on the current master
> 
> Note that all 4.1.17 changelog is now under 4.1.16, please change the changelog so that only the changes related to this branch are added as 4.1.17, and the rest is left untouched

Hi, @dbeltrankyl. I'm rebasing. Thanks for letting me know

dbeltrankyl:
In the logs, I see the fc3_1_9_APP_HYDROLAND completed:

> 2026-02-11 17:30:25,880 Job: a370_20200101_fc3_1_9_APP_HYDROLAND submitted with job_id: 36174571 and workflow commit: 7470ebcf62a5c745fec47b3229f5963ca45de0f0


2026-02-11 18:29:08,690 Job a370_20200101_fc3_1_9_APP_HYDROLAND is COMPLETED
2026-02-11 18:29:08,696 Job a370_20200101_fc3_1_9_APP_HYDROLAND finished at 2026-02-11 18:27:58

And the wrapper too: 

> 2026-02-11 18:29:11,364 Wrapper job a370_ASThread_OPA_HYDROLAND_17708242481073_1_15 changed from RUNNING to status COMPLETED


I see this failed: 

> 

2026-02-11 18:00:31,899 Wrapper job a370_ASThread_OPA_ENERGYTDIG1_OPA_ENERGYTDIG2_OPA_ENERGY_INDICATORS_17708201383709_0_138 changed from RUNNING to status FAILED


And finally, a failed submission

> sbatch: error: Batch job submission failed: Unexpected message received

Which is likely an slurm temporal issue

------------

```
Checking [dbeltran@climatedt-wf ASLOGS]$ cat jobs_failed_status.log

Job Name                           Job Id         Job Status     Job Platform        Job Queue      
a370_20200101_fc0_1_1_DN           2651832        FAILED         MARENOSTRUM5-LOGIN  gp_interactive 
a370_20200101_fc3_1_9_APP_HYDROLAND36174571       FAILED         MARENOSTRUM5        gp_bsces       

```
I see the APP_HYDROLAND failed.. which is very strange as it didn't fail  according to  the logs :thinking: 

--------

Going back to checking when it was marked to failed...

First occurrence

> 52489 of 52755 jobs remaining (17:09)
2026-02-11 18:31:07,674 2 jobs has been  failed (18:31)

This couldn't be possible to run if the 09 failed

> 2026-02-11 18:29:31,344 Job a370_20200101_fc3_1_30_OPA_HYDROLAND is COMPLETED
2026-02-11 18:29:31,349 Job a370_20200101_fc3_1_30_OPA_HYDROLAND finished at 2026-02-11 18:28:57



> 2026-02-11 18:28:59,767 Job a370_20200101_fc3_1_8_APP_HYDROLAND is COMPLETED
2026-02-11 18:28:59,772 Job a370_20200101_fc3_1_8_APP_HYDROLAND finished at 2026-02-11 18:22:44
[WARNING] 2026-02-11 18:29:04,002 File a370_20200101_fc3_1_8_APP_HYDROLAND_FAILED couldn't be found
[WARNING] 2026-02-11 18:29:04,002 Wrapper a370_ASThread_APP_HYDROLAND_17708238055257_1_15 failed, cancelling it

Ok this doesn't make any sense, the FAILED file is expected to not be found if the 1_8 is completed :thinking: I'll revise that part of the code..
------------------------

I think what happened is that: 

1) The full wrapper ended already, but AS is not aware

2) Autosubmit falsely detects that the a370_20200101_fc3_1_8_APP_HYDROLAND has to have a FAILED file instead of a completed file ( bug is here, I believe), but still it marks the job as completed 

3) Autosubmit cancels the wrappe,r but it is already finished, or at least up to 30 jobs are completed

4) Autosubmit marks as failed the _9 instead of _8. Not sure if it is a bug, it could be derived from the malfunction of the _8. Check as the _8 is marked as completed, so for autosubmit, only the _9 is currently running, so when it is cancelled in the middle of the run, it is set to Failed


Do you think it could be derived from your code additions, or is there a bug in master?

dbeltrankyl:
Also, maybe this bug is very rare... in 4.2 we changed how we check whether an inner job is completed. If this happens very often, maybe it's not worth having a look, but if it happens all the time,y es

VindeeR:
> Hi @dbeltrankyl! I have read this last week but I'm answering now because I was working in the report. Thanks for taking some time for analyzing these cases.
> 
> > Then all the retries of the vertical wrapper are managed by the wrapper itself, but reusing the same allocation. This means that if you set the amount to 10, you are likely to reach max_wallclock. Also, keep in mind that by default, Autosubmit only adds enough wallclock time to run the tasks once. You have to set the extend_wallclock as well
> 
> You're right. To be honest, I was not considering a case when the wallclock was exceeded because of retries, since the inner jobs are really small, with a wide walltime. But this is a great point if someday we discuss maintaining _retrials_ inside the vertical wrappers.
> 
> > Also, maybe it's worth discussing: should we allow a vertical wrapper to be retried externally if the user wants? This implies a lot of changes to the code tho cc: @kinow
> 
> Sure. Also, it would be great if users were advised (e.g. in the docs) abuot the fact that walltime of possible retrials is not being aggregated for the submission by default. Anyway, keep in mind that this "issue" concerned me because I wanted to run a very large workflow at once with no errors, which is complicated sometimes, but in other cases it is just a matter of re-run the workflow from that point to restore the _retrial_ count, as you propose.
> 
> > This is interesting because we have energy to 0 issues cc: @VindeeR , did this make the job to crash tho?
> 
> Here, what I can comment by now is that I have not seen that error again. It's probably just a one-off thing.
> 
> > Reacing oom in this case, I think is normal if the number of nodes didn't match,
> > I think you can do
> 
> Oh, thanks! I have stopped testing with horizontal wrappers but I will definitely make a note for this.
> 
> > Do you think it could be derived from your code additions, or is there a bug in master?
> 
> It's quite suspicious. In principle my new additions do not interfere with the code that checks the completeness of the tasks. The Flux wrappers imitates the behavior of ASThread in creating the `_COMPLETED` and `_FAILED` files. Last friday, after reading this, I have done some changes to make them even more similar, at least in the vertical case. I did this just in case, but I believe that the code was already correct. I can't say with certainty if it's a bug in master.
> 
> > Also, maybe this bug is very rare... in 4.2 we changed how we check whether an inner job is completed. If this happens very often, maybe it's not worth having a look, but if it happens all the time,y es
> 
> I'll keep you updated if it happens again in new runs.
> 
> > @pablogoitia Can you check the remote platform , inside the LOG_$EXPID dir for any _FAILED file?
> 
> ```shell
> (base) [bsc032177@glogin1 LOG_a370]$ ll *_FAILED
> -rw-r--r-- 1 bsc032177 ehpc01 0 Feb 11 16:09 a370_20200101_fc3_1_13_OPA_ENERGYTDIG1_FAILED
> -rw-r--r-- 1 bsc032177 ehpc01 0 Feb 11 16:09 a370_20200101_fc3_1_13_OPA_ENERGYTDIG2_FAILED
> ```

Regarding the energy issue mentioned above, we noticed that there were quite a few experiments that should have stored energy. 

We couldn’t identify any issues that would prevent it from being properly stored, yet the energy was stored as 0.

A provisional solution was developed to minimize failures, but we believe it won't completely resolve the issue. Therefore, it’s interesting that you haven’t encountered this problem.

pablogoitia:
image

Delegated adaptive wrappers have been configured for the postprocess tasks of the workflow (POST1, POST2, and POST3), which have interdependencies. Wrapper sizes increases throughout the execution because the SIM tasks' execution rate is higher.

pablogoitia:
These last days I have been working on implementing a new type of `delegate` wrapper that is executed through *wrapper engines* in cluster resource allocations (see #2968). This new type has enabled us to create more flexible and scalable configurations. For example, it supports dependencies between tasks from different sections within the same wrapper. This can be particularly beneficial for grouping the post-processing of simulation data.

The idea is that this simplifies the workflow configuration process for users, who no longer have to think about which type of wrapper best suits the execution flow, but rather about more general issues such as policy or size. On the development side, it abstracts us from having to plan the resources onto which tasks are mapped, for example.

## Relevant information

- This PR has undergone some significant changes. Previously, there was a `Flux` method for each type of original wrapper (*vertical*, *horizontal*, and *hybrid*). Now, instead, there is a new type of wrapper, which we are calling `delegate` for now (because it *delegates* execution), that differs from the previous ones in that it can adapt to the workflow topology. In other words, it creates a subworkflow with tasks that can belong to different sections regardless of dependencies between them.

- The new `JobPackagerDelegated` uses the `networkx` library, already used in Autosubmit, to calculate the wrapped subworkflow while respecting the specified limits, calculate critical paths (to determine the number of vertical tasks in the wrapper and the `max_wallclock`), as well as calculate the width (number of tasks in the largest horizontal level) and the `max_processors`.

- The exploration of groupable tasks is performed using an algorithm that prioritizes horizontal exploration to make the best use of the requested resources (somewhat similar to BFS)

- Task retries are handled in the same way as in the *horizontal*, *vertical-horizontal*, and *horizontal-vertical* types.

- The new wrapper type interacts correctly with other Autosubmit components, such as the visual graph generator.

## The interface

The `DelegatedWrapper` graph is serialized and exported to a JSON file (`subworkflow_***.json`) with a consistent structure. This file is unique to each wrapper and is uploaded to the platform along with the *wrapper script*. The name of the *wrapper script* in delegated wrappers has the structure `<EXPID>_WRAPPER_***.cmd`. The JSON file contains:

- A global configuration for the wrapper under `graph`>`wrapper_defaults` that contains the `cwd`.

- The task specifications under `tasks`. Each one contains the `nodes`, `threads`, `processors`, `wallclock`, `exclusive`, and the `id` (equivalent to the task name), according to the Autosubmit task resource specification.

- Dependencies are located under `dependencies`, and each one contains the task `from` (source), `to` (target), and `weight`, which is the “weight” of the link in the graph and corresponds to the *wallclock* of the source task.

With this, integrating a wrapper engine into Autosubmit is very straightforward, requiring only a few changes to `WrapperFactory` and `DelegatedWrapperBuilder`. The first, to assign a new wrapper builder specific to the new engine on a particular platform, and the second to design the new wrapper builder (which must implement the abstract class `DelegatedWrapperBuilder`). The structure of this class facilitates the inclusion of the script responsible for managing the workflow execution through the binding provided by the tool.

This final script must perform several general steps, which should be adapted to the specific needs of the tool:

1. Load the JSON file and reconstruct the task graph with the dependencies and weights between them.
2. Send the tasks in the correct order.
3. Monitor the tasks and create SUCCESS or FAILED files as appropriate, so that Autosubmit can track the status of the wrapper's execution.

## Flux as *wrapper engine*

Flux, which has already been integrated, is the default method for delegate wrappers (at least for now). The script follows the structure defined above, launching all tasks asynchronously while respecting dependencies. The completion of each task is managed through a reactor that launches a function responsible for its post-processing. Additionally, after submission, all data structures that are no longer in use are forcibly deleted to minimize memory consumption within the resource allocation.

## What we have so far

- A consistent and optimized interface for communicating with wrapper engines.

- Flux integrated as the default method.

- Unit tests for new and modified components.

- Updated documentation, including:
    - Detailed configuration information for users.
	- A step-by-step guide for developers on integrating new wrapper engines using the implemented interface, with links to the literature for further motivation and technical details.

## Pending

- How would we store the timestamps for the delegated wrappers? Should we store them in the top-level tasks? If I'm not wrong, that would be the best way to ensure the GUI continues to calculate statistics correctly without having to make changes to it.

- [EDITED] I have detected a case where the Flux instance terminates immediately after starting because the time required to execute the tasks is less than the requested allocation time, due to the script's processing overhead. This occurs in most cases when we have a single horizontal level (when there are more, the user's overestimation of the wallclock for individual tasks hides this problem). I think we could implement the `EXTEND_WALLCLOCK` directive in the delegate wrappers as well, with a default value of >1 (yet to be determined).

- Development of integration and regression tests.

- Review some minor comments in the code (just some open questions for the review process).

## Notes

- [EDITED] Due to a change in terminology between `networkx` versions 3.5 and 3.6 (see https://github.com/networkx/networkx/pull/8282) that affected the implementation, I had to adapt the graph serialization function in order to give support to all networkx versions, because of our need to test Python 3.9 (that Python version forces using networkx v3.2.1 if I recall well). Dani opened an issue regarding dropping the support for this Python version (#3045).

---

CC: @dbeltrankyl @mcastril

mcastril:
> * I have detected a case where the Flux instance terminates immediately after starting because the time required to execute the tasks is less than the required time, due to the script's processing time. This occurs in most cases when we have a single horizontal level (when there are more, the user's overestimation of the wallclock for individual tasks hides this problem). I think we could implement the `EXTEND_WALLCLOCK` directive in the delegate wrappers as well, with a default value of >1 (yet to be determined).

Thank you for the update, Pablo, it reads very interesting and sound.

Could you clarify the previous paragraph? There were some inconsistencies for me but maybe I just didn't understand

pablogoitia:
Thanks, Miguel.

> Could you clarify the previous paragraph? There were some inconsistencies for me but maybe I just didn't understand

Sure. It's my fault. I meant to say that "the time required to execute the tasks is less than the requested time for each one". This situation occurs when you have wrappers composed by just parallel (horizontal) tasks: if you create a wrapper for a couple of parallel tasks requesting, let's say, 10 minutes (your allocation will also have a wallclock of 10 minutes), you need to subtract from that time the pre-processing overhead. When Flux tries to schedule the jobs after that pre-processing, it raises an error saying something like you are trying to submit a task that needs 10 minutes to run but only 9.8 minutes are left.

So, the idea would be to leave a margin for that overhead to not to interfere with the wrapper execution, because it will always be present (in varying degrees) independently of the wrapper engine employed. To do that I propose forcing a default "minimum" extended wallclock to leave a small room, but I would appreciate external opinion on this

pablogoitia:
> I think we could implement the EXTEND_WALLCLOCK directive in the delegate wrappers as well, with a default value of >1 (yet to be determined).

Yesterday I implemented the `EXTEND_WALLCLOCK` directive replicating the behavior of the vertical wrappers'. Here is how it's calculated:

If user specified `EXTEND_WALLCLOCK` > 1 (it's always integer), then the wallclock by level is calculated (such as an average wallclock per vertical level). This average is multiplied `EXTEND_WALLCLOCK` times and finally added to the original wrapper wallclock.

By now, I have not set a default value, but this should be considered since users might not be aware of the risks of not having some extra time window if the configured policy allows for the generation of horizontal wrappers. Neither Flux or HQ allows scheduling tasks when the remaining time inside the allocation is smaller than the required by the wrapped tasks.

pablogoitia:
I've been performing some testing with Auto-MONARCH. I tried to create wrappers for `SIM` and `POSTPROC`, but they didn't work due to some issues that arise when using Flux as the wrapper engine. On the positive side, this has also helped identify areas for improvement and necessary fixes when facing the integration of any other engine. I'll describe everything below.

## Detected Problems (sequentially)

### PROBLEM 1: Unable to satisfy the requested resource reservation

For `SIM`, 260 `PROCESSORS` (tasks) are requested, which, combined with the 48 processors required by `POSTPROC`, totals 308. This results in a request to the wrapper for `--cpus-per-task=308` and `--ntasks=1`. Although one might naturally assume this equates to 3 nodes, with one of them operating below its full capacity (MN5’s Slurm configuration does not allow requesting multiple nodes without exclusivity), the scheduler cannot launch a single task spanning more than one node, so the request fails.

**Possible solution:** Specify one task per node. Each will be a Flux broker. Autosubmit must ensure that when the number of processors exceeds the maximum capacity of a node on the platform, as many complete, exclusive nodes as necessary are requested. This requires a slight change to the logic in `WrapperFactory`, but it is necessary for any wrapper engine to be able to run a worker on each node.

### PROBLEM 2: Flux removes Slurm environment variables used by Auto-MONARCH

According to the Flux documentation, brokers actively remove environment variables they consider conflictive, such as those from Slurm. https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-environment.html#index-20

**Implemented solution:** Manually replace the values “on the fly” in all scripts containing occurrences of `SLURM_`.

**Alternative solutions:**

1. Replace these variables with their “on-the-fly” values within the wrapper script itself.
2. Knowing that they are removed, copy all of them by replacing “SLURM_” with “WRAPPER_” and then replace the names in the files.

For now, I don’t think it’s reasonable to go to such lengths. Variable names can be enclosed in several ways, such as `$VAR`, `${VAR}`... And besides, executables don’t necessarily have to be limited (though they should be) to the scope of the project.
	
### PROBLEM 3: Multiple brokers launch multiple wrapper scripts


In multi-node configurations, the current logic is to launch one broker/worker instance on each node. In Flux, this also means launching the wrapper script (which coordinates execution) multiple times. For now, although this is the expected behavior, I have no evidence that it is actually happening, so this remains an open issue pending further investigation.

### PROBLEM 4: Some of the scripts use `srun`

I understand that Flux does not inherit the entire Slurm execution environment, so it is not permitted to invoke `srun` within a job running on Flux. The problem has been detected in both `SIM` and `POSTPROC`. I have tried manually replacing it with `mpirun` in `SIM`, but “mpiexec” returns an error.

<details><summary>Details</summary>
<p>
[INFO] Executing nmmb-monarch model

Wed Sep  2 13:13:10 CEST 2026
[mpiexec@gs08r2b20] check_exit_codes (../../../../../src/pm/i_hydra/libhydra/demux/hydra_demux_poll.c:117): unable to run bstrap_proxy on gs08r2b20 (pid 1946658, exit code 256)
[mpiexec@gs08r2b20] poll_for_event (../../../../../src/pm/i_hydra/libhydra/demux/hydra_demux_poll.c:159): check exit codes error
[mpiexec@gs08r2b20] HYD_dmx_poll_wait_for_proxy_event (../../../../../src/pm/i_hydra/libhydra/demux/hydra_demux_poll.c:212): poll for event error
[mpiexec@gs08r2b20] HYD_bstrap_setup (../../../../../src/pm/i_hydra/libhydra/bstrap/src/intel/i_hydra_bstrap.c:1065): error waiting for event
[mpiexec@gs08r2b20] HYD_print_bstrap_setup_error_message (../../../../../src/pm/i_hydra/mpiexec/intel/i_mpiexec.c:1026): error setting up the bootstrap proxies
[mpiexec@gs08r2b20] Possible reasons:
[mpiexec@gs08r2b20] 1. Host is unavailable. Please check that all hosts are available.
[mpiexec@gs08r2b20] 2. Cannot launch hydra_bstrap_proxy or it crashed on one of the hosts. Make sure hydra_bstrap_proxy is available on all hosts and it has right permissions.
[mpiexec@gs08r2b20] 3. Firewall refused connection. Check that enough ports are allowed in the firewall and specify them with the I_MPI_PORT_RANGE variable.
[mpiexec@gs08r2b20] 4. slurm bootstrap cannot launch processes on remote host. You may try using -bootstrap option to select alternative launcher.

</p>
</details> 

## Upcoming actions

For now, to keep the momentum going, I'm going to continue testing Flux with other auto-models. Soon, once I've resolved the issues with HyperQueue in multi-node configurations, I'll repeat the tests.

manuel-g-castro:
Hi, Pablo! Thanks for documenting the development. 

I have some comments

> Flux removes Slurm environment variables used by Auto-MONARCH

It is great that you identified this issue. And I think we have nothing that we can do about it besides recommending the users to not use Slurm variables in their scripts. Not sure why the auto-monarch developers use them for. 

> Some of the scripts use srun

The MPI bootstrapping has to be done with Flux's command, otherwise Flux will not be able to keep track of where the SIM's ranks are being executed and could eventually collide with POSTPROC's CPUs. 

For all the MPI stuff, `srun` should be changed to flux run. Here you have the [documentation](https://flux-framework.readthedocs.io/en/latest/faqs.html#mpi-questions).

Issue #2657, opened 2025-10-17
[Bug] Unused code to store energy


VindeeR:
### Description

Might be related to #2418 

As I was going through some tests, to complement and/or implement I found a test that led me to this piece of code, upon deeper analysis here's what I found.

At the function `retrieve_times` in the `job_list` Class, some of the statistics data is generated, and the variable that stores it is the `job_data_collection` I notice this function has only one call at `autosubmit/autosubmir.py:2949`, the energy attribute is never used in the `autosubmit.py` file, and the variable that is used to get the metrics is sent as `None`, another variable `job_times` is sent as `None` too which impacts a part of the code that comes later.

I found other places where the energy attribute is used, so it can't be assured with precision that this is where the energy issue is coming from but, it this seems to be related.

Another thing is despite being called `job_data_collection` I think it might be related to the `SlurmMonitorItem`  Class, but I'm not too confident on that, since the attributes it tries to use are only found there.

Because of all that, there's about 45 lines of the code that is never touched.

I'm not sure which should be the proper behavior, but at the end of the day that piece of code should gather the statistics has not being used at least since 12/2021
`autosubmit/job/job_list.py | retrieve_times | 3157-3192  (unused code)`
`autosubmit/job/job_list.py | retrieve_times | 3224-3234  (unused code)`

Apparently this was the commit that changed&created this feature `Implemented #644. A new module that handles statistics has been created (f7b27446)`

Issue #2418, opened 2025-06-12
Missing energy in o-suite experiments


orioltinto:
Hi,

we were checking metrics from the operational runs from DestinE and the energy metrics are missing:

https://climatedt-wf.csc.fi/experiment/o007/performance

I'm not the owner of the experiment so I cannot directly check the outputs of sacct.

Also, we realized that under the statistics tab the CPU hours doesn't seem to be correct as the "Real Consumption (h)" and the "CPU Consumption (h)" report the same number. My understanding is that the CPU hours should be the real consumption multiplied by the parallelization. 

I looked at a random experiment in the earth gui and it seemed to be correct, so I don't know if this is happening because the climatedt gui version is the **Pre-release v4.0.1-beta.2 - Release date: 2024-10-01** while  the earth one is **Pre-release v4.1.0-beta.1 - Release date: 2025-05-21**.

dbeltrankyl:
> I think we could create a Jupyter notebook, copy the experiment sqlite database to our local notebook, and execute it to see if it produces the correct values. Then with some help from @LuiggiTenorioK & @dbeltrankyl verify that the data looks right and can be used by Oriol/others to create the report they need for their deliverable.


Looks good

Issue #1455, opened 2024-11-19
Energy calculations possible error


kinow:
In GitLab by @eferre1 on Nov 19, 2024, 09:46

Hello @dbeltrankyl and @kinow,

### Autosubmit Version

v3.15.19

### Expid affected(If applicable)

Almost any experiment from my testing suite (range from t0n2 to t0ni). And I guess any experiment would be affected if the config has jobs running with less than full nodes. Found and most visible on t0n8 as explained below.

### Which task has issues? Where is the log(If applicable)

* **Full_name**: t0nX_REMOTE_SETUP

No logs, not exactly an issue with the job itself.

### Summary

In the image, you can see that the job is running mainly (except last run) with 48 processors (24x2 due to using multithreading). But then, on the nodes column, you can see that it mostly uses 1, but some times 2 or 3. This is (I'm guessing) due to using less than one node, then being distributed among shared nodes, and usually it is all together but sometimes it isn't. But I'm still using only 24 "real" processors, which means that the energy calculations done, are not working properly for the jobs running in shared nodes, as it is counting more or less the same in the last job with half the resources, or double or triple on the cases where the processors where distributed among 2 or 3 nodes.

![image](https://earth.bsc.es/gitlab/es/autosubmit/uploads/5e508b06903a6f8f800da0d965993aef/image.png)

### Steps to reproduce

Run jobs with less than one node usage, and then having luck (not sure if it can be controlled) that it is distributed among more than one node.

### What is the current bug behavior?

The energy calculations are not correct (GUI is the tool to see them, but they are done in autosubmit if I'm not mistaken).

### What is the expected correct behavior?

When using shared nodes, energy calculation should take that into account and just use one node energy usage and divide it by the proportion of the node used (24/112 in this case). Not sure how feasible to check this is, moreover with the case where it is using multiple shared nodes, but it is something that could throw off our estimations of energy consumption (this is a job using ~25% of the node usually, but some jobs using less than 10 processors will get the same energy usage).

kinow:
In GitLab by @LuiggiTenorioK on Nov 19, 2024, 10:21

> fyi @LuiggiTenorioK ( correct me if it is wrong as you're more updated than me in metric related stuff )
> 
> I think that value comes directly from slurm, https://earth.bsc.es/gitlab/es/autosubmit/-/issues/1424#note_315745

Yes, from the API/GUI side, there is no transformation done from what we have inside the database which is written by Autosubmit and is something we get directly from SLURM.

[...]

PR #2719, opened 2025-11-26
[Bug] - Energy data no being saved

VindeeR:
Related #2657 #2418 
Fixing Threads that were being closed before the `Energy Data` being stored, resulting in jobs not storing the data and a loss of information.

This fix provide a wilder consistency in saving the data but might not solve for all scenarios

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -152,7 +152,7 @@ class Job(object):
         'ec_queue', 'platform_name', '_serial_platform',
         'submitter', '_shape', '_x11', '_x11_options', '_hyperthreading',
         '_scratch_free_space', '_delay_retrials', '_custom_directives',
-        '_log_recovered', 'packed_during_building', 'workflow_commit', 'thread_write_finish'
+        '_log_recovered', 'packed_during_building', 'workflow_commit'
     )
 
     def __setstate__(self, state):
--- a/autosubmit/job/job_list.py
+++ b/autosubmit/job/job_list.py
@@ -88,8 +88,7 @@ class JobList(object):
 
     @property
     def expid(self):
-        """
-        Returns the experiment identifier
+        """Returns the experiment identifier
 
         :return: experiment's identifier
         :rtype: str
@@ -151,8 +150,7 @@ class JobList(object):
     def generate(self, as_conf, date_list, member_list, num_chunks, chunk_ini, parameters,
                  date_format, default_retrials, default_job_type, wrapper_jobs=dict(), new=True,
                  run_only_members=[], show_log=True, monitor=False, force=False, create=False):
-        """
-        Creates all jobs needed for the current workflow.
+        """Creates all jobs needed for the current workflow.
 
         :param create:
         :type create: bool
@@ -308,8 +306,7 @@ class JobList(object):
         self.graph = None
 
     def split_by_platform(self):
-        """
-        Splits the job list by platform name
+        """Splits the job list by platform name
 
         :return: job list per platform
         :rtype: dict
@@ -371,8 +368,7 @@ class JobList(object):
             dic_jobs: DicJobs,
             option: str = "DEPENDENCIES",
     ) -> None:
-        """
-        Build dependency maps and populate the dependency graph for all jobs.
+        """Build dependency maps and populate the dependency graph for all jobs.
 
         Iterate experiment `JOBS` sections to:
         - build deep dependency maps (with and without distance metadata),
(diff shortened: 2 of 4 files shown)


manuel-g-castro:
Related to https://github.com/BSC-ES/autosubmit/issues/2418?

VindeeR:
> Related to #2418?

yeap :+1: I believe it is related to both #2657 #2418

LuiggiTenorioK:
Related to #1455?

autosubmit/history/platform_monitor/slurm_monitor.py
VindeeR: This file was just reformatted since it was formatted in a different way from the rest of the project [Standardizing]

test/unit/test_job.py
```diff
@@ -2327,3 +2327,46 @@ def test_update_status(create_jobs: list[Job], status: Status, failed_file,
     assert job.status != status
     job.update_status(as_conf=as_conf, failed_file=failed_file)
     assert job.status == status
+
+@pytest.mark.parametrize(
```
kinow: Two lines between functions, missing one blank line above ☝️

autosubmit/history/platform_monitor/slurm_monitor.py
kinow: Thank you!

autosubmit/history/experiment_history.py
```diff
@@ -182,14 +184,12 @@ def write_platform_data_after_finish(self, job_data_dc, platform_obj):
         Call it in a thread.
         """
         try:
-            sleep(SECONDS_WAIT_PLATFORM)
```
kinow: No idea why we had this sleep here, but my wild guess would be that at some point, AS was writing too fast so this was added to slow things down and avoid I/O issues? Probably fine if we allow users to control it via total jobs, or some other config? cc @dbeltrankyl

autosubmit/history/strategies.py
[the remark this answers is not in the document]
kinow: Thank you! I'm skipping the files you said were just reformatted 👍

kinow (commented):
@VindeeR the only file I didn't review is `job.py`. Leaving the partial review here while I go to have lunch 👍 Feel free to address any comments in the meantime, but I will try to complete `job.py` today.

Great job investigating this nasty bug! 🐛 🥾

autosubmit/job/job.py
```diff
@@ -2715,12 +2715,12 @@ def write_end_time(self, completed, count=-1):
                                                     job_id=self.id, out_file=out, err_file=err)
 
         # Launch second as threaded function only for slurm
-        if job_data_dc and type(self.platform) is not str and self.platform.type == "slurm":
-            thread_write_finish = Thread(target=ExperimentHistory(self.expid, jobdata_dir_path=BasicConfig.JOBDATA_DIR,
+        if job_data_dc and self.platform and self.platform.type == "slurm":
+            self.thread_write_finish = Thread(target=ExperimentHistory(self.expid, jobdata_dir_path=BasicConfig.JOBDATA_DIR,
```
kinow: Huh, are we sure this is fixing it, @VindeeR ? I just checked and the thread is not a daemon thread here, so in theory I think the main process shouldn't exit until all non-daemon threads are gone?

https://docs.python.org/3/library/threading.html#:~:text=The%20significance%20of%20this%20flag%20is%20that%20the%20entire%20Python%20program%20exits%20when%20only%20daemon%20threads%20are%20left.

autosubmit/history/experiment_history.py
VindeeR: Yeap, from what @LuiggiTenorioK and I spoke about, apparently sometime ago they thought maybe that was the issue that didn't let the them retrieve and store data; but because the `Thread` was being killed that would just worsen the problem, and I notice that the function would only be run if the job was `COMPLETED` or `FAILED`, which means that the energy would have been stored by then, at least that the behavior I managed to replicated during my tests.

autosubmit/job/job.py
```diff
@@ -152,7 +152,7 @@ class Job(object):
         'ec_queue', 'platform_name', '_serial_platform',
         'submitter', '_shape', '_x11', '_x11_options', '_hyperthreading',
         '_scratch_free_space', '_delay_retrials', '_custom_directives',
-        '_log_recovered', 'packed_during_building', 'workflow_commit'
+        '_log_recovered', 'packed_during_building', 'workflow_commit', 'thread_write_finish'
```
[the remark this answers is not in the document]
VindeeR: At some point during the development I made it a property, trying to make it keep running even outside of that function, and I forgot to remove this part

autosubmit/job/job_list.py
```diff
@@ -884,7 +893,8 @@ def _unify_to_filters(self, filter_to_apply, splits=None):
         return unified_filter
 
     def _filter_current_job(self, current_job: Job, relationships: dict) -> dict:
-        """This function will filter the current job based on the relationships given.
+        """
+        This function will filter the current job based on the relationships given.
```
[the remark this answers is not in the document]
VindeeR: This file was using the other pattern and I got confused, I'll change all of them in this file so it doesn't happen again

VindeeR:
Merging after CI workflow approves the rebase

PR #2720, opened 2025-11-26
Integration of Flux as a wrapper engine (YAML Jobspecs method)

pablogoitia:
This PR addresses #2697. While implementing the wrappers, I realized that there are Slurm directives that cannot be mapped directly onto Flux ones, because the set of Flux batch directives is more limited. As a solution to this problem, I have been exploring the option of launching jobs on the HPC by manually generating Jobspecs (represented in YAML files), which are processed using the Flux Python API. Jobspecs allow us to control resources in a more detailed and advanced way. So far, I have not found a more direct way to launch jobs using their specification.

In this issue, I provide a prototype for solving this problem. The implementation is not yet fully functional: I only provide limited support for vertical wrappers. However, extending the implementation to other types of wrappers would be relatively straightforward.

The part that is not easy and would require more dedication is the correct construction of the Jobspecs, ensuring that the scheduling parameters (e.g., processors, tasks, threads...) are accurately mapped.

This method would facilitate the introduction of equivalents to Slurm's _hetjobs_ (there are still no direct alternatives in Flux) in the future.

**Note:** this PR would overwrite a significant part of the implementation carried out in PR #2708.

**Check List**
_Not applies, as the branch would not be merged into master yet._

### DIFF
--- a/autosubmit/platforms/wrappers/wrapper_factory.py
+++ b/autosubmit/platforms/wrappers/wrapper_factory.py
@@ -56,6 +56,15 @@ class WrapperFactory(object):
             kwargs['threads'] = self.threads(wrapper_data.threads)
             kwargs['reservation'] = self.reservation(wrapper_data.reservation)
 
+            # When wrapping with Flux, give it control over all resources
+            if wrapper_data.method.lower() == "flux":
+                if wrapper_data.nodes == '':
+                    kwargs['threads'] = self.threads(str(int(wrapper_data.threads) * int(kwargs['num_processors_value'])))
+                else:
+                    kwargs['threads'] = self.threads('')
+                kwargs['num_processors'] = self.processors('')
+                kwargs['tasks'] = self.tasks('')
+
         kwargs["executable"] = wrapper_data.executable
 
         kwargs['header_directive'] = self.header_directives(**kwargs)


dbeltrankyl:
> Maybe the PROCESSORS is being misinterpreted as the number of nodes.

Didn't check how it is coded, but another possibility is that you don't have the PLATFORMS.PLATFORM.PROCESSORS_PER_NODE defined, and instead of 112 for mn5, it is taking the value as 1?

pablogoitia:
Hi again, @manuel-g-castro. After some testing I have concluded that the job specification is right. There are some examples in [RFC 14](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_14.html#section-2-general-requests). Specifically, Use Case 1.6 shows an example of min count of nodes.

However, something is leading Flux to fail, and it does not matter if the Jobspec V1 or the normal Jobspec is being used because it fails anyway.

I will search for a way to handle this specific case so that I could avoid the usage of the `min` key. I will let you know the solution.

Thanks again for reporting the bug. If you want to keep testing, remember that I expect any other case to work. By now, do not test with cases where tasks per node are specified without specifying the node count too.

PR #2738, opened 2025-12-12
[Enhencement] - Removing the generation of SLURM HEADERS with --cpus-per-task

### LINKED ISSUE #2380: cpus-per-task automatically set to 1
Hello!

### Description

We (@ainagaya and I) are trying to enable NEMO standalone again for Marc Batlle and the PREDDYCT project using [this MR](https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/merge_requests/807). We expected the number of threads/cpus-per-task to be set by sim_nemo.sh but instead we are seeing it set to 1 in the SLURM header. We think this is similar to [this issue](https://github.com/BSC-ES/autosubmit/issues/1407).

* Autosubmit version: 4.1.14
* Machine/VM/environment name (if applicable): MN5
* Experiment ID (if applicable): a2h2
  * /appl/AS/AUTOSUBMIT_DATA/a2h2/tmp/a2h2_19930101_fc0_1_SIM.cmd

### Reproducible Example

Copy experiment a2h2 and try to run it.

### Expected Behaviour

We would expect `cpus-per-task` to not be defined in the SLURM header.

VindeeR:
Close #2380

First commit removing the mandatory aspect of the --cpus-per-task Directive to SLURM, ensuring that the value can be substituted or added when needed by the user

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2025 Earth Sciences Department, BSC-CNS
+# Copyright 2015-2026 Earth Sciences Department, BSC-CNS
 #
 # This file is part of Autosubmit.
 #
@@ -1387,8 +1387,7 @@ class Job(object):
         return None
 
     def _time_in_seconds_and_margin(self, wallclock: datetime.timedelta) -> int:
-        """
-        Calculate the total wallclock time in seconds and the wallclock time with a margin.
+        """Calculate the total wallclock time in seconds and the wallclock time with a margin.
 
         This method increases the given wallclock time by 30%.
         It then converts the total wallclock time to seconds and returns both the total
@@ -1414,7 +1413,8 @@ class Job(object):
 
     @staticmethod
     def parse_time(wallclock):
-        if not isinstance(wallclock, str):  # TODO This is a workaround for the time being, just defined for tests passing without more issues
+        # TODO This is a workaround for the time being, just defined for tests passing without more issues
+        if type(wallclock) is not str:
             return datetime.timedelta(24 * 60 * 60)
         regex = re.compile(r'(((?P<hours>\d+):)((?P<minutes>\d+)))(:(?P<seconds>\d+))?')
         parts = regex.match(wallclock)
@@ -1427,14 +1427,10 @@ class Job(object):
                 time_params[name] = int(param)
         return datetime.timedelta(**time_params)
 
-    # TODO Duplicated for wrappers and jobs to fix in 4.1.X but in wrappers is called _is_over_wallclock for unknown reasons
-    def is_over_wallclock(self):
-        """
-        Check if the job is over the wallclock time, it is an alternative method to avoid platform issues
-
-        :return:
-        :rtype: bool
-        """
+    # TODO: Duplicated for wrappers and jobs to fix in 4.1.X but in wrappers is called
+    #       _is_over_wallclock for unknown reasons.
+    def is_over_wallclock(self) -> bool:
+        """Check if the job is over the wallclock time, it is an alternative method to avoid platform issues."""
         elapsed = datetime.datetime.now() - self.start_time
         if int(elapsed.total_seconds()) > self.wallclock_in_seconds:
             Log.warning(f"Job {self.name} is over wallclock time, Autosubmit will check if it is completed")
@@ -1511,13 +1507,6 @@ class Job(object):
                 )
 
         return self.status
-
-    @staticmethod
-    def _get_submitter(as_conf):
-        """
-        Returns the submitter corresponding to the communication defined on Autosubmit's config file
-        """
-        return ParamikoSubmitter(as_conf=as_conf)
 
     def update_children_status(self):
         children = list(self.children)
@@ -1625,12 +1614,8 @@ class Job(object):
 
         return parameters
 
-    def process_scheduler_parameters(self, job_platform, chunk):
-        """
-        Parsers yaml data stored in the dictionary and calculates the components of the heterogeneous job if any
-
-        :return:
-        """
+    def process_scheduler_parameters(self, job_platform: 'Platform', chunk: int) -> None:
+        """Parsers yaml data stored in the dictionary and calculates the components of the heterogeneous job if any."""
         if type(self.processors) is list:
             hetsize = (len(self.processors))
         else:
@@ -2157,7 +2142,7 @@ class Job(object):
             self.x11 = False if str(parameters.get("CURRENT_X11", False)).lower() == "false" else True
             self.notify_on = parameters.get("CURRENT_NOTIFY_ON", [])
             self.update_stat_file()
-            if self.checkpoint:  # To activate placeholder sustitution per <empty> in the template
+            if self.checkpoint:  # To activate placeholder substitution per <empty> in the template
                 parameters["AS_CHECKPOINT"] = self.checkpoint
             self.wchunkinc = as_conf.get_wchunkinc(self.section)
 
@@ -2292,7 +2277,7 @@ class Job(object):
             self.update_dict_parameters(as_conf)
         self.init_platform(as_conf)
         parameters = as_conf.load_parameters()
-        # TODO: This shouldn't be neccesary aims to fix 2432 issue
+        # TODO: This shouldn't be necessary aims to fix 2432 issue
         as_conf.load_current_hpcarch_parameters(parameters)
         parameters = self.update_current_parameters(as_conf, parameters)
         parameters = self.update_job_parameters(as_conf, parameters, set_attributes)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@
 
 ### 4.1.16: Unreleased
 
+**Bug fixes:**
+
+- Fixed bug where placeholder variables were not removed when they referenced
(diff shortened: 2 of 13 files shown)


autosubmit/platforms/wrappers/wrapper_factory.py
```diff
@@ -204,17 +204,21 @@ def partition_directive(self, partition):
...
     def nodes_directive(self, nodes):
         return f'#SBATCH -N {nodes}'
 
     def processors_directive(self, processors):
         return f'#SBATCH -n {processors}'
 
-    def threads_directive(self, threads):
-        return f'#SBATCH --cpus-per-task={threads}'
+    def threads_directive(self, threads) -> str:
+        if threads and threads != '':
```
kinow: Same here?

autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -882,19 +882,19 @@ def build_imports(self):
         """).format(str(scripts_bash), '\n'.ljust(13))
 
     def build_srun_launcher(self, jobs_list, footer=True):
-        srun_launcher = textwrap.dedent("""
+        srun_launcher = textwrap.dedent(f"""
         i=0
         suffix=".cmd"
-        for template in "${{{0}[@]}}"; do
+        for template in "${{{jobs_list}[@]}}"; do
```
kinow: Interesting, I thought for a second jobs_list was a `JobList` object. Maybe we can add the type to the function too ☝️ (I suspect it's a `str`?).

autosubmit/platforms/wrappers/wrapper_factory.py
```diff
@@ -204,17 +204,21 @@ def partition_directive(self, partition):
     def exclusive_directive(self, exclusive):
         return '#SBATCH --exclusive'
 
-    def tasks_directive(self, tasks):
-        return f'#SBATCH --ntasks-per-node={tasks}'
+    def tasks_directive(self, tasks) -> str:
+        if tasks and tasks != '':
```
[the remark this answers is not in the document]
VindeeR: It's to make sure that there's a value inside the variable and the value is not an empty string, otherwise return an empty string without defining the `--ntasks-per-node`

pablogoitia: Hi! Just to recall that the passed tasks value has been already checked in the tasks function of the `WrapperFactory`. Maybe a refactor would let us to not to have two functions (tasks in `WrapperFactory` and tasks_directive in `{Scheduler}WrapperFactory`) to perform a single simple operation. But, this could be another PR. WDYT?

autosubmit/platforms/wrappers/wrapper_builder.py
pablogoitia: This code is confusing, but it should work because by inserting "scripts" in the loop `for template in "${{{0}[@]}}"`, you are iterating a bash array of .cmd filenames called `scripts`, previously defined in the `build_imports` function.

Probably a clearer way to do this would be to have `for template in "${{scripts[@]}}"`, and not passing the name as an argument.

The wrapper script generation is a bit complex to understand at first, but it is done this way because there are reusable pieces of code.

dbeltrankyl:
Hello! 

Just to mention, to avoid merging this one until 4.1.16 is released, please

VindeeR:
@kinow I think there might be something going on with the tests yesterday there were 2 time where a test failed or was running for really long but after rerunning it passed fine, and locally it didn't happen, maybe something worth keeping an eye on.

@kinow @pablogoitia thanks for the feedback, comments were addressed

autosubmit/job/job.py
```diff
@@ -1673,11 +1673,15 @@ def process_scheduler_parameters(self, job_platform, chunk):
             # Get the max threads, each element can be a str or int
             self.het['NUMTHREADS'] = list()
             if len(self.threads) == 1:
-                for x in range(self.het['HETSIZE']):
-                    self.het['NUMTHREADS'].append(self.threads)
+                if(self.threads > 1):
```
kinow: No need for parentheses in Python 👍 

And isn't what you are doing invalid for a list?

The code above has a `if type(self.threads) is list`, so we know that `self.threads` is a list. Above we check the length, but here you compare the list to `> 1`?

autosubmit/job/job.py
```diff
@@ -1673,11 +1673,15 @@ def process_scheduler_parameters(self, job_platform, chunk):
...
+                    for x in range(self.het['HETSIZE']):
+                        self.het['NUMTHREADS'].append(self.threads)
             else:
                 for x in self.threads:
-                    self.het['NUMTHREADS'].append(str(x))
+                    if(x > 1):
```
kinow: No parentheses 😬

kinow (commented):
Reviewed a few more parts as I was taking a break after reviewing other PRs 👍

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -141,11 +141,11 @@ def get_nodes_directive(self, job: 'Job', parameters, het=-1) -> str:
         :rtype: str
         """
         if het > -1 and len(job.het['NODES']) > 0:
-            if job.het['NODES'][het] != '':
-                return "SBATCH --nodes={0}".format(job.het['NODES'][het])
+            if job.het['NODES'][het] != '' and int(job.het['NODES'][het]) > 1:
```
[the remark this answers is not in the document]
VindeeR: I'm pretty sure it will be initialized but not necessarily filled with data.
the `het` is initialized at `process_scheduler_parameters`

autosubmit/job/job.py
VindeeR: old habits die hard :rofl:

autosubmit/job/job.py
```diff
@@ -1673,11 +1673,15 @@ def process_scheduler_parameters(self, job_platform, chunk):
...
+                if self.threads > 1:
+                    for x in range(self.het['HETSIZE']):
+                        self.het['NUMTHREADS'].append(self.threads)
             else:
                 for x in self.threads:
-                    self.het['NUMTHREADS'].append(str(x))
+                    if x > 1:
+                        self.het['NUMTHREADS'].append(str(x))
+                    else:
+                        self.het['NUMTHREADS'].append('')
```
kinow: Do we need empty strings for when x is not greater than 1?

autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -159,31 +160,31 @@ def writelines(self, datas):
            def __getattr__(self, attr):
                return getattr(self.stream, attr)
         sys.stdout = Unbuffered(sys.stdout)
-        node_id = "{1}"
-        wrapper_id = "{1}_FAILED"
+        node_id = "{self.get_random_alphanumeric_string(5, 5)}"
+        wrapper_id = "{self.get_random_alphanumeric_string(5, 5)}_FAILED"
```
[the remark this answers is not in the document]
VindeeR: I just reviewed this because I found it was weird it not being in the master, and yeap that was me, I removed the `.format()` and put the function on the string, but I believe doing it this way might generate a problem in case the number is different so I'll roll this one back.

sorry for the confusion, I think I took too long of a break 😆

autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -159,31 +160,31 @@ def writelines(self, datas):
...
         sys.stdout = Unbuffered(sys.stdout)
-        node_id = "{1}"
-        wrapper_id = "{1}_FAILED"
+        node_id = "{0}"
+        wrapper_id = "{0}_FAILED"
         # Defining scripts to be run
-        scripts= {0}
-        """).format(str(self.job_scripts), self.get_random_alphanumeric_string(5,5),'\n'.ljust(13))
+        scripts= {str(self.job_scripts)}
+        {os.linesep.ljust(13)}""").format(self.get_random_alphanumeric_string(5,5))
```
kinow: I think here we are mixing f-strings and `''.format`. I think it'd be simpler to move the code in `.format` out, before as a variable. Then just use only f-strings.

kinow (commented):
I think there's just one place where you left f-strings + str.format. Fixing that, I think it should be ready for a quick review if CICD passes, then approve & merge 🎉 Thanks Erick!

autosubmit/platforms/wrappers/wrapper_builder.py
```diff
@@ -744,31 +765,31 @@ def build_joblist_thread(self):
...
-
-        return textwrap.dedent("""
+        return textwrap.dedent(f"""
         # Getting the list of allocated nodes
-        {0}
+        {self.allocated_nodes}
         os.system("mkdir -p machinefiles")
 
-        with open("node_list_{{0}}".format(node_id), 'r') as file:
+        with open("node_list_{{{self.allocated_nodes}}}".format(node_id), 'r') as file:
```
kinow: .format not needed here?

kinow (commented):
Using GitHub UI I went searching for ".format" and I think there are a few ones that I **think** are not needed too (sorry, missed those before). Also left a comment asking if we are trying to remove all format calls... because there are a few left-overs in other functions (but not in the scope of the PR, so those can be ignored if you are not aiming at removing all format calls).

Thanks @VindeeR !

Issue #1199, opened 2024-01-05
Improve how we log errors


kinow:
Last one for today from !395.

I noticed it some months ago, but only now I had some time to carefully look at the code. I think it might be due to our custom `Log` class, or some other thing, but our exception stack traces are not showing everything. See this error for example.

```bash
Configuration files OK

Traceback (most recent call last):
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 2814, in recovery
    raise AutosubmitCritical(
log.log.AutosubmitCritical:  

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/kinow/Development/python/workspace/autosubmit/bin/autosubmit", line 37, in main
    return_value = Autosubmit.parse_args()
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 685, in parse_args
    return Autosubmit.recovery(args.expid, args.noplot, args.save, args.all, args.hide, args.group_by,
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/autosubmit.py", line 2819, in recovery
    raise AutosubmitCritical(
log.log.AutosubmitCritical:  
[ERROR] Trace:  
 [CRITICAL] Couldn't restore the job_list or packages, check if the filesystem is having issues [eCode=7040]
More info at https://autosubmit.readthedocs.io/en/master/troubleshooting/error-codes.html
```

The user sees the last message, "Couldn't restore the job_list or packages, check if the filesystem is having issues [eCode=7040]". The traceback is, however, hiding a more useful message, created where the initial exception is raised.

![image](https://earth.bsc.es/gitlab/es/autosubmit/uploads/ad598209270bfd9af0b1050b1439bfd0/image.png)

In the image above, you should be able to see the original error: "Experiment can't be recovered due being 1 active jobs in your experiment, If you want to recover the experiment, please use the flag -f and all active jobs will be cancelled".

But that error message is never displayed to the user.

This means users will open an issue in Autosubmit, a developer will open the code, see that message when following the traceback back to where it originated, and then tell the user something that could have been printed.

This is not very efficient (for both users and devs), and I think we can do better at printing the complete traceback for our users, empowering them and allowing them to troubleshoot issues much faster (i.e. a better user experience for Autosubmit users).

kinow:
mentioned in issue digital-twins/de_340/project_management#632

kinow:
[...]

PR #2745, opened 2025-12-16
Improve exception logging, fix unit test randomness, fix Slurm intermittent failure

### LINKED ISSUE #2744: Intermittent failures in test-slurm (infinite loop?
Somehow, we are getting what appears to be an infinite loop in test-slurm.

@VindeeR noticed this, but I had had the same in #2474.

I tested setting the `JobList.save_wrappers`back to the original code, instead of using `side_effects` in https://github.com/BSC-ES/autosubmit/blob/b2156ae4501c18fa5cb922bd2cf9d6bcf92d517d/test/integration/platforms/test_slurmplatform.py#L1226

But I still got the same error on my PR in #2474 

Created this issue as we had a discussion in Slack, and another in #2474 .

cc @VindeeR , @dbeltrankyl

kinow:
Closes #2744 
Related #1199

What this PR does:

- [x] Fix intermittent failure due to a random used in tests (not it's deterministic)
- [x] Changes so we run integration & regression tests only with oldest and newest versions
- [x] Add __str__ and __repr__ to Autosubmit exceptions to avoid hiding useful logging information
- [x] Deleted `test/unit/test_log.py` and combined it with `test/unit/log/test_log.py`
- ~[x] Spawns only 1 container for all test session for Git/Postgres/Slurm/SSH, even when using pytest-xdist workers (default 1 per core, I think, so my machine spawned 16 containers)~ This did not work as expected, at least we now know we need to improve the SSH containers first
- [x] Isolates the tests by calling next_expid when no expid given, does not use `autosubmit expid` directly as that causes duplicate IDs (each xdist process is an isolated python process), and does not re-use expid directories
- [x] Improve test_log_recovery that before relied on sleep(2) and could fail randomly (as it did for me), now it waits 10 seconds checking every second if the log recovery process is still alive
  - Note, working on this branch, when I had duplicate expid's across different pytest-xdist processes (i.e. 2 or more Python processes using the same expid) I got the Slurm container stuck, just looping and doing something, but without logging anything -- I believe this test isolation will fix the infinite loop we saw in our tests, the good news is that apparently it is not related to any code in Autosubmit & platforms
- [x] Makes the xz test more portable as it didn't work on my Ubuntu LTS nor on my newer Ubuntu 25.10

**Check List**

- [ ] Documentation updated.


kinow:
The only commit here, for now, is removing the test added in #2474 . Let's first confirm that test is the culprit.

kinow:
And what's crazier. Now the unit test failed, and test-slurm passed. All I did in the last commit was to increase the wallclock of platform and test in the integration test (no changes in the code related to unit test / joblist).

kinow:
Managed to get a single container (the other one is the monitoring container). However, it only works when a single test runs each time :thinking: When pytest-xdist is enabled I get several errors with SSH connections. Weird it works without xdist, but fails with it.

image

kinow:
Now Git server and SSH server and Postgres server need the same approach done in Slurm to avoid 1 container for each xdist worker. I'll also disable integration/regression on intermediary versions, leaving only 3.9 and 3.12 (oldest, newest supported) to save some time & resources in GH Actions.

test/integration/commands/run/conftest.py
```diff
@@ -63,17 +60,19 @@ def prepare_scratch(tmp_path: Path) -> Any:
...
-    dummy_dir.mkdir(parents=True)
-    real_data.mkdir(parents=True)
+    def _prepare_scratch(expid: str) -> None:
+        dummy_dir = tmp_path / f"scratch/whatever/{tmp_path.owner()}/{expid}/dummy_dir"
+        real_data = tmp_path / f"scratch/whatever/{tmp_path.owner()}/{expid}/real_data"
+        dummy_dir.mkdir(parents=True)
+        real_data.mkdir(parents=True)
+
+        with open(dummy_dir / 'dummy_file', 'w') as f:
+            f.write('dummy data')
```
kinow: Changed to return a factory function instead. That's because I needed it to be executed after other fixtures.

test/integration/test_scheduler_general.py
```diff
@@ -230,6 +230,9 @@ def test_scheduler_job_types(scheduler, job_type, autosubmit, autosubmit_exp: Ca
     if not expected_data:
         assert False, f"Could not find the expected data for {scheduler} and {job_type}"
 
+    # Replace the expid
+    expected_data = expected_data.replace('t000', expid)
```
kinow: Our example test files, in `test/files` have hard-coded expids, but instead of using placeholders or trying anything complicated, I thought it'd be simpler to just replace them as needed (in the end, `t000` behaves like a placeholder...)

test/unit/log/test_log.py
```diff
@@ -329,3 +340,70 @@ def test_set_file_more_than_10_files(test_tmp_path: Path, mocker):
...
+    output_file = test_path / "test-compressed.xz"
+    compress_xz(input_file, output_file, preset=9, extreme=True)
+
+    assert Path(output_file).exists()
+    assert is_xz_file(output_file)
+    assert len(find_uncompressed_files(str(test_path))) == 1
+
+    # Cover nonexistent path
+    with pytest.raises(FileNotFoundError):
+        find_uncompressed_files(str(tmp_path.joinpath("nonexistent_path")))
```
kinow: These were in `test/unit/test_log`. Moved them here to combine everything in a single `test_log.py`.

kinow:
Git tests now running in seconds, and using a single container! :tada: 

I've also added a marker `git` for pytest tests that use Git, in case we need to run only git tests, or anything by git, etc.

Now next SSH, 2FA, and Postgres.

test/unit/log/test_log.py
```diff
@@ -33,7 +39,7 @@ def test_autosubmit_error_default_values():
     assert as_error.code == 6000
     assert as_error.trace is None
 
-    assert str(as_error) == " "
+    assert str(as_error) == "Unhandled Error\nError code: 6000"
```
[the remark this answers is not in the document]
kinow: In my case I had to debug it myself. I configured an experiment for an integration test, and ran it a few times and couldn't find anything on the experiment log files (or those log files that do not have an expid, in the AS installation folder). The pytest logs had nothing (even enabling output with `-v -s`), and even after setting up a breakpoint and looking inside the container to look at the remote logs, nada.

This is one place where this happens, when AS tries to connect to a platform like MN5 or LUMI (remote, via SSH):

https://github.com/BSC-ES/autosubmit/blob/29af393b483992d4c78ccda0249357d938497ed8/autosubmit/platforms/paramiko_platform.py#L265-L280

Here's what happens when we do a simple `print` without re-raising the exception.

```python
>>> from autosubmit.log.log import AutosubmitError
>>> e = AutosubmitError("An error happened opening a configuration file ABC.yaml")
>>> print(f'There was an error: {str(e)}')
There was an error:  
```

And on this branch,

```python
>>> from autosubmit.log.log import AutosubmitError
>>> e = AutosubmitError("An error happened opening a configuration file ABC.yaml")
>>> print(f'There was an error: {str(e)}')
There was an error: An error happened opening a configuration file ABC.yaml
Error code: 6000
>>> 
```

Normally you either use the Python parent class' functions for __str__/__repr__, or you implement something with more meaningful information about your exception. From Python docs:

image

And an exception from Cylc source code:

https://github.com/cylc/cylc-flow/blob/9d50fc02fce5898c2ed86fd636c7a70a02620233/cylc/flow/exceptions.py#L119-L127

image

I never had to make an exception return empty or anything that was not something to be logged in Python, but maybe there was a reason for this initially, I am not sure.

When I added a new __str__ function, then I could see the problem in my experiment, which I fixed and the test then passed.

manuel-g-castro:
Hi @kinow , I am reading this issue but I cannot understand what is the issue with the slurm container. 

What I read so far is that you want to reuse the same container for multiple tests, right? And it randomly gives these "timeout banners" to autosubmit which make the test fail. 

In this case, I am thinking of an ssh configuration that triggers timeouts if too many requests are made. 

I could try to take a look into this, but without a reproducible example I can only assist you with more [shotgun debugging](https://en.wikipedia.org/wiki/Shotgun_debugging) hahahahah

manuel-g-castro:
Hi @kinow !

> The execution time is about the same now.

Okay! That is great! 

> The Slurm container and our error handling needs some improvement.

Yesterday I was looking into how to improve quality of life with a unified log to the stdout of the container. I looked into switching the init process to systemd. Then the entrypoint would be `journalctld -f`.

Since we are using ubuntu:slim, the image already has sytemd installed. It is just a matter of configuring it. 

Other option would be to have an infinite loop tailing all the logs (`/var/log/slurm/*` and wherever we put the sshd log) but I think that this solution maybe causes, well, infinite loops and too much verbosity with duplicity of lines of logs.

Maybe Pradeep knows a better solution still using tini.

manuel-g-castro:
Hi, just following on the meeting today.

Besides Pradeep, in the computer sciences department, and specifically the Edge to Cloud Computing Paradigms group, there is [Vicente Masip](mailto:vicente.masip@bsc.es) who could help us.

.github/workflows/ci.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.12"]
         pytest-markers: ["not postgres and not slurm"]
         # Here we include an extra build with the lowest version of Python and Postgres.
```
dbeltrankyl: Maybe we can keep only 3.9 and 3.11? 

* 3.9 is deployed in Destine
* 3.11 is deployed in the BSC environment. 

And personally, to me 3.12 debugger is buggy so I avoid working with that version

test/integration/git/test_autosubmit_git.py
```diff
@@ -156,6 +155,7 @@ def test_git_local_dirty(
         'OK: Committed and pushed the submodule, and committed and pushed the repository of a common experiment',
     ]
 )
+@pytest.mark.git
```
dbeltrankyl: This is also new, no?

dbeltrankyl (approved):
Hello @kinow 

Thanks for the intensive work!. There is a lot of new stuff for the tests

I did some comments, but I feel like that may be outscope of what this branch is trying to fix which I don't have much to add regarding that, so we can merge as it is

test/integration/git/test_autosubmit_git.py
kinow: Yup. This one too. It uses a separate contaiber too. So we cqn isolate it now.

test/integration/test_utils/pytest.py
```diff
@@ -0,0 +1,112 @@
...
+    """Check if a marker is used in the test.
+
+    Returns ``True`` if the caller test is decorated with a
+    marker that matches the given text. Otherwise, ``False``.
+    """
+    markers = request.node.iter_markers()
+    return any(marker.name == txt for marker in markers)
+
+
+def get_next_pytest_base_temp() -> Path:
```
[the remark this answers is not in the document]
kinow: Ah I think this can be remov3d. I had it for accessing the pytest folder before pytest created it (so containers would be started before pytest and pytest-xdist). Will che k and confirm if we can remove it. Thanks!

Issue #973, opened 2023-03-15
[enhancement] Simplify autosubmit.py, improving AS performance, code maintenance, and developer experience


kinow:
Hi,

This is an issue I wanted to created and remembered after @mcastril 's [comment on a MR](https://earth.bsc.es/gitlab/es/autosubmit/-/merge_requests/320).

At the moment our `autosubmit.py` module is growing, nearly reaching 6000 lines of code.

All the subcommands, such as `autosubmit monitor`, `autosubmit create`, etc., are created on that file. That means that as we modify these subcommands, or add new ones, that line number keeps changing (probably increasing).

This is hard for developer experience as you have to modify a really large file, with some changes at the top of the file, others at the end, and sometimes you may need to copy code from other commands too. On my IDE I normally divide the screen to have two `autosubmit.py` views, one at each part of that file, and it's still not enough at times.

![image](https://earth.bsc.es/gitlab/es/autosubmit/uploads/156aa9f2e7ec77291abeb2692d44917b/image.png)

Still on developer experience, having a large file like this means that if two or more developers are working on it, it's likely to have some merging conflicts in merge requests.

If you measure test coverage, or run static analysis/code inspection, you will also get **a lot** of warnings all over the code. And you may not notice that something was missing at one part of that file due to its length (see the MR linked at the top of this description for a real-world example). I disable all inspections on that file, as otherwise my right sidebar, where inspection warnings/errors are displayed looks almost multi-color, instead of gray with a few points to be corrected.

Finally, that file contains a lot of imports for **all** of the Autosubmit subcommands. That means when you run `autosubmit create`, the Python interpreter has had to load imports for `monitor`, `report`, `expid`, `updatedesc`, etc. Even though Python tries to be clever about caching imports, it's still often a big offender when benchmarking and profiling code for performance issues [^1] [^2].

I suggest we:

1. Create a directory `autosubmit/commands` (or `subcommands`, or `scripts`, etc.)
2. Create a separate module for each subcommand, i.e. `autosubmit/commands/create.py`, `autosubmit/commands/report.py`, and so on
3. Load these modules dynamically, as needed, i.e. if the user typed `autosubmit create`, then in `autosubmit.py` we will parse the args and upon realizing we want the `create` subcommand, we load its module and ask it to be executed

With this approach, coverage would be easier to be visualized as it would involve a small portion of `autosubmit.py` (if any) and a module in the `subcommands/` directory.

Developers working on separate subcommands would not have merge conflicts.

[...]

PR #2778, opened 2026-01-19
Remove files arg from upgrade_scripts, and move the code out of Autosubmit class

### LINKED ISSUE #2711: files arg is never used in Autosubmit.upgrade_scripts
### Description

Found while rebasing #2474 

Image

An internal function has an arg with the same name. But the `files` from the outer function is never used. It's, however, passed from `args.files` in the command line... so I guess that's just never used.

Haven't checked if that breaks the code. But I guess if the `files` given by the user are to be processed, they are probably being ignored and only the conf files are being upgraded.

https://github.com/BSC-ES/autosubmit/blob/6534303f98584a204810d1d1a5e408eb56a67a06/autosubmit/autosubmit.py#L4016C37-L4029

### Reproducible Example

NA

### Expected Behaviour

We should probably delete the arg if not used, or use it as it was intended.

kinow:
Closes #2711

The code to upgrade scripts is self-contained and easy to move and improve. A good way to kill some technical debt, especially reducing the Autosubmit class (related #973).

- [x] Adds more types
- [x] Remove dead code
- [x] Replaces some `os.path` by `Path` in the code changed
- [x] fixes some pydoc comments
- [x] moves code changed out of Autosubmit
- [x] the main function, `upgrade_scripts` was moved out of Autosubmit, but it had dependency to other function in Autosubmit that could be moved to another file too (as it's only used by `upgrade_scripts`, or it's related to configuration). Doing so reduced configcommon.py and autosubmit.py lines of code.
- [x] Added new tests to cover the code changed.
- [x] Replaced the recursive `_update_dict` by an iterative `_update_dict` ([ref](https://www.geeksforgeeks.org/python/when-not-to-use-recursion-while-programming-in-python/))
- [x] Removed print statements
- [x] In upgrade_scripts.py, removed a repeated created dict by re-using the same dict
- [x] Removed unnecessary enumerate call
- [x] Replaced the loop + re.sub by a single call to re.sub with a callback (less code, less places to err, might be faster if using C under the hood)
- [x] ini_to_yaml would read the INI file, replace lists, write back to INI file, convert to YAML, write to INI file, rename the file to YAML... now it just reads INI, convert content, and writes to YAML....
- [x] Fixed Optional[] ruff issues, using Type | None now

**Check List**

- [ ] Documentation updated.


test/unit/test_ecplatform.py
kinow: I opened this file by accident but noticed a small lint warning, so fixed that here 👍

autosubmit/autosubmit.py
kinow: Changes here are mainly removing things, and renaming a function call. I believe this should be easy to rebase/fix conflicts in the joblist PR, @dbeltrankyl 👍

VindeeR (commented):
I've done a review of what you've done so far and things seems already to be getting much simpler, and cleaner thanks for the great step towards improving the quality of code 🙌

kinow:
Rebased fixing conflicts. Will start working on ruff, mypy, finish the python changes, and then add final tests.

kinow:
Rebased.

kinow:
Locally `autosubmit.py` function `expid` is now being 100% covered, and I fixed all mypy issues in `db_common.py` and `experiment_common.py` while fixing the issues in this `expid` function. Mypy will fail, but that's because of the other files linked to this change.

autosubmit/database/db_common.py
kinow, line 1: Typing and docs updated.

autosubmit/database/db_structure.py
kinow, line 1: Typing and docs updated.

autosubmit/experiment/experiment_common.py
kinow, line 1: Typing and docs updated.

autosubmit/job/job_dict.py
kinow, line 1: Typing and docs updated.

autosubmit/provenance/rocrate.py
kinow, line 1: Typing and docs updated.

test/integration/commands/run/test_run_slurm_scenarios.py
kinow, line 1: Typing and docs updated.

test/integration/test_utils/ssh.py
kinow, line 1: Typing and docs updated.

test/integration/conftest.py
kinow, line 1: Typing and docs updated.

test/integration/test_autosubmit.py
kinow, line 1: We don't need an expid, and it's actually dangerous to have intermittent failures. Leave it to be randomly calculated.

test/unit/test_dependencies.py
kinow, line 1: Remvoved dependency on `mock`. Just one file missing so we can drop it from requirements.

test/unit/test_packages.py
kinow, line 1: Remvoved dependency on `mock`. Just one file missing so we can drop it from requirements.

pyproject.toml
```diff
@@ -87,10 +87,18 @@ tests = [
     "testcontainers",
     "pytest-timeout",
     "roc-validator",
-    "types-requests",
     "typing_extensions==4.*",
     "pytest-timestamper"
 ]
+mypy = [
```
kinow, line 92: We had a GH Actions job with the dependencies listed individually there, plus some in the test group.

I've moved all to the mypy group, and updated the GH Actions to now install this optional group so it brings all typing dependencies. When adding new deps, please check that in pypi the package is maintained by that typeshed GitHub org. If it's maintained by some random dev not member of the main project, better be safe and type: ignore instead (security, etc.).

kinow:
Stashing what I have for today. It's missing just a few tests for `upgrade_scripts.py` now, then it should bring coverage a bit up in the report, but also fix several mypy and pydoc warnings/issues. :pray:

kinow:
Nearly there. The build will likely fail again. But I covered the exception parts. Just need to finish with or two more tests for the `upgrade_scripts` function, and then this should be ready for review. Unfortunately, I think I won't be able to complete it before I go on holidays and come back on June 1st.

kinow:
Rebased fixing conflicts.

kinow:
Looks like most of the lines that are not covered are `Log.info`, `Log.warning` lines. In another PR I added `# pragma: no cover` to those lines. But adding to every log line now seems a bit of overkill.

I tried a couple of settings in coveragerc, and also in codecov config. But nothing fixed it. 

I ran the report locally, and the HTML & XML generated by pytest-cov + coverage is correct. It shows the line as excluded. It appears to be a bug in Codecov, possibly this one: https://github.com/codecov/codecov-action/issues/1326

Maybe @VindeeR or @ntorqulu can review this issue? The idea here is to add tests to the code that upgrades AS3 to AS4, verify that it works (it was not tested, and it was not working as expected -- it did work at some point, years ago). I moved the code out of the `Autosubmit` class, and restructured the code to upgrade code so it didn't suffer from infinite recursion or stack limits, added more tests, types, docs, etc. There are 2 or 3 lines that are not tested, but they are related to this code which is not critical and was not tested at all. I think the lines missing are harmless, but feel free to have a look and let me know if you think it's necessary to have a test for those :+1: 

I'll start syncing things to start reviewing a PR by Suraj and another one from Luiggi today/tomorrow. Thanks!

kinow:
Whew, quite a few conflicts, but in the end it was a ~5 minutes job. I'll check if the build fails in case I missed anything, and if so I will just use reflog &|| GH UI to fix it. Should be ready to review again tomorrow morning.

EDIT: ready for review

kinow:
@VindeeR no need to review it for now. I will wait for 4.2.0 joblist PR to be merged, so Dani doesn't have to rebase his branch one more time.

.github/workflows/lint.yml
```diff
@@ -37,12 +37,8 @@ jobs:
       - uses: actions/setup-python@v7
         with:
           python-version: ${{ matrix.python-version }}
-      - run: pip install mypy
       - run: |
-          python3 -m pip install \
-            types-mock \
-            types-requests \
-            types-paramiko  
+          pip install -e .[mypy]
```
kinow, line 41: @VindeeR ☝️ I think we may need a similar install using a group in lint.yml later, for ruff. Now, the imported types are controlled in pyproject.toml, and we can run mypy the same way as in CI with pip install -e .[mypy] then the mypy commands.

kinow (commented):
@VindeeR this should be ready for review too, if CICD passes. I think docs/sphinx will fail because of py3.9. I believe you have a fix for it in another branch, so I'll leave that failing here 👍

autosubmit/config/upgrade_scripts.py
```diff
@@ -0,0 +1,299 @@
...
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Code to handle upgrading Autosubmit scripts between AS versions."""
+
+import locale
+import re
+import shutil
+from io import StringIO
+from pathlib import Path
+from typing import Any, Optional
```
VindeeR: Are we gonna keep using Optional? I believe it is better to migrate to the new standard of
`Type | None`

This could be easily addressed in a PR just focusing on that, and there's no need to use the `__future__ from annotations`, if we are using `3.10>`, just checking with you what you think would be the best approach to this.

autosubmit/config/upgrade_scripts.py
```diff
@@ -0,0 +1,299 @@
...
+
+from configobj import ConfigObj
+from ruamel.yaml import YAML
+
+from autosubmit.config.basicconfig import BasicConfig
+from autosubmit.config.configcommon import AutosubmitConfig
+from autosubmit.config.yamlparser import YAMLParserFactory
+from autosubmit.database.db_common import (
+    update_experiment_description_version
+)
```
VindeeR:
```suggestion
from autosubmit.database.db_common import update_experiment_description_version
```

autosubmit/config/upgrade_scripts.py
```diff
@@ -0,0 +1,299 @@
...
+    If the file name contains "jobs" or "platform" in ANY part of its name,
+    then the generated YAML data will become ``{"JOBS": yaml_data}`` or
+    ``{"PLATFORMS": yaml_data}``.
+
+    :param ini_file: Path to the AS3 conf file to convert.
+    :return: The YAML file.
+    """
+    yaml_file_path = ini_file.with_suffix(".yml")
+    if yaml_file_path.exists():
+        Log.debug(f'AS3 conf file {ini_file} not upgraded. YAML file already exists: {yaml_file_path}')
```
VindeeR, line 123: ruff doesn't complain about `' or "`, pycharm converts them into `"`, but I believe you said that vscode might do `'` should we decide on one and make it standard?

autosubmit/config/upgrade_scripts.py
```diff
@@ -0,0 +1,299 @@
...
+
+    if not template_or_script_path.exists():
+        Log.warning(f"Skipping template not found: {template_or_script_path}")
+        return warnings, success
+
+    with open(template_or_script_path, 'r', encoding=locale.getlocale()[1]) as f:
+        template_or_script_content = f.read()
+    # TODO: quite sure this is duplicating work done in the config module, we
+    #       can reuse the same code.
+    # Look for %_%, and make them all uppercase (Autosubmit 4 default format)
```
VindeeR, line 259: was this addressed at some point? not a priority just wondering

autosubmit/config/upgrade_scripts.py
```diff
@@ -0,0 +1,299 @@
...
+    variables = re.findall('%(?<!%%)[a-zA-Z0-9_.-]+%(?!%%)', template_or_script_content, flags=re.IGNORECASE)
+    variables = [variable[1:-1].upper() for variable in variables]
+    results: dict[str, set] = {}
+    # Change format
+    parameters = as_conf.load_parameters()
+    for old_format_key in variables:
+        for key in parameters.keys():
+            last_key = key.split(".")[-1]
+            if last_key == old_format_key:
```
VindeeR: Maybe it would be better optimized to create the list outside and check it once inside the for loop, this way we would avoid this nested for that might affect bigger projects.

Not sure how much of an impact that would be tho since I'm not sure how big the AS3 projects are.

autosubmit/job/job_dict.py
```diff
@@ -210,9 +210,9 @@ def _create_jobs_split(
         self,
         splits: Union[int, dict[str, list[int]]],
         section: str,
-        date: Optional[datetime.datetime],
-        member: Optional[str],
-        chunk: Optional[int],
+        date: datetime.datetime | None,
```
VindeeR, line 213: I see you're using the new "Optional" here but there are some instances where you're using the older version, I guess just answering the other one should be fine, just wanted to point out here in case we decide on using the other one.

test/integration/test_expid.py
```diff
@@ -751,13 +778,85 @@ def test_autosubmit_generate_config_resource_listdir_order(autosubmit, mocker, m
...
+def test_autosubmit_generate_config_non_dict_yaml(
+    mocker,
+    monkeypatch,
+    autosubmit: Autosubmit,
+    tmp_path,
+    get_next_expid,
+):
+    """generate_as_config handles YAML documents that are not dictionaries."""
+    expid = get_next_expid()
+    monkeypatch.setattr(BasicConfig, "LOCAL_ROOT_DIR", tmp_path)
```
VindeeR, line 794: correct me if I'm wrong but, wouldn't be better to do like you did above and change this to 

`BasicConfig.LOCAL_ROOT_DIR = tmp_path`

This way we wouldn't need to use the `monkeypatch` fixture

test/integration/test_expid.py
```diff
@@ -751,13 +778,85 @@ def test_autosubmit_generate_config_resource_listdir_order(autosubmit, mocker, m
...
+def test_autosubmit_generate_config_non_dict_parameters(
+    mocker,
+    monkeypatch,
+    autosubmit: Autosubmit,
+    tmp_path,
+    get_next_expid,
+):
+    """generate_as_config ignores invalid non-dict parameter structures."""
+    expid = get_next_expid()
+    monkeypatch.setattr(BasicConfig, "LOCAL_ROOT_DIR", tmp_path)
```
VindeeR, line 823: this one too

VindeeR (approved):
I'm confused with the coverage on this branch, probably it is weird due to the merge of 4.2, but I don't see enough of a justification to why it is dropping almost 1%, Do you have any idea?

Regardless of that everything looks fine, I did mostly some questioning on some of the things you did to clarify it for future reference.

Thanks for the improvement on the quality of the code 🙌

dbeltrankyl:
> I'm confused with the coverage on this branch, probably it is weird due to the merge of 4.2, but I don't see enough of a justification to why it is dropping almost 1%, Do you have any idea?

It is weird indeed because before merging 4.2 I checked it and I had 82.21%

image

dbeltrankyl:
Ah it is because the comment was not updated reflecting the new coverage ( not sure why, because the coverage check is green  ) 

This is the last update: 

image

autosubmit/config/upgrade_scripts.py
kinow: Nope, I updated another PR, buy I am afraid we still have it in a lot of places. I can update all optionals here, or we can do the batch update to fix ruff.

About coverage, it might be because I am deleting code.In some cases deleting tested code reduces the coverage a bit.

autosubmit/config/upgrade_scripts.py
kinow: My pycharm doesn't convert them. I told @ntorqulu that if we decide to enforce double quotes everywhere, then once cicd is checlking them we can enforce that in code reviews too.

autosubmit/config/upgrade_scripts.py
kinow: Thanks!

autosubmit/config/upgrade_scripts.py
kinow: Not yet. But IMO we should. No point in having that done in two functions.

autosubmit/config/upgrade_scripts.py
kinow: Let me check that with the IDE. Sounds like a good suggestion!

test/integration/test_expid.py
kinow: Not really. Changing that might affect other tests. Whereas monkeypatch only applies the change during this one test. I replaced some calls by monkeypatch exactly to fix some intermittent failures (there are more, related to pypi install and docker containers)

VindeeR:
Maybe kicking the build again could fix the coverage strange behavior, as @dbeltrankyl mentioned, it was last updated 3 weeks ago

autosubmit/config/upgrade_scripts.py
kinow: `ruff check --select UP045 --fix` applied to code.

test/integration/test_expid.py
kinow: Same as above, kept to avoid changing BasicConfig and affecting other tests in the same pytest session.

autosubmit/job/job_dict.py
kinow: Fixed across project, `ruff check --select UP045 --fix`.

autosubmit/config/upgrade_scripts.py
kinow: I hope I got it right. Here's the updated loops

```py
    # Change format
    parameters = as_conf.load_parameters()

    # Build an index of parameters by their last part to avoid
    # iterating over all the parameters for each variable.
    by_last_key: dict[str, set[str]] = {}
    for key in parameters:
        last_key = key.split(".")[-1]
        value = "%" + key.strip("'") + "%"
        by_last_key.setdefault(last_key, set()).add(value)

    # Look up matching parameters directly (without performing a nested search).
    for old_format_key in variables:
        matches = by_last_key.get(old_format_key)
        if matches is not None:
            results.setdefault(old_format_key, set()).update(matches)
```

kinow:
I think RTD might be reading the 3.9 from pyproject.toml. @VindeeR once we merge #3152  I will rebase this one, and hopefully the docs builds will pass.

kinow:
Rebased. Let's see if the docs are fixed now.

kinow:
@VindeeR assuming CICD passes (two test-integration running, and the test-slurm, which may need a kick if they fail), this should be ready for review again! I've fixed all `Optional[]` with ruff (and a few manually). Thanks!

kinow:
Done @VindeeR . Coverage has some issues with the `Log.` calls, which is why (in part) the upgrade_scripts.py is not 100%. There are other lines there that were not tested as it was a bit hard to come up with the configuration to reach those lines.

I could write a unit + mock, but it wouldn't guarantee that the code actually works (i.e. I'd prefer to take the longer path, and investigate if a config can be used to reach that or not first). But given the code is rarely used, and was not tested before, I think what we have is already an improvement over status quo :sweat_smile:

autosubmit/database/db_manager_job_list.py
```diff
@@ -150,7 +150,7 @@ def load_jobs(
             self,
             full_load: bool = False,
             load_failed_jobs: bool = False,
-            members: Optional[List[Any]] = None
+            members: List[Any] | None = None
     ) -> List[Dict[str, Any]]:
```
VindeeR, line 154:
```suggestion
            members: list[Any] | None = None
) -> list[dict[str, Any]]:
```

autosubmit/database/db_manager_job_list.py
```diff
@@ -34,9 +34,9 @@
```
VindeeR, line 34: My ruff is suggesting to organize this like this


```
import datetime
from pathlib import Path
from typing import TYPE_CHECKING, Any, Tuple, Union

from sqlalchemy import and_, exists, func, not_, or_, select, update
from sqlalchemy.exc import IntegrityError

from autosubmit.config.basicconfig import BasicConfig
from autosubmit.database.db_common import get_connection_url
from autosubmit.database.db_manager import DbManager
from autosubmit.database.tables import (
    ExperimentStructureTable,
    JobsTable,
    PreviewWrapperInfoTable,
    PreviewWrapperJobsTable,
    SectionsStructureTable,
    Table,
    WrapperInfoTable,
    WrapperJobsTable,
)
from autosubmit.log.log import Log
```

The order is changing quite a bit

autosubmit/database/db_manager_job_list.py
```diff
@@ -199,9 +199,9 @@ def get_job_list_size(self) -> Tuple[int, int, int]:
     def select_job_names_by_sections(
             self,
             sections: List[str],
-            exclude_names: Optional[Set[str]] = None,
+            exclude_names: Set[str] | None = None,
             exclude_completed: bool = False,
-            status_filter: Optional[str] = None,
+            status_filter: str | None = None,
     ) -> Set[str]:
```
VindeeR, line 205:
```suggestion
            exclude_names: set[str] | None = None,
exclude_completed: bool = False,
            status_filter: str | None = None,
) -> set[str]:
```

autosubmit/database/db_manager_job_list.py
```diff
@@ -376,7 +376,7 @@ def select_loadable_inner_jobs(
     def select_active_jobs(
             self,
             include_failed: bool = False,
-            members: Optional[List[Any]] = None
+            members: List[Any] | None = None
     ) -> List[Union[str, Any]]:
```
VindeeR:
```suggestion
            members: list[Any] | None = None
) -> list[Union[str, Any]]:
```

autosubmit/database/db_manager_job_list.py
```diff
@@ -421,17 +421,14 @@ def select_finished_jobs_needing_log_recovery(self) -> List[Dict[str, Any]]:
     def select_children_jobs(
             self,
             job_list: List[Union[str, Any]],
-            members: Optional[List[Any]] = None
+            members: List[Any] | None = None
```
VindeeR, line 424:
```suggestion
            job_list: list[Union[str, Any]],
            members: list[Any] | None = None
```

autosubmit/database/db_manager_job_list.py
```diff
@@ -596,7 +593,7 @@ def load_edges(self, job_list: List[dict[str, Any]] = None, full_load: bool = Tr
             graph = self.select_edges(job_list)
         return graph
 
-    def select_edges(self, job_list: Optional[List[dict[str, Any]]] = None, only_parents: bool = False) -> List[dict[str, Any]]:
+    def select_edges(self, job_list: List[dict[str, Any]] | None = None, only_parents: bool = False) -> List[dict[str, Any]]:
```
VindeeR, line 596:
```suggestion
    def select_edges(self, job_list: list[dict[str, Any]] | None = None, only_parents: bool = False) -> list[dict[str, Any]]:
```

autosubmit/database/db_manager_job_list.py
```diff
@@ -647,17 +644,14 @@ def save_wrappers(
             self,
             wrappers: Tuple[List[Dict[str, Any]], List[Dict[str, Any]]],
```
VindeeR:
```suggestion
            wrappers: tuple[list[dict[str, Any]], list[dict[str, Any]]],
```

autosubmit/helpers/processes.py
```diff
@@ -50,7 +50,7 @@ def _match_expid(token: str, expid=None) -> bool:
     return re.search(_EXPID_REGEX, token) is not None
 
 
-def _match_autosubmit_cmdline(cmdline: List[str], command='run', expid: Optional[str] = None) -> Optional[str]:
+def _match_autosubmit_cmdline(cmdline: List[str], command='run', expid: str | None = None) -> str | None:
```
VindeeR, line 53:
```suggestion
def _match_autosubmit_cmdline(cmdline: list[str], command='run', expid: str | None = None) -> str | None:
```

autosubmit/job/job_list.py
```diff
@@ -1926,36 +1926,28 @@ def _manage_job_dependencies(
...
             chunk_list: List[int],
-            member: Optional[str],
+            member: str | None,
             member_list: List[str],
-            date: Optional[str],
+            date: str | None,
             date_list: List[str],
             dependency: Any
-    ) -> Tuple[bool, Tuple[Optional[int], Optional[str], Optional[str]]]:
+    ) -> Tuple[bool, Tuple[int | None, str | None, str | None]]:
```
VindeeR, line 1936:
```suggestion
            chunk: int | None,
            chunk_list: list[int],
            member: str | None,
            member_list: list[str],
            date: str | None,
            date_list: list[str],
            dependency: Any
    ) -> tuple[bool, tuple[int | None, str | None, str | None]]:
```

autosubmit/job/job_list.py
```diff
@@ -2911,14 +2903,11 @@ def sort_by_status(self):
         """
         return sorted(self.job_list, key=lambda k: k.status)
 
-    def save_jobs(self, jobs_to_save: Optional[List[Job]] = None, reset_log_counters: bool = False):
+    def save_jobs(self, jobs_to_save: List[Job] | None = None, reset_log_counters: bool = False):
```
VindeeR:
```suggestion
    def save_jobs(self, jobs_to_save: list[Job] | None = None, reset_log_counters: bool = False):
```

autosubmit/job/job_packages.py
```diff
@@ -280,9 +280,9 @@ class JobPackageThread(JobPackageBase):
     """
     FILE_PREFIX = 'ASThread'
 
-    def __init__(self, jobs: list[Job], dependency=None, jobs_resources: Optional[dict] = None,
-                 method: str = 'ASThread', configuration: Optional['AutosubmitConfig'] = None,
-                 wrapper_section: str = "WRAPPERS", wrapper_info: Optional[list] = None):
+    def __init__(self, jobs: list[Job], dependency=None, jobs_resources: dict | None = None,
+                 method: str = 'ASThread', configuration: 'AutosubmitConfig | None' = None,
```
VindeeR, line 284: I didn't know this is how to use forward reference with optional, kinda cool but I feel like it can get messy

test/unit/helpers/test_processes.py
```diff
@@ -60,7 +60,7 @@ def _create_process(mocker, expid, username: Optional[str] = None, command='run'
         'Two expids, none are valid, so none is retrieved',
     ]
 )
-def test_retrieve_expids(mocker, expids: List[str], expected_retrieved: int, username: Optional[str]):
+def test_retrieve_expids(mocker, expids: List[str], expected_retrieved: int, username: str | None):
```
VindeeR:
```suggestion
def test_retrieve_expids(mocker, expids: list[str], expected_retrieved: int, username: str | None):
```

test/unit/test_stop.py
```diff
@@ -17,7 +17,7 @@
 
 import signal
 from itertools import chain
-from typing import Optional, List
+from typing import List
```
VindeeR:
```suggestion
```

test/unit/test_stop.py
```diff
@@ -200,7 +200,7 @@ def test_stop_expids(autosubmit, mocker, expids: str, num_expids: int, cancel: b
         # (as we passed ``force=False``), so we avoid the ``sleep`` and loop by returning
         # a ``None``, which is interpreted as if the process was correctly killed by
         # ``os.kill``.
-        returned_pids: List[Optional[int]] = [pid + n for n in range(num_expids)]
+        returned_pids: List[int | None] = [pid + n for n in range(num_expids)]
```
VindeeR, line 203:
```suggestion
        returned_pids: list[int | None] = [pid + n for n in range(num_expids)]
```

autosubmit/config/upgrade_scripts.py
VindeeR: This looks like it is working fine to me, Thanks for the improvement :raised_hands:

VindeeR (approved):
Great work in here thanks for all the improvement and cleaning of the code it is very helpful and appreciated work <3 I just had a few suggestions related to the use of List or Tuple, but nothing major.

Again thanks for the hard work 🙌

autosubmit/database/db_manager_job_list.py
kinow: Good catch! Thanks!

autosubmit/database/db_manager_job_list.py
kinow: Yup. Organising with PyCharm gives me the same order. I hadn´t applied that formatting here. Just did it :+1: Thanks!

autosubmit/job/job_packages.py
kinow: Yeah, this is one place where using `Optional` looks "cleaner". At least for compiler/ASM and IDE other tools :sweat_smile: but that appears to be the only way with `|`.

kinow:
@VindeeR I started applying your changes, but realised some were left in the code (Union/Tuples, maybe because of the order of the changes applied in GH, a couple of comments GH UI refused to merge). So I started applying manually, but the number of lines changed in db_manage_list and job_list were quite high, so I reverted the changes. I think we can do a single PR for ruff/mypy after this one, fixing these files. I also found that pydocs need to be fixed. In some cases the `:type`, `:rtype` is not matching, or it has information that's not in type hints (so mypy won't recognise it, I believe).

VindeeR (approved):
No problem, addressing as we go is generally easier but since we had the bump up in version it makes things more complicated, I agree on having one big commit addressing it, it won't be great impact despite affecting many different files.

Issue #2617, opened 2025-09-30
Unarchive/ Migrate using Autosubmit 4 in an Autosubmit 3 experiment


dbeltrankyl:
Hello @kinow , @VindeeR 

We need to revise how we check if the experiment is an Autosubmit 3 experiment or it is an Autosubmit 4. 

### Description

Using Autosubmit 3.15..

Marga had an issue while migrating an experiment, not sure if it was a misconfiguration or not.

The problem was that the person using the pickup was not able of doing so locally. And after that, Autosubmit migrate was complaining that the experiment was not archived. 

At this point, the experiment was still under /esarchive/autosubmit/2025/$expid.tar

I told them to use `autosubmit unarchive $expid`. However it was launched with Autosubmit 4 by mistake and the experiment ( locally) was erased somehow.

```bash

The migrate with the issue:

[dbeltran@bsceshub03 Aslogs]$ cat 20250929_130620_migrate_a7w1.log
2025-09-29 15:55:45,316 Autosubmit is running with 3.15.19
2025-09-29 15:55:45,318 Migrating experiment a7w1
2025-09-29 15:55:45,319 Moving local files/dirs
2025-09-29 15:55:45,321 Experiment located in 2025 archive
2025-09-29 15:55:45,322 Unpacking tar file ...
[ERROR] 2025-09-29 16:13:36,326 Can not extract tar file: [Errno 13] Permission denied: u'/esarchive/autosubmit/a7w1/proj/auto-mars/.git/objects/c7/d439127421aea344b2074d0a8962b982ef0c81'[eCode=6012]
[CRITICAL] 2025-09-29 16:13:36,345 The experiment cannot be picked up [eCode=7012]

The first unarchive was using 4.1.15..

[dbeltran@bsceshub03 Aslogs]$ cat *unarchive*a7w1*
[ERROR] 2025-09-30 10:48:21,256 Experiment a7w1 is not archived
2025-09-30 10:48:21,235 Autosubmit is running with 4.1.15
[ERROR] 2025-09-30 10:48:21,256 Experiment a7w1 is not archived
[ERROR] 2025-09-30 10:48:47,313 Experiment a7w1 is not archived
2025-09-30 10:48:47,300 Autosubmit is running with 3.15.19
[ERROR] 2025-09-30 10:48:47,313 Experiment a7w1 is not archived
[ERROR] 2025-09-30 10:53:28,143 Experiment a7w1 is not archived
[...]

PR #2787, opened 2026-01-26
Remove the AS migrate command temporarily until it is fixed

kinow:
Another PR mainly removing things (shouldn't interfere with rebases too much).

Related to #2617 

**Check List**

- [x] Documentation updated.


VindeeR (approved):
The removal of the migrate command seems to be complete, with the maybe exception of the following.

https://github.com/BSC-ES/autosubmit/blob/29ea1f5489c9c6bb3f076db3ba6498a792aa59f3/autosubmit/autosubmit.py#L371-L374

I think this line doesn't make sense being kept, even if we were to add it back again later.

Regardless of my comment the change is complete, and you can merge it after removing this piece of code if agree.

PR #2788, opened 2026-01-26
Use a trap in Shell instead of heredoc to maintain the error line number

### LINKED ISSUE #2722: Fix Error Line Numbering
### Description

The line reported in the `.err` will always be where the bash interpreter was called due to the redirect of HEREDOC to the stdin of another bash call (see below snippet) if an error is thrown by the new bash interpreter. 

https://github.com/BSC-ES/autosubmit/blob/0531b9ec9fb1a85fd250e33fd3a34da67898f540/autosubmit/job/template/bash.py#L35

* Autosubmit version:  [0531b9ec9fb1a85fd250e33fd3a34da67898f540](https://github.com/BSC-ES/autosubmit/commits/fix-variables-being-interpreted-before-the-script-runs/)

### Reproducible Example

This is my script, and because of the `set -xuve`, it fails because of unbounded variables:

```
#!/bin/bash

set -xuve
echo ${UNBOUNDED_VARIABLE}
```

If I execute run this script with Autosubmit 4.15, i get the following error on the `.err`:

```
/app/autosubmit/experiments/a000/tmp/LOG_a000/a000_LOCAL_SETUP.cmd: line 45: UNBOUNDED_VARIABLE: unbound variable
```

which correctly indicates the faulty line at the 45ft line of the `.cmd`. Below you have the snipped of the script (with numbering).

```
    38	###################
    39	# Autosubmit job
    40	###################
    41	
    42	#!/bin/bash
    43	
    44	set -xuve
    45	echo ${UNBOUNDED_VARIABLE}
    46	
    47	###################
    48	# Autosubmit tailer
    49	###################
```

Now, if I do the same with the changes in master (`[c80654](https://github.com/BSC-ES/autosubmit/commit/c80654)`), the error points to the line that executes bash -- 46 --, but not the line where the error actually happens in the script  -- 51 --:

```
/app/autosubmit/experiments/a000/tmp/LOG_a000/a000_LOCAL_SETUP.cmd: line 46: UNBOUNDED_VARIABLE: unbound variable
```

Here is the snipped of the final `.cmd` script (with line numbers).

```
    46	bash -e <<__AS_CMD__
    47	set -xuve
    48	#!/bin/bash
    49	
    50	set -xuve
    51	echo ${UNBOUNDED_VARIABLE}
    52	
    53	__AS_CMD__
```

### Expected Behaviour

That the error line pointed by bash is matches the faulty code in the `.cmd` file.

kinow:
Closes #2722 

@manuel-g-castro I will probably need your help if you have time to verify this change. But the idea is to use a `trap` for the exit signal, and create the stat/completed file there. This way we don't need a heredoc/subshell, and the error line number should be displayed correctly.

**Check List**

- [ ] Documentation updated.


VindeeR (approved):
Review finished all good thanks for the hard work fixing the tests 🙌 

I just was a bit unsure regarding the execution time of 2 tests, but other than that everything is perfect and I believe it to be ready to merge

kinow:
@VindeeR review feedback addressed. Can you have another look at it, please? Thanks!

VindeeR (approved):
Everything looks good, thanks for the quick reply, it looks good to go 🙌

PR #2807, opened 2026-02-05
Update how we handle signals with trap in Shell scripts

### LINKED ISSUE #2806: Current shell trap on master may fail with Slurm timeout
### Description

Found by @dbeltrankyl and @dmcclainbsc on DestinE. When a job is killed by Slurm due to reaching the wallclock limit, out current shell trap function is executed and may create the _COMPLETED file even though the job is being cancelled.

I believe this is due to the trap function being called on EXIT, which I **think** is called always, even when there is a timeout. And the timeout in Slurm should raise both/or SIGTERM SIGKILL, I think.

https://github.com/SchedMD/slurm/blob/cead8b9d2c2360f976c77d9e9e7ab875de9d867d/src/common/run_command.c#L215

I have to test with Slurm, but I suspect that will not trigger the trap function immediately, as it's not "listening" for those signals, and only for the special EXIT call 

> Signal terminations are not caught by EXIT.
https://www.spinics.net/lists/dash/msg02231.html

@dbeltrankyl is working on a fix in #2725. This issue is to make sure it's added to master so we can start testing it. And it'd be nice to have some tests with slurm cancellations due to timeouts too, to avoid this issue repeating.

I think the fix might be a combination of what Dani suggests (moving the extended tailer out, and executing stat later in the trap, if I understood correctly), and then adding a second trap to signals which changes the exit code. This shouldn't be too hard to test on Slurm, but might take  a bit of time to test a few different scenarios.

### Reproducible Example

TODO

### Expected Behaviour

TODO

kinow:
Closes #2806

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/job/template/bash.py
+++ b/autosubmit/job/template/bash.py
@@ -46,24 +46,101 @@ fi
 echo "$(date +%s)" > "${job_name_ptrn}_STAT_%FAIL_COUNT%"
 
 ################### 
-# AS TRAP FUNCTION
+# AS TRAP FUNCTIONS
 ###################
+# Used in trap functions to control the exit code of the script.
+AS_EXIT_CODE=0
+
+# as_signals_handler
+#
+# This function is intended to be used as a trap handler and
+# terminates the script with a signal-derived exit code.
+#
+# List of captured signals:
+# Ref: https://services.criann.fr/en/services/hpc/cluster-myria/guide/signals-sent-by-slurm/
+#
+# - SIGHUP,  terminal/session closed
+# - SIGINT,  user interrupt, e.g., CTRL+C
+# - SIGQUIT, quiet/core dump, e.g., CTRL-\
+# - SIGPIPE, broken pipe
+# - SIGTERM, graceful termination request, e.g., Slurm wallclock (then a SIGKILL shortly afterwards)
+# - SIGXCPU, CPU time limit exceeded, e.g., job exceeded Slurm/PBS CPU time
+# - SIGXFSZ, File size limit exceeded, e.g., Kernel detected a process is exceeding `ulimit -f`/quota
+#
+# We do NOT capture SIGUSR1 and SIGUSR2. These are user signals that can be enabled in
+# Slurm via `--signal`. But since these will be executed seconds before a SIGTERM, it
+# is safer to capture SIGTERM only (imagine if we receive the SIGUSR1, and in 2 seconds
+# the SIM job completes successfully -- we do not want to capture it and mark as an error).
+#
+# We also do not capture SIGPIPE. SIGPIPE would work with our `-o pipefail`, but it
+# could trigger the trap function if any command in the user template raises a pipe
+# error, even if the user disabled the pipe fail. The `as_exit_handler` function
+# handles actual broken pipe errors.
+#
+# Arguments:
+#   $1 - the Unix signal
+#
+# Globals:
+#   AS_EXIT_CODE
+#
+# Side effects:
+#   Exits the script with the computed exit code (use with care if calling this directly).
+#
+# Exit codes:
+#   1            if a signal was received but we do not have a match case 
+#   128 + signal based on the signal received
+function as_signals_handler() {
+    local signal="$1"
+    # Convert signal to standard Unix exit code:
+    # exit code = 128 + signal number
+    # Ref: https://www.ditig.com/linux-exit-status-codes#avoid-confusion-with-signal-numbers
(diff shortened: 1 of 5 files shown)


autosubmit/job/template/bash.py
```diff
@@ -87,7 +119,9 @@
 # Autosubmit tailer
 ###################
 # Job completed successfully
-# The exit trap will handle the tailer
+
+%EXTENDED_TAILER%
```
kinow: @dbeltrankyl moved the extended tailer here, after the user template (body).

autosubmit/job/template/bash.py
```diff
@@ -46,25 +46,103 @@
...
+trap 'as_signals_handler SIGHUP'  SIGHUP
+trap 'as_signals_handler SIGINT'  SIGINT
+trap 'as_signals_handler SIGQUIT' SIGQUIT
+trap 'as_signals_handler SIGTERM' SIGTERM
+trap 'as_signals_handler SIGXCPU' SIGXCPU
+trap 'as_signals_handler SIGXFSZ' SIGXFSZ
+
+# Set up the exit trap to ensure exit code always runs
+trap as_exit_handler EXIT
+
```
kinow: Explaining the logic above:

- We have a logic where the default exit code is `0`, set in `AS_EXIT_CODE`
- If we have a signal HUP/INT/QUIT/TERM raised by Slurm/OS/user, then we capture that and set the `AS_EXIT_CODE` to 128+signal (linux default behaviour), and exit with that value
- The trap now checks if `AS_EXIT_CODE` is not zero, and in this case, it uses that exit code, otherwise it uses the last command's exit status (Slurm may send a SIGTERM but our last command like `hres` or `htset.pl` succeeded)
- The trap exits with the exit code, which is either the error with signal, or the last executed command's exit code

We may still miss the _STAT file is someone sends a sigkill to the script, or if the computer crashes, or someone hits the power button... but I think we have no choice in these cases.

I added the trap for common signals, that should cover 98, 99% of our cases I think: HUP, INT, TERM, QUIT.

The XCPU and XFSZ may be triggered, but I think the wallclock at least only raised SIGTERM. It should still be safe to capture those if we ever see them in PBS or Slurm.

SIGUSR1, SIGUSR2, SIGPIPE may happen, but the first two are custom signals added by users with --signal, and are executed before SIGTERM. So they are ignored as a simulation could complete right after they are raised (if the user wants to handle that, they can do it so in their scripts). SIGPIPE is tricky as I think that can be raised in a normal bash shell due to some `script | bla` where bla didn't receive data, or script closed the stream, etc. Our old trap should be able to handle that.

dbeltrankyl (approved):
Hello @kinow 

Just one comment about a commented line, but it looks good! 

I've run an experiment with wrappers to see if it was working after the trap changes, and it is

autosubmit/job/template/bash.py
```diff
@@ -46,25 +46,103 @@
...
+        # If the user-provided script failed, we exit here with the same exit code.
     fi
     
-    exit $exit_code
+    trap - EXIT
+    exit $AS_EXIT_CODE
 }
 
+# Set up the signals trap to ensure the job is killed on signals
+# trap as_signals_handler HUP INT QUIT STP TERM
```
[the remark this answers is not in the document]
kinow: The last one was just UP, INT, QUIT, STP, TERM to list the signals. Somehow "trap as_signals_handler" was added there. But I think it's better not to list them as it's already documented elsewhere and we may have to change it. I will remove the second line and end the sentence with a period after "on signals". Thanks Dani!

dbeltrankyl:
Thanks @kinow, all tests are ok, so I think it can be merged :) I'll cherry-pick into 4.1.16-release-branch-trap & merge into 4.1.16-release-branch + upload a new module once it is

dbeltrankyl:
It will just be the same bug as in 4.1.14 in these cases, I guess ( I expect a delay of >1min)

4.1.14 

Completed jobs have always: 
```

------autosubmit log process
submit
----template
start
end
------
```

Failed jobs have always:

```
------autosubmit log process
submit
----template
start
------autosubmit log process
Approx end_time ( after recovering the logs )
```

There were other bugs where the timestamp was defaulted to 1970...


4.1.16 

Completed jobs have always: 

```
------autosubmit log process
submit
----template
start
end
------
```

Failed jobs have always:

```
------autosubmit log process
submit
----template
start
end
------If there is no end autosubmit log process
Approx end_time ( after recovering the logs )
```

PR #2836, opened 2026-02-24
[Refactoring] Removing the ENABLE_WRAPPER_THREADS

### LINKED ISSUE #2691: ENABLE_WRAPPER_THREADS setting may not be properly implemented.
Hello @kinow , @VindeeR 

While working on #2468, I noticed that there is this setting:

```yaml
CONFIG:
 ENABLE_WRAPPER_THREADS: False
```

By default, it is disabled, and I think nobody has enabled it.

I was checking the associated code in JobPackageBase.submit, and I don't believe it works correctly, as it is using a different function than usual, which hasn't been updated for a while. 

I think that part needs a refactor, and see if it is useful to keep, and whether it actually improves performance when building wrappers.

For normal jobs it is never updated because len(self.jobs) == 1

VindeeR:
Closes #2691

When forced to enter the thread creation it would create the thread but since each job run in isolation it would never execute with the exception of when mocked.
So the deleting the best approach would be delete the inactive piece of code 

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/job/job_packages.py
+++ b/autosubmit/job/job_packages.py
@@ -19,7 +19,6 @@
 import datetime
 import json
 import locale
-import multiprocessing
 import os
 import random
 import re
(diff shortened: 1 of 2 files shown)


kinow (approved):
Thank you!

PR #2837, opened 2026-02-24
[Enhancement] Removing unused variable

### LINKED ISSUE #2420: Review ExperimentHistory constructor
I am writing a jupyter notebook for @orioltinto #2418 , using the `ExperimentHistory` class. I was setting the constructor arguments when I realized they were never used. It's quite confusing.

The current code:

Image

`ExperimentHistory` takes three parameters. The first, `expid`, is mandatory, whereas the remaining ones are optional.

The `jobdata_dir_path` has the default value `DEFAULT_JOBDATA_DIR`, which is `os.path.join('/esarchive', 'autosubmit', 'as_metadata', 'data')` (err, probably better to remove uses of esarchive too… but that's another issue).

And `historiclog_dir_path` has the default value set by `DEFAULT_HISTORICAL_LOGS_DIR`, `os.path.join('/esarchive', 'autosubmit', 'as_metadata', 'logs')`.

These default values come from `autosubmit.history.database_managers.database_manager.py`, from constants in that Python module.

But now comes the interesting part.

These two parameters, `jobdata_dir_path` and `historiclog_dir_path` are just never used 😵 💫 😖 

And what's worse, it initializes several attributes using `BasicConfig` directly (harder to test, we keep having to mock things because of this…). And those values from `BasicConfig` have the same values from the constants in `autosubmit.history.database_managers.database_manager.py`.

But `BasicConfig` constants like `JOBDATA_DIR` are not really constants… they are re-defined in `__read_file_config` of the same file the defines `BasicConfig`.

We can fix this issue by:

* Fix the constructor to have parameters to what's modifiable by external callers, trying to write it as simple as possible, and thinking in how to make it easier to be tested;
* Add documentation;
* Add tests (`experiment_history.py` has less than 50% of its code covered by tests)

Image

The rest can be fixed later.

VindeeR:
Closes #2420 

Removing variables unused `Experiment History` variables

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1798,8 +1798,7 @@ class Autosubmit:
         exp_history = None
         try:
             # Historical Database: Can create a new run if there is a difference in the number of jobs or if the current run does not exist.
-            exp_history = ExperimentHistory(expid, jobdata_dir_path=BasicConfig.JOBDATA_DIR,
-                                            historiclog_dir_path=BasicConfig.HISTORICAL_LOG_DIR)
+            exp_history = ExperimentHistory(expid)
             exp_history.initialize_database()
             exp_history.process_status_changes(job_list.get_job_list(), as_conf.get_chunk_size_unit(),
                                                as_conf.get_chunk_size(),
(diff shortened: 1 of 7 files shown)


kinow (approved):
Small PR, I'm reviewing it while I wait for my GH Actions build to finish. Looks good to me. +1 if CICD build passes. Leaving approved.

PR #2839, opened 2026-02-25
[Enhancement] - refactoring connect Paramiko

### LINKED ISSUE #2748: Rework Paramiko platform connect function & reconnect
In https://github.com/BSC-ES/autosubmit/pull/2428#discussion_r2623066655, I removed an `AutosubmitCritical` in the `reconnect` function.

@dbeltrankyl reviewed it and confirmed that the new behavior looked alright as on the first connection it would still raise an `AutosubmitCritical` if it failed, and with the change the `AutosubmitError` would be raised only in the reconnect failures.

This issue is so we can review the code and make it clearer the behavior, and avoid the risk of someone moving code around, or replacing an exception by the other.

The idea (feel free to correct me here @dbeltrankyl ) is this:

- When AS tries to connect for the first time, if it gets an Authentication error, then we abort with an error
- When AS tries to **re**connect, if it gets an Authentication error, then we retry the same way we would in a networking timeout

VindeeR:
Closes #2748

Refactoring connect Paramiko

**Check List**

- [ ] Documentation updated.


autosubmit/platforms/paramiko_platform.py
```diff
@@ -198,13 +198,13 @@ def wrapper(self):
     def reset(self):
         self.close_connection()
         self.connected = False
-        self._ssh = None
+        self._ssh = paramiko.SSHClient()
         self._ssh_config = None
         self._ssh_output = None
-        self._host_config = None
+        self._host_config = {}
```
kinow: Not a dict, exactly, it's an optional `paramiko.SSHConfigDict`. Might be easier to use `None` or check if we can instantiate that object empty.

IDE was raising a type error here. Maybe worth inspecting this with mypy.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -378,66 +428,19 @@ def connect(
...
                                                                   max_packet_size=pow(4, 12))
             self._ftpChannel.get_channel().settimeout(120)
             self.connected = True
             if not log_recovery_process:
                 self.spawn_log_retrieval_process(as_conf)
-        except SSHException:
+        except SSHException as e:
             self.connected = False
-            raise
+            raise SSHException(f"SSH exception occurred: {e}")
```
kinow: I think this is not needed. Probably easier to just raise as that will re-raise the same exception (except here we had a chance to mark connected = false).

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1273,7 +1271,8 @@ def exec_command(
                 stdin = chan.makefile('wb', bufsize)
                 stdout = chan.makefile('rb', bufsize)
                 stderr = chan.makefile_stderr('rb', bufsize)
-                return stdin, stdout, stderr
+                if (stdin, stdout, stderr) != (False, False, False):
+                    return stdin, stdout, stderr
```
kinow: I think it can never be `False`here? The `ChannelFile.makefile` will return a `ChannelFile`. Or maybe crash.

Here we must just return the three values.

kinow: I think we can change the function signature to channel only as you did, and raise the exception like you did.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1611,12 +1607,12 @@ def close_connection(self):
...
-                self._ssh._agent.close()
+            if self._agent:  # May not be in all runs
+                self._agent.close()
         with suppress(Exception):
-            if self._ssh._transport:
-                self._ssh._transport.close()
-                self._ssh._transport.stop_thread()
+            if self._transport:
+                self._transport.close()
+                self._transport.stop_thread()
```
kinow: Interesting. Looks like we never closed the transport & agent created with 2FA, only the SSHClient created with ssh. 👏 Good catch!

kinow (commented):
@VindeeR I think it's improving a bit the code that does the connection with Paramiko. I will review again next week with calm because I think there's more things we can improve in this function (the timeouts should be used, I think? maybe a sleep time between retries, we seem to have a retry for send_command and for connect... maybe that needs different configs, etc).

autosubmit/platforms/paramiko_platform.py
VindeeR: I tried implement as an `Optional[paramiko.SSHConfigDict]` but I started running into other issues, and I wasn't sure if I should have applied it with all the changes, I'll add it again and try to solve the mypy issues once more, they're a bit tricky to fix.

autosubmit/platforms/paramiko_platform.py
VindeeR: There's a test, `test_send_command_timeout_error_exec_command`, that set those values as `False` forcibly, hence I had to do it this way, I believe that simply changing it on the test wasn't simple so I opted to work around it this way, since previously it was already making this validation elsewhere

VindeeR: I altered the test to raise the new error, but I believe it might not replicate the previously expected behavior exactly as intended

kinow:
Kick the integration tests again :grimacing:

kinow:
Can you rebase this, @VindeeR ? Please :grimacing:

autosubmit/platforms/paramiko_platform.py
kinow: This appears to be missing testing here. Maybe we can include one, @VindeeR ? Even if we have to mock, just to make sure there won't be any runtime error here (if we need to mock transport, maybe use the `spec=realclass` to make sure we call the right methods).

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1299,7 +1297,6 @@ def exec_command(
                 #        https://github.com/BSC-ES/autosubmit/issues/2439
                 # chan.settimeout(timeout)
 
-        return False, False, False
```
kinow, line 1302: If an exception happens, what's returned in this function then? Right now the signature for the return is `Union[tuple[paramiko.ChannelFile, paramiko.ChannelFile, paramiko.ChannelFile]]`. I think IDEs/mypy might be warning about this removal here.

kinow:
CI needs a kick or two, I think?

VindeeR:
> CI needs a kick or two, I think?

I actually forgot to fix one of the tests, on top of the intermittent issue with the image.

The report is a bit weird tho, I'll try to rerun the tests to see if it is fixed.

PR #2853, opened 2026-03-04
Changes yaml init mode from safe to rt

### LINKED ISSUE #2851: AS config parser to ignore `!` YAML tags instead of crashing
Hi @kinow @dbeltrankyl @VindeeR,

We are currently working on a new auto-ece4 workflow and we would like to use Autosubmit to directly parse ECE4 YAML files that contain custom ! tags like !rrule !noparse and !noparse_jinja

At the moment, Autosubmit crashes when encountering these tags. For our use case, Autosubmit does not need to interpret or process these tags. We only need it to **avoid crashing** when they are present.

Please, implement a minimal fix `!` tags are safely ignored (treated effectively as comments or plain values), instead of causing a parsing failure.

Thanks in advance!

dbeltrankyl:
fixes #2851

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   now Autosubmit crates conform to RO-Crate 1.1, process 0.5, workflow run
   crate 0.5, and workflow ro crate 1.0. A new integration test has been
   added to check that we produce valid crates with the `rocrate-validator` #2816
+- Changed Yaml load mode from 'safe' to 'rt' #2851
 
 ### 4.1.16: Postgres (experimental) support, bug fixes, and enhancements
 


kinow (commented):
Kicked the integration test that failed, @dbeltrankyl . I think it was not caused by this change.

Do you know if the round-trip parser protects against attacks like serialization/RCE/etc. as the safe parser, @dbeltrankyl ?

dbeltrankyl:
Maybe we can merge without it passing @kinow ? 😅 It is the 8th try 

Locally, it works and 

https://github.com/BSC-ES/autosubmit/actions/runs/22574995939

is based on the last master that worked...

PR #2871, opened 2026-03-09
Use a Slurm Docker image maintained by giovtorres

kinow:
This PR replaces our custom Slurm image made by @manuel-g-castro by the one used by him, I believe, initially as reference: https://hub.docker.com/r/giovtorres/slurm-docker/

Huge thanks to @giovtorres who opened [a discussion asking for feature requests](https://github.com/giovtorres/slurm-docker-cluster/discussions/88#discussioncomment-16044448) and promptly answered my question about having a self-contained Slurm for CICD.

I got one test from `test_slurm` passing locally after some trial and error, so I think this might be ready to be tested on GH Actions.

We had some intermittent failures on Slurm, and I also had some during winter break while trying to use a single container. In most errors, I saw a banner timeout that I believe might be related to throttling on the SSH server. @giovtorres included the SSH settings to disable throttling (thanks again!) so once this works, we can monitor the tests and check if that banner timeout error is gone, or if that persists (in which case, we now will know that it must be something else).

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -74,7 +74,7 @@ jobs:
     env:
       COVERAGE_PROCESS_START: .coveragerc
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
 
     strategy:
       matrix:
@@ -147,13 +147,32 @@ jobs:
           retention-days: 7
 
       - name: Tar logs
+        id: tar_logs
         if: failure()
         run: |
-          mv ../container_logs_ssh_server/ container_logs_integration 
-          mv ../container_logs_git_server/ container_logs_integration_git
+          logs_found=false
+          
+          if [ -d ../container_logs_ssh_server ]; then
+            echo "Found container logs for SSH server, moving them to container_logs_integration"
+            mv ../container_logs_ssh_server/ container_logs_integration
+            logs_found=true
+          else
+            echo "No container logs found for SSH server"
+          fi
+          
+          if [ -d ../container_logs_git_server ]; then
+            echo "Found container logs for Git server, moving them to container_logs_integration_git"
+            mv ../container_logs_git_server/ container_logs_integration_git
+            logs_found=true
+          else
+            echo "No container logs found for Git server"
+          fi
+          
+          # Write the value to the GH Actions step output, so other steps can use it.
+          echo "logs_found=$logs_found" >> $GITHUB_OUTPUT
 
       - name: Upload logs to GitHub
-        if: failure()
+        if: failure() && steps.tar_logs.outputs.logs_found == 'true'
         uses: actions/upload-artifact@master
         with:
           name: container_logs_integration_${{ github.run_id }}
@@ -228,9 +247,6 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install system dependencies
-        run: sudo apt-get install -y curl git graphviz rsync
-
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip packaging
@@ -246,8 +262,8 @@ jobs:
 
       - name: Integration tests
         run: |
-          pytest \
-            -n $(nproc) \
+          pytest -x \
+            -n 3 \
             --cov=autosubmit --cov-config=.coveragerc \
             --cov-report=xml:test/coverage.xml --cov-append \
             test/integration \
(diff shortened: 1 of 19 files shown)


dbeltrankyl:
Hello @kinow ! 

Thanks, all, for this

Does this container support ACC and HET jobs? related to https://github.com/BSC-ES/autosubmit/pull/2873 

cc: @VindeeR

VindeeR (approved):
This test locally worked
`pytest test/integration/ -m "not postgres and slurm"`

But when I tried 
`pytest test/integration -m "not postgres and not slurm"`
it never actually finished even after 10 minutes it gets stuck locally, considering it works here in the CI, and the same happens for master, I believe it might be something related to my machine

Other than this observation everything looks perfect, thanks for the hard work 🙌

kinow:
> i added xz to the slurm-docker container. v0.1.7 is releasing to docker hub now. should be ready in like an hour. using this version includes two changes: i added resource tracking in v0.1.6 and now xz is pre-installed for you. So, you can remove the EXTRA_PACKAGES line, as it is no longer need to install xz. This should fix the wait_for_logs timeout issue.

Thank you @giovtorres !

> i took a peek at the test failures more closely (i couldn't resist :) ).

You are welcome to do it! :laughing:  (and I'm grateful!)

> run_experiment appears to return 0 when the loop completes even if every job failed. I saw "Run successful" which means to me that the scheduler ran to completion, not that all jobs succeeded. I think it ends up with assert 0 > 0. I think the fix could be to just remove/comment the _assert_exit_code(final_status, exit_code) in test_run_ps_scenarios.py.

Yeah, but the question remains why triggering this again (and sometimes again and again :angry: ) passes. I've been trying to understand if it's because the test is badly designed, or if it's our main code, and am a bit concerned of removing assertions like that and hiding  real bug, but I will try to start analyzing what happens from that assertion backwards. Maybe it's clear that that test needs to be removed but I just cannot see for working so long on this. Thanks for the unbiased review, @giovtorres !

> The xdist you mentioned definitely seems like it was causing race conditions where jobs from different tests were interfering with the shared slurm state. I think you could add @pytest.mark.xdist_group("slurm_container") or something to that effect to all tests in test_run_ps_scenarios.py and test_run_slurm_scenarios.py that use slurm_server. I believe this will make xdist run them sequentially on one worker, eliminating the race condition. You'll have to double check me on this one.

We had that before, and I removed the dist group to speed up the build. I'll try a bit more with the single container/docker service here in GitHub Actions, to see if that speeds up the build by a few minutes. If not, then I think we might just have to accept the slurm tests will run above 10 minutes because it needs to run real Slurm cluster + jobs.

> Not sure if this is a bug, but in ci.yaml, there is a condition to check for the container_logs_slurm_server directory, but the mv command inside that if block references a different directory entirely.

Ha! A copy-pasta bug I added yesterday. Thanks for spotting that!

Cheers!

kinow:
'right, now it's failing due to exit code. Going to have a look...

kinow:
Thanks @VindeeR !

kinow:
Yeah, I had tried installing it manually in the previous image (I think I had to use dnf?) and it appeared to be quite fast, but I guess it might be quite random at times. I moved some tests off the slurm fixture, but I saw one that was intentionally written for xz + slurm (which makes sense, as we will have to compress our logs on Slurm).

So thanks a lot for adding that to the image, as otherwise we'd have to push our own image with your image as base + whatever packages we'd need :bow: 

Second run after fixing ruff passed as well! And again in 6 minutes :pray: Now, squash commits.

kinow:
Fixed conflicts and rebased it. Tests will likely fail, but will restart looking at what's causing the flakiness. Note-to-self: @giovtorres had some improvements in this commit: 
https://github.com/giovtorres/autosubmit/pull/2/changes/e7a4057a721f508f147d0a2d1eff0cec897425ab, to be compared with current code and review the differences (thanks again, Giovanni)

kinow:
Last failure here was due to the integration tests without Slurm taking more than 10 minutes. I'm bumping the timeout to 15 minutes, but it'd be good to investigate later if we can reduce it back to 10 minutes and make our integration tests faster.

Other integration test jobs, and the slurm job passed. Let's see the next run.

kinow:
One successful run. Let's see how it goes launching everything again.

kinow:
Failed in the apt-get installation, for taking too long.

image


It appears to be some internet issue. I'll kick the build.

kinow:
Build passed. I didn´t see that failure expecting 6 entries found only 3. So maybe when I saw it it was on an older commit, @dbeltrankyl , @VindeeR .

kinow:
Rebasing again as master changed. But this is ready for review. If you see the build red, please kick it :grimacing: (I'll try to kick it if it fails in the next minutes)

Only if it's a Slurm failure then it might be something related to this change.

kinow:
Build passed (missing only Sphinx docs build).

There are a couple of Git conflicts, so I'm rebasing and fixing changelog.md and test/integration/conftest.py conflicts. Let's see if all tests pass again.

kinow:
Build successful again. Only RTD step now. It looks like the tests are a bit more stable now. I haven´t seen any Slurm failures today :pray:

kinow:
You approved it, but I updated the PR with changes from Giovanni. Can you do a final review, please, @VindeeR ?

VindeeR (approved):
Things seems to be in order with the PR.

I believe you changes might conflict with Dani's changes, if not, and there are no issues with the CI you can marge the changes from this branch.

Thanks for the hard work 🙌

kinow:
Thanks @VindeeR . I fixed the conflicts but there was a change in an unit test that I will have to adjust so it passes here. Will merge once CICD passes. Thanks!

PR #2872, opened 2026-03-09
ECMWF fixes ( there is no issue, created) 

dbeltrankyl:
**Check List**

- [x] Documentation updated.


dbeltrankyl:
This fixes the issue with ecmwf. 

I've added a test, but it can only be run locally, with the ecaccess package installed and (new mark: local) and with a valid eccert.crt in your home directory. At first, the test was failing because I was unable to change where the ecaccess command looks for eccert.

* Fixes an issue when the scratch dir doesn't exist ( there is no mkdir -p equivalent )
* Fixes an issue with the online_recovery / during running gathering completed files

kinow (approved):
Left two comments about changelog and the TODO, but these are not blockers! This one is a tricky one to test. Good idea the local test + new marker. Looks good to me, thanks @dbeltrankyl !

PR #2873, opened 2026-03-10
[Bug] - Slurm header broken when HETSIZE > 0

### LINKED ISSUE #2660: Slurm header broken when HETSIZE > 0 without a `CURRENT_QUEUE`
### Description

I'm writing tests for the Submitter in this PR #2604 . At one part of the tests & linting/mypy changes, I found another part of the code that was called in `wrapper_factory.py`. In principle I was reluctant to touch it to reduce the number of changes, but then I had to alter another line in `wrapper_factory` that was going to that file, so I decided to just update the code in that file, and then write a few tests if there wasn't a lot of coverage missing.

Codecov reported that only a few lines that I modified were not tested. But to cover those lines, I had to create objects that call `slurm_header.py` (which I am not touching in my PR). So I prepared some objects so that the `SlurmHeader` could be created.

However, passing `hetsize > 0` results in a bug in runtime, which I believe will happen in real workflows.

https://github.com/BSC-ES/autosubmit/blob/7672b54aae7bd6f760cdfe93252a435a6f771506/autosubmit/platforms/headers/slurm_header.py#L337-L341

The code above gets the `HETSIZE` from the job config, an integer. If you pass a number like `2`, then it'll be used to create a `component` for the range of zero up to `2`.

Then, the first line below sets the value of `header` by calling a function `self.get_queue_directive`, see below.

https://github.com/BSC-ES/autosubmit/blob/7672b54aae7bd6f760cdfe93252a435a6f771506/autosubmit/platforms/headers/slurm_header.py#L27-L41C13

This is the bug 🐛 

The code never uses the parameter `parameters` when `het` is greater than `-1` and there's a current queue in the job. Otherwise, it will use the `parameters` to retrieve the `CURRENT_QUEUE`. However, note that `parameters` is the second parameter. If you look at the previous code, you'll see the second parameter we are passing is the component/het.

I think the fix is to just adjust the call to pass the het/component in the right location, and then figure out if we have a `parameters` object to pass.

* Autosubmit version: master
* Machine/VM/environment name (if applicable): NA

### Reproducible Example

This is the WIP test I had when the bug happened.

```py
def test_get_wrapper(slurm_platform: SlurmPlatform, wrapper_builder_kwargs: dict, mocker):
    wrapper_factory = SlurmWrapperFactory(slurm_platform)

    wrapper_data = mocker.MagicMock()
    wrapper_data.het = {
        'HETSIZE': 2
    }
    wallclock = '00:30'

    wrapper_builder_kwargs['wrapper_data'] = wrapper_data
    wrapper_builder_kwargs['wallclock'] = wallclock

    wrapper_factory.get_wrapper(BashVerticalWrapperBuilder, **wrapper_builder_kwargs)
````

I'll change the test to avoid going through that line where the bug happens as I need to finish the tests for Submitter.

### Expected Behaviour

No runtime error.

### LINKED ISSUE #2897: THREADS key and new Slurm version causing as many tasks as CPUs to be executed
### Description

With the latest change of Slurm version (now we are in 24.11.6), the behavior of the key `--cpus-per-task`, which is set by the key `THREADS` in Autosubmit, changed. If this key is not set, Autosubmit will default to one.

Where before it was either ignored by srun or behaved differently, this flag instructs Slurm to launch one task per cpu, as defined by Autosubmit's default policy. 

This was also reported by Eric Ferrer and the fix is just setting the flag to the empty string, which works because the Slurm header code checks for the empty string. And only the empty string. Setting to 0 or -1 won't affect the flag being added to the SBATCH header.

https://github.com/BSC-ES/autosubmit/blob/v4.1.15/autosubmit/platforms/headers/slurm_header.py#L191-L192

* Autosubmit version: 4 (all)

### Reproducible Example

Create an experiment and set the following to the job file:

```
JOBS:
  TEST_NO_EXTRA_DIRECTIVE:
    SCRIPT: srun -v -l hostname
    PLATFORM: marenostrum5 
    RUNNING: once
    WALLCLOCK: 00:01
    PROCESSORS: 10
```

If you execute this you will get 

```
1: gs09r1b62
3: gs09r1b62
9: gs09r1b62
5: gs09r1b62
4: gs09r1b62
7: gs09r1b62
0: gs09r1b62
2: gs09r1b62
6: gs09r1b62
8: gs09r1b62
```

And if you check the `.err` you see 

```
(...)
srun: launching StepId=38034010.0 on host gs09r1b62, 10 tasks: [0-9]
(...)
```

### Expected Behaviour

That a single task should be executed or that this behavior is clarified.

### PR DESCRIPTION
**Pull request by VindeeR**
Closes #2660
Closes #2897

Changes to ensure that the 2 different ways a header is generated doesn't conflict

**Check List**

- [ ] Documentation updated.

### CODE REVIEW

**Comment by VindeeR** [comment]
@dbeltrankyl This PR is ready for review

**Comment by dbeltrankyl** [comment]
Thanks @VindeeR 

tomorrow morning i'll revise all my pending pr

File: test/integration/test_scheduler_general.py [hunk 1]
```diff
@@ -169,16 +169,16 @@ def _write_test_files(expid, local_root_dir: Path):
 
 
 @pytest.mark.parametrize("scheduler, job_type", [
-    ('pjm', 'DEFAULT'),
-    ('slurm', 'DEFAULT'),
-    ('ecaccess', 'DEFAULT'),
-    ('ps', 'DEFAULT'),
-    ('pjm', 'NODES'),
+    # ('pjm', 'DEFAULT'),
```
dbeltrankyl: hm, why are these tests commented?

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -39,43 +39,46 @@ def get_queue_directive(self, job: 'Job', parameters, het=-1):
...
+            if isinstance(job.het['CURRENT_QUEUE'], dict) and het in job.het['CURRENT_QUEUE'] and job.het['CURRENT_QUEUE'][het] != '':
                 return f"SBATCH --qos={job.het['CURRENT_QUEUE'][het]}"
-        else:
-            if parameters['CURRENT_QUEUE'] != '':
-                return f"SBATCH --qos={parameters['CURRENT_QUEUE']}"
+            else:
+                return f"SBATCH --qos={job.het['CURRENT_QUEUE']}"
+        elif parameters is not None and 'CURRENT_QUEUE' in parameters and parameters['CURRENT_QUEUE'] != '':
+            return f"SBATCH --qos={parameters['CURRENT_QUEUE']}"
         return ""
```
dbeltrankyl, line 77: I would add a warning if the queue is not found for a job that uses the Slurm platform, as without a queue, the job won't likely be able to be submitted (unless the hpc admin defaults it to something)

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -39,43 +39,46 @@ def get_queue_directive(self, job: 'Job', parameters, het=-1):
...
             if len(job.het['PROCESSORS']) == 0 or job.het['PROCESSORS'][het] == '' or job.het['PROCESSORS'][
                 het] == '1' and int(job_nodes) > 0:
                 return ""
             else:
-                return "SBATCH -n {0}".format(job.het['PROCESSORS'][het])
+                return f"SBATCH -n {job.het['PROCESSORS'][het]}"
         if job.nodes == "":
             job_nodes = 0
         else:
             job_nodes = job.nodes
```
dbeltrankyl, line 101: I believe this `else` is redundant

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -84,16 +87,17 @@ def get_partition_directive(self, job: 'Job', parameters, het: int = -1) -> str:
...
         """
-        if het > -1 and len(job.het['PARTITION']) > 0:
-            if job.het['PARTITION'][het] != '':
-                return "SBATCH --partition={0}".format(job.het['PARTITION'][het])
-        else:
-            if job.partition != '':
-                return "SBATCH --partition={0}".format(job.partition)
+        if (
+                het > -1 and 'PARTITION' in job.het
+                and isinstance(job.het['PARTITION'], dict) and job.het['PARTITION'].index(het)
```
dbeltrankyl: Why are we looking for the **value** "het" (`index`) here? Shouldn't we use it to access a specific position in the list in the rest of the code? Is that intended?

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -122,15 +127,13 @@ def get_exclusive_directive(self, job: 'Job', parameters, het=-1) -> str:
         :return: account directive
         :rtype: str
         """
-        if het > -1 and len(job.het['EXCLUSIVE']) > 0:
-            if str(parameters['EXCLUSIVE']).lower() == 'true':
-                return "SBATCH --exclusive"
-        else:
-            if str(parameters['EXCLUSIVE']).lower() == 'true':
-                return "SBATCH --exclusive"
+        if het > -1 and 'EXCLUSIVE' in job.het and bool(job.het['EXCLUSIVE']):
```
dbeltrankyl: I think this is not correct

Exclusive will likely be "true" or "False" and derivatives
```

"true", "True" → True
"false", "FALSE" → True
"" → False

>> exclusive=""
>>> bool(exclusive)
False
>>> exclusive="false"
>>> bool(exclusive)
True
```

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -140,16 +143,17 @@ def get_nodes_directive(self, job: 'Job', parameters, het=-1) -> str:
         :return: nodes directive
         :rtype: str
         """
-        if het > -1 and len(job.het['NODES']) > 0:
-            if job.het['NODES'][het] != '' and int(job.het['NODES'][het]) > 1:
-                return f"SBATCH --nodes={job.het['NODES'][het]}"
-        else:
-            if parameters['NODES'] != '' and int(parameters['NODES']) > 1:
-                return f"SBATCH --nodes={parameters['NODES']}"
+        if het > -1 and 'NODES' in job.het and isinstance(job.het['NODES'], dict) and str(het) in job.het['NODES']:
```
dbeltrankyl: In the rest of the code, we didn't convert the het value to str. Is this correct? If so, why the difference here? Something feels wrong

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -159,16 +163,14 @@ def get_memory_directive(self, job: 'Job', parameters, het=-1):
         :return: memory directive
         :rtype: str
         """
-        if het > -1 and len(job.het['MEMORY']) > 0:
-            if job.het['MEMORY'][het] != '':
-                return "SBATCH --mem={0}".format(job.het['MEMORY'][het])
-        else:
-            if parameters['MEMORY'] != '':
-                return "SBATCH --mem={0}".format(parameters['MEMORY'])
+        if het > -1 and 'MEMORY' in job.het and isinstance(job.het['MEMORY'], dict) and job.het['MEMORY'].index(het):
```
dbeltrankyl: again index here, in some yes in some no

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1555,11 +1555,13 @@ def get_header(self, job: 'Job', parameters: dict) -> str:
...
             if hasattr(self.header, 'get_processors_directive'):
+                hetsize = job.het['HETSIZE'] if 'HETSIZE' in job.het else -1
                 header = header.replace(
-                    '%NUMPROC_DIRECTIVE%', self.header.get_processors_directive(job, parameters))
+                    '%NUMPROC_DIRECTIVE%', self.header.get_processors_directive(job, hetsize))
             if hasattr(self.header, 'get_partition_directive'):
+                partition = job.het['PARTITION'] if 'PARTITION' in job.het else -1
                 header = header.replace(
-                    '%PARTITION_DIRECTIVE%', self.header.get_partition_directive(job, parameters))
+                    '%PARTITION_DIRECTIVE%', self.header.get_partition_directive(job, partition))
```
dbeltrankyl: I think this call is not correct, it expects the het value

`    def get_partition_directive(self, job: 'Job', het: int = -1) -> str:
`

dbeltrankyl (commented):
Hello!

Thanks for the work @VindeeR 

Do you mind writing an integration test? Just add one in the run Slurm scenarios that uses a hetjob. ( it should be possible, at least with the new container, I asked two weeks ago )  I feel that the unit tests are not enough since I found some inconsistencies  and wrong call arguments

test/integration/test_scheduler_general.py
VindeeR: I was having an issue only with one of those tests so I commented and forgot to uncomment

autosubmit/platforms/headers/slurm_header.py
VindeeR: I rewrote all the validation and simplified them into a function, forcing the same validation and behavior, this way I believe we fix make it standard which avoid the confusion I was having with each of them validation the same things each their own way

autosubmit/platforms/paramiko_platform.py
VindeeR: I was doing some tests regarding the `het` job and the `parameters` and got confused here sorry for that

VindeeR:
@dbeltrankyl I realized that the test could be simplified and reduced to just one function, I was having a hard time following all the differences in the validation, because each was written in a different way, but all were doing virtually the same thing.

I believe some validations, might have had some edge cases not validated too.

I believe it should be simpler now, and I believe this sort the issue I was having with `dict/list` causing me to change some behavior on the test `aka: index | directive in`.

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -18,247 +18,291 @@
...
+
+from autosubmit.log.log import Log
 
 if TYPE_CHECKING:
     from autosubmit.job.job import Job
 
 
+def check_directive(directive: str, job_parameters: Union[dict, list, str], het: int = -1) -> bool:
+    """
+    Returns if directive the directive exists and has value
```
dbeltrankyl: -> """Returns...

autosubmit/platforms/headers/slurm_header.py
```diff
@@ -355,31 +399,32 @@ def calculate_wrapper_het_header(self, wr_job) -> str:
         header = self.hetjob_common_header(hetsize, wr_job)
         for components in range(hetsize):
             header = header.replace(
-                f'%QUEUE_DIRECTIVE_{components}%', self.get_queue_directive(wr_job, components))
+                f'%QUEUE_DIRECTIVE_{components}%', self.get_queue_directive(wr_job, het=components))
             header = header.replace(
-                f'%PARTITION_DIRECTIVE_{components}%', self.get_partition_directive(wr_job, components))
+                f'%PARTITION_DIRECTIVE_{components}%', self.get_partition_directive(job=wr_job, het=components))
```
dbeltrankyl: (sorry nitpicky), why job=wr_job while the rest of the calls is just wr_jobs?

dbeltrankyl (commented):
Much clearer now! Thanks

I just have one last request, which is an integration test for HetJobs, if possible


Under:

```python
 ], ids=["Success", "Success with wrapper", "Failure", "Failure with wrapper", "Success with horizontal wrapper"])
def test_run_uninterrupted(

```
from a`utosubmit->test-integration->commands-run->test_run_slurm_scenarios`

Just copy the Success conf , name it success-with-hetjobs and add some hetjob parameters

```yaml
    # Success
    (dedent("""\

    EXPERIMENT:
        NUMCHUNKS: '3'
    JOBS:
        job:
            SCRIPT: |
                echo "Hello World with id=Success"
                sleep 1
            PLATFORM: TEST_SLURM
            RUNNING: chunk
            wallclock: 00:01
            # add hetjobs parameters
    """), 3, "COMPLETED", "simple"),  # No wrappers, simple type
````

autosubmit/platforms/headers/slurm_header.py
VindeeR: no real reason, probably just added it because I was adding other as I was tackling a issue on that function

VindeeR:
> Much clearer now! Thanks
> 
> I just have one last request, which is an integration test for HetJobs, if possible
> 
> Under:
> 
> ```python
>  ], ids=["Success", "Success with wrapper", "Failure", "Failure with wrapper", "Success with horizontal wrapper"])
> def test_run_uninterrupted(
> ```
> 
> from a`utosubmit->test-integration->commands-run->test_run_slurm_scenarios`
> 
> Just copy the Success conf , name it success-with-hetjobs and add some hetjob parameters
> 
> ```yaml
>     # Success
>     (dedent("""\
> 
>     EXPERIMENT:
>         NUMCHUNKS: '3'
>     JOBS:
>         job:
>             SCRIPT: |
>                 echo "Hello World with id=Success"
>                 sleep 1
>             PLATFORM: TEST_SLURM
>             RUNNING: chunk
>             wallclock: 00:01
>             # add hetjobs parameters
>     """), 3, "COMPLETED", "simple"),  # No wrappers, simple type
> ```

Test added Thanks for the review :raised_hands:

VindeeR:
@dbeltrankyl Rebase already dealt with

dbeltrankyl (approved):
Sorry for the late review, with wrappers and holidays, I lost track of this one. Seems good!

PR #2884, opened 2026-03-13
Overwrite parameter DEFAULT.EXPID from config files

### LINKED ISSUE #2877: Infinite loop in ```autosubmit run``` with missing LOCAL_SETUP.md under a different experiment path
### Description

Running a minimal experiment created from ```auto-advanced_config_example``` enters an infinite warning loop.
```autosubmit run a00q``` warns about a missing ```.cmd``` file under a different experiment path: ```/home/ntorquet/autosubmit/a04c/tmp/a04c_LOCAL_SETUP.cmd``` and never exits the loop until timeout.
Looks like a missmatch between the experiment ID executed (a00q) and the configuration path a04c.

### Reproducible Example
1. Create the experiment:
```autosubmit expid -min -repo https://earth.bsc.es/gitlab/ces/auto-advanced_config_example -b main -conf as_conf -d "test"```
2. Create job list
```autosubmit create a00q```
3. Run experiment
```autosubmit run a00q```

Currently expid and create commands are executed correctly. When executing run:
- enter a repeated loop
- keeps displaying
```
8 of 8 jobs remaining (17:25)
No jobs to check for platform local
Section:LOCAL_SETUP can submit 20 jobs at this time
[WARNING] An unexpected error happened while submitting the package: [Errno 2] No such file or directory: '/home/ntorquet/autosubmit/a04c/tmp/a04c_LOCAL_SETUP.cmd'
```
Warning repeats until timeout.

### Expected Behaviour
Run command should only use the files under a00q path and the experiment should execute all the jobs, not stay in an infinite loop of warnings.

### PR DESCRIPTION
**Pull request by ntorqulu**
Closes [#2877](https://github.com/BSC-ES/autosubmit/issues/2877)

**Things Done**
- Prevented project configuration yaml files from overwritting parameter DEFAULT.EXPID
- Implemented method ```_pin_immutable_variables``` to prevent predefined parameters from being overwritten. Called this method after loading the PRE configuration and after loading the POST configuration.
- Added unit tests ```test_immutable_variables_overwrites_default_values``` and ```test_immutable_variables_adds_missing_sections``` in ```test_configcommon.py``` to check that DEFAULT.EXPID  is not overwritten while other custom project configuration in DEFAULT section is kept. 
- Added integration test ```test_override_immutable_variables_in_custom_config``` in ```test_reload.py``` to check integration with reload() method.

**Check List**

- [x] Documentation updated.

### CODE REVIEW

**Review by VindeeR — CHANGES_REQUESTED** [review]
I felt it was just missing a test with the `load_custom_config` function since there were a few changes in there.
Maybe we can have a simple unity test running there.

I did a few other comments to be looked into too, and I believe you have a conflict, but everything looks really well done 🙌

Good job with the changes, thanks for the hard work.

File: test/regression/config/test_reload.py [hunk 1]
```diff
@@ -297,3 +297,51 @@ def test_destine_workflows(temp_folder: Path, mocker, prepare_basic_config: Any)
...
+    as_conf.conf_folder_yaml = Path(temp_folder)
+    as_conf.load_workflow_commit = MagicMock()
+    as_conf.reload(True)
+
+    # Custom file merged
+    assert as_conf.experiment_data["CUSTOM_MARKER"] == "loaded"
+
+    # Inmutable variables not overridden
+    assert as_conf.experiment_data["DEFAULT"]["EXPID"] == "a000"
+    assert as_conf.experiment_data["DEFAULT"]["HPCARCH"] == "LOCAL"
```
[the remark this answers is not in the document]
VindeeR: It is a convention adopted since C/UNIX systems that we keep using to this day
I found this [stack overflow](https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log) thread that explains it in more detail in case you're interested

VindeeR (changes requested):

kinow:
Hmm, interesting. I have a, b, c, d files in my experiment conf dir. With b including t and g (in this order). The order the files are loaded from disk are:

```
Loading YAML: /home/bdepaula/autosubmit/a004/conf/a.yml
Loading YAML: /home/bdepaula/autosubmit/a004/conf/as_misc.yml
Loading YAML: /home/bdepaula/autosubmit/a004/conf/b.yml
Loading YAML: /home/bdepaula/autosubmit/a004/conf/c.yml
Loading YAML: /home/bdepaula/autosubmit/a004/conf/d.yml
Loading YAML: /home/bdepaula/autosubmit/a004/proj/local_dir/t.yml
Loading YAML: /home/bdepaula/autosubmit/a004/proj/local_dir/g.yml
Loading YAML: /home/bdepaula/autosubmit/a004/conf/as_misc.yml
```

So:

- The files are found and loaded in alphabetical order
- PRE and POST are indeed the last to be loaded, even if found in `b.yml`, it will be loaded only after the other files are loaded.
- Users can control the order with the PRE/POST order, so in that case it is not alphabetical

ntorqulu:
I've moved it again to 'draft' because after the rebase I've the following tests failing:
- `test_load_config_file`
- `test_inspect_local`
- `test_placeholders`

ntorqulu:
Decided only to pin as immutable variable `DEFAULT.EXPID`. This variable will not be modified when the configuration files are loaded, even if some configuration files have this variable defined. It will always take the value of the experiment id given when created.

On the other hand, initially I also pinned `HPCARCH` variable, but I've seen that the following tests fail when I pin it:
- `test/integration/commands/inspect/test_inspect_local.py`
- `test/integration/config/test_placeholders.py::test_job_platform_is_a_placeholder`

So, for now I've decided only to pin EXPID variable. I'm not sure if users should be able to modify HPCARCH variable from the config files.

VindeeR:
> Decided only to pin as immutable variable `DEFAULT.EXPID`. This variable will not be modified when the configuration files are loaded, even if some configuration files have this variable defined. It will always take the value of the experiment id given when created.
> 
> On the other hand, initially I also pinned `HPCARCH` variable, but I've seen that the following tests fail when I pin it:
> 
>     * `test/integration/commands/inspect/test_inspect_local.py`
> 
>     * `test/integration/config/test_placeholders.py::test_job_platform_is_a_placeholder`
> 
> 
> So, for now I've decided only to pin EXPID variable. I'm not sure if users should be able to modify HPCARCH variable from the config files.

I believe that whenever the `HPCARCH` differs between the repository and the experiment, the value defined within the experiment should take priority, since the repository configuration may contain a different `EXPID` and `HPCARCH`, which could raise errors during the run.

I believe this is what you were trying to achieve, but you encountered errors in the tests.
So it might be needed to check if the tests have the intended proper behavioral check.

@kinow @dbeltrankyl could you confirm this to me please?

test/regression/config/test_reload.py
```diff
@@ -297,3 +297,50 @@ def test_destine_workflows(temp_folder: Path, mocker, prepare_basic_config: Any)
...
         stats = pstats.Stats(profiler).sort_stats('cumtime')
         stats.print_stats()
+
+
+@pytest.mark.parametrize("custom_section", ["PRE", "POST"])
+def test_override_immutable_variables_in_custom_config(
+    temp_folder: Path, mocker, custom_section: str
+) -> None:
+    """
+    Test that immutable variables (DEFAULT.EXPID, DEFAULT.HPCARCH) cannot be overridden
```
VindeeR:
```suggestion
    """Test that immutable variables (DEFAULT.EXPID, DEFAULT.HPCARCH) cannot be overridden
```

autosubmit/config/configcommon.py
```diff
@@ -1625,6 +1626,30 @@ def _delete_autosubmit_calculated_variables(yaml_data: dict):
...
 
+    def _pin_immutable_variables(self, parameters: dict) -> dict:
+        """Keep default variables regardless of the experiment configuration files
+        :param parameters: dict with current parameters
+        :type parameters: dict
+        :return: dict with updated parameters
+        :rtype: dict
+        """
+        # Variables that should be fixed regardless of the configuration file
+        PINNED_VARIABLES = ["EXPID"]
```
VindeeR:
```suggestion
        pinned_variables = ["EXPID"]
```

autosubmit/config/configcommon.py
```diff
@@ -1625,6 +1626,30 @@ def _delete_autosubmit_calculated_variables(yaml_data: dict):
...
+
+        starter_default = self.starter_conf.get("DEFAULT", {})
+
+        # Add default section only if it exists in starter_conf
+        if starter_default:
+            default_section = parameters.setdefault("DEFAULT", {})
+
+            for key in PINNED_VARIABLES:
```
VindeeR:
```suggestion
            for key in pinned_variables:
```

VindeeR (changes requested):
As mentioned in the meeting, we need to prioritize the info on the main experiment instead of the one found on the project sub module.

Regarding the mentioned tests, it needs to be validated the impact of the change, we might have to check why it is conflicting with your change, since the tests don't use the sub module.

autosubmit/config/configcommon.py
ntorqulu: I was thinking about `PINNED_VARIABLES` as a constant, that cannot be modified, thus using the upper-case convention. The idea is that it cannot be changed once is set.
But I don't know if it's better to get rid of this convention if it's not used elsewhere, it's just to indicate that cannot be overwritten nor modified once set.

ntorqulu:
Errors obtained in tests when pinning `DEFAULT.HPCARCH` from experiment creation. Only integration tests fail.
- failed integration tests:
```
FAILED test/integration/commands/inspect/test_inspect_local.py::test_inspect[HPC*_TEST_SCRIPT] - AssertionError: assert 'LOCAL' == 'TEST_PS'
FAILED test/integration/commands/inspect/test_inspect_local.py::test_inspect[HPC*_TEST_FILE] - AssertionError: assert 'LOCAL' == 'TEST_PS'
FAILED test/integration/config/test_placeholders.py::test_job_platform_is_a_placeholder[True] - assert False
FAILED test/integration/config/test_placeholders.py::test_job_platform_is_a_placeholder[False] - assert False
```

I'll go one by one:

1. **test/integration/config/test_placeholders.py::test_job_platform_is_a_placeholder**
```
        'JOBS': {
            'DN': {
                'SCRIPT': 'echo "%DEFAULT.HPCARCH% %DEFAULT.DESCRIPTION%"',
                'PLATFORM': '%HPCOFFLINE_DN_PLATFORM%',
                'EXPERIMENT_DESCRIPTION': '%default.description%'
            }
        }
    }, wrapper=create_wrappers)

    assert exp.as_conf.jobs_data['DN']['SCRIPT'].startswith('echo "MARENOSTRUM5')
```
`JOBS.DN.SCRIPT` is set as `DEFAULT.HPCARCH`. When a new autosubmit experiment is initialized in the test, it sets `hpc = "local"`:

```
expid = autosubmit.expid(
            description="Pytest experiment (delete me)",
            hpc="local",
            copy_id="",
            dummy=True,
            minimal_configuration=False,
            git_repo="",
            git_branch="",
            git_as_conf="",
            operational=operational,
            testcase=testcase,
            evaluation=evaluation,
            use_local_minimal=False
        )
```
And then in the test `test_job_platform_is_a_placeholder` it sets in the experiment_data `DEFAULT.HPCARCH = MARENOSTRUM5`. So the experiment is first created with local HPCARCH and then it tries to overwrite it with MARENOSTRUM5 HPCARCH.

2. **test/integration/commands/inspect/test_inspect_local.py::test_inspect**

```
for hpcarch in ['TEST_PS', 'TEST_SLURM', 'LOCAL']:
        general_data['DEFAULT']['HPCARCH'] = hpcarch
        general_data['PLATFORMS'][hpcarch]['CUSTOM_DIR'] = 'test'
        general_data['PLATFORMS'][hpcarch]['CUSTOM_DIR_POINTS_TO_OTHER_DIR'] = '%TEST_REFERENCE%'
        as_exp = autosubmit_exp(experiment_data=general_data | yaml.load(additional_data), include_jobs=False, create=True)
```
After initialization of the experiment with autosubmit_exp fixture (sets hpcarch to local) it tries to set the DEFAULT.HPCARCH to other values, and then do an assert with those values. Tests fail because DEFAULT.HPCARCH still keep value local.

autosubmit/config/configcommon.py
VindeeR: In Python, uppercase variable names are conventionally reserved for module-level constants, or global values that are not expected to change (as you suggested).

Since these variables are being initialized inside a function, they are local variables, not constants or global. Using uppercase for local variables can be misleading, because it suggests the values are immutable or globally shared when they are not.

In this case, it would be more consistent with Python conventions to use lowercase (or snake_case) variable names instead.

And since those variable are not used elsewhere it makes no sense to make them global

ntorqulu: Okay, thanks!

ntorqulu:
From AS meeting 20260604, only pin `DEFAULT.EXPID`.

ntorqulu:
I'm missing @VindeeR approval, so this pr will be in stand by until he returns. I think there's no way of unassigning him once he have made a review.

test/unit/config/test_configcommon.py
```diff
@@ -622,3 +623,151 @@ def test_is_section_in_any_wrapper(
...
+) -> None:
+    """Test that the _pin_immutable_variables method adds missing sections and keys."""
+    as_conf: AutosubmitConfig = autosubmit_config(expid="a000", experiment_data={})
+    as_conf.starter_conf = {"DEFAULT": {"EXPID": "a000", "HPCARCH": "LOCAL"}}
+
+    parameters = {}
+    pinned = as_conf._pin_immutable_variables(parameters)
+
+    # Check that missing DEFAULT section is added with original values
+    assert pinned["DEFAULT"]["EXPID"] == "a000"
```
VindeeR, line 656:
```suggestion
    assert pinned["DEFAULT"]["EXPID"] == "a000"
    assert not pinned["DEFAULT"].get("HPCARCH")
```

test/regression/config/test_reload.py
```diff
@@ -297,3 +297,50 @@ def test_destine_workflows(temp_folder: Path, mocker, prepare_basic_config: Any)
...
+    )
+    prepare_yaml_files(default_yaml_file, temp_folder)
+
+    as_conf = AutosubmitConfig("test")
+    as_conf.conf_folder_yaml = Path(temp_folder)
+    as_conf.load_workflow_commit = MagicMock()
+    as_conf.reload(True)
+
+    # Custom file merged
+    assert as_conf.experiment_data["CUSTOM_MARKER"] == "loaded"
```
VindeeR, line 342:
```suggestion
    assert as_conf.experiment_data["CUSTOM_MARKER"] == "loaded"
    if custom_section == 'POST':
        assert as_conf.experiment_data["DEFAULT"]["HPCARCH"] == "DIFFERENT_HPC"
    else:
        assert as_conf.experiment_data["DEFAULT"]["HPCARCH"] == "LOCAL"
```

VindeeR (approved):
This was a brilliant work, very good, and thank for such a well structured code and test.

I just did requested one small addition to your tests,
One the test wasn't validating the different behaviors in each `PRE, POST`, and for the other I realized no test was checking if the key was being eliminated from the array.

But it is a very small thing, thanks for the great improvement 🙌

VindeeR:
Once those additions are dealt with you can squash and merge this PR thanks for the hard work.

ntorqulu:
Great, thanks for the review! Added the assertions to check that HPCARCH is not pinned.

PR #2890, opened 2026-03-18
Fix vertical queue

dbeltrankyl:
@pablogoitia, this should fix the issue with the wrapper timers reported in #2708 

**Check List**

- [x] Documentation updated.


manuel-g-castro:
Sorry to jump in!

> Can we emulate some seconds of queuing time in the new container?

I think the most straight forward is to launch a job that occupies all of the resources. In gio's image that would be a `sbatch -N 3 --wrap sleep 20`, for a job that sleeps for 20 seconds and takes all of the three nodes available there. That would give at least 20 seconds of wait time.

pablogoitia (approved):
Hi, @dbeltrankyl.

The new logic looks fine, and a test with wrappers of 6 splits running on MN5 worked for me (see attached image, where only the first job of the wrapper accounts for the submit time).

Image

I am approving the changes, thanks!

dbeltrankyl:
This was already approved, rebased + merging

Issue #2226, opened 2025-03-14
Experiment not starting (AS4)


marcelcases:
Hi,

I am unable to start an OpenIFS 48 autosubmit experiment ([a8r7](https://earth.bsc.es/autosubmitapp/experiment/a8r7/graph)). The first job (SYNCHRONIZE) does not change the status and no .CMD files are created. I've tried with 4.1.9 and 4.1.11. Could you take a look at this?

Thank you,

@dbeltrankyl @eferre1

dbeltrankyl:
Hello @marcelcases 

I see it stopped ( intentionally )

/esarchive/autosubmit/a8r7/tmp/ASLOGS/20250314_120105_run.log

```bash

2025-03-14 12:10:17,391 Autosubmit will interrupt at the next safe occasion
2025-03-14 12:10:19,632 Looking for active threads before closing Autosubmit. Ending the program before these threads finish may result in unexpected behavior. This procedure will last until all threads have finished or the program has waited for more than 30 seconds.
```

marcelcases:
Hi @dbeltrankyl,

Yes, I stopped it manually after some minutes repeating attempts:

```
5 of 5 jobs remaining (12:09)
2025-03-14 12:09:48,612 Sleep: 10
2025-03-14 12:09:48,612 Number of retrials: 0
2025-03-14 12:09:48,612 Checking jobs for platform=MN5_GLOGIN4
2025-03-14 12:09:48,612 Checking jobs for platform=BSC-MARENOSTRUM5
2025-03-14 12:09:48,612 No jobs to check for platform MN5_GLOGIN4
...
2025-03-14 12:09:58,849 Loading parameters...
2025-03-14 12:09:58,850 Parameters load.
2025-03-14 12:09:58,850 
```

```
5 of 5 jobs remaining (12:09)
No jobs to check for platform MN5_GLOGIN4
No jobs to check for platform BSC-MARENOSTRUM5
[32mFile /esarchive/autosubmit/a8r7/proj/auto-ecearth4/templates/synchronize.tmpl.sh exists[0m[39m
Calculating possible ready jobs for MN5_GLOGIN4
[32mSection:SYNCHRONIZE can submit 20 jobs at this time[0m[39m
Calculating possible ready jobs for BSC-MARENOSTRUM5
```

Might be a wrong configuration or something simple but we did not find what makes it behave like this.

dbeltrankyl:
I'll have a proper look on Monday. Which version are you using? 

[...]

PR #2908, opened 2026-03-30
[Enhencement] - Set AS variables as upper or lower case

### LINKED ISSUE #2504: Possibility to set AS variables as lowercase/uppercase
In GitLab by @ainagaya on Jul 19, 2024, 12:34

Hi!! 

I think that it would be useful to be able to change the variables to uppercase/lowercase. For example,

```
SECTION:
  VARIABLE: "Watermelon"
```

But then, when using the variable, if in the template/other config file when it is used:

```
BASH_VARIABLE1=%SECTION.VARIABLE^^%
BASH_VARIABLE2=%SECTION.VARIABLE,,%

```
That would translate in `BASH_VARIABLE1=WATERMELON`, `BASH_VARIABLE2=watermelon`. In the templates, we can do this in bash, its true. But my usecase is more in the configuration files, for example, the bootstrap. Here we are defining some paths:

```
DEFAULT:
  CUSTOM_CONFIG:
    PRE:
      # Basic configuration: platforms, jobs and defaults
      - "%PROJDIR%/conf/platforms.yml"
      - "%PROJDIR%/conf/jobs_%RUN.WORKFLOW%.yml" 
      - "%PROJDIR%/conf/defaults_%RUN.WORKFLOW%.yml"
```

.To avoid user errors, it would be amazing to have

```
DEFAULT:
  CUSTOM_CONFIG:
    PRE:
      # Basic configuration: platforms, jobs and defaults
      - "%PROJDIR%/conf/platforms.yml"
      - "%PROJDIR%/conf/jobs_%RUN.WORKFLOW,,%.yml" 
      - "%PROJDIR%/conf/defaults_%RUN.WORKFLOW,,%.yml"
```

In that way, this would work with RUN.WORKFLOW: MODEL, Model, MoDeL, or model. Right now, if the word is not capitalized correctly, the file is not found. It also happens a lot with "true", "True" values, and others.

VindeeR:
Closes #2504
Related #2226

Changing behavior of the AS variables to be able to set them as upper or lower case based on `^^` or `,,` at end of a variable

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/config/configcommon.py
+++ b/autosubmit/config/configcommon.py
@@ -921,7 +921,7 @@ class AutosubmitConfig(object):
         :returns: A tuple containing the dynamic variables, the regex pattern, and the start index.
         :rtype: tuple
         """
-        return copy.deepcopy(self.dynamic_variables), '%[a-zA-Z0-9_.-]*%', 1
+        return copy.deepcopy(self.dynamic_variables), "%[a-zA-Z0-9_.-]*(\^\^|,,)?%", 1
 
     def _process_dynamic_variables(
             self,
@@ -1028,12 +1028,10 @@ class AutosubmitConfig(object):
         :rtype: tuple[dict[str, Any], dict[str, Any]]
         """
         for i, key in enumerate(filter(None, keys)):
-            matches = list()
-            if in_the_end and ("^" in key or "," in key):
-                pattern_special_variables = "%[\^]?[a-zA-Z0-9_.-]*(\^\^|,,)?%"
+            matches = list(re.finditer(pattern, key, flags=re.IGNORECASE))[::-1]
+            if in_the_end and "^" in key:
+                pattern_special_variables = "%\^[a-zA-Z0-9_.-]*(\^\^|,,)?%"
                 matches.extend(list(re.finditer(pattern_special_variables, key, flags=re.IGNORECASE))[::-1])
-            else:
-                matches = list(re.finditer(pattern, key, flags=re.IGNORECASE))[::-1]
             for match in matches:
                 value = self._get_substituted_value(key, match, parameters, start_long, dict_keys_type)
                 if value:
@@ -1054,7 +1052,7 @@ class AutosubmitConfig(object):
             parameters: dict[str, Any],
             start_long: int,
             dict_keys_type: str
-    ) -> str:
+    ) -> Optional[str]:
         """
         Get the substituted value for a dynamic variable in the key.
 
@@ -1074,21 +1072,30 @@ class AutosubmitConfig(object):
         rest_of_key_start = key[:match.start()]
         rest_of_key_end = key[match.end():]
         key_parts = key[match.start():match.end()]
+        param = parameters
+        upper_validate, lower_validate = False, False
+
+        if "^^" == key[match.start():match.end()][-3:-1]:
+            key_parts = key_parts.replace("^^", "")
+            upper_validate = True
+        if ",," == key[match.start():match.end()][-3:-1]:
+            key_parts = key_parts.replace(",,", "")
+            lower_validate = True
+
         key_parts = key_parts[start_long:-1].split(".") if "." in key_parts and dict_keys_type != "long" else [
             key_parts[start_long:-1]]
-        param = parameters
+
         for k in key_parts:
-            upper_validate = True if "^^" in k else False
-            lower_validate = True if ",," in k else False
             k = k.strip("^")
             k = k.strip(",")
             param = param.get(k.upper(), {})
             if isinstance(param, int):
                 param = str(param)
-            if upper_validate:
-                param = param.upper()
-            if lower_validate:
-                param = param.lower()
+            if isinstance(param, str):
+                if upper_validate:
+                    param = param.upper()
+                if lower_validate:
+                    param = param.lower()
         return str(rest_of_key_start) + str(param) + str(rest_of_key_end) if param else None
 
     def _update_parameters(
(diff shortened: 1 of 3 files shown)


LuiggiTenorioK (changes requested):
Here is a first review (I'll check it deeper later).

In general, the testing is good, but I would add additional test cases that don't use the new `,,` and `^^` operators to be sure everything that was working before also works after the patch.

About the `_substitute_keys` method, I left some comments in the pattern construction. I also have some doubts about how the matches list is populated, but I will leave that to my next deeper review.

LuiggiTenorioK (approved):
Now it seems pretty good. Great job!

However, I would wait for a final check from @dbeltrankyl before merging it to master.

autosubmit/config/configcommon.py
```diff
@@ -1027,10 +1027,12 @@ def _substitute_keys(
         :rtype: tuple[dict[str, Any], dict[str, Any]]
         """
         for i, key in enumerate(filter(None, keys)):
-            matches = list(re.finditer(pattern, key, flags=re.IGNORECASE))[::-1]
-            if in_the_end and "^" in key:
-                pattern_special_variables = '%\^[a-zA-Z0-9_.-]*%'
+            matches = list()
+            if in_the_end and ("^" in key or "," in key):
```
dbeltrankyl: Hm, I think this behavior is not correct

A variable that has a placeholder,, or placeholder^^, should be the same regarding the Autosubmit substitution rules. So if the same placeholder is found without the trailing ^^ or ,, it should still be the same placeholder for Autosubmit


ONLY when you have a ^ in the beggining so (^placeholder) (^placeholder^^) or (^placeholder,,) will be substituted at the end

dbeltrankyl: So, for the first point, can you check if the placeholders are substituted correctly when they're referenced as placeholder , placeholder,, or placeholder^^ across multiple files?

What I would do... is that if ,, or ^^ is found for that particular placeholder, store it and normalize the variable after all the files are loaded. 

If ^^ and ,, are found for the same variable, add a log.warning and use the last one loaded 

And regarding the second point it should be something similar to:

            if in_the_end and key[0] == "^" :

autosubmit/config/configcommon.py
```diff
@@ -1027,10 +1027,12 @@ def _substitute_keys(
...
+                pattern_special_variables = "%[\^]?[a-zA-Z0-9_.-]*(\^\^|,,)?%"
```
dbeltrankyl: special_variable should be still only the ^ at the beginning I believe ( see my previous comment)

dbeltrankyl (commented):
Hello @VindeeR 

Thanks for the work

Can you revise the loading behavior? 

I miss one integration test that loads multiple files and sees if the behavior when loading is the correct one

autosubmit/config/configcommon.py
LuiggiTenorioK: > ONLY when you have a ^ in the beggining so (^placeholder) (^placeholder^^) or (^placeholder,,) will be substituted at the end

This is what I was trying to search for. I wasn't sure if the `in_the_end=True` was exclusive to special prefixes/suffixes or just to the `^` prefix.

The other doubt I had was how the matches list should be built. Before Erick changes, it was incremental, using the `pattern` first and then extending it with the special prefix pattern. Now it uses one pattern or another, but not both. I didn't raise it before because the `pattern` is fixed now in the `initialize_variables` method (I remember there were 2 patterns before), and the new special prefix/suffix pattern already matches the fixed `pattern`.

VindeeR:
@dbeltrankyl the upper&lower case was detached from just the special character, and apply to all scenarios, adding the changelog now

autosubmit/config/configcommon.py
```diff
@@ -1071,14 +1071,30 @@ def _get_substituted_value(
...
         key_parts = key[match.start():match.end()]
+        param = parameters
+        upper_validate, lower_validate = False, False
+
+        if "^^" == key[match.start():match.end()][-3:-1]:
+            key_parts = key_parts.replace("^^", "")
+            upper_validate = True
+        if ",," == key[match.start():match.end()][-3:-1]:
+            key_parts = key_parts.replace(",,", "")
+            lower_validate = True
```
dbeltrankyl: I would change it to:

```python
if part.endswith("^^"):
    key_parts = key_parts.replace("^^", "")
    upper_validate = True

(also ^^ should exclude ,,) 
elif part.endswith(",,"):
    key_parts = key_parts.replace(",,", "")
    lower_validate = True
```

Also, is this checked: What if the user puts `%bla^^,,%` ?

This is to avoid an error when the placeholder length is less than 3 characters ( and easier to read)

autosubmit/config/configcommon.py
```diff
@@ -1071,14 +1071,30 @@ def _get_substituted_value(
...
+
         key_parts = key_parts[start_long:-1].split(".") if "." in key_parts and dict_keys_type != "long" else [
             key_parts[start_long:-1]]
-        param = parameters
+
         for k in key_parts:
             k = k.strip("^")
+            k = k.strip(",")
```
dbeltrankyl: No need for this if you include the other comment

dbeltrankyl: Also, k = k.strip("^") would strip the starter ^ as well, no? if any

dbeltrankyl (commented):
Some minor comment about the code

I'll check later manually whether everything is working as expected, but it looks good so far

autosubmit/config/configcommon.py
VindeeR: yes, it will be removed, but this is by design since it tries to get the value from the `param` variable, it wasn't part of my change

dbeltrankyl:
Hello @VindeeR 

> I'll check later manually whether everything is working as expected, but it looks good so far

I did not find anything strange, safe to merge i believe. Thanks for the work!

dbeltrankyl (approved):

PR #2925, opened 2026-04-07
Set `noplot` argument to `True` as default for `recovery`, `create` and `setstatus`

### LINKED ISSUE #1347: Provide a separate command for the PDF, or allow users to disable the PDF generation
Feedback from DestinE users; but I also have the same request.

For experienced users, it's probably automatic to use `-np` or `--noplot` with `autosubmit create`. But it brings the problem that users may forget that, and then have to angrily hit CTRL + C, and then repeat the command with `-np`. Or they may forget it, leave and come back to find the PDF displayed unnecessarily (assuming they do not want to look at the PDF).

We have a configuration setting to choose the OUTPUT of experiments, and txt is a valid value, if I recall correctly. I think that used to work, but the last time I looked at the code, there were parts of the code that respected it, and others that did not. I believe the plot ignores that setting.

With cylc you only get a plot if you ask for a plot (`cylc graph`), as you can use the GUI for that, or the TUI.

I think we could either make the plot optional, or at least make the OUTPUT configuration work for every command that outputs non-text data. So users that do not want a PDF, just need to set that option (**but**, that means users would have to modify a setting to get a plot, which is not good user experience...)

ntorqulu:
Closes #1347

**Things done**
- Set by default `noplot` to `True`.
- Added new cli argument `-plt/--plot` in commands `setstatus`, `create` and `recovery`. It references the variable `noplot` and sets it to `False` when selected.
- Added mututally exclusive groups for arguments `-np/-plt` in order to prevent the user selecting both arguments.
-  Modified documentation to explain how the noplot variable now behaves + modified arguments for automatically generated images, to include -plt in order to render the plots

**Check List**

- [x] Documentation updated.


ntorqulu:
About this pr, at the beginning I was thinking of two possible solutions:

- the one implemented, which is keep the `--noplot` argument as legacy setting it to `True` by default and just add another mutually exclusive argument (`-plt`) in order to change the variable value to `False` when users want to plot.
This creates legacy code, but does not modify too much the codebase and is compatible with past autosubmit versions.

- the second solution (not implemented) is get rid of variable `noplot` and create a new argument just for plotting. This would imply that is no longer compatible with past commands if they included the -np flag, but we ensure that we do not keep legacy code.

dbeltrankyl:
> This creates legacy code, but does not modify too much the codebase and is compatible with past autosubmit versions.

I wouldn't introduce changes that break compatibility in a minor version (4.1.17, 4.1.X...). Also, introducing something that breaks the compatibility must be announced in the release communication, so the users can adapt their scripts if they have any


 If this were set for 4.2, it would be fine

ntorqulu:
Okay, thanks @dbeltrankyl! I leave it like this then, implemented as the first option.

VindeeR:
> I wouldn't introduce changes that break compatibility in a minor version (4.1.17, 4.1.X...). Also, introducing something that breaks the compatibility must be announced in the release communication, so the users can adapt their scripts if they have any
> 
> If this were set for 4.2, it would be fine

I believe her current change kept compatibility keeping the legacy code.

So her change should be fine, haven't reviewed it yet starting it now.

VindeeR (changes requested):
I think we could do one more test similar to the one in:

`test/integration/test_run_command.py::test_run_command`

Where you can submit the command, "manually", and check the result.

This way we can even check the scenarios where both `-np` and `-plt` are used, validating edge cases, for example someone deleted the `add_mutually_exclusive_group` by mistake.

Other than that everything looks really good to me. 

Thanks for the hard work :raised_hands:

ntorqulu:
Sure, I didn't know about this tests. I'll take a look, thanks!

ntorqulu:
One question, why the file `test_run_command.py` is outside the folder `test/integration/commands`?

For me, it would make more sense to have in `commands/` one folder for each command and then file `test_run_command.py` to test them all

VindeeR:
> One question, why the file `test_run_command.py` is outside the folder `test/integration/commands`?
> 
> For me, it would make more sense to have in `commands/` one folder for each command and then file `test_run_command.py` to test them all

This folder was created after the creation of the file, and the person who created it the file, me, didn't think about creating the folder, and the person who create the folder probably didn't realize this file existed.

I totally agree with you there, could you move this file to inside the `commands` folder please :pray:

ntorqulu:
jajaj it seems like a riddle; sure, no problem!

ntorqulu:
I've moved file `test_run_command.py` into the test/integration/commands folder and added `test_run_command_plot_behaviour` to test diff behaviours when executing commands with `plot/noplot` arguments.

VindeeR (approved):
Thanks for the new test, it looks good to go to me 🙌

PR #2927, opened 2026-04-08
Enhancement: Detect inefficiencies in SIM jobs with CPMIPS thresholds - Part 3

MissterP:
Hi AS team!

This is the third part of a large [PR](https://github.com/MissterP/autosubmit/pull/1) with the objective to implement the detection of inefficiencies once the SIM job is completed (or section indicated) using CPMIPS and thresholds. If a threshold is violated, it will send a notification to the user indicated in the config. 

Last PR: #2918

This third part consists in the computation of the CPMIPS and selection of violated metrics. 

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/metrics/__init__.py
+++ b/autosubmit/metrics/__init__.py
@@ -1,0 +1,18 @@
+# Copyright 2015-2026 Earth Sciences Department, BSC-CNS
+#
+# This file is part of Autosubmit.
+#
+# Autosubmit is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Autosubmit is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Metrics for Autosubmit Jobs."""
--- a/autosubmit/metrics/cpmip_metrics.py
+++ b/autosubmit/metrics/cpmip_metrics.py
@@ -1,0 +1,538 @@
+# Copyright 2015-2026 Earth Sciences Department, BSC-CNS
+#
+# This file is part of Autosubmit.
+#
+# Autosubmit is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Autosubmit is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Autosubmit. If not, see <http://www.gnu.org/licenses/>.
+
+from dataclasses import dataclass
+from typing import Any, Optional, Union
+
+from autosubmit.log.log import Log
+
+
+@dataclass
+class CPMIPMetricsData:
+    """
+    Container for CPMIP metrics computation inputs.
+    
+    Holds the essential runtime and simulation metadata needed to compute CPMIP metrics.
+    This decouples metrics computation from job object structure.
+    
+    :param start_time: Unix timestamp (seconds) when the job started.
+    :type start_time: Optional[int]
+    :param end_time: Unix timestamp (seconds) when the job ended.
+    :type end_time: Optional[int]
+    :param run_years: Length of the simulation in years. Can be None.
+    :type run_years: Optional[float]
+    :param ncpus: Number of CPUs used by the job. Can be None if unavailable.
+    :type ncpus: Optional[int]
+    """
+    start_time: Optional[int] = None
+    end_time: Optional[int] = None
+    run_years: Optional[float] = None
+    ncpus: Optional[int] = None
+
+
+class CPMIPMetrics:
+    """Evaluates CPMIP performance metrics against configured thresholds."""
+
+    METRIC_NAME_ALIASES = {
+        "SYPD": ("SYPD", "simulated_years_per_day"),
+        "CORE_HOURS": ("CORE_HOURS", "core_hours"),
+        "CHSY": ("CHSY", "core_hours_per_simulated_year"),
+    }
+    _METRIC_NAME_LOOKUP = {
+        alias.strip().lower(): metric_name
+        for metric_name, aliases in METRIC_NAME_ALIASES.items()
+        for alias in aliases
+    }
+
+    _VALIDATION_SCHEMA = {
+        "start_time": {
+            "type": int,
+            "type_label": "an integer value",
+            "required": True,
+            "positive": False,
+        },
+        "end_time": {
+            "type": int,
+            "type_label": "an integer value",
+            "required": True,
+            "positive": False,
+        },
+        "run_years": {
+            "type": float,
+            "type_label": "a float value",
+            "required": False,
+            "positive": True,
+        },
+        "ncpus": {
+            "type": int,
+            "type_label": "an integer value",
+            "required": False,
+            "positive": True,
+        },
+    }
+
+    @staticmethod
+    def _validate_positive(
+        value: Union[int, float],
+        var_name: str,
+        error_message: Optional[str] = None,
+        allow_zero: bool = False,
+    ) -> None:
+        """
+        Validate that a value is positive, optionally allowing zero.
+
+        :param value: Value to validate.
+        :type value: Union[int, float]
+        :param var_name: Variable name for error messages.
+        :type var_name: str
+        :param error_message: Optional custom message for invalid values.
+        :type error_message: Optional[str]
+        :param allow_zero: Whether value can be equal to zero.
+        :type allow_zero: bool
+        :raises ValueError: If value is invalid for the configured positivity rule.
+        """
+        if allow_zero:
+            if value < 0:
+                raise ValueError(error_message or f"{var_name} must be >= 0")
+            return
+
+        if value <= 0:
+            raise ValueError(error_message or f"{var_name} must be > 0")
+
+    @staticmethod
+    def _validate_against_schema(value: Any, field_name: str, schema_rules: dict[str, Any]) -> Any:
+        """
+        Validate a value against schema rules and return normalized value.
+
+        :param value: Value to validate (can be None).
+        :type value: any
+        :param field_name: Field name for error messages.
+        :type field_name: str
+        :param schema_rules: Dictionary with validation rules (type, allowed_values, positive, etc.).
+        :type schema_rules: dict
+        :return: Normalized value (e.g., lowercased strings if case_insensitive=True).
+        :rtype: any
+        :raises TypeError: If type is invalid.
+        :raises ValueError: If value violates constraints.
+        """
+
+        if value is None:
+            if schema_rules.get("required", False):
+                raise ValueError(f"{field_name} is required")
+            return None
+
+
+        expected_type = schema_rules.get("type")
+        if expected_type and not isinstance(value, expected_type):
+            type_label = schema_rules.get("type_label", f"{expected_type.__name__}")
+            raise TypeError(f"{field_name} must be {type_label}")
+
+
+        if schema_rules.get("positive", False):
+            CPMIPMetrics._validate_positive(
+                value,
+                field_name,
+                error_message=f"{field_name} must be > 0",
+            )
+
+
+        allowed_values = schema_rules.get("allowed_values")
+        if allowed_values:
+            case_insensitive = schema_rules.get("case_insensitive", False)
+            check_value = value.strip().lower() if case_insensitive and isinstance(value, str) else value
+
+            if case_insensitive and isinstance(value, str):
+                allowed_lower = [v.lower() for v in allowed_values]
+                if check_value not in allowed_lower:
+                    raise ValueError(
+                        f"{field_name} must be one of {allowed_values}, got: {value}"
+                    )
+            else:
+                if value not in allowed_values:
+                    raise ValueError(
+                        f"{field_name} must be one of {allowed_values}, got: {value}"
+                    )
+
+        if schema_rules.get("case_insensitive", False) and isinstance(value, str):
+            return value.strip().lower()
+
+        return value
+
+    @staticmethod
+    def validate_metrics_data(data: CPMIPMetricsData) -> None:
+        """
+        Validate all fields of CPMIPMetricsData using the schema.
+
+        :param data: CPMIPMetricsData object to validate.
+        :type data: CPMIPMetricsData
+        :raises TypeError: If data is not a CPMIPMetricsData instance.
+        :raises ValueError: If any field violates schema constraints.
+        """
+        if not isinstance(data, CPMIPMetricsData):
+            raise TypeError("data must be a CPMIPMetricsData object")
+
+        for field_name, schema_rules in CPMIPMetrics._VALIDATION_SCHEMA.items():
+            field_value = getattr(data, field_name)
+            try:
+                normalized_value = CPMIPMetrics._validate_against_schema(field_value, field_name, schema_rules)
+                setattr(data, field_name, normalized_value)
+            except (TypeError, ValueError):
+                if schema_rules.get("required", False):
+                    raise
+                # Optional fields degrade to None so non-required metric paths can continue.
+                setattr(data, field_name, None)
+
+    @staticmethod
+    def _validate_timestamp_pair(start_time: Optional[int], end_time: Optional[int]) -> bool:
+        """
+        Validate start_time and end_time, ensuring end_time > start_time.
+
+        :param start_time: Unix timestamp (seconds) when job started.
+        :type start_time: Optional[int]
+        :param end_time: Unix timestamp (seconds) when job ended.
+        :type end_time: Optional[int]
+        :return: True when timestamps are valid for runtime computation, otherwise False.
+        :rtype: bool
+        """
+        if start_time is None or end_time is None:
+            Log.warning(
+                "Unable to compute CPMIP metrics due to invalid runtime metadata: "
+                "start_time and end_time are required"
+            )
+            return False
+
+        if end_time <= start_time:
+            Log.warning(
+                "Unable to compute CPMIP metrics due to invalid runtime metadata: "
+                "end_time must be greater than start_time"
+            )
+            return False
+
+        return True
+
+    @staticmethod
+    def _validate_threshold_config(metric_name: str, threshold_config: dict) -> dict:
+        """
+        Validate and normalize threshold configuration for a metric.
+
+        :param metric_name: Metric name from thresholds dict.
+        :type metric_name: str
+        :param threshold_config: Configuration dict with threshold, comparison, %_accepted_error.
+        :type threshold_config: dict
+        :return: Normalized config dict with parsed values.
+        :rtype: dict
+        :raises ValueError: If config is invalid or missing required keys.
+        :raises TypeError: If values have invalid types.
+        """
+        try:
+            normalized_config = {
+                str(key).strip().lower(): value
+                for key, value in threshold_config.items()
+            }
+
+            threshold = float(normalized_config["threshold"])
+            CPMIPMetrics._validate_positive(
+                threshold,
+                f"{metric_name} threshold",
+                error_message=f"{metric_name} threshold must be > 0",
+            )
+
+            comparison = str(normalized_config["comparison"]).strip().lower()
+
+            accepted_error = float(normalized_config["%_accepted_error"])
+            CPMIPMetrics._validate_positive(
+                accepted_error,
+                f"{metric_name} accepted error",
+                error_message=f"{metric_name} accepted error must be >= 0",
+                allow_zero=True,
+            )
+
+            if comparison not in ["greater_than", "less_than"]:
+                raise ValueError(
+                    f"Unsupported comparison operator: {comparison}"
+                )
+
+            return {
+                "threshold": threshold,
+                "comparison": comparison,
+                "accepted_error": accepted_error,
+            }
+        except (ValueError, TypeError, KeyError) as error:
+            raise ValueError(
+                f"Invalid threshold configuration for '{metric_name}': {error}"
+            ) from error
+
+    @staticmethod
+    def _simulated_years_per_day(runtime: float, simulated_years: float) -> float:
+        """
+        Compute Simulated Years Per Day from runtime and simulated years.
+
+        Assumes inputs are pre-validated. Called from _fetch_metrics() only.
+
+        :param runtime: Runtime of a Job in hours.
+        :type runtime: float
+        :param simulated_years: Simulated years.
+        :type simulated_years: float
+        :return: Simulated years per day.
+        :rtype: float
+        """
+        return simulated_years * 24.0 / runtime
+
+    @staticmethod
+    def _core_hours(runtime: float, ncpus: int) -> float:
+        """
+        Compute core-hours from runtime and CPU count.
+
+        Assumes inputs are pre-validated. Called from _fetch_metrics() and _core_hours_per_simulated_year() only.
+
+        :param runtime: Runtime of a Job in hours.
+        :type runtime: float
+        :param ncpus: Number of CPUs used by the job.
+        :type ncpus: int
+        :return: Core-hours.
+        :rtype: float
+        """
+        return ncpus * runtime
+
+    @staticmethod
+    def _core_hours_per_simulated_year(runtime: float, simulated_years: float, ncpus: int) -> float:
+        """
+        Compute Core-Hours per Simulated Year.
+
+        Assumes inputs are pre-validated. Called from _fetch_metrics() only.
+
+        :param runtime: Runtime of a Job in hours.
+        :type runtime: float
+        :param simulated_years: Simulated years.
+        :type simulated_years: float
+        :param ncpus: Number of CPUs used by the job.
+        :type ncpus: int
+        :return: Core-hours per simulated year.
+        :rtype: float
+        """
+        return CPMIPMetrics._core_hours(runtime, ncpus) / simulated_years
+
+    @staticmethod
+    def _get_runtime_hours(data: CPMIPMetricsData) -> float:
+        """
+        Extract runtime in hours from metrics data timestamps.
+
+        :param data: CPMIPMetricsData object with start/end times.
+        :type data: CPMIPMetricsData
+        :return: Runtime in hours.
+        :rtype: float
+        """
+        if data.start_time is None or data.end_time is None:
+            raise ValueError("start_time and end_time are required to compute runtime")
+
+        runtime_seconds = data.end_time - data.start_time
+        return runtime_seconds / 3600
+
+    @staticmethod
+    def _fetch_metrics(data: CPMIPMetricsData) -> dict[str, float]:
+        """
+        Build CPMIP metrics from runtime and simulation metadata.
+
+        :param data: CPMIPMetricsData object.
+        :type data: CPMIPMetricsData
+        :return: Dictionary with computed metrics. Empty dict when required metadata is missing or invalid.
+               If run_years metadata is missing/invalid, run_years-dependent metrics can fail while
+             runtime-dependent metrics (for example CORE_HOURS) can still be returned.
+        :rtype: dict[str, float]
+           :raises ValueError: If timestamp or run_years metadata is invalid.
+        :raises TypeError: If data types are invalid.
+        """
+
+        try:
+            CPMIPMetrics.validate_metrics_data(data)
+        except (ValueError, TypeError) as error:
+            Log.warning(f"Unable to compute CPMIP metrics due to invalid data: {error}")
+            return {}
+
+        if not CPMIPMetrics._validate_timestamp_pair(data.start_time, data.end_time):
+            return {}
+
+        runtime_hours = CPMIPMetrics._get_runtime_hours(data)
+        simulated_years = data.run_years
+
+        metrics: dict[str, float] = {}
+        ncpus = data.ncpus
+
+        if ncpus is None:
+            Log.warning("Skipping CORE_HOURS and CHSY metric computation because ncpus is missing")
+
+        if simulated_years is not None:
+            try:
+                metrics["SYPD"] = CPMIPMetrics._simulated_years_per_day(runtime_hours, simulated_years)
+            except (ValueError, TypeError) as error:
+                Log.warning(f"Unable to compute SYPD metric: {error}")
+
+        if ncpus is not None:
+            try:
+                metrics["CORE_HOURS"] = CPMIPMetrics._core_hours(runtime_hours, ncpus)
+            except (ValueError, TypeError) as error:
+                Log.warning(f"Unable to compute CORE_HOURS metric: {error}")
+
+        if ncpus is not None and simulated_years is not None and "CORE_HOURS" in metrics:
+            try:
+                metrics["CHSY"] = CPMIPMetrics._core_hours_per_simulated_year(
+                    runtime_hours,
+                    simulated_years,
+                    ncpus,
+                )
+            except (ValueError, TypeError) as error:
+                Log.warning(f"Unable to compute CHSY metric: {error}")
+
+        return metrics
+
+    @staticmethod
+    def _resolve_metric_for_evaluation(metric_name: str, metrics: dict) -> Optional[str]:
+        """
+        Resolve a metric alias to its canonical name and ensure it was computed.
+
+        :param metric_name: Metric alias provided by the threshold configuration.
+        :type metric_name: str
+        :param metrics: Dictionary of already computed metrics.
+        :type metrics: dict
+        :return: Canonical metric name if resolvable and present in computed metrics, otherwise None.
+        :rtype: Optional[str]
+        """
+        canonical_metric_name = CPMIPMetrics._METRIC_NAME_LOOKUP.get(metric_name.strip().lower())
+        if canonical_metric_name is None:
+            Log.warning(f"Ignoring unknown CPMIP metric alias: {metric_name}")
+            return None
+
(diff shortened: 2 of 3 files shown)


MissterP:
What I changed from the documentation of last PR is that now, The CPMIPS computation class will only handle a dataclass with the key attributes to compute the CPMIPS. That way is avoided the coupling between Job and CPMIPS computation.

MissterP:
Checking missing lines from coverage

autosubmit/metrics/__init__.py
VindeeR, line 1: I believe this is a new file, to check how to create this file you can use the file bellow as a guide

`autosubmit/job/__init__.py`

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,241 @@
...
+# along with Autosubmit. If not, see <http://www.gnu.org/licenses/>.
+
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class CPMIPMetricsData:
+    """Container for CPMIP metrics computation inputs.
+
```
VindeeR:
```suggestion
    """
    Container for CPMIP metrics computation inputs.
    
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,241 @@
...
+    
+    Holds the essential runtime and chunk metadata needed to compute CPMIP metrics
+    (SYPD and CHSY). This decouples metrics computation from job object structure.
+    
+    Attributes:
+        start_time: Unix timestamp (seconds) when the job started. Can be None.
+        end_time: Unix timestamp (seconds) when the job ended. Can be None.
+        chunk_size: Size of the simulation chunk (numeric). Can be None.
+        chunk_size_unit: Unit of chunk_size. Supported: "year", "month", "day", "hour". Can be None.
+        ncpus: Number of CPUs used by the job. Can be None if unavailable.
```
VindeeR: The standard for AS docstring is sphinx, because of our documentation be generated by it.
Below I created the structure on how to change it to follow this pattern.

I don't know the types, although some can be deduced, I don't want write the wrong type.

*I did one as an example for the other, but would ask you to double check it please

```suggestion
    :param start_time: Unix timestamp (seconds) when the job started. Can be None.
    :type: 
    :param end_time: Unix timestamp (seconds) when the job ended. Can be None.
    :type: 
    :param chunk_size: Size of the simulation chunk (numeric). Can be None.
    :type: 
    :param chunk_size_unit: Unit of chunk_size. Supported: "year", "month", "day", "hour". Can be None.
    :type: 
    :param ncpus: Number of CPUs used by the job. Can be None if unavailable.
    :type: None | int
```

VindeeR: *copied from another PR just to give you a bit more context on how to work the docstring

We are trying to follow the Sphinx or reST style, since we use the Sphinx to generate our documentation. an example would be the following.

```
def function_name(param1, param2):
    """
    Short summary line.

    More detailed description if needed.

    :param param1: Description of param1
    :type param1: type
    :param param2: Description of param2
    :type param2: type
    :return: Description of the return value
    :rtype: return_type
    :raises ExceptionType: Description of the exception
    """
```

Despite this being the standard we adopted we still need many changes as we go, since it was adopted not that long ago.

[reference](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html)

VindeeR (changes requested):
I just had a glance over this PR, but I already seen some things to fix, I will comeback to this PR tomorrow

test/unit/test_cpmip_metrics.py
```diff
@@ -0,0 +1,532 @@
...
+        end_time=2000,
+        chunk_size=1,
+        chunk_size_unit="year",
+        ncpus=64,
+    )
+
+
+class TestEvaluateCPMIPMetrics:
+
+    def test_evaluate_returns_empty_when_thresholds_is_empty(self, metrics_data):
```
ntorqulu: Hi @MissterP, great work with the tests!
I was checking the evaluations of the thresholds and I think we should also ensure that the thresholds provided by the user are not negative.
What i mean by that:
I think that simulated years per day, core hours and chsy must be always positive in the thresholds of the yaml files. 
The user could make an error without noticing and put a negative value, and the threshold per se does not make sense anymore but the validation it provides yes:
SYPD > -10 for instance will always be true, undercovered mistake.
What do you think?

test/unit/test_cpmip_metrics.py
```diff
@@ -0,0 +1,532 @@
...
+        mocker,
+        metrics_data,
+        metric_name,
+        threshold,
+        comparison,
+        accepted_error,
+        real_value,
+        expected_bound,
+    ):
+        thresholds = {
```
ntorqulu: Also, about the standards we follow in the tests of AS, we try to ensure that:
- between each test there's a separation of two lines
- there's a blank line at the end of the file
- all tests are shortly explained with a brief comment. E.g.
"""Test that `evaluate` correctly identifies violations when the fetched metric value exceeds the threshold + tolerance."""

test/unit/test_cpmip_metrics.py
paparici: Thank you so much for the comments and suggestion!

test/unit/test_cpmip_metrics.py
paparici: Great! On it!

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,413 @@
...
+            float,
+            expected_label="a float value",
+        )
+        CPMIPMetrics._validate_positive(runtime, "runtime", positive_error_message="runtime must be > 0 hours")
+        CPMIPMetrics._validate_positive(simulated_years, "simulated_years")
+
+        return simulated_years * 24.0 / runtime
+
+    @staticmethod
+    def CORE_HOURS(runtime, ncpus) -> float:
```
ntorqulu: Thanks for the modifications in the tests @MissterP ! Another thing I'm noticing is the names of some of the methods.
In python is usually recommended to follow PEP 8 style, in which functions/methods use `snake_case`. Upper case is usually reserved only for constants. For functions CORE_HOURS, SYPD, SY, CHSY I would recommend following the PEP 8 style and use `snake_case` instead.

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,413 @@
...
+
+        for metric_name, threshold_config in thresholds.items():
+            if metric_name not in metrics:
+                continue
+            
+            try:
+                threshold = float(threshold_config["THRESHOLD"])
+                CPMIPMetrics._validate_positive(threshold, f"{metric_name} threshold", positive_error_message=f"{metric_name} threshold must be > 0")
+                comparison = str(threshold_config["COMPARISON"]).strip().lower()
+                accepted_error = int(threshold_config.get("%_ACCEPTED_ERROR"))
```
ntorqulu: Question: `accepted_error` can be float? I see that in test `test_evaluate_violation_cases`, `accepted_error` is defined as a float. Should be cased into a float instead of int?

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,413 @@
...
+            ncpus,
+            "ncpus",
+            int,
+            expected_label="an integer value",
+        )
+        CPMIPMetrics._validate_positive(runtime, "runtime", positive_error_message="runtime must be > 0 hours")
+        CPMIPMetrics._validate_positive(simulated_years, "simulated_years")
+        CPMIPMetrics._validate_positive(ncpus, "ncpus")
+
+        return ncpus * runtime / simulated_years
```
ntorqulu: Why don't use the CORE_HOURS function from before instead of repeating code?

`return CPMIPMetrics.CORE_HOURS(runtime, cpus) / simulated_years`

This guarantees consistency and reduces errors later when we have to modify functions.

ntorqulu (commented):
Hi! Thanks for the modifications of the tests.
Checking it again I've found the following things. Some are just suggestions, the only real issue is the conversion of "%_ACCEPTED_UNITS" into integer.

MissterP:
Hi @ntorqulu! Thank you so much for the suggestions! Applying them now.

autosubmit/metrics/cpmip_metrics.py
paparici: Thank you for pointing that out! At first, I cased it as a float. Then change it as an int (e.g. it really make sense to have a 45.5% of acceptance error?). However, thanks to your question maybe it makes sense to case it back as a float, because someone could want to use  2.5% (example). Just in case, the first decimal will be collected (to avoid the possibility of someone adding many decimals xD).

ntorqulu: Sure, just one thing. If a value is modified from the configuration file to perform the metric computations I would log the change to keep traceability and document it in the documentation (for the future).

MissterP:
Added metric aliases to enhance the readability of the metrics (at the config) (accept SYPD and simulated_years_per_day)

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+@dataclass
+class CPMIPMetricsData:
+    """
+    Container for CPMIP metrics computation inputs.
+    
+    Holds the essential runtime and chunk metadata needed to compute CPMIP metrics. 
+    This decouples metrics computation from job object structure.
+    
+    :param start_time: Unix timestamp (seconds) when the job started. Can be None.
+    :type: None | int
```
VindeeR:
```suggestion
    :type start_time: Optional[int]
```

Whenever you have a variable that can return as `None` in python you can use the `Optional` type hint `e.g.: Optional[int]` in this case or you could use `Union`, but that is usually used for more than one type instead of None and 1 type.

[reference material](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/consider-alternative-union-syntax.html)

edit: I just realized you used `Optional` on the variables but not on the docstring, so yeah, you can write it as `Optional` there too

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+        "CHSY": ("CHSY", "core_hours_per_simulated_year"),
+    }
+    _METRIC_NAME_LOOKUP = {
+        alias.lower(): metric_name
+        for metric_name, aliases in METRIC_NAME_ALIASES.items()
+        for alias in aliases
+    }
+
+    @staticmethod
+    def _validate_type(value, var_name, expected_type, expected_label: Optional[str] = None) -> None:
```
VindeeR: don't forget the type hint

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+    def _validate_positive(
+        value,
+        var_name,
+        positive_error_message: Optional[str] = None,
+    ) -> None:
+        """
+        Validate that a value is greater than zero.
+
+        :param value: Value to validate.
+        :type value: any
```
VindeeR, line 122: your value is `Any` here but you treat it as int down there

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+        if not isinstance(value, expected_type):
+            label = expected_label or str(expected_type)
+            raise TypeError(f"{var_name} must be {label}")
+
+    @staticmethod
+    def _validate_positive(
+        value,
+        var_name,
+        positive_error_message: Optional[str] = None,
+    ) -> None:
```
VindeeR, line 94: Maybe you could implement this function like this, avoiding duplicating you function and still keep the behaviors separated this way you can check which message to display based on which error message you get or maybe even both.

```suggestion
    def _validate_positive(
        value,
        var_name,
        positive_error_message: Optional[str] = None,
        zero_or_positive_error_message: Optional[str] = None,
    ) -> None:
```

what do you think?

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+        """
+        Normalize a metric name used in thresholds to the canonical metric key.
+
+        :param metric_name: Threshold metric name or alias.
+        :type metric_name: str
+        :return: Canonical metric key when known, otherwise None.
+        :rtype: Optional[str]
+        """
+        if not isinstance(metric_name, str):
+            return None
```
VindeeR: Should this exhibit a message to the user?

It just occurred to me that it could be interesting have some info logged to the user using something like

`Log.warning()`

Whenever one data couldn't be used for any reason, this can help an eventual debugging later on too.

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+        :param var_name: Variable name for error messages.
+        :type var_name: str
+        :param expected_type: Accepted type.
+        :type expected_type: type
+        :param expected_label: Optional custom label for expected type in error messages.
+        :type expected_label: Optional[str]
+        :raises TypeError: If value type is invalid.
+        """
+        if not isinstance(value, expected_type):
+            label = expected_label or str(expected_type)
```
VindeeR: As I was checking the call for this functions I realized that you always use a string, so I believe there's no point in having this line

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+        :rtype: float
+        :raises ValueError: If runtime or ncpus is not positive.
+        :raises TypeError: If runtime is not a float or ncpus is not an int.
+        """
+        CPMIPMetrics._validate_type(
+            runtime,
+            "runtime",
+            float,
+            expected_label="a float value",
+        )
```
VindeeR: I just notice this is used in a bunch of different functions, wouldn't it be better to calculate it before and pass on the function, this way you would be avoiding repeating the call.

This is more of a suggestion since it might not make sense with the design and the variable might change in between functions in a way that wouldn't allow this.

On the same note, would be possible to encapsulate them, I notice you have a bunch of them being called at all functions, so I was wondering if there's the possibility of rearranging this validation inside one function, that could be an alternative to the other proposed change.

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+        :type data: CPMIPMetricsData
+        :return: Dictionary with computed metrics. Empty dict when required metadata is missing or invalid.
+        :rtype: dict
+        :raises ValueError: If timestamp or chunk metadata is invalid.
+        :raises TypeError: If data types are invalid.
+        """
+        try:
+            runtime_hours = CPMIPMetrics._get_runtime_hours(data)
+        except (ValueError, TypeError):
+            return {}
```
VindeeR, line 364: Maybe adding a message so the user can know where the error came from and why it returned an empty dict?

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+            except (ValueError, TypeError):
+                pass
+
+            if core_hours_computed and simulated_years is not None:
+                try:
+                    metrics["CHSY"] = CPMIPMetrics.core_hours_per_simulated_year(runtime_hours, simulated_years, ncpus)
+                except (ValueError, TypeError):
+                    pass
+
+        return metrics
```
VindeeR, line 400: Maybe a message for when the metrics are empty

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+
+        for metric_name, threshold_config in thresholds.items():
+            canonical_metric_name = CPMIPMetrics._resolve_metric_name(metric_name)
+            if canonical_metric_name is None or canonical_metric_name not in metrics:
+                continue
+            
+            try:
+                threshold = float(threshold_config["THRESHOLD"])
+                CPMIPMetrics._validate_positive(threshold, f"{metric_name} threshold", positive_error_message=f"{metric_name} threshold must be > 0")
+                comparison = str(threshold_config["COMPARISON"]).strip().lower()
```
VindeeR: This is something that will be coming from the configuration files am I correct? If so you might have to ensure the normalization of the spelling, iirc, you are already did it when creating this dictionary, but I'm not 100% sure.

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,442 @@
...
+                bound = lower_bound
+                if real_value < lower_bound:
+                    is_violation = True
+            elif comparison == "less_than":
+                upper_bound = threshold * (1 + tolerance_factor)
+                bound = upper_bound
+                if real_value > upper_bound:
+                    is_violation = True
+            else:
+                continue
```
VindeeR, line 527: Maybe printing a warning here so the user can identify that the comparison variable fulfill none of the expected values?

VindeeR (commented):
I haven't gone through the tests yet.

Here a some considerations, and some small adjustments.

Other than this small things good job overall on developing this feature, we are almost done 🙌

MissterP:
Hi @VindeeR! Thank you so much for the suggestions! I will start working on them now

autosubmit/metrics/cpmip_metrics.py
paparici: The key thing is that these are static methods, so any class could use them in the future. That's why I'm repeating the validation even if it is duplicated. While I apply the other suggestions, I will think of what to do for these validations.

ntorqulu:
Hi @MissterP! We're having problems with the rocrate version compatibility with AS.
In order for the integration tests to pass, can you change the rocrate version in `project.toml` and enforce version 
```
"rocrate==0.14",
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,448 @@
...
+
+                threshold = float(normalized_threshold_config["threshold"])
+                CPMIPMetrics._validate_positive(
+                    threshold,
+                    f"{metric_name} threshold",
+                    error_message=f"{metric_name} threshold must be > 0",
+                )
+                comparison = str(normalized_threshold_config["comparison"]).strip().lower()
+                accepted_error = float(normalized_threshold_config.get("%_accepted_error"))
+                accepted_error = round(accepted_error, 1)
```
ntorqulu: About the rounding of the accepted error, I'm not sure if this is the correct approach (for discussion). I would:
- round and log a message saying that the accepted_error has been rounded to one decimal place if num decimals > 1
- do not round and use the float value as it is
But I would notify the user that the value entered has been modified if we round

test/unit/test_cpmip_metrics.py
```diff
@@ -0,0 +1,776 @@
...
+            return_value={"SYPD": 3.9, "LATENCY": 10.4},
+        )
+
+        violations = CPMIPMetrics.evaluate(metrics_data, thresholds)
+
+        assert list(violations.keys()) == ["SYPD"]
+
+
+    def test_evaluate_rounds_accepted_error_to_first_decimal(self, mocker, metrics_data):
+        """Test that accepted error is rounded to one decimal place."""
```
ntorqulu: test here that a log info message is logged if we decide to round

test/unit/test_cpmip_metrics.py
```diff
@@ -0,0 +1,776 @@
...
+            "invalid_threshold_type",
+            "zero_threshold",
+            "negative_threshold",
+            "invalid_accepted_error_type",
+            "negative_accepted_error",
+            "accepted_error_zero",
+            "non_string_metric_name",
+        ],
+    )
+    def test_evaluate_non_violation_paths(self, mocker, metrics_data, thresholds, fetched_metrics):
```
ntorqulu: I would also test that the warning are being log. Not the warning test per se, because it could change in the future. More that Log.warning is being called.

test/unit/test_cpmip_metrics.py
```diff
@@ -0,0 +1,776 @@
...
+
+        assert metrics == expected_metrics
+
+
+    @pytest.mark.parametrize(
+        "simulated_years_error",
+        [ValueError("invalid chunk metadata"), TypeError("invalid chunk metadata type")],
+        ids=["simulated_years_value_error", "simulated_years_type_error"],
+    )
+    def test_fetch_metrics_returns_core_hours_when_simulated_years_extraction_fails(self, mocker, simulated_years_error):
```
ntorqulu: Here the same, I would test that logs are called for the failed metrics to ensure that they are being displayed correctly.

test/unit/test_cpmip_metrics.py
```diff
@@ -0,0 +1,776 @@
...
+        chsy_mock.assert_not_called()
+        assert metrics == {"CORE_HOURS": 2880.0}
+
+
+    @pytest.mark.parametrize(
+        "chsy_error",
+        [ValueError("invalid chsy"), TypeError("invalid chsy type")],
+        ids=["chsy_value_error", "chsy_type_error"],
+    )
+    def test_fetch_metrics_returns_only_sypd_and_ch_when_chsy_computation_fails(self, mocker, chsy_error):
```
ntorqulu: Here the same, I would test that logs are called for the failed metrics to ensure that they are being displayed correctly.

test/unit/test_cpmip_metrics.py
```diff
@@ -0,0 +1,776 @@
...
+        chsy_mock.assert_called_once_with(12.0, 0.5, 64)
+        assert metrics == {"SYPD": 1.0, "CORE_HOURS": 2880.0}
+
+
+    @pytest.mark.parametrize(
+        "core_hours_error",
+        [ValueError("invalid core hours"), TypeError("invalid core hours type")],
+        ids=["core_hours_value_error", "core_hours_type_error"],
+    )
+    def test_fetch_metrics_returns_only_sypd_when_core_hours_computation_fails(self, mocker, core_hours_error):
```
ntorqulu: Here the same, I would test that logs are called for the failed metrics to ensure that they are being displayed correctly.

ntorqulu (commented):
Hi! Just a few suggestions about testing that warning logs are being displayed, and a discussion on how to treat the rounding of % accepted error for the metrics.

MissterP:
Hi @ntorqulu ! Thank you so much for the suggestions! Applying them between today and tomorrow

MissterP:
Rebased branch with master

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+from dataclasses import dataclass
+from typing import Optional, Union
+
+from autosubmit.log.log import Log
+
+
+@dataclass
+class CPMIPMetricsData:
+    """
+    Container for CPMIP metrics computation inputs.
```
VindeeR, line 27:
```suggestion
    """Container for CPMIP metrics computation inputs.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+
+    @staticmethod
+    def _validate_positive(
+        value: Union[int, float],
+        var_name: str,
+        error_message: Optional[str] = None,
+        allow_zero: bool = False,
+    ) -> None:
+        """
+        Validate that a value is positive, optionally allowing zero.
```
VindeeR, line 96:
```suggestion
        """Validate that a value is positive, optionally allowing zero.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+                raise ValueError(error_message or f"{var_name} must be >= 0")
+            return
+
+        if value <= 0:
+            raise ValueError(error_message or f"{var_name} must be > 0")
+
+    @staticmethod
+    def _validate_against_schema(value: any, field_name: str, schema_rules: dict) -> any:
+        """
+        Validate a value against schema rules and return normalized value.
```
VindeeR, line 119:
```suggestion
        """Validate a value against schema rules and return normalized value.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+
+        if schema_rules.get("case_insensitive", False) and isinstance(value, str):
+            return value.strip().lower()
+
+        return value
+
+    @staticmethod
+    def validate_metrics_data(data: CPMIPMetricsData) -> None:
+        """
+        Validate all fields of CPMIPMetricsData using the schema.
```
VindeeR, line 178:
```suggestion
        """Validate all fields of CPMIPMetricsData using the schema.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+            except (TypeError, ValueError):
+                if schema_rules.get("required", False):
+                    raise
+                # Optional fields degrade to None so non-required metric paths can continue.
+                setattr(data, field_name, None)
+
+    @staticmethod
+    def _validate_timestamp_pair(start_time: Optional[int], end_time: Optional[int]) -> bool:
+        """
+        Validate start_time and end_time, ensuring end_time > start_time.
```
VindeeR, line 202:
```suggestion
        """Validate start_time and end_time, ensuring end_time > start_time.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+                "end_time must be greater than start_time"
+            )
+            return False
+
+        return True
+
+    @staticmethod
+    def _validate_threshold_config(metric_name: str, threshold_config: dict) -> dict:
+        """
+        Validate and normalize threshold configuration for a metric.
```
VindeeR, line 230:
```suggestion
        """Validate and normalize threshold configuration for a metric.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+        :rtype: float
+        :raises TypeError: If chunk_size_unit is not a string or chunk_size is not an int.
+        :raises ValueError: If chunk_size_unit is unsupported or chunk_size is not positive.
+        """
+
+        unit_to_years = {
+            "year": 1.0,
+            "month": 1.0 / 12.0,
+            "day": 1.0 / 365.0,
+            "hour": 1.0 / 8760.0,
```
VindeeR: I was just wondering, but what about the leap year? with 366 days in a year?
Could this cause any sort of problem? Both in `days` and `hours` field

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+        }
+
+        unit = chunk_size_unit.strip().lower()
+        return chunk_size * unit_to_years[unit]
+
+    @staticmethod
+    def _simulated_years_per_day(runtime: float, simulated_years: float) -> float:
+        """
+        Compute Simulated Years Per Day from runtime and simulated years.
```
VindeeR, line 282:
```suggestion
        """Compute Simulated Years Per Day from runtime and simulated years.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+        :type simulated_years: float
+        :return: Simulated years per day.
+        :rtype: float
+        """
+        return simulated_years * 24.0 / runtime
+
+    @staticmethod
+    def _core_hours(runtime: float, ncpus: int) -> float:
+        """
+        Compute core-hours from runtime and CPU count.
```
VindeeR, line 298:
```suggestion
        """Compute core-hours from runtime and CPU count.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+        :type ncpus: int
+        :return: Core-hours.
+        :rtype: float
+        """
+        return ncpus * runtime
+
+    @staticmethod
+    def _core_hours_per_simulated_year(runtime: float, simulated_years: float, ncpus: int) -> float:
+        """
+        Compute Core-Hours per Simulated Year.
```
VindeeR, line 314:
```suggestion
        """Compute Core-Hours per Simulated Year.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+        :type ncpus: int
+        :return: Core-hours per simulated year.
+        :rtype: float
+        """
+        return CPMIPMetrics._core_hours(runtime, ncpus) / simulated_years
+
+    @staticmethod
+    def _get_runtime_hours(data: CPMIPMetricsData) -> float:
+        """
+        Extract runtime in hours from metrics data timestamps.
```
VindeeR, line 332:
```suggestion
        """Extract runtime in hours from metrics data timestamps.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+        :return: Runtime in hours.
+        :rtype: float
+        """
+        runtime_seconds = data.end_time - data.start_time
+        return runtime_seconds / 3600
+
+    @staticmethod
+    def _fetch_metrics(data: CPMIPMetricsData) -> dict:
+        """
+        Build CPMIP metrics from metrics data runtime/chunk metadata.
```
VindeeR:
```suggestion
        """Build CPMIP metrics from metrics data runtime/chunk metadata.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+                metrics[spec["name"]] = spec["func"](*spec["args"]())
+            except (ValueError, TypeError) as error:
+                Log.warning(f"Unable to compute {spec['name']} metric: {error}")
+
+        return metrics
+
+    @staticmethod
+    def _resolve_metric_for_evaluation(metric_name: str, metrics: dict) -> Optional[str]:
+        """
+        Resolve a metric alias to its canonical name and ensure it was computed.
```
VindeeR, line 405:
```suggestion
        """Resolve a metric alias to its canonical name and ensure it was computed.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+                f"Skipping CPMIP metric '{metric_name}' ({canonical_metric_name}) because it was not computed"
+            )
+            return None
+
+        return canonical_metric_name
+
+    @staticmethod
+    def _extract_threshold_for_evaluation(metric_name: str, threshold_config: dict) -> Optional[dict]:
+        """
+        Validate threshold configuration and normalize it for evaluation.
```
VindeeR, line 430:
```suggestion
        """Validate threshold configuration and normalize it for evaluation.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+
+    @staticmethod
+    def _compute_violation(
+        real_value: float,
+        threshold: float,
+        comparison: str,
+        accepted_error: float,
+    ) -> Optional[dict]:
+        """
+        Compute violation payload for a single metric.
```
VindeeR, line 453:
```suggestion
        """Compute violation payload for a single metric.
```

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -0,0 +1,605 @@
...
+            "accepted_error": accepted_error,
+            "comparison": comparison,
+            "bound": bound,
+            "real_value": real_value,
+        }
+
+    @staticmethod
+    def evaluate(data: CPMIPMetricsData, thresholds: dict[str, dict]) -> dict:
+        """
+        Evaluate metrics against threshold definitions.
```
VindeeR, line 496:
```suggestion
        """Evaluate metrics against threshold definitions.
```

test/unit/test_cpmip_metrics.py
VindeeR, line 1: I'm a bit skeptical of the need of creating 3 classes here just for the test, at least as it is now, do you intend to use it in the future to test other things in future updates? if not I believe they all could be their own functions.

Sorry if I miss something regarding that.

VindeeR (commented):
The features looks great I just have 2 concerns, the docstring is negligible so you don't need to worry about fixing it unless you need to change something else pointed out by my questions.

Thanks for the hard-work, the code/validation is very impressive 🙌

MissterP:
Thank you so much for the review @VindeeR! I will answer to your questions.

test/unit/test_cpmip_metrics.py
paparici: It was just for organizing, for each class it tests a function and the rest are mocked. I will delete the classes and leave functions.

autosubmit/metrics/cpmip_metrics.py
paparici: The best thing to do is taking into account the date of the chunks. Is there any function in AS that I could use to compute the real days, hours ...?

VindeeR: If you know the year being calculate you can always use the `calendar` library
https://www.geeksforgeeks.org/python/python-program-to-check-if-given-year-is-leap-year/
https://realpython.com/ref/stdlib/calendar/

paparici: In that case, I would need the start chunk date, right? From the job

paparici: Maybe I could use parameters['RUN_DAYS'] from the Job class? Instead of using the chunk size and chunk size unit? That way redundancy is avoided?

ntorqulu: Hi @MissterP! Yes, you can find how chunks and splits are automatically computed in `job_utils.py` -> `calendar_chunk_section()`.
I'm not sure if it's over-complicating things, as I don't know which will be the final purpose of this. 
Check functions `chunk_start_date` and `chunk_end_date` for your case.

autosubmit/metrics/cpmip_metrics.py
```diff
@@ -74,19 +71,12 @@ class CPMIPMetrics:
...
         },
-        "chunk_size": {
-            "type": int,
-            "type_label": "an integer value",
+        "run_years": {
+            "type": float,
+            "type_label": "a float value",
             "required": False,
             "positive": True,
         },
```
VindeeR, line 79: Are you sure you want to have this as a float? how would you know if the year is a leap year and such? Will you be validating it elsewhere?

VindeeR (commented):
I just have one more question before approving you changes

autosubmit/metrics/cpmip_metrics.py
paparici: The validation will be done outside the class. For example, in the Job class there is the run_days parameter that takes into account the start and end of the chunk. The only thing, would be to check if the years is leap or not to convert from run_days to run_years.

Issue #2696, opened 2025-11-06
CPU Performance comparison 4.1.16 vs 4.2+ for a DestinE multimember stress test experiment


dbeltrankyl:
Hello @kinow , @VindeeR 

We already did some tests for the Coding sprint regarding 4.2 postgres vs sqlite and vs 4.1.16. #2627 

This issue is only for CPU, memory and size was checked already and it was fine. 

Opening this issue to do a comparison once the 4.2 branch is fully finished and new a PR ( to link when it is open and needed) for the optimization part.

I was testing this for @orioltinto due the Operational issue where the experiment is taking 6min on submit a new job.

Tests based on 
https://gitlab.earth.bsc.es/digital-twins/de_340-2/workflow/-/issues/1323#note_406835

Since it is a stress multimember experiment created by @franra9 

4.1.16: 

10 splits 136807 

2025-11-06 13:53:40,350 Iteration time: 57.527984380722046 seconds
2025-11-06 13:54:32,379 Iteration time: 51.84905290603638 seconds
2025-11-06 13:55:20,726 Iteration time: 48.169872999191284 seconds
2025-11-06 13:56:13,518 Iteration time: 52.617791175842285 seconds
2025-11-06 13:56:58,579 Iteration time: 44.86899137496948 seconds

2 splits 33127 jobs

2025-11-06 12:32:51,131 Iteration time: 6.9863433837890625 seconds
2025-11-06 12:33:03,522 Iteration time: 12.348278045654297 seconds
2025-11-06 12:33:18,438 Iteration time: 14.872258186340332 seconds
2025-11-06 12:33:29,457 Iteration time: 10.975615739822388 seconds
2025-11-06 12:33:44,755 Iteration time: 15.253679752349854 seconds
2025-11-06 12:33:51,342 Iteration time: 6.543314456939697 seconds
2025-11-06 12:33:57,825 Iteration time: 6.440295934677124 seconds
2025-11-06 12:34:04,223 Iteration time: 6.362769365310669 seconds
2025-11-06 12:34:10,752 Iteration time: 6.493175268173218 seconds
2025-11-06 12:34:17,682 Iteration time: 6.895272970199585 seconds
2025-11-06 12:34:24,235 Iteration time: 6.4666314125061035 seconds

4.2 - (TODO)

4.2 (optimized if needed) - (TODO)

------------------

TODO list:

(set safetysleeptime to 0 for the tests)

[...]

PR #2933, opened 2026-04-09
Better profiler from 4.2.0

dbeltrankyl:
**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/profiler/profiler.py
+++ b/autosubmit/profiler/profiler.py
@@ -72,6 +72,7 @@ class Profiler:
 
         # File descriptor / handle profiling variables
         self._fd_iteration: list = []
+        self._fd_names_iteration: list = []
         self._fd_grow: list = []
 
         # Workflow stats
@@ -132,11 +133,11 @@ class Profiler:
         :rtype: bool
         """
         gc.collect()
-
         self._mem_iteration.append(_get_current_memory())
         self._obj_iteration.append(_get_current_object_count())
 
         self._fd_iteration.append(_get_current_open_fds())
+        self._fd_names_iteration.append(_get_current_open_fds_names())
         if self._trace_enabled and tracemalloc.is_tracing():
             snapshot = tracemalloc.take_snapshot()
             self._trace_stats_by_iter.append(
@@ -148,7 +149,7 @@ class Profiler:
         self._edges_iteration.append(loaded_edges)
 
         self._mem_iteration[-1] -= sys.getsizeof(self._mem_iteration) + sys.getsizeof(self._obj_iteration) + sys.getsizeof(self._fd_iteration) + sys.getsizeof(
-            self._jobs_iteration) + sys.getsizeof(self._edges_iteration)
+            self._jobs_iteration) + sys.getsizeof(self._edges_iteration) + sys.getsizeof(self._fd_names_iteration)
         if self.max_checkpoints != 0:
             self.checkpoints += 1
             if self.checkpoints > self.max_checkpoints:
@@ -209,6 +210,7 @@ class Profiler:
             mem = self._mem_iteration[i]
             obj = self._obj_iteration[i]
             fd = self._fd_iteration[i]
+            fd_names = self._fd_names_iteration[i]
 
             mem_unit = 0
             while mem >= 1024 and mem_unit <= len(_UNITS):
@@ -220,6 +222,15 @@ class Profiler:
             report += f"{current_iter} File Descriptors: {fd}\n"
             report += f"{current_iter} Loaded jobs: {self._jobs_iteration[i]}\n"
             report += f"{current_iter} Loaded edges: {self._edges_iteration[i]}\n"
+            if i > 0:
+                for names in fd_names:
+                    if names not in self._fd_names_iteration[i - 1]:
+                        report += f"{current_iter} Opened file descriptor: {names}\n"
+
+                for names in self._fd_names_iteration[i - 1]:
+                    if names not in fd_names:
+                        report += f"{current_iter} Closed file descriptor: {names}\n"
+
 
             if i < len(self._trace_stats_by_iter):
                 report += self._format_top_allocations(self._trace_stats_by_iter[i])
@@ -382,19 +393,25 @@ def _get_current_object_count() -> int:
 def _get_current_open_fds() -> int:
     """Return count of open file descriptors.
 
-    Falls back to approximating using open files and connections
-    if direct FD/handle count is unavailable.
-
     :return: The number of open file descriptors or handles.
     :rtype: int
     """
+
     proc = Process(os.getpid())
     if hasattr(proc, "num_fds"):
         return proc.num_fds()
     if hasattr(proc, "num_handles"):
         return proc.num_handles()
 
-    # Fallback: approximate using open files + connections
-    open_files = len(proc.open_files())
-    connections = len(proc.net_connections(kind="all"))
-    return open_files + connections
+
+def _get_current_open_fds_names() -> list:
+    """Return a list of names of open file descriptors.
+
+    :return: A list of names of open file descriptors or handles.
+    :rtype: list
+    """
+    proc = Process(os.getpid())
+    if hasattr(proc, "open_files"):
+        return [f.path for f in proc.open_files()]
+    if hasattr(proc, "num_handles"):
+        return [h.path for h in proc.open_files()]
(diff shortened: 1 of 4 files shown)


dbeltrankyl:
Hello @VindeeR 

This is the profiler enhancements. Can you review it? 

Afterward, you can rebase your branch.

The test will be added later for issue #2696 in 4.1.17 since I need to adapt some test code. The tests are already done in the 4.2 branch 

Usage ex: 

autosubmit run $expid -p 20  # stops after 20 iterations of the main loop
autosubmit run $expid -p
autosubmit run $expid -p -trace # shows which objects are growing ( This is very expensive and adds a lot of noise, for fd you don't need to run this)

ex output:


```
Iteration 17: Memory: 114.63 MiB
Iteration 17: Objects: 130057
Iteration 17: File Descriptors: 17
Iteration 17: Loaded jobs: 30
Iteration 17: Loaded edges: 29
Iteration 17: Closed file descriptor: [fd=15] pipe (read) pipe:[846914]
Iteration 17: Closed file descriptor: [fd=16] pipe (write) pipe:[846914]
Iteration 18: Memory: 114.63 MiB
Iteration 18: Objects: 130058
Iteration 18: File Descriptors: 17
Iteration 18: Loaded jobs: 30
Iteration 18: Loaded edges: 29
Iteration 19: Memory: 114.63 MiB
Iteration 19: Objects: 130060
Iteration 19: File Descriptors: 19
Iteration 19: Loaded jobs: 30
Iteration 19: Loaded edges: 29
Iteration 19: Opened file descriptor: [fd=15] pipe (read) pipe:[846941]
Iteration 19: Opened file descriptor: [fd=16] pipe (write) pipe:[846941]

================================================================================
               Overall Memory, Object and File Descriptor Growth                
================================================================================

MEMORY GROW: 5.32 MiB.
INITIAL MEMORY: 109.31 MiB.
FINAL MEMORY: 114.63 MiB.
OBJECTS GROW: 16 objects.
FILE DESCRIPTORS GROW: 0 file descriptors.

FINAL OPEN FILE DESCRIPTORS:
  [fd=0] stdin (pipe:[842740])
  [fd=1] stdout (pipe:[842741])
  [fd=2] stderr (pipe:[842742])
  [fd=3] /home/dbeltran/autosubmit/a028/tmp/ASLOGS/20260409_122842_run.log
  [fd=4] /home/dbeltran/autosubmit/a028/tmp/ASLOGS/20260409_122842_run_err.log
  [fd=5] /home/dbeltran/autosubmit/a028/tmp/ASLOGS/jobs_active_status.log
  [fd=6] /home/dbeltran/autosubmit/a028/tmp/ASLOGS/jobs_failed_status.log
  [fd=8] unix-socket: (unnamed)
  [fd=9] socket: 127.0.0.1:53458 -> 127.0.0.1:2222 [ESTABLISHED]
  [fd=10] socket: 127.0.0.1:53462 -> 127.0.0.1:2222 [ESTABLISHED]
  [fd=11] pipe (read) pipe:[845418]
  [fd=12] pipe (write) pipe:[845416]
  [fd=13] pipe (write) pipe:[845418]
  [fd=14] pipe (read) pipe:[845419]
  [fd=17] pipe (write) pipe:[845420]
```
The Iteration 17: stuff ... is to be more machine readable, the complete benchmark tests ( not included in this branch) parse this  to be more human readable

autosubmit/profiler/profiler.py
```diff
@@ -41,59 +46,251 @@ class ProfilerState(Enum):
 class Profiler:
     """Class to profile the execution of experiments."""
 
-    def __init__(self, expid: str):
+    def __init__(self, expid: str, trace_enabled: bool = False, max_checkpoints: int = 0):
+        """Initialize the profiler with an experiment ID.
+
+        :param expid: The experiment identifier.
+        :type expid: str
+        """
```
VindeeR, line 58: Could you fill in the missing variable in the docstring please

```suggestion
        """Initialize the profiler with an experiment ID.

        :param expid: The experiment identifier.
        :type expid: str
        :param trace_enabled:
        :type trace_enabled: bool
        :param max_checkpoints:
        :type max_checkpoints: int
        """
```

autosubmit/profiler/profiler.py
```diff
@@ -41,59 +46,251 @@ class ProfilerState(Enum):
...
+        self._obj_grow = [self._obj_iteration[i] - self._obj_iteration[i - 1]
+                          for i in range(1, len(self._obj_iteration))]
+        self._fd_grow = [self._fd_iteration[i] - self._fd_iteration[i - 1]
+                         for i in range(1, len(self._fd_iteration))]
+
+        # total grow
+        self._mem_total_grow = self._mem_iteration[-1] - self._mem_iteration[0] if self._mem_iteration else 0
+        if self.checkpoints > 3:
+            self._obj_total_grow = self._obj_iteration[-1] - self._obj_iteration[3] if self._obj_iteration else 0
+            self._fd_total_grow = self._fd_iteration[-1] - self._fd_iteration[3] if self._fd_iteration else 0
```
VindeeR, line 203: I'm a bit confused by this is calculation, its first interaction will be with with the 4th checkpoint, the first interaction will be 0 always since the [-1] will be the 4th value in the dict, just like [3], was this intended to be like this?

I'm saying this because  this intends to calculate the total growth them shouldn't it always compare with the 1 value stored?

autosubmit/profiler/profiler.py
```diff
@@ -110,24 +307,55 @@ def _report(self) -> None:
...
+            absolute_mem_total /= 1024
+            mem_total /= 1024
+        unit = 0
+        while mem_init >= 1024 and unit <= len(_UNITS):
+            unit += 1
+            mem_init /= 1024
+        unit = 0
+        while mem_final >= 1024 and unit <= len(_UNITS):
+            unit += 1
+            mem_final /= 1024
```
VindeeR, line 344: Maybe we could convert your function to something like the `parse_output_number` and leave it at the utils??

I went to looking for your function because I thought we already had it implemented but I got confused with the function that does the exact opposite.

But looking into it I realized we already have 2 instances of the `parse_output_number`.
`
autosubmit/history/platform_monitor/platform_utils.py`
`autosubmit/job/job_common.py`

I'll create an issue so we keep trach of it to fix later

VindeeR (changes requested):
I believe one of the calculation might not be working as expected everything else is pretty minimal and can be approved without the changes.

I was wondering if you have any test available for the new parameter you could add here too.

Thanks for the profiler it'll be very helpful debugging why there are leaks. 🙌

autosubmit/profiler/profiler.py
dbeltrankyl: Yup, it is intended. It may seem unusual, but it was added mainly to reduce noise. 

This mainly comes from the try-and-error runs using 4.2.0

In 4.2.0, we first load everything to check whether the workflow has changed, dependencies, etc. Then discard what we don't need. 

We also have the prepare run function

And during the first iterations of the autosubmit run, the log recovery process is being initializating etc which can also add noise

The delay of the growth is intended to not add all this noise without needing to add anything to the run itself that could alter the profiler result


In 4.1.16, I want to be as close as possible to 4.2 so we're comparing the same code (there is already some overhead in calculating the total number of edges...).

autosubmit/profiler/profiler.py
dbeltrankyl: I think this is only useful for the profiler, I can move it to a diferent file/function tho but not sure if it will be useful

autosubmit/profiler/profiler.py
VindeeR: ooohh okay that clarifies it than thanks for the explanation.

I'll keep it in mind. :raised_hands:

dbeltrankyl:
Added a test for the argparser and the docstring, thanks!

autosubmit/profiler/profiler.py
VindeeR: I was thinking it could eventually be used in a different places, especially now that we are implementing the metrics on AS, that is the main reason that I had to transfer to a `Utils` file.

But if we won't be using it elsewhere we could leave it there I believe, and in case we do add something that might make use of it we can always move later

dbeltrankyl: I'll create an issue to move it then, thanks!

VindeeR (approved):
Everything looks good to me and things make more sense now.

If the CI run complete successfully you can marge it.

Thanks for the hard work 😄

Issue #2415, opened 2025-06-11
EXPORT job variable


lauherrero:
### Description

* Autosubmit version: 4.1.14-conda
* Machine/VM/environment name (if applicable): HUB03 and oper01
* Experiment ID (if applicable): a9bh

Hi 😸 

I am using the EXPORT job variable, but it seems not to be working properly.  The file `/esarchive/autosubmit/a9bh/tmp/ASLOGS/submit_****.sh ` does not exist and my jobs are failing because my conda env is not activated.

My job list is:
```yml
JOBS:
  DOWNLOAD_ENAGAS_distribution:
    FILE:
      - templates/download.sh
      - templates/configuration_files/ENAGAS_distribution.yml
    PLATFORM: local
    FILE_TO_DOWNLOAD: ENAGAS_distribution
    RUNNING: chunk
    # DOWNLOAD_FLAG: "True"
    ORIGINAL_FILE_PATH: "%INPUTS.ENAGAS_DISTRIBUTION.ORIGINAL_FILE_PATH%"
    # DOWNLOAD_DELAY: "%INPUTS.ENAGAS_DISTRIBUTION.DELAY%"
    NOTIFY_ON: FAILED
    EXPORT: "source %PROJDIR%/platforms/local/env.sh"
```

and my `env.sh` looks like:
```bash
module purge
# Load modules and activate conda environment
module load Miniconda3
# shellcheck source=/dev/null
source activate base
conda deactivate
conda activate /esarchive/scratch/cpinero/conda_envs/PHENOMENA_v0.2.3

```

I don't know if this is a bug or if I am doing something wrong 🙏

dbeltrankyl:
Export should work after #2939 is merged

PR #2939, opened 2026-04-13
Concurrent submission

### LINKED ISSUE #2921: Improve submission code
### Problem

While working on #2894, to automate the tests, I need to fill a large database. At first, I thought that I just had to submit 1000+ jobs to fill all the stuff necessary and be closer to what would be the real scenario, but I noticed multiple issues: 

1) Local. PS platform jobs are being submitted sequentially. I think that improving this can greatly improve the transfer jobs of DestinE
2) Batch submission for a large number of jobs is hammering the scheduler of the localDocker container
3) The submission exceeded the timeout set to 60 seconds
4) When the submission exceeded the timeout, it didn't exit ( The idea is to remove the timeout, or use the bash timeout )
5) The code quality of the submission is very bad as it has been hotfixed a lot of times, having ifs to do different stuff depending the platform etc
6) The export parameter doesn't work as expected, I think ( not confirmed)

### Proposed Solution

Refactor it so all platforms follow the same logic, submitting in batch, and try to add some parallelism to the script submission, but I'm not 100% sure of it being the right choice at the moment of writting this, as then I won't know which job.id match which submission, so there is an option of moving the parallelism to the python instead of the bash command

I'm working on this already, and I expect to have it for early next week ~~ ( considering the holidays)

cc: @VindeeR, as this will impact Miyabi, and hopefully it will be easier to debug errors like #2900

dbeltrankyl:
This fixes #2921

**Check List**

- [x] Documentation updated.


dbeltrankyl:
Oops, I was messing with copilot this weekend in a private project for investigation/study proporses and forgot to disable the auto-review, welp let's see what it tells

dbeltrankyl:
This is still a WIP, I'm missing to re-add the recovery mechanm against failed submission, and also I want to send all the files of a section at once

kinow:
> Oops, I was messing with copilot this weekend in a private project for investigation/study proporses and forgot to disable the auto-review, welp let's see what it tells

Curious to hear later how did it go. I think @LuiggiTenorioK is using it too for the API pull requests, but I haven't used it yet.

dbeltrankyl:
This affects to all our current platforms...

Unfortunately, I just spoke with @Lerriola, and we don't have access to PJM/Fugaku right now because it abruptly ended last week.

So I can't test it. In principle, the changes are the same for both platforms (slurm, pjm), but one never knows... in anycase in case of issues I'll fix it asap

dbeltrankyl:
Hello @kinow 

```
FAILED test/integration/provenance/test_rocrate.py::test_rocrate_validator - AssertionError: [CheckIssue(severity=Severity.REQUIRED, check=<rocrate_validator.requirements.shacl.checks.SHACLCheck object at 0x7fb1273dad50>, message=The RO-Crate metadata file descriptor MUST have a `conformsTo` property with the RO-Crate specification version)]
assert not True
 +  where True = has_issues()
 +    where has_issues = ValidationResult(passed=False,issues=[CheckIssue(severity=Severity.REQUIRED, check=<rocrate_validator.requirements.shacl.checks.SHACLCheck object at 0x7fb1273dad50>, message=The RO-Crate metadata file descriptor MUST have a `conformsTo` property with the RO-Crate specification version)]).has_issues
= 1 failed, 419 passed, 18 skipped, 3 xfailed, 59917 warnings in 472.60s (0:07:52) =
Error: Process completed with exit code 1.
```

This test passes locally but not in CI/CD. Do you know if it's a general thing or if it's something in my branch that is affecting it ( I can't think of anything tho)? 🤔 

thanks!

kinow:
Hi @dbeltrankyl , can you try to reinstall rocratepy locally? I think they released a new version today, and it could be causing this test to fail.

dbeltrankyl:
Regarding PJM, I was trying to follow this guide https://www.r-ccs.riken.jp/assets/uploads/2025/04/Private_Fugaku_Installation_Guide_en.pdf to do a local installation for test (since we don't currently have Fugaku access) but https://virtual.fugaku.r-ccs.riken.jp/container/ doesn't load 😓

dbeltrankyl:
Regarding sending files in parallel, I've discarded it since the Paramiko FTP channel doesn't support mput, anyway, I think it is acceptable fast already

I'll check the coverage and check that X11 is working before asking for a review

dbeltrankyl:
image

image


Randomly getting canceled during the installation of the dependencies, not running the tests, I guess that there are some tmp issues somewhere, not related to AS itself

dbeltrankyl:
Hello @kinow, @LuiggiTenorioK , @VindeeR , @ntorqulu 

I think this is ready for review!

I'm tagging everyone, as this changes the submission flow, it is worth taking a look, but since this will affect Miyabi #2668 I'll assign @VindeeR to review it 

Motivation is mainly to:

- Improve the readability so all platforms follow the same flow (even ecaccess that has some limitations)
- Improve performance 
- Improve the scalability of the code when adding new platforms

These features: 

- All platforms now follow the same flow for the submission process, which includes timeout, export, X11(mn5), and X11_options(mn5) working on all platforms that support it 
- Concurrent submission for all platforms by section. (so all sim jobs submitted at once, then all DA, etc ) 
- Packager class standardized, ARRAY packages removed ( not used)
- Error checks are now unified in a function 
- A lot of tests 
- _do_submission and other single-submission functions have been removed 


**New flow which must be followed by new platforms** 


`autosubmit run/inspect/-cw flag -> autosubmit.submit_ready_jobs -> platform.prepare_submission`
* `-> (inspect, cw only) platform.prepare_dry_run`
* `-> (inspect, run) package.generate_scripts`
* `-> (run) -> send_files`

`
autosubmit run/inspect/-cw flag -> autosubmit.submit_ready_jobs -> process_ready_jobs(grouped_by_platform_section//seq(x11 jobs) )
-> platform.submit_multiple_jobs`

* `-> platform.get_multi_submit_cmd -> get_call ( includes all the export timeout x11... parameters) -> _construct_final_call`
* `-> platform.send_command`
* `-> platform.get_submitted_job_id`
If that doesn't work, or the amount of retrieved job.id doesn't match the length of packages, the fallback is triggered (platform.get_submitted_jobs_by_name)

If that is still getting weird outputs, all the platform-submitted jobs will be canceled, and an AutosubmitError will be raised, so it restores the connection and tries again

After that, if squeue ( or equivalents) reports the same job twice, the oldest one will be canceled (platform._check_and_cancel_duplicated_job_names) ( same behavior as the current code)


send_command has a new call to a function that checks in order:

-> right output (continues)
-> transient errors (ssh timeout, etc.) -> raise AutosubmitError
-> unrecoverable errors (ex. wrong configuration) -> raise AutosubmitCritical



Warnings: 

Some stuff may appear to be uncovered, but it is covered by the integration test of the ecplatform, which only I have a user 

`    def _snapshot_job_ids_before_submission(self, script_names: list[str]) -> None:
`
This function is used in particular because I noticed that the ecaccess-submit API sometimes doesn't work as expected. 

A job could be silently not submitted, but other jobs in the same script sent yes. This platform doesn't have a sacct equivalent, and the timezone is a bit odd ( it mismatches the hour but also the minutes), so I've just added a snapshot of what the output was before submitting the command to find the jobs that weren't submitted

autosubmit/job/job.py
```diff
@@ -2950,6 +2940,14 @@ def recover_last_ready_date(self) -> None:
...
                     Log.debug(f"Failed to recover ready date for the job {self.name}")
 
+    def _calculate_executable(self):
+        """Calculate the executable for the job based on the platform type and job type.
+         This setting only makes sense when the remote platform is not being used within a scheduler.
+        """
+        if self.platform.type.upper() in ["PS", "LOCAL"]:
+            self.executable = self.executable if self.executable else Language.get_executable(self.type)
+        else:
+            self.executable = None
```
VindeeR: Since I can't make a 1 to 1 on this change, I just want to clear it, before we were getting the parameter from the `.yaml` file `EXECUTABLE` variable but now it is getting from the `TYPE` of the job and getting with format of executable it will be.

I understand it was intended but were the values expected from `EXECUTABLE` the same ones that the ones from `TYPE` and are the users aware of this change?

autosubmit/job/job_packages.py
```diff
@@ -154,58 +173,39 @@ def build_scripts(self, configuration: 'AutosubmitConfig') -> None:
...
+            job.update_local_logs()
+            # Clean previous run logs on local
+            log_completed = os.path.join(self._tmp_path, job.name + '_COMPLETED')
+            log_stat = os.path.join(self._tmp_path, job.name + '_STAT')
+            if Path(log_completed).exists():
+                os.remove(log_completed)
+            if Path(log_stat).exists():
+                os.remove(log_stat)
+            self.platform.remove_stat_file(job)
+            self.platform.remove_completed_file(job.name)
```
VindeeR:
```suggestion
            log_completed = Path(f"{self._tmp_path}/{job.name}_COMPLETED'")
            log_stat = Path(f"{self._tmp_path}/{job.name}_STAT'")
            if log_completed.exists():
                log_completed.unlink(missing_ok=True)
            if log_stat.exists():
                log_stat.unlink(missing_ok=True)
            self.platform.remove_stat_file(job)
            self.platform.remove_completed_file(job.name)
```

autosubmit/job/job_utils.py
```diff
@@ -480,15 +480,21 @@ def cancel_jobs(job_list: "JobList", active_jobs_filter=None, target_status=Opti
...
+            platform.cancel_jobs(job_ids)
         except Exception as e:
-            Log.warning(f"Failed to cancel job {job.name} on platform {job.platform.name}: {str(e)}")
+            Log.warning(f"Failed to cancel jobs {', '.join(job_ids)} on platform {platform.name}: {str(e)}")
 
-        Log.info(f"Changing status of job {job.name} to {target_status}")
-        job.status = Status.KEY_TO_VALUE[target_status]
+        for job in jobs:
+            Log.info(f"Changing status of job {job.name} to {target_status}")
+            job.status = Status.KEY_TO_VALUE[target_status]
```
VindeeR, line 498: If the jobs are being canceled shouldn't they be converted to the 'FAILED' status always?

I'm wondering because it seems like the code would allow any status as long as it is valid, wouldn't be better to set it as canceled already, instead of doing it in a roundabout way?

autosubmit/platforms/ecplatform.py
```diff
@@ -119,18 +133,33 @@ def update_cmds(self):
...
     def get_mkdir_cmd(self):
         return self.mkdir_cmd
 
-    def set_submit_cmd(self, ec_queue):
-        self._submit_cmd = ("ecaccess-job-submit -distant -queueName " + ec_queue + " " + self.host + ":" +
-                            self.remote_log_dir + "/")
+    def _set_submit_cmd(self, ec_queue: str):
+        """Sets the command to submit a job, for the current platform, with the given parameters."""
+        # Sometimes ecaccess-job-submit doesn't submit anything due the SSL connection being unstable
+        # Don't confuse retry by the amount of retries of the job itself.
```
VindeeR: This sentence is not very clear. 

Which retry are you talking about?

I'm not sure what you mean by that, did you mean to say `"there's a difference between the retries for submitting a command, and the retries of the submit of a job"` is it right?

autosubmit/job/job_list.py
```diff
@@ -2849,20 +2849,17 @@ def update_genealogy(self):
         except Exception as exp:
             Log.warning(str(exp))
 
-    def save_wrappers(self, packages_to_save, failed_packages, as_conf, packages_persistence,
+    def save_wrappers(self, submitted_scripts, as_conf, packages_persistence,
```
VindeeR: Could you please add the `Docstring` to this function?

autosubmit/platforms/ecplatform.py
```diff
@@ -144,20 +173,18 @@ def parse_job_output(self, output):
...
-        self.set_submit_cmd(job.ec_queue)
-        if export == "none" or export == "None" or export is None or export == "":
-            export = ""
-        else:
-            export += " ; "
-        return export + self._submit_cmd + job_script
-
     def connect(self, as_conf: 'AutosubmitConfig', reconnect: bool = False, log_recovery_process: bool = False) -> None:
         """
         Establishes an SSH connection to the host.
```
VindeeR:
```suggestion
"""Establishes an SSH connection to the host.
```

autosubmit/platforms/ecplatform.py
```diff
@@ -391,3 +420,147 @@ def get_completed_job_names(self, job_names: Optional[list[str]] = None) -> list
...
+        if not job_ids:
+            return
+        command = " ; ".join(f"{self.cancel_cmd} {job_id}" for job_id in job_ids)
+        self.send_command(command)
+
+    def _check_and_cancel_duplicated_job_names(self, scripts_to_submit: dict) -> None:
+        """Check for duplicated job names in the submitted packages.
+        :param scripts_to_submit: Package script names and their info.
+        :type: dict
+        """
```
VindeeR, line 451:
```suggestion
        """Check for duplicated job names in the submitted packages.

        :param scripts_to_submit: Package script names and their info.
        :type scripts_to_submit: dict
        """
```

autosubmit/platforms/ecplatform.py
```diff
@@ -391,3 +420,147 @@ def get_completed_job_names(self, job_names: Optional[list[str]] = None) -> list
...
+        stems = {Path(name).stem for name in script_names}
+        with suppress(Exception):
+            self.send_command("ecaccess-job-list")
+            output = self.get_ssh_output()
+            pre: dict[str, set[int]] = {}
+            for line in output.splitlines():
+                parts = line.split()
+                if len(parts) >= 2 and parts[0].isdigit() and parts[-1] in stems:
+                    pre.setdefault(parts[-1], set()).add(int(parts[0]))
+            self._pre_submission_ids = pre
```
VindeeR, line 532: Since I'm not confident I understand the return values, I can't be sure if it work as intended, maybe asking someone else to double check this? Or you can explain to me what are the output of the `ec platform` for this scenario

autosubmit/platforms/locplatform.py
```diff
@@ -92,21 +87,13 @@ def get_mkdir_cmd(self):
...
-        else:
-            export += " ; "
-        command = self.get_call(job_script, job, export=export, timeout=seconds)
-        return f"cd {self.remote_log_dir} ; {command}"
+    def get_submitted_job_id(self, raw_output: str, x11: bool = False) -> list[str]:
+        """Parses the output of the submit command to get the job ID.
+        :param raw_output: output of the submit command.
+        :param x11: whether the job is an x11 job, which has a different output format.
+        :return: job ID of the submitted job.
+        """
```
VindeeR, line 96: I believe x11 is not being used in here despite having a different format no validation, or other treatment is done.

should it be removed?

```suggestion
        """Parses the output of the submit command to get the job ID.

        :param raw_output: output of the submit command.
        :type raw_output:
        :param x11: whether the job is an x11 job, which has a different output format.
        :type x11:
        :return: job ID of the submitted job.
```

autosubmit/platforms/locplatform.py
```diff
@@ -385,3 +373,28 @@ def compress_file(self, file_path: str) -> None:
         except Exception as exc:
             Log.error(f"Error compressing file {file_path}: {exc}")
         return None
+
+    def cancel_jobs(self, job_ids: list[str]) -> None:
+        """Cancel local processes by their PIDs.
+        :param job_ids: List of local process IDs to cancel.
+        :type job_ids: list[str]
+        """
```
VindeeR, line 383:
```suggestion
        """Cancel local processes by their PIDs.

        :param job_ids: List of local process IDs to cancel.
        :type job_ids: list[str]
        """
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -735,13 +704,54 @@ def check_job_energy(self, job_id):
...
+
+        :param script_names: Script filenames to submit on the remote
+            platform.
+        :type script_names: list[str]
+        :return: Submitted Slurm job identifiers in submission order.
+        :rtype: list[int]
+        :raises AutosubmitError: If the submission command fails or no submit
+            output can be parsed.
+        :raises AutosubmitCritical: If Slurm reports a critical submission
+            failure.
```
VindeeR: raises comes first, sorry for the nitpick 
```suggestion
        :raises AutosubmitError: If the submission command fails or no submit
            output can be parsed.
        :raises AutosubmitCritical: If Slurm reports a critical submission
            failure.
        :return: Submitted Slurm job identifiers in submission order.
        :rtype: list[int]
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -735,13 +704,54 @@ def check_job_energy(self, job_id):
...
+        """
+
```
VindeeR, line 723:
```suggestion
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1247,12 +1257,7 @@ def send_command(self, command: str, ignore_log=False, x11=False) -> bool:
         :return: True if executed, False if failed
         """
         lang = locale.getlocale()[1] or locale.getdefaultlocale()[1] or 'UTF-8'
-        if "rsync" in command or "find" in command or "convertLink" in command:
-            timeout = None  # infinite timeout on migrate command  # TODO: still needed?
-        elif "rm" in command:
-            timeout = 60
-        else:
-            timeout = 60 * 2
+        timeout = None
```
VindeeR: Isn't it better to delete this variable if it will be None?
I don't think it makes sense to keep this or the log under it if none is always `None` now.

There're 2 lines under this that uses timeout, maybe we can change that too.

Log.debug(f"send_command timeout used: {timeout} seconds (None = infinity)")

channel.settimeout(timeout)

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1427,26 +1368,27 @@ def parse_all_jobs_output(self, output, job_id):
...
-        :param job_script: path to job script
-        :param job_script: str
-        :param hold: submit a job in a held status
-        :param hold: boolean
-        :param export: modules that should've downloaded
-        :param export: string
-        :return: command to submit job to platforms
+    def get_multi_submit_cmd(self, job_scripts: dict) -> str:
+        """Gets command to submit all the current active jobs on HPC
+        :param job_scripts: dict of job names and their info (export, x11_options)
```
VindeeR, line 1446:
```suggestion
        """Gets command to submit all the current active jobs on HPC

        :param job_scripts: dict of job names and their info (export, x11_options)
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1472,38 +1414,40 @@ def get_ssh_output(self):
...
+        :type post: str
+        :param x11_options: x11 options to run the script, if any
+        :type x11_options: str
+        :return: command to submit a job
+        """
+        return f"{pre} {x11_options} {script_name} {post}" if x11_options else f"nohup {pre} {script_name} {post} & echo $!"
+
+    def get_call(self, script_name, timeout: float, export: str, executable: str, x11_options: str, fail_count: int, sub_queue: str, redirect_out_err: bool = False) -> str:
+        """Gets execution command for given job. it builds the command to execute the script on the remote platform.
+        :param script_name: script to run
```
VindeeR, line 1509:
```suggestion
        """Gets execution command for given job. it builds the command to execute the script on the remote platform.

        :param script_name: script to run
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1472,38 +1414,40 @@ def get_ssh_output(self):
...
+        :param timeout: timeout for the execution
+        :param export: export command to set environment variables
+        :param executable: executable to run the script
+        :param x11_options: x11 options to run the script
+        :param fail_count: number of times the job has failed, used to name the output and error files
+        :param sub_queue: alternative queue to run the job, used for some platforms like ECaccess
+        :param redirect_out_err: whether to redirect stdout and stderr to files
         :return: command to execute script
```
VindeeR, line 1517:
```suggestion
        :return: command to execute script
        :rtype: str
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1472,38 +1414,40 @@ def get_ssh_output(self):
...
-                                                         f'{os.path.join(self.remote_log_dir, remote_logs[1])} & '
-                                                         f'echo $!')
-        else:
-            command = (export + f"timeout {timeout}" + ' nohup ' + executable +
-                       f' {os.path.join(self.remote_log_dir, job_script)} > '
-                       f'{os.path.join(self.remote_log_dir, remote_logs[0])} 2> '
-                       f'{os.path.join(self.remote_log_dir, remote_logs[1])} & echo $!')
-        return command
+        # Some platforms (right now only ECaccess) has a dual queue system, one to run the job and another to submit the job.
+        self._set_submit_cmd(sub_queue)
```
VindeeR, line 1521: wouldn't make more sense to implement it only at the `ECaccess` level?
This way this function wouldn't transpire to other classes to which it wouldn't have any effect

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1516,14 +1460,11 @@ def get_pscall(job_id):
...
-        :param x11:
-        :param output: output to parse
-        :type output: str
-        :return: job id
-        :rtype: str
+    def get_submitted_job_id(self, output: str, x11: bool = False) -> list[str]:
+        """Parses the output of the submit command to get the job ID.
+        :param output: output of the submit command.
+        :param x11: whether the job is an x11 job, which has a different output format.
+        :return: job ID of the submitted job.
```
VindeeR, line 1544:
```suggestion
        """Parses the output of the submit command to get the job ID.

        :param output: output of the submit command.
        :type output: str
        :param x11: whether the job is an x11 job, which has a different output format.
        :type x11: bool
        :return: job ID of the submitted job.
        :rtype: list[str]
```

autosubmit/platforms/ecplatform.py
```diff
@@ -144,20 +173,18 @@ def parse_job_output(self, output):
                 return job_state[1]
         return 'DONE'
 
-    def get_submitted_job_id(self, output, x11=False):
-        return output
+    def get_submitted_job_id(self, output: str, x11: bool = False) -> list[str]:
```
VindeeR, line 187:
```suggestion
    @overload
    def get_submitted_job_id(self, output: str) -> list[str]:
```
I believe x11 is not being used in here despite having a different format no validation, or other treatment is done.

shouldn't it be removed with overload?

autosubmit/platforms/psplatform.py
```diff
@@ -84,15 +79,13 @@ def get_mkdir_cmd(self):
...
-            export = ""
-        else:
-            export += " ; "
-        return self.get_call(job_script, job, export=export, timeout=job.wallclock_in_seconds)
+    def get_submitted_job_id(self, raw_output: str, x11: bool = False) -> list[str]:
+        """Parses the output of the submit command to get the job ID.
+        :param raw_output: output of the submit command.
+        :param x11: whether the job is an x11 job, which has a different output format.
+        :return: job ID of the submitted job.
+        """
```
VindeeR, line 89:
```suggestion
    @overload
    def get_submitted_job_id(self, raw_output: str) -> list[str]:
        """Parses the output of the submit command to get the job ID.

        :param raw_output: output of the submit command.
        :type raw_output:
        :return: job ID of the submitted job.
        :rtype: list[str]
        """
```
I believe x11 is not being used in here despite having a different format no validation, or other treatment is done.

shouldn't it be removed with overload?

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1742,6 +1683,93 @@ def update_cmds(self):
         """Updates commands for this platform. """
         pass  # pragma: no cover
 
+    def _check_and_cancel_duplicated_job_names(self, scripts_to_submit: dict) -> None:
+        """Check for duplicated job names in the submitted packages.
+        :param scripts_to_submit: Package script names and their info.
+        :type: dict
+        """
```
VindeeR, line 1767:
```suggestion
    def _check_and_cancel_duplicated_job_names(self, scripts_to_submit: dict) -> None:
        """Check for duplicated job names in the submitted packages.

        :param scripts_to_submit: Package script names and their info.
        :type scripts_to_submit: dict
        """
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1742,6 +1683,93 @@ def update_cmds(self):
...
+                parsed_job_names[job_name.strip()] = [ids]
+
+        return parsed_job_names
+
+    def _get_job_names_cmd(self, job_names: list) -> str:
+        """Return a command that groups job IDs by job name.
+
+        The command returns one line per job name using the format
+        ``JobName:id,id2,id3``.
+        """
```
VindeeR, line 1815:
```suggestion
        """Return a command that groups job IDs by job name.

        The command returns one line per job name using the format
        ``JobName:id,id2,id3``.

        :param job_names:
        :type job_names: list
        :return:
        :rtype: str
        """
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1742,6 +1683,93 @@ def update_cmds(self):
...
+        raise NotImplementedError  # pragma: no cover
+
+    def cancel_jobs(self, job_ids: list[str]) -> None:
+        """Cancel jobs with given job ids.
+
+        :param job_ids: List of job ids to cancel
+        :type job_ids: list
+
```
VindeeR, line 1823:
```suggestion
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1742,6 +1683,93 @@ def update_cmds(self):
...
+        raise NotImplementedError  # pragma: no cover
+
+    def process_ready_jobs(self, scripts_to_submit: dict[str, 'JobPackageBase']) -> tuple[bool, list[Any]]:
+        """Retrieve multiple jobs identifiers.
+
+        :param scripts_to_submit: List of valid Job Packages to be processes
+        :type scripts_to_submit: List[Any]
+
+        :return: retrieve the ID of the Jobs
+        :rtype: tuple[bool, list[Any]]
```
VindeeR, line 1835:
```suggestion
        :param scripts_to_submit: List of valid Job Packages to be processes
        :type scripts_to_submit: dict[str, 'JobPackageBase']
        :return: retrieve the ID of the Jobs
        :rtype: tuple[bool, list[Any]]
```

dbeltrankyl:
closes #2415

autosubmit/platforms/pjmplatform.py
```diff
@@ -480,3 +331,138 @@ def check_file_exists(self, filename: str, wrapper_failed: bool = False, sleepti
...
+        :param script_names: Submitted script filenames.
+        :type script_names: list[str]
+        :return: Matching PJM job IDs in the same order as ``script_names``.
+        :rtype: list[int]
+        """
+        submitted_job_ids: list[int] = []
+
+        for script_name in script_names:
+            job_name = Path(script_name).stem
+            command = f'pjstat -v --choose jid,jnam --filter "jnam={job_name}"'
```
VindeeR, line 350: I was wondering if the filter is supposed to be `jnam` I notice you repeat it later on so I believe it is right but just for piece of mind, I would like to double check

autosubmit/platforms/platform.py
```diff
@@ -367,103 +367,95 @@ def root_dir(self):
...
+        :type inspect: bool
+        :param only_wrappers: If ``True``, prepare only wrapper-related metadata
+            and skip the regular package submission flow.
+        :type only_wrappers: bool
+        :return: A list containing the jobs gathered while preparing the given
+            packages for submission.
+        :rtype: list
+        :raises Exception: Propagate any exception raised while preparing
+            packages, generating scripts, or transferring files.
         """
```
VindeeR, line 404:
```suggestion
        :raises Exception: Propagate any exception raised while preparing
            packages, generating scripts, or transferring files.
        :return: A list containing the jobs gathered while preparing the given
            packages for submission.
        :rtype: list
       """
```

autosubmit/platforms/platform.py
```diff
@@ -367,103 +367,95 @@ def root_dir(self):
...
+        :type only_wrappers: bool
+        :param inspect: If ``True``, prepare package metadata for inspect mode.
+        :type inspect: bool
+        :param packages_persistence: Persistence helper used to store package
+            data for later recovery or visualization.
+        :type packages_persistence: JobPackagePersistence
+        :param as_conf: Autosubmit configuration for the current experiment.
+        :type as_conf: AutosubmitConfig
+        :return: None
+        :rtype: None
```
VindeeR:
```suggestion
```

autosubmit/platforms/platform.py
```diff
@@ -367,103 +367,95 @@ def root_dir(self):
...
+        :raises Exception: Propagate any exception raised while creating wrapper
+            job metadata or saving the package.
+        """
+        if only_wrappers or inspect:
+            # Now name is used for submit scripts, before this was used to determine if a package has a wrapper or not
```
VindeeR:
```suggestion
            # Now name is used for submit scripts, before it was used to determine if a package had a wrapper or not
```

autosubmit/platforms/platform.py
```diff
@@ -749,27 +740,11 @@ def get_files_path(self):
...
-        :type hold: bool
-        :param export: export environment variables
-        :type export: str
-        :return: job id for the submitted job
-        :rtype: int
-        """
-        raise NotImplementedError  # pragma: no cover
-
-    def check_all_jobs(self, job_list: list['Job'], as_conf:'AutosubmitConfig', retries: int = 5):
+    def check_all_jobs(self, job_list: list['Job'], as_conf: 'AutosubmitConfig', retries: int = 5):
```
VindeeR:
```suggestion
    def check_all_jobs(self, job_list: list['Job'], as_conf: 'AutosubmitConfig'):
```

autosubmit/platforms/psplatform.py
```diff
@@ -114,3 +107,77 @@ def check_remote_permissions(self) -> bool:
...
+        pids = " ".join(str(job_id) for job_id in job_ids)
+        self.send_command(f"{self.cancel_cmd} {pids}")
+
+    def _get_job_names_cmd(self, job_names: list) -> str:
+        """Gets command to check for duplicated job names on remote platforms (UNIX)
+        It receives a list of job names, and returns a command that:
+
+        1) Checks if there is a job name already present on the remote platform.
+        2) If there any, it returns the oldest ID ( process or job_id) entry for each duplicated job name, separated by commas.
+        """
```
VindeeR, line 135:
```suggestion
        2) If there any, it returns the oldest ID ( process or job_id) entry for each duplicated job name, separated by commas.

        :param job_names:
        :type job_names: list
        :return:
        :rtype: str
        """
```

autosubmit/platforms/slurmplatform.py
```diff
@@ -371,54 +144,16 @@ def update_cmds(self) -> None:
...
-        except BaseException as e:
-            try:
-                self.send_command(f"scancel {job.id}")
-                raise AutosubmitError(f"Can't hold jobid:{job.id}, canceling job", 6000, str(e)) from e
-            except BaseException as err:
-                raise AutosubmitError(f"Can't cancel the jobid: {job.id}", 6000, str(err)) from err
-            except AutosubmitError as as_err:
-                raise as_err
+    def _construct_final_call(self, script_name: str, pre: str, post: str, x11_options: str):
+        return f"cd {self.remote_log_dir} ; {pre} salloc {x11_options} {script_name} {post}" if x11_options else f"{pre} {self._submit_cmd} {script_name} {post}"
```
VindeeR, line 165:
```suggestion
    def _construct_final_call(self, script_name: str, pre: str, post: str, x11_options: str) -> str:
    """Docstring"""
        return f"cd {self.remote_log_dir} ; {pre} salloc {x11_options} {script_name} {post}" if x11_options else f"{pre} {self._submit_cmd} {script_name} {post}"
```

autosubmit/platforms/slurmplatform.py
```diff
@@ -458,43 +193,48 @@ def parse_all_jobs_output(self, output: str, job_id: int) -> Union[list[str], st
...
 
-    def get_submitted_job_id(self, output: str, x11: bool = False) -> Union[list[int], int]:
+    def get_submitted_job_id(self, output: str, x11: bool = False) -> list[str]:
+        """Parses the output of the submit command to get the job ID.
+        :param output: output of the submit command.
+        :param x11: whether the job is an x11 job, which has a different output format.
+        :return: job ID of the submitted job.
+        """
+        jobs_id = []
         try:
```
VindeeR, line 213:
```suggestion
    def get_submitted_job_id(self, output: str, x11: bool = False) -> list[str]:
        """Parses the output of the submit command to get the job ID.

        :param output: output of the submit command.
        :type output: str
        :param x11: whether the job is an x11 job, which has a different output format.
        :type x11: bool
        :return: job ID of the submitted job.
        :rtype: list[str]
        """
        jobs_id = []
        try:
```

autosubmit/platforms/slurmplatform.py
```diff
@@ -458,43 +193,48 @@ def parse_all_jobs_output(self, output: str, job_id: int) -> Union[list[str], st
...
+        :type script_names: list[str]
+        :return: Matching Slurm job IDs in submission order.
+        :rtype: list[int]
+        """
+        submitted_job_ids: list[int] = []
+
+        for script_name in script_names:
+            job_name = Path(script_name).stem
+
+            matched_ids = [int(job_id) for job_id in self.get_jobid_by_jobname(job_name)]
```
VindeeR:
```suggestion
            matched_ids = [int(job_id) for job_id in self.get_job_id_by_job_name(job_name)]
```

autosubmit/platforms/slurmplatform.py
VindeeR:
```suggestion
    def get_job_id_by_job_name_cmd(self, job_name: str) -> str:
```

test/unit/test_slurm_parallel_submission.py
```diff
@@ -0,0 +1,226 @@
...
+
+    :param slurm_platform: Slurm platform under test.
+    :param monkeypatch: Pytest monkeypatch fixture.
+    """
+    job_ids_by_name = {
+        "job_a": ["101", "103"],
+        "job_b": ["202"],
+    }
+
+    def _get_jobid_by_jobname(job_name: str) -> list[str]:
```
VindeeR:
```suggestion
    def _get_job_id_by_job_name(job_name: str) -> list[str]:
```

test/unit/test_slurm_parallel_submission.py
```diff
@@ -0,0 +1,226 @@
...
+        return job_ids_by_name[job_name]
+
+    monkeypatch.setattr(slurm_platform, "get_jobid_by_jobname", _get_jobid_by_jobname)
```
VindeeR:
```suggestion
    monkeypatch.setattr(slurm_platform, "get_job_id_by_job_name", _get_job_id_by_job_name)
```

test/unit/test_slurm_parallel_submission.py
```diff
@@ -0,0 +1,226 @@
...
+
+    :param slurm_platform: Slurm platform under test.
+    :param monkeypatch: Pytest monkeypatch fixture.
+    """
+    job_ids_by_name = {
+        "job_a": ["101"],
+        "job_b": [],
+    }
+
+    def _get_jobid_by_jobname(job_name: str) -> list[str]:
```
VindeeR:
```suggestion
    def _get_job_id_by_job_name(job_name: str) -> list[str]:
```

test/unit/test_slurm_parallel_submission.py
```diff
@@ -0,0 +1,226 @@
...
+        return job_ids_by_name[job_name]
+
+    monkeypatch.setattr(slurm_platform, "get_jobid_by_jobname", _get_jobid_by_jobname)
```
VindeeR:
```suggestion
    monkeypatch.setattr(slurm_platform, "get_job_id_by_job_name", _get_job_id_by_job_name)
```

autosubmit/autosubmit.py
```diff
@@ -2493,12 +2490,11 @@ def restore_platforms(platforms_to_test, mail_notify=False, as_conf=None, expid=
                                          7010, issues + "\n" + ssh_config_issues)
 
     @staticmethod
-    def submit_ready_jobs(as_conf: AutosubmitConfig, job_list: JobList, platforms_to_test: list[Platform],
+    def submit_ready_jobs(as_conf: AutosubmitConfig, job_list: JobList, platforms_to_test: list[ParamikoPlatform],
                           packages_persistence: JobPackagePersistence, inspect=False,
-                          only_wrappers=False, hold=False) -> bool:
+                          only_wrappers=False) -> None:
         """Gets READY jobs and send them to the platforms if there is available space on the queues.
```
VindeeR, line 2499: I think it would be nice to move this function to outside of the AS class in the utils.py.
this would make the code simpler to read and more straight forward.

wdyt?

autosubmit/autosubmit.py
```diff
@@ -2524,33 +2518,43 @@ def submit_ready_jobs(as_conf: AutosubmitConfig, job_list: JobList, platforms_to
...
-                job_list.save_wrappers(valid_packages_to_submit, failed_packages, as_conf, packages_persistence,
-                                       hold=hold, inspect=inspect)
-                if error_message != "":
-                    raise AutosubmitCritical(f"Submission Failed due wrong configuration:{error_message}", 7014)
-            Autosubmit.check_deadlock(wrapper_errors, any_job_submitted, job_list)
-            return save_1 or save_2
-        except Exception:
-            raise
+
+        for p in platforms_to_test:
```
VindeeR: I think if you keep this variable as `platform` the readability of the function improves a bit IMO.

```suggestion
        for platform in platforms_to_test:
```

autosubmit/autosubmit.py
```diff
@@ -2524,33 +2518,43 @@ def submit_ready_jobs(as_conf: AutosubmitConfig, job_list: JobList, platforms_to
...
+            packager = JobPackager(as_conf, p, job_list)
+            packages_to_submit = packager.build_packages()
+
+            scripts_to_submit_by_section, x11_scripts_to_submit_by_section = p.prepare_submission(as_conf, job_list, packages_persistence, packages_to_submit, inspect=inspect,
+                                                                                                  only_wrappers=only_wrappers)
+
```
VindeeR:
```suggestion
            scripts_to_submit_by_section, x11_scripts_to_submit_by_section = p.prepare_submission(as_conf, job_list, 
                                                                                                  packages_persistence, 
                                                                                                  packages_to_submit, 
                                                                                                  inspect=inspect,
                                                                                                  only_wrappers=only_wrappers)
```

test/unit/platforms/test_cancel_jobs.py
```diff
@@ -0,0 +1,211 @@
...
+        expected_command: str,
+) -> None:
+    """Send one comma-joined scancel command for multiple Slurm job IDs."""
+    sent: list[str] = []
+    monkeypatch.setattr(slurm_platform, "send_command", lambda cmd, **_: sent.append(cmd))
+
+    slurm_platform.cancel_jobs(job_ids)
+
+    assert len(sent) == 1
+    assert sent[0] == expected_command
```
VindeeR: I believe that the 3 jobs could be merged into 1 with minimal changes

test/unit/platforms/test_cancel_jobs.py
```diff
@@ -0,0 +1,211 @@
...
+        ec_platform: EcPlatform,
+        monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Send a single ecaccess-job-delete command for one job ID."""
+    sent: list[str] = []
+    monkeypatch.setattr(ec_platform, "send_command", lambda cmd, **_: sent.append(cmd))
+
+    ec_platform.cancel_jobs(["9001"])
+
+    assert sent == ["ecaccess-job-delete 9001"]
```
VindeeR: I believe these 2 jobs could be merged into 1 with minimal changes

test/unit/platforms/test_cancel_jobs.py
```diff
@@ -0,0 +1,211 @@
...
+        expected_command: str,
+) -> None:
+    """Send one space-joined kill command for multiple local PIDs."""
+    sent: list[str] = []
+    monkeypatch.setattr(local_platform, "send_command", lambda cmd, **_: sent.append(cmd))
+
+    local_platform.cancel_jobs(job_ids)
+
+    assert len(sent) == 1
+    assert sent[0] == expected_command
```
VindeeR: ditto

test/unit/platforms/test_cancel_jobs.py
```diff
@@ -0,0 +1,211 @@
...
+        expected_command: str,
+) -> None:
+    """Send one space-joined kill command for multiple remote PIDs."""
+    sent: list[str] = []
+    monkeypatch.setattr(ps_platform, "send_command", lambda cmd, **_: sent.append(cmd))
+
+    ps_platform.cancel_jobs(job_ids)
+
+    assert len(sent) == 1
+    assert sent[0] == expected_command
```
VindeeR: ditto

test/unit/test_check_for_unrecoverable_errors.py
VindeeR, line 1: Maybe move this file to under the `platform` folder since all these test are related to the treatment of outputs of the platform

test/unit/platforms/test_ecplatform.py
VindeeR, line 1: Maybe move this file to under the `platform` folder

test/unit/test_job_package.py
VindeeR, line 1: Maybe we should move this file under the `job` folder

test/unit/platforms/test_paramiko_platform.py
VindeeR, line 1: Move file to folder `platform`

test/unit/platforms/test_pjm.py
VindeeR, line 1: Move file to folder `platform`

test/unit/platforms/test_slurm_parallel_submission.py
VindeeR, line 1: Move file to folder `platform`

test/unit/platforms/test_slurm_platform.py
VindeeR, line 1: Move file to folder `platform`

autosubmit/platforms/ecplatform.py
```diff
@@ -144,20 +173,18 @@ def parse_job_output(self, output):
...
+        """Parses the output of the submit command to get the job ID.
+        :param output: output of the submit command.
+        :param x11: whether the job is an x11 job, which has a different output format.
+        :return: job ID of the submitted job.
```
VindeeR, line 192:
```suggestion
        """Parses the output of the submit command to get the job ID.

        :param output: output of the submit command.
        :type output: str
        :param x11: whether the job is an x11 job, which has a different output format.
        :type x11: bool
        :return: job ID of the submitted job.
        :rtype: list[str]
```

VindeeR (changes requested):
Sorry for the delay, I tried to catch as much as possible, that's why so many comments, but most things are pretty small docstring and etc, just a couple of them were changes related to quality of the code or doubts I had as I reviewed. 

I don't think there anything particular notorious in regard my comments, there were a few small things I was a bit confused but nothing that seemed like an issue.

Thanks for the hard work, the platforms are so much better and more organized, thanks for the improvement. 🙌

VindeeR (changes requested):
Sorry for the delay, I tried to catch as much as possible, that's why so many comments, but most things are pretty small docstring and etc, just a couple of them were changes related to quality of the code or doubts I had as I reviewed. 

I don't think there anything particular notorious in regard my comments, there were a few small things I was a bit confused but nothing that seemed like an issue.

Thanks for the hard work, the platforms are so much better and more organized, thanks for the improvement. 🙌

autosubmit/job/job.py
dbeltrankyl: Hello @VindeeR 

The behavior didn't change; this is more of a bug fix for the executable because it was not working properly for a while

What this does is that when it is a platform without a scheduler, like ps and local ( I'll add a new variable to avoid this if type), it changes the command for launching the template

per example python3 something -> python3custom something 

For the user, this is transparent

Having said that, I'm thinking that this should also change the first line of the templates (shebang) for jobs inside a scheduler, so... I'll address that

autosubmit/job/job_packages.py
dbeltrankyl: Thanks! I just moved code what we had and forgot to upgrade it

autosubmit/job/job_utils.py
dbeltrankyl: A job is only considered failed if it was in running status previously 

So, if it were, let's say, QUEUING, when you cancel it, the proper status should be WAITING

If a Job fails, it means that the template processed something.

autosubmit/platforms/ecplatform.py
dbeltrankyl: > This sentence is not very clear.


Yes, I'll commit a better explanation 

https://confluence.ecmwf.int/display/ECAC/ecaccess-job-submit 
`-retry count`


`Number of SSL connection retries per 5s to ECMWF. This parameter only apply to the initial SSL connection initiated by the command to the ECMWF server. It does not apply to all the subsequent requests made afteward as it is mainly targeting errors that can happen from time to time during the SSL handshake. Default is no retry.`

The -retry of the command, is not the retry parameter of the job

dbeltrankyl: What happens in this platform is that SSL connection sometimes doesn't go through and the job is not submitted at all the -retry is to retry only that action

autosubmit/platforms/ecplatform.py
dbeltrankyl: Overloads are for different call signatures, but this function is always called as p.get_submitted_job_id(out,x11)

and    def get_submitted_job_id(self, output: str) -> list[str]:

would violate 
https://en.wikipedia.org/wiki/Liskov_substitution_principle#Principle 

If your base class defines all the args, then all the subclasses must accept the same arguments, even if they're not used. The x11 is optional due that

dbeltrankyl: (also the overload in python is only used for typing, I believe as python always accepts it)

autosubmit/platforms/ecplatform.py
dbeltrankyl: The output is  this 
Image 


It gathers all the existing job_ids before trying to submit new ones. After that, check which jobs haven't been submitted when the ecaccess-job-submit fails due to the SSL stuff I commented on the other 

( I noticed a bug tho, it stems the script_name when it shouldn't, as the output is with the .cmd, fixing that) 

Then it fills a dict with 

[script_name] = [job_id,...job_id2...]

autosubmit/platforms/locplatform.py
dbeltrankyl: same as ECplatform

autosubmit/platforms/paramiko_platform.py
dbeltrankyl: Yeah.. well I removed that earlier this PR, and now I'm considering keeping it, but to a really high value in case it totally freezes Autosubmit for some reason

At the same time, during some testing, I noticed that the timeout, wrongly set to 60, was not finishing the remote command at all (atleast during the submission, other commands may work?), but it may work with other Paramiko versions 🤔 

what do you think?

autosubmit/platforms/paramiko_platform.py
dbeltrankyl: I put it in the paramiko_platform level in case we need it for other platforms in the future. After all, this is the flow that new platforms should follow, so it is good to have this already placed here


The other issue is that if I do what you propose, then I would have to override this function with redundant code just not to call the function, so it would be harder to maintain  

The other solution is to add the if self.type... clause, but the changes here are motivated by a desire to avoid these kind of if's and to remove duplicate code... So I think it makes sense to leave it as it is now

autosubmit/platforms/pjmplatform.py
dbeltrankyl: I'll double check.. I don't have access to Fugaku right now so it is not tested manually 😅

autosubmit/platforms/platform.py
dbeltrankyl: Does it matter tho? Isn't it right both ways? I can change it tho

autosubmit/platforms/platform.py
dbeltrankyl: Ok in this case is correct, because then the subclasses will add additional behaviour which won't end with a runtime error

dbeltrankyl: Although this shouldn't have code at all in platform.py

dbeltrankyl: Anyway, I think I'll leave the changes regarding this to a minimum because in 4.2.0 there is a rework...

autosubmit/platforms/psplatform.py
dbeltrankyl: same as other comment

autosubmit/autosubmit.py
dbeltrankyl: I don't think this belongs to utils.py, but rather to something like 

runtime/common.py
runtime/run.py
runtime/inspect.py
runtime/check_wrappers.py 

We need a separate PR to refactor all the autosubmit.run_experiment function for that, I believe

autosubmit/autosubmit.py
dbeltrankyl: I had a shadow warning when it was named like that, but we can choose another name. Any suggestions?

dbeltrankyl (commented):
Thanks for the feedback! , I did a first commit with the docstrings etc.. I'll address the two bugs I saw and the test part next

autosubmit/job/job_utils.py
VindeeR: Maybe we can rewrite either the name of this function and/or the docstring because I think it is not very clear this behavior.

autosubmit/platforms/ecplatform.py
VindeeR: Thanks it is much clearer now thanks :raised_hands:

autosubmit/platforms/ecplatform.py
VindeeR: Okay this behavior makes sense to me now, and thanks for noticing the bug :raised_hands: 
Once this is fixed, I think it should be fine to marge.

autosubmit/platforms/paramiko_platform.py
VindeeR: In my opinion, It’s probably best to remove it. Even if we reintroduce a timeout with a high value, we still can’t guarantee consistent behavior over the lifetime of the connection. That means we wouldn’t be able to tell whether the connection is frozen, still processing, or just slow because the platform’s resources are tied up elsewhere.

Unless we are able to identify and handle the above points I believe removing is the best solution.

autosubmit/platforms/paramiko_platform.py
VindeeR: Oh that's very true, thanks for pointing it out :raised_hands:

autosubmit/platforms/pjmplatform.py
VindeeR: Thanks for checking it, it just feels unnatural to be spelled like this, so I worry.

autosubmit/platforms/platform.py
VindeeR: I just pointed it out for the sake of using the same format with all the changes, it's not important, but since I notice thought it wouldn't harm pointing it

autosubmit/autosubmit.py
VindeeR: Yeah, I wasn't sure if we had somewhere else to move it to, so I suggested the `utils`, I can create a new issue to restructure this function into a new `runtime/common.py` file

autosubmit/autosubmit.py
VindeeR: this is a tricky name to change if you get the shadow warning, maybe we can use `platform_interface`, or `platform_service`.

what do you think?

VindeeR (commented):
Sorry for nitpick some stuff, I just thought it was better point out even if it wasn't important, I'll avoid unimportant/very minimal things from now on.

I replied to your questions, thanks for the hard work 🙌

autosubmit/job/job_utils.py
dbeltrankyl: I was about to change this, but it's only related to the stop command, so to avoid overcomplicating this, Pr I'll do it in another issue. 

I think this should be changed to 

Stop -> admit (true/false)
Cancel_jobs -> optional change status? if yes, the running jobs will be set to "Failed" and the active ones except running to "WAITING"

test/unit/platforms/test_cancel_jobs.py
dbeltrankyl: I've simplified this test file overall

test/unit/platforms/test_cancel_jobs.py
dbeltrankyl: same

test/unit/platforms/test_cancel_jobs.py
dbeltrankyl: same

test/unit/platforms/test_cancel_jobs.py
dbeltrankyl: same

test/unit/platforms/test_ecplatform.py
dbeltrankyl: done

dbeltrankyl:
Hello @VindeeR 

Thanks for the feedback

Sorry for the delay, I've fixed the bugs

* Addressed your last feedback

* Added platform.has_scheduler 

* Fixed shebang and normalized to use the loaded env instead of a hardcoded path (this was more complicated than anticipated) 
* Fixed executable parameter
* Fixed the snapshot name bug

* Added unimplemented methods for PS, local, for getting the name of the submitted jobs that is are similar to the EC one, which caused another refactor for these platforms

* added/moved/simplificated tests 

I believe this is the final update 🤞

dbeltrankyl:
Tests where not working after removing the timeout, now it should pass 👍

autosubmit/job/job_packages.py
```diff
@@ -82,10 +90,23 @@ def __init__(self, jobs: list[Job]):
                     raise Exception('Only one valid platform per package')
         except IndexError:
             raise Exception('No jobs given')
+        self.fail_count = jobs[0].fail_count
+        self.ec_queue = jobs[0].ec_queue if hasattr(jobs[0], "ec_queue") and jobs[0].ec_queue else None
+        self.sections = set(job.section for job in jobs if job.section)
+        self.sections = "&".join(sorted(self.sections))
```
VindeeR:
```suggestion
        self.sections = "&".join(
            sorted(
                set(job.section for job in jobs if job.section)
            ))
```

autosubmit/job/job_packages.py
```diff
@@ -689,45 +532,19 @@ def _create_scripts(self, configuration: 'AutosubmitConfig'):
             self._job_scripts[self.jobs[i].name] = self.jobs[i].create_script(configuration)
         self._common_script = self._create_common_script()
 
-
-    #TODO: I believe this might not be working as intended since _common_script_content is not defined (code from 31/03/2017)
+    # TODO: I believe this might not be working as intended since _common_script_content is not defined (code from 31/03/2017)
```
VindeeR: I believe this line can be deleted

```suggestion
```

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1472,38 +1487,44 @@ def get_ssh_output(self):
...
+        :param timeout: timeout for the execution
+        :param export: export command to set environment variables
+        :param executable: executable to run the script
+        :param x11_options: x11 options to run the script
+        :param fail_count: number of times the job has failed, used to name the output and error files
+        :param sub_queue: alternative queue to run the job, used for some platforms like ECaccess
+        :param redirect_out_err: whether to redirect stdout and stderr to files
         :return: command to execute script
+        :rtype: str
+        :rtype: str
```
VindeeR:
```suggestion
        :rtype: str
```

autosubmit/job/job_utils.py
VindeeR: Maybe we should change the name of the function to something that is clearer like, `stop_or_cancel_jobs` and change the description to match this behavior.

test/unit/platforms/conftest.py
```diff
@@ -0,0 +1,109 @@
...
+from autosubmit.platforms.locplatform import LocalPlatform
+from autosubmit.platforms.pjmplatform import PJMPlatform
+from autosubmit.platforms.psplatform import PsPlatform
+from autosubmit.platforms.slurmplatform import SlurmPlatform
+
+"""Fixtures for unit tests."""
+
+
+@pytest.fixture
+def slurm_platform(autosubmit_config, tmp_path):
```
VindeeR, line 32: Maybe we can use this fixture at others tests too like `test/unit/test_slurm_platform.py` and `test/unit/platforms/wrappers/test_wrapper_factory.py` since they have their own fixtures, instead of rewriting the same fixture

VindeeR (changes requested):
Just the one from `autosubmit/job/job_utils.py` feels more important due to clarity of the code, everything else is pretty minor

dbeltrankyl:
Hello @VindeeR!

Thanks for the feedback! (again)

> Maybe we should change the name of the function to something clearer, like stop_or_cancel_jobs, and update the description to match this behavior.

Did you read my previous comment on this?

I've created a new issue to handle this https://github.com/BSC-ES/autosubmit/issues/2964 because it's not just changing a description; it also involves changing the flags and behavior, I believe. And it would be out-scope of this pr

In this PR, the change was to adapt the function to use the new cancel_jobs call from platforms, and nothing more. the behaviour is the same as before this PR


> Maybe we can use this fixture in other tests too, like test/unit/test_slurm_platform.py and test/unit/platforms/wrappers/test_wrapper_factory.py, since they have their own fixtures, instead of rewriting the same fixture.

I would prefer to move this to another PR because it looks like a refactor ( maybe more tests can benefit from it or reorder some of them), and this PR is large enough. What do you think?


Addressed the rest of the feedback, thanks!

VindeeR:
> Hello @VindeeR!
> 
> Thanks for the feedback! (again)
> 
> > Maybe we should change the name of the function to something clearer, like stop_or_cancel_jobs, and update the description to match this behavior.
> 
> Did you read my previous comment on this?
> 
> I've created a new issue to handle this #2964 because it's not just changing a description; it also involves changing the flags and behavior, I believe. And it would be out-scope of this pr
> 
> In this PR, the change was to adapt the function to use the new cancel_jobs call from platforms, and nothing more. the behaviour is the same as before this PR
> 
> > Maybe we can use this fixture in other tests too, like test/unit/test_slurm_platform.py and test/unit/platforms/wrappers/test_wrapper_factory.py, since they have their own fixtures, instead of rewriting the same fixture.
> 
> I would prefer to move this to another PR because it looks like a refactor ( maybe more tests can benefit from it or reorder some of them), and this PR is large enough. What do you think?
> 
> Addressed the rest of the feedback, thanks!

Oh I've read, but I got confused during the review if this will be done elsewhere then no problem :+1:

VindeeR (approved):
Thanks for the good work refactoring this issue 🙌 now we can move on with the other PRs

PR #2949, opened 2026-04-14
Correct conversion in num_max_splits when SPLITSIZEUNIT = 'month'

### LINKED ISSUE #2944: Error in `num_max_splits` computation when `split_unit` is `month`
### Description

I was checking the code for the updated documentation on splits, and I've found the following inconsistency.
In file `job_utils.py`, method `calendar_chunk_section`.

When user defines in the yml files:
`SPLITSIZEUNIT: "month"`

The maximum number of splits is set as:
`num_max_splits = run_days / 12`

Is this correct? Assuming run_days is in unit=day, and we want to convert run_days to unit=month should not be divided by 30.0?
E.g.
`num_max_splits = run_days / 30`

ntorqulu:
Closes #2944

**Things done**
- Fixed the units conversion mismatch in variable `num_max_splits` of file job_utils.py. Previously it was computed as:
```{python}
num_max_splits = run_days / 12
```
But run_days is in `days` unit, so the conversion factor to months should be
```{python}
num_max_splits = run_days / 30
```
- Extended `test_auto_calendar_split` to check that the split auto conversion is computed correctly for diff chunk_size_unit and spit_size_unit.
---

Edit: better explanation.
We want to compute the number of splits given a chunk. The error arises when `CHUNK_UNIT = year` and `SPLIT_UNIT = month`. Example to illustrate the problem.
1. Chunk defined with date 19900101. Chunk unit is year. Chunk size is 1 -> we have one chunk that lasts one year.
2. We want to auto split this chunk into multiple months. We define auto-splits with `SPLIT_UNIT = month` and `SPLIT_SIZE = one` -> we want AS to automatically generate 12 splits given the yearly chunk.

**What is currently doing**
1. Compute the number of days in the chunk -> 365 days
2. Divide the number of days in the chunk by 12 -> 30.41
3. With flexible policy, we obtain **31 months**. With strict policy we obtain **30 month**
4. But we wanted 12...

**Which are the issues**
Two main issues:
- Unit factors are not correct. If we want the number of months in a year, we should divide by 30 something (depending on the month, which raises the next issue) instead of 12.
- The number of months are computed without taking into account the days in each month. The splits should take into account the calendar.

----

**Check List**

- [x] Documentation updated.


autosubmit/job/job_utils.py
```diff
@@ -171,7 +171,7 @@ def calendar_chunk_section(exp_data, section, date, chunk) -> int:
         if split_unit == "hour":
             num_max_splits = run_days * 24
         elif split_unit == "month":
-            num_max_splits = run_days / 12
+            num_max_splits = run_days / 30
         elif split_unit == "year":
             if not is_leap_year(chunk_start.year) or cal == "noleap":
```
ntorqulu: Why it is only checked if the first year is leap or not? Should we check all the years between chunk_start.year and chunk_end.year?

autosubmit/job/job_utils.py
```diff
@@ -171,7 +171,7 @@ def calendar_chunk_section(exp_data, section, date, chunk) -> int:
         if split_unit == "hour":
             num_max_splits = run_days * 24
         elif split_unit == "month":
-            num_max_splits = run_days / 12
+            num_max_splits = run_days / 30
```
ntorqulu: Is this a correct approximation or should be more strict (e.g. months with 31, 28 or 29 days)

ntorqulu (commented):
Hi, I was modifying the units to convert num_max_splits for monthly splits, and I realized that for years we take into account if the year is leap or not, but just for the first year.... Should we enforce strict calendar computations or is accepted to use approximations (e.g. use 30 days as standard month measure)

autosubmit/job/job_utils.py
VindeeR: I agree with you, it is weird that it was dividing by 12.

we need to wait for @dbeltrankyl reply, but I have a suggestion


```suggestion
            num_max_splits = run_days / calendar.monthrange(chunk_start.year, chunk_start.month)[1]
```

maybe we can do something like the above, this way it calculates the split base on the month it uses as a reference to start.

I'm not confident that it would be the best solution since it might conflict if it goes between months, but it is likely to happen with any other approach with the current design.

wdyt?

VindeeR (commented):
I don't feel confident enough to approve this change but I did add a comment

autosubmit/job/job_utils.py
ntorqulu: mmm I find it a good approach if the chunk spans only for one month.

For instance, if a chunk spans February to March, `run_days` can be 59 (28+31) but `calendar.monthrange(chunk_start.year, chunk_start.month)[1] == 28`.

So `num_max_splits` = 59/28 = 2.1 which rounds up to 3 in flexible (default) mode. So splits is set to 3 where in reality should be 2.
The thing is that years I think is already treated like this (only consider the first year) but with months there is more date variability.

VindeeR: Maybe we can iterate between all the possible months, calculating the day, and divide by the amount of months rounding it (up maybe?)

ntorqulu: Yes, I think this would be the solution... 
We can then reuse the logic on how we compute the days from the months to compute the days form the years in order to avoid just checking if the first year is leap or not.
But then all this section would need a refactor :upside_down_face:

VindeeR: lets wait and see what does @dbeltrankyl think is the best solution.

Thanks for the hard work :raised_hands:

autosubmit/job/job_utils.py
dbeltrankyl: I think we didn't consider to have a range of years when writting it 

although I think it is not possible to have an split_unit as we don't have implemented any higher unit than year for chunks

autosubmit/job/job_utils.py
dbeltrankyl: Hello! 

Sorry for the delay, today I had a few issues with my environment ( due to an update of PyCharm), and then I was dealing with other issues 😓 

chunk_unit == "year"
split_unit == "month"

chunk_start = 1/01/01
chunk_end = 31/12/12 
run_days = 365 ( depends on leap tho)

So.. 365/12 can be correct (although simple), as we are splitting the amount of work into 12 equitable parts, but it is not the aim of the "auto" feature 


The current behavior can work if the template code is aware that the distribution is always 31 days per month

The approach of @ntorqulu is much better as the templates won't have to add extra stuff to handle that

autosubmit/job/job_utils.py
dbeltrankyl: so tthe split_unit == "year" can be just removed

autosubmit/job/job_utils.py
dbeltrankyl: Having said so, any change regarding that should be communicated to DestinE workflow developers ( and maybe auto-ecearth4 I'll put @franra9 in the loop, as he is usually dealing with splits in DestinE

ntorqulu: Sure, thanks for the explanation @dbeltrankyl.
Sorry to follow with questions, splits are a little bit confusing.
Then in your example:

chunk_unit == "year"
split_unit == "month"

chunk_size = 1
split_size = 1
We expect one month per split, 12 splits per chunk

chunk_start = 1/01/01
chunk_end = 31/12/12

`num_max_splits` = 365/12 = 30.41 rounded up to 31 due to flexible policy.
and `splits` = 31/1 = 31. So is expected AS to compute 31 splits per year when the split unit is month?

dbeltrankyl: > 31/1 = 31. So is expected AS to compute 31 splits per year when the split unit is month?

Hm, no(not looking at the current implementation, just what it should be doing right now, and if it is not doing this, it is bugged).

With 1 year chunk and monthly splits:

What I meant is that with the current code (at least how it was thought), we should have a total of 12 jobs splits, and each split should have 31 days SPLIT_START_DATE -> SPLIT_END_DATE

The problem with that is:

1) Some splits will process two different months, and the offset will shortly start after the first month of the first year
2) The last split:
2a) It may break the boundary of chunk_start_date and begin to process the next year ( 372 days)
2b) or it may process fewer days than other splits 

It is better to have 12 splits with the appropriate split_start_date -> split_end_date as you proposed, so the templates don't have to calculate anything

ntorqulu: Okay, thanks! I think it's bugged then. I've been testing it with the following configuration:

- `expdef.yml`
```
DEFAULT:
  EXPID: a03l
  HPCARCH: local
EXPERIMENT:
  DATELIST: 19600101
  MEMBERS: "00"
  CHUNKSIZEUNIT: year
  CHUNKSIZE: '1'
  NUMCHUNKS: '1'
  SPLITSIZEUNIT: month
  CALENDAR: standard
PROJECT:
  PROJECT_TYPE: none
  PROJECT_DESTINATION: ''
GIT:
  PROJECT_ORIGIN: ''
  PROJECT_BRANCH: ''
  PROJECT_COMMIT: ''
  PROJECT_SUBMODULES: ''
  FETCH_SINGLE_BRANCH: true
SVN:
  PROJECT_URL: ''
  PROJECT_REVISION: ''
LOCAL:
  PROJECT_PATH: ''
PROJECT_FILES:
  FILE_PROJECT_CONF: ''
  FILE_JOBS_CONF: ''
  JOB_SCRIPTS_TYPE: ''
RERUN:
  RERUN: false
  RERUN_JOBLIST: ''
```

- `jobs.yml`
```
JOBS:
  TEST_MONTH_AUTO:
    FILE: sleep 0
    RUNNING: chunk
    SPLITS: auto
    WALLCLOCK: '00:05'
    PLATFORM: local
```

- and obtained the following plot (31 splits instead of 12):

[a03l_20260420_0949.pdf](https://github.com/user-attachments/files/26887838/a03l_20260420_0949.pdf)

ntorqulu:
In theory you can compute auto splits when job is RUNNING in a level different from 'once'. In the documentation was stated that the level should be 'chunk' so I'll change it in the code to match

[if jobs_data.get(section, {}).get("RUNNING", "once") != "once":](https://github.com/BSC-ES/autosubmit/blob/78f9a377f50b044f4adfd0e57859c08c82de0f6a/autosubmit/job/job_utils.py#L216)

LuiggiTenorioK:
Hi @ntorqulu, the final approval of this issue will remain with @dbeltrankyl once he finishes with the AS 4.1.17 and comes back from holidays

Since it is planned to be in AS 4.2.0, there is no hurry to be approved until 4.1.17 is released.

autosubmit/job/job_utils.py
```diff
@@ -34,18 +34,33 @@
...
 
+VALID_CALENDARS = ["standard", "noleap"]
+SPLIT_POLICIES = ["flexible", "strict"]
 
-def is_leap_year(year) -> bool:
-    """Determine whether a year is a leap year."""
+
+def is_leap_year(year: int) -> bool:
+    """
+    Determine whether a year is a leap year.
```
dbeltrankyl: It seems that this was well indented originally:

```
    """
        Determine whether a year is a leap year.
-> 
 """Determine whether a year is a leap year.

```

autosubmit/job/job_utils.py
```diff
@@ -34,18 +34,33 @@
...
 
 
-def calendar_unitsize_isgreater(split_unit, chunk_unit) -> bool:
+def calendar_unitsize_isgreater(split_unit: str, chunk_unit: str) -> bool:
     """
-    Check if the split unit is greater than the chunk unit
-    :param split_unit:
-    :param chunk_unit:
-    :return: boolean
+    Check if the split unit is greater than the chunk unit.
```
dbeltrankyl: same, at some point, we decided to change the docstring standard to

"""bla 

instead of 
"""
bla

autosubmit/job/job_utils.py
```diff
@@ -55,63 +70,97 @@ def calendar_unitsize_isgreater(split_unit, chunk_unit) -> bool:
         raise AutosubmitCritical("Invalid calendar unit size")
 
 
-def calendar_unitsize_getlowersize(unitsize) -> str:
+def calendar_unitsize_getlowersize(unitsize: str) -> str:
     """
-    Get the lower size of a calendar unit
-    :return: str
+    Given a calendar unit size, return the lower unit size.
```
dbeltrankyl: same

autosubmit/job/job_utils.py
```diff
@@ -55,63 +70,97 @@ def calendar_unitsize_isgreater(split_unit, chunk_unit) -> bool:
...
+    :param cal: Calendar type (standard, noleap)
+    :dtype cal: str
+    :return: Number of days in the month
+    :rtype: int
     """
     year = int(date_str[0:4])
     month = int(date_str[4:6])
     if month == 2:
-        if is_leap_year(year):
+        if cal != "noleap" and is_leap_year(year):
```
dbeltrankyl: couldn't we move the `cal != "noleap"` inside the function is_leap_year(year) ? is_leap_year(year,cal)

This function is called in 5 places

In 2 of these 5 places, it is being called without the `cal != "noleap"` which seems a bug. Also, it is called in the future, the developer in charge may not know that the `cal != "noleap" is necessary to have

(lines without it 132, 165 )

autosubmit/job/job_utils.py
```diff
@@ -55,63 +70,97 @@ def calendar_unitsize_isgreater(split_unit, chunk_unit) -> bool:
...
         return "hour"
     else:
         return list(CALENDAR_UNITSIZE_ENUM.keys())[unit_value - 1]
 
 
-def calendar_get_month_days(date_str) -> int:
+def calendar_get_month_days(date_str: str, cal: str = "standard") -> int:
     """
-    Get the number of days in a month
+    Get the number of days in a month, respecting the calendar type (leap year or not).
```
dbeltrankyl: same ""bla...

autosubmit/job/job_utils.py
```diff
@@ -55,63 +70,97 @@ def calendar_unitsize_isgreater(split_unit, chunk_unit) -> bool:
...
+        return 28
     elif month in [4, 6, 9, 11]:
         return 30
     else:
         return 31
 
 
-def get_chunksize_in_hours(date_str, chunk_unit, chunk_length) -> int:
+def get_chunksize_in_hours(date_str: str, chunk_unit: str, chunk_length: int, cal: str = "standard") -> int:
+    """
```
dbeltrankyl: same ""bla...

autosubmit/job/job_utils.py
```diff
@@ -55,63 +70,97 @@ def calendar_unitsize_isgreater(split_unit, chunk_unit) -> bool:
...
+    :param chunk_unit: Unit of the chunk
+    :dtype chunk_unit: str
+    :param chunk_length: Length of the chunk
+    :dtype chunk_length: int
+    :param cal: Calendar type (standard, noleap)
+    :dtype cal: str
+    :return: The chunk size in hours
+    :rtype: int
+    """
     if is_leap_year(int(date_str[0:4])):
```
dbeltrankyl: missing cal != "noleap"?

autosubmit/job/job_utils.py
```diff
@@ -55,63 +70,97 @@ def calendar_unitsize_isgreater(split_unit, chunk_unit) -> bool:
...
     return chunk_size_in_hours
 
 
-def calendar_split_size_isvalid(date_str, split_size, split_unit,
-                                chunk_size_in_hours) -> bool:
+def calendar_split_size_isvalid(date_str: str, split_size: int, split_unit: str,
+                                chunk_size_in_hours: int, cal: str = "standard") -> bool:
     """
-    Check if the split size is valid for the calendar
+    Check if the split size is valid for the calendar.
```
dbeltrankyl: same "bla...

autosubmit/job/job_utils.py
```diff
@@ -135,62 +184,195 @@ def calendar_split_size_isvalid(date_str, split_size, split_unit,
...
-def calendar_chunk_section(exp_data, section, date, chunk) -> int:
+def _validate_calendar_inputs(
+    cal: str, chunk_unit: str, split_unit: str, split_policy: str
+) -> None:
     """
-    Calendar for chunks
-    :param section:
-    :param parameters:
-    :return: int
+    Validate the calendar inputs.
```
dbeltrankyl: same """bla..

autosubmit/job/job_utils.py
```diff
@@ -55,63 +70,97 @@ def calendar_unitsize_isgreater(split_unit, chunk_unit) -> bool:
...
+    :dtype split_unit: str
     :param chunk_size_in_hours: chunk size in hours
-    :return: bool
+    :dtype chunk_size_in_hours: int
+    :param cal: Calendar type (standard, noleap)
+    :dtype cal: str
+    :return: True if the split size is valid for the calendar
+    :rtype: bool
     """
     if is_leap_year(int(date_str[0:4])):
```
dbeltrankyl: missing cal != "noleap"?

autosubmit/job/job_utils.py
```diff
@@ -135,62 +184,195 @@ def calendar_split_size_isvalid(date_str, split_size, split_unit,
...
+        )
+    if split_policy not in SPLIT_POLICIES:
+        raise AutosubmitCritical(
+            f"Invalid split policy {split_policy}. Autosubmit only supports 'flexible' and 'strict' split policies."
+        )
+
+
+def _count_units_between_dates(start_date, end_date, unit, cal) -> float:
+    """
+    Count the number of units between two dates.
```
dbeltrankyl: same """bla...

autosubmit/job/job_utils.py
```diff
@@ -135,62 +184,195 @@ def calendar_split_size_isvalid(date_str, split_size, split_unit,
...
+    elif unit == "month":
+        total = 0.0
+        current = start_date.replace(day=1)
+        while current < end_date:
+            days_in_month = calendar_get_month_days(f"{current.year}{current.month:02d}01", cal)
+            next_year = current.year + current.month // 12
+            next_month_num = (current.month % 12) + 1
+            next_month = current.replace(year=next_year, month=next_month_num, day=1)
+
+            # Compute start and end day within the month
```
dbeltrankyl: (nitpicky) I think the code is self-explanatory enough, but feel free to leaving it 👍

autosubmit/job/job_utils.py
```diff
@@ -135,62 +184,195 @@ def calendar_split_size_isvalid(date_str, split_size, split_unit,
...
+        raise AutosubmitCritical(
+            f"Split unit {split_unit} is greater than chunk unit {chunk_unit}. Autosubmit doesn't support this configuration. Please change the split unit to day or lower. Or don't use calendar splits."
+        )
+    if split_policy not in SPLIT_POLICIES:
+        raise AutosubmitCritical(
+            f"Invalid split policy {split_policy}. Autosubmit only supports 'flexible' and 'strict' split policies."
+        )
+
+
+def _count_units_between_dates(start_date, end_date, unit, cal) -> float:
```
dbeltrankyl: missing type hints

autosubmit/job/job_utils.py
```diff
@@ -135,62 +184,195 @@ def calendar_split_size_isvalid(date_str, split_size, split_unit,
...
+        return total
+    else:
+        raise AutosubmitCritical(
+            f"Invalid split unit {unit}. Autosubmit only supports 'hour', 'day', 'month' and 'year' units, or 'none' to automatically select the lower unit of the chunk unit."
+        )
+
+
+def calendar_chunk_section(exp_data: dict, section: str, date: str, chunk: int) -> int:
+    """
+    Calculate the number of splits for a chunk based on the calendar configuration.
```
dbeltrankyl: same """bla...

autosubmit/job/job_utils.py
```diff
@@ -135,62 +184,195 @@ def calendar_split_size_isvalid(date_str, split_size, split_unit,
...
+    :dtype date: str
+    :param chunk: Chunk number
+    :dtype chunk: int
+    :return: Number of splits for the chunk
+    :rtype: int
+    :raises AutosubmitCritical: If the calendar configuration is invalid or if the split size doesn't fit in the chunk size.
+    """
+    jobs_data = exp_data.get("JOBS", {})
+
+    # Return early
```
dbeltrankyl: (nitpicky) In line with the other comment regarding # comments, I think some comments are redundant. I'm fine with it, tho if you think that they're really neccesary

test/unit/job/test_job_utils.py
```diff
@@ -164,4 +171,291 @@ def test_cancel_jobs(create_job_list):
...
+def test_calendar_unitsize_isgreater_raises():
+    """Test that an invalid unit in calendar_unitsize_isgreater raises AutosubmitCritical."""
+    with pytest.raises(AutosubmitCritical):
+        calendar_unitsize_isgreater("invalid", "hour")
+
+
+@pytest.mark.parametrize(
+    "date_str, cal, expected_days",
+    [
+        # Februray with standard calendar (leap year)
```
dbeltrankyl: typo > Februray

test/unit/job/test_job_utils.py
```diff
@@ -164,4 +171,291 @@ def test_cancel_jobs(create_job_list):
...
+    # for year chunk_size_unit (2000), run_days = 366 (leap year)
+    # So:
+    # chunk day + split hour -> 24 splits
+    # chunk day + split day -> 1 split
+    # chunk month + split hour -> 744 splits (31 days * 24 hours)
+    # chunk month + split day -> 31 splits
+    # chunk year + split hour -> 8784 splits (366 days * 24 hours)
+    # chunk year + split day -> 366 splits
+    # chunk year + split month -> 12 splits
+    # chunk year + split year -> 1 split
```
dbeltrankyl: Instead of comments, you can add id's like in the other tests. So if some of them fail at some point, we will have a better time figuring out which one is failing 

Also, if the id_name is good enough, we don't need the comments, but if you want to keep them, maybe it is better to have them in-line? , to reduce human mistakes mainly ( pinpointing which comment points to which tests)

dbeltrankyl: The id could be the params, or the expected behavior encoded

test/unit/job/test_job_utils.py
```diff
@@ -164,4 +171,291 @@ def test_cancel_jobs(create_job_list):
...
+        },
+    }
+    date = datetime.strptime("20000101", "%Y%m%d")
+    chunk = 1
+    with pytest.raises(AutosubmitCritical):
+        calendar_chunk_section(experiment_data, "A", date, chunk)
+
+
+def test_calendar_chunk_section_running_not_chunk():
+    """Test that calendar_chunk_section raises AutosubmitCritical when the job is not set to run by chunk."""
```
dbeltrankyl: Should it raise AutosubmitCritical or > `assert calendar_chunk_section(experiment_data, "A", date, chunk) == 0` ?

 There is a mismatch

autosubmit/job/job_utils.py
dbeltrankyl: > date: str

I think this is a `date: datetime` ? (according to the test)

dbeltrankyl (commented):
Great work @ntorqulu !

Some very minor things about docstrings/comments. And I think there is a bug in two lines with the cal != "noleap" 

I've tested it, and it seems to work as expected... example: 

```

(split: auto , splitsize:2 , chunksize: 1, chunk: 1)
################
# Autosubmit job
################
chunk=1
chunk_first=TRUE
chunk_last=TRUE
chunk_start_date=20000101
chunk_start_year=2000
chunk_start_month=01
chunk_start_day=01
chunk_start_hour=00
chunk_end_date=20010101
chunk_end_year=2001
chunk_end_month=01
chunk_end_day=01
chunk_end_hour=00
chunk_second_to_last_date=20001231
chunk_second_to_last_year=2000
chunk_second_to_last_month=12
chunk_second_to_last_day=31
chunk_second_to_last_hour=00
chunk_end_in_days=366
run_days=366
prev=0
day_before=19991231

splits=6
split=2
split_first=FALSE
split_last=FALSE
splitsize=2
splitsizeunit=month
splitscalendar=standard
split_start_date=20000301
split_start_year=2000
split_start_month=03
split_start_day=01
split_start_hour=00
split_end_date=20000501
split_end_year=2000
split_end_month=05

split_end_day=01
split_end_hour=00
split_second_to_last_date=20000430
split_second_to_last_year=2000
split_second_to_last_month=04
split_second_to_last_day=30
split_second_to_last_hour=00
```

autosubmit/job/job_utils.py
ntorqulu: Sure, I keep forgetting the docstring standard because there're still places that use the newline style. I'll change them to the original one for all the docstrings in `job_utils.py`.
I also had doubts about the type declaration.  Should we use `:type` or `:dtype`? 
Do you think is a good idea to have it written it down somewhere in the repo? In the `.github` folder, in the docs or in the discussion section for instance.

autosubmit/job/job_utils.py
ntorqulu: Yes, it's true; it's a bug. I'll change it, thanks!

autosubmit/job/job_utils.py
ntorqulu: Sure, it was more for ordering my mind 😅. No problem, I'll delete them.

autosubmit/job/job_utils.py
ntorqulu: Yes, it should be `datetime`. Changing it!

test/unit/job/test_job_utils.py
ntorqulu: Yes, it should return 0; not `AutosubmitCritical`. Changing the comment!

autosubmit/job/job_utils.py
ntorqulu: I'll also add a unit test for `is_leap_year`, as it has been refactored and carried bugs.

autosubmit/job/job_utils.py
dbeltrankyl: hm now that you mention it... I think `dtype` won't be rendered in readthedocs, no?

https://www.sphinx-doc.org/en/master/usage/domains/python.html#info-field-lists 

So I would use` :type`

> Do you think is a good idea to have it written it down somewhere in the repo?

We can add a small section on CONTRIBUTING.md

dbeltrankyl (approved):
Looks good! thanks!

Just consider the change of dtype for type, as it seems that readthedocs won't recognize dtype as a valid keyword

ntorqulu:
Yes, already fixed the ones in `job_utils.py`

autosubmit/job/job_utils.py
ntorqulu: I'll open a new issue for this and work on it there; more for me to remember which docstring standard we're using than anything else :laughing:

PR #2950, opened 2026-04-15
[Enhancement] - Fixing leaks of semaphore

### LINKED ISSUE #2900: [Bug] - Leaked semaphore objects to clean up
### Description
As @Lerriola was making use of the `PBS platform` he kept running into an issue that would terminate the workflow.

He said he executes the workflow normally, making use of tmux, and after some time he finds the workflow killed with the following Message

```
375 of 413 jobs remaining (20:44)
Checking 3 jobs for platform MIYABI
Job aapa_20120101_fc05_3_SIMULATION is RUNNING
Job aapa_20120101_fc06_3_SIMULATION is RUNNING
Job aapa_20120101_fc07_3_SIMULATION is RUNNING
File /esarchive/autosubmit/aapa/proj/git_project/src/jobs/nicoco/simulation_nicoco.sh exists
Calculating possible ready jobs for MIYABI
/shared/earth/easybuild/rocky/9.5/x86_64/software/Miniconda3/24.7.1-0/lib/python3.12/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 13 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
Killed
```

No error can be found at the `ASLOG` or `LOG_aapa`.

After looking it up we notice it could be due to opening and closing libraries or resource consumption.

[Stack Overflow](https://stackoverflow.com/questions/64515797/there-appear-to-be-6-leaked-semaphore-objects-to-clean-up-at-shutdown-warnings-w)

[Python Org](https://discuss.python.org/t/leaked-semaphore-objects-to-clean-up-error-when-using-concurrent-futures/54510)

we are not sure if this could be some memory leak happening or maybe something else that could be impacting the execution.

* Autosubmit version: branch::miyabi_pbs_platform
* Machine/VM/environment name (if applicable): bsc04
* Experiment ID (if applicable): aapa

### Reproducible Example

Running the workflow, after a few minutes to hours it happens

VindeeR:
Closes #2900

Part of the development fixing this issues was implemented at #2668 

We identified that some semaphore leaks were happening altogether with memory leaks.
In order to fix it we identified that the problem was due to the `Processes` and `Queues` not being directly and promptly terminated when associated a new one to the variable.

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - Slurm tests in CICD are using a new container by @giovtorres #2871
 - Fixed main-loop iteration message handling singular/plural and reducing number of line breaks #2969
 - Updated `Defining the workflow` documentation section #2954
+- Improving the handling of leakages that were happening when dealing with queues and processes #2900
 
 ### 4.1.16: Postgres (experimental) support, bug fixes, and enhancements
 
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2374,10 +2374,16 @@ class Autosubmit:
                     Autosubmit.restore_platforms(platforms_to_test, as_conf=as_conf, expid=expid)
                 Log.info("Waiting for all logs to be updated")
                 for p in platforms_to_test:
-                    if p.log_recovery_process and p.log_recovery_process.is_alive():
+                    if (p.log_recovery_process is not None and p.log_recovery_process.is_alive()
+                            and p.cleanup_event is not None):
                         p.cleanup_event.set()  # Send cleanup event
                         p.log_recovery_process.join(timeout=60)
                         if p.log_recovery_process.is_alive():
+                            p.log_recovery_process.terminate()
+                            p.log_recovery_process.join(timeout=60)
+                            if p.log_recovery_process.is_alive():
+                                Log.error("Log recovery process refused to terminate")
+                                p.log_recovery_process.kill()
                             Log.warning(f"Process for platform {p.name} failed to terminate within the timeout.")
                 Autosubmit.check_logs_status(job_list, as_conf, new_run=False)
                 job_list.save()
@@ -2986,13 +2992,13 @@ class Autosubmit:
         submitter.load_platforms(as_conf)
         platforms = submitter.platforms
 
-        platforms_to_test = set()
+        platforms_to_test: list['ParamikoPlatform'] = list()
         for job in job_list.get_job_list():
             job.submitter = submitter
             if not job.platform_name:
                 job.platform_name = hpcarch
             job.platform = platforms[job.platform_name]
-            platforms_to_test.add(job.platform)
+            platforms_to_test.append(job.platform)
 
         completed_jobnames = Autosubmit.online_recovery(as_conf, platforms_to_test, job_list, offline)
         current_active_jobs = job_list.get_in_queue()
@@ -3314,7 +3320,7 @@ class Autosubmit:
 
                 as_conf = AutosubmitConfig(experiment_id, BasicConfig, YAMLParserFactory())
                 as_conf.check_conf_files(False, no_log=True)
-                user = os.stat(as_conf.conf_folder_yaml).st_uid
+                user = int(Path(as_conf.conf_folder_yaml).stat().st_uid)
                 try:
                     user = pwd.getpwuid(user).pw_name
                 except Exception:
@@ -3322,7 +3328,7 @@ class Autosubmit:
                         "The user does not exist anymore in the system, using id instead")
                     continue
 
-                created = datetime.datetime.fromtimestamp(os.path.getmtime(as_conf.conf_folder_yaml))
+                created = datetime.datetime.fromtimestamp(Path(as_conf.conf_folder_yaml).stat().st_mtime)
 
                 if as_conf.get_svn_project_url():
                     model = as_conf.get_svn_project_url()
@@ -3502,8 +3508,10 @@ class Autosubmit:
 
                     sep = '\n\t'
                     Log.result(sep.join(['Directories configured successfully:'] + [str(d) for d in dirs]))
-                except (IOError, OSError) as e:
+                except (IOError) as e:
                     raise AutosubmitCritical(f"Can not write config file: {e.message}", 7012)
+                except (OSError) as e:
+                    raise AutosubmitCritical(f"Can not write config file: {e}", 7012)
         except (AutosubmitCritical, AutosubmitError):
             raise
         except BaseException as e:
@@ -3606,7 +3614,7 @@ class Autosubmit:
             # Look for %_%
             variables = re.findall('%(?<!%%)[a-zA-Z0-9_.-]+%(?!%%)', template_content, flags=re.IGNORECASE)
             variables = [variable[1:-1].upper() for variable in variables]
-            results = {}
+            results: dict = dict()
             # Change format
             for old_format_key in variables:
                 for key in as_conf.load_parameters().keys():
@@ -3996,7 +4004,7 @@ class Autosubmit:
         :return: ``True`` if the experiment has been successfully archived. ``False`` otherwise.
         :rtype: bool
         """
-        exp_folder = os.path.join(BasicConfig.LOCAL_ROOT_DIR, expid)
+        exp_folder = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(expid)
 
         if not noclean:
             # Cleaning to reduce file size.
@@ -4007,29 +4015,27 @@ class Autosubmit:
 
         # Getting year of last completed. If not, year of expid folder
         year = None
-        tmp_folder = os.path.join(exp_folder, BasicConfig.LOCAL_TMP_DIR)
-        if os.path.isdir(tmp_folder):
-            for filename in os.listdir(tmp_folder):
-                if filename.endswith("COMPLETED"):
-                    file_year = time.localtime(os.path.getmtime(
-                        os.path.join(tmp_folder, filename))).tm_year
+        tmp_folder = exp_folder.joinpath(BasicConfig.LOCAL_TMP_DIR)
+        if tmp_folder.is_dir():
+            for filename in tmp_folder.iterdir():
+                if filename.name.endswith("COMPLETED"):
+                    file_year = time.localtime(tmp_folder.joinpath(filename).stat().st_mtime).tm_year
                     if year is None or year < file_year:
                         year = file_year
 
         if year is None:
             year = time.localtime(os.path.getmtime(exp_folder)).tm_year
         try:
-            year_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, str(year))
-            if not os.path.exists(year_path):
-                os.mkdir(year_path)
-                os.chmod(year_path, 0o775)
+            year_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(str(year))
+            if not year_path.exists():
+                year_path.mkdir(mode=0o775, parents=True)
         except Exception as e:
             raise AutosubmitCritical(f"Failed to create year-directory {str(year)} for experiment {expid}", 7012,
                                      str(e))
         Log.info(f"Archiving in year {str(year)}")
 
         if rocrate:
-            Autosubmit.rocrate(expid, Path(year_path))
+            Autosubmit.rocrate(expid, year_path)
             Log.info('RO-Crate ZIP file created!')
         else:
             # Creating tar file
@@ -4041,10 +4047,11 @@ class Autosubmit:
                 else:
                     compress_type = "w"
                     output_filepath = f'{expid}.tar'
-                with tarfile.open(os.path.join(year_path, output_filepath), compress_type) as tar:
+                year_path_file = year_path.joinpath(output_filepath)
+                with tarfile.open(year_path_file, compress_type) as tar:
                     tar.add(exp_folder, arcname='')
                     tar.close()
-                    os.chmod(os.path.join(year_path, output_filepath), 0o775)
+                    year_path_file.chmod(mode=0o775)
             except Exception as e:
                 raise AutosubmitCritical("Can not write tar file", 7012, str(e))
 
@@ -4080,11 +4087,11 @@ class Autosubmit:
         :param rocrate: flag to enable RO-Crate
         :type rocrate: bool
         """
-        exp_folder = os.path.join(BasicConfig.LOCAL_ROOT_DIR, experiment_id)
+        exp_folder = Path(BasicConfig.LOCAL_ROOT_DIR, experiment_id)
 
         # Searching by year. We will store it on database
         year = datetime.datetime.today().year
-        archive_path = None
+        archive_path = Path()
         if rocrate:
             compress_type = None
             output_pathfile = f'{experiment_id}.zip'
@@ -4095,9 +4102,8 @@ class Autosubmit:
             compress_type = "r:"
             output_pathfile = f'{experiment_id}.tar'
         while year > 2000:
-            archive_path = os.path.join(
-                BasicConfig.LOCAL_ROOT_DIR, str(year), output_pathfile)
-            if os.path.exists(archive_path):
+            archive_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(str(year), output_pathfile)
+            if archive_path.exists():
                 break
             year -= 1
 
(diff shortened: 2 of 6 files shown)


VindeeR:
@LuiggiTenorioK I believe this is fixing all the critical leaks appointed by the agent at #2668 could you please re run the agent here just to make sure no other leak was created and all the critical issues were fixed please :pray:

LuiggiTenorioK:
From Copilot:

## PR #2950 Analysis: Remaining Leaks and New Issues

### ✅ What PR #2950 Successfully Fixes

Comparing against the issues identified in the prior review of PR #2668:

| Prior Issue | Status |
|---|---|
| `send_cleanup_signal()` crashes on `None` | ✅ Fixed — null guards added (L941-942) |
| Unbounded atexit handler accumulation | ✅ Fixed — `_atexit_registered` flag prevents re-registration (L886-889) |
| `del self.recovery_queue` without `close()` in `prepare_process()` | ✅ Fixed — `close()` + `join_thread()` called before `del` (L879-882) |
| `err.message` AttributeError in `remove_workers()` | ✅ Fixed — rewrote to catch `AttributeError` and use `str(e)` (L254-255) |
| `remove_workers()` not protected by lock | ✅ Fixed — `with cls.lock:` added (L248) |
| `ctx` may be `None` in `clean_log_recovery_process()` | ✅ Fixed — guard added (L815-816) |
| `join()` without timeout at experiment end (autosubmit.py) | ✅ Fixed — `timeout=60` added (L2379) |
| `send_cleanup_signal()` doesn't call `terminate()` after join timeout | ✅ Fixed — added SIGTERM fallback (L947-950) |

---

### 🔴 Critical Remaining Issues

#### Issue 1: Double-start race condition in `create_new_process()` — NEW BUG (L892-904)

```python
def create_new_process(self, new_platform, as_conf) -> None:
    try:
        self.log_recovery_process = self.ctx.Process(...)
        if self.log_recovery_process is not None:  # Dead code: ctx.Process() never returns None
            self.log_recovery_process.start()
    except Exception as e:
        Log.error(f"Failed to start log recovery process: {e}")
        self.log_recovery_process = None
    if self.log_recovery_process is not None and not self.log_recovery_process.is_alive():
        self.log_recovery_process.start()  # ← BUG: tries to start an already-started process!
```

The final two lines constitute a regression bug introduced by this PR. After `start()` succeeds inside the `try` block, `is_alive()` is checked immediately. If the spawned process crashes very quickly during startup (e.g., due to misconfiguration, missing resources), `is_alive()` returns `False`, and the code tries to call `.start()` a **second time** on an already-started process. This raises `AssertionError: cannot start a process twice` in CPython's multiprocessing module.

Additionally, `ctx.Process()` **never** returns `None` — it either returns a Process object or raises an exception. The `if self.log_recovery_process is not None:` check inside the try block is dead code.

**Fix**: Remove the final `if not is_alive()` block entirely. The re-start logic is both incorrect and dangerous:
```python
def create_new_process(self, new_platform, as_conf) -> None:
    try:
        self.log_recovery_process = self.ctx.Process(
            target=recover_platform_job_logs_wrapper,
            args=(new_platform, self.recovery_queue, self.work_event, self.cleanup_event, as_conf),
            name=f"{self.name}_log_recovery")
        self.log_recovery_process.daemon = True
        self.log_recovery_process.start()
    except Exception as e:
        Log.error(f"Failed to start log recovery process: {e}")
        self.log_recovery_process = None
```

---

#### Issue 2: Process not killed at experiment end when `join()` times out (autosubmit.py L2376-2381)

```python
for p in platforms_to_test:
    if p.log_recovery_process and p.log_recovery_process.is_alive():
        p.cleanup_event.set()
        p.log_recovery_process.join(timeout=60)
        if p.log_recovery_process.is_alive():
            Log.warning(f"Process for platform {p.name} failed to terminate within the timeout.")
            # ← Missing: p.log_recovery_process.terminate()
```

When the process doesn't stop within 60 seconds, only a warning is logged but the process is left running indefinitely. This is an orphan process leak. `clean_log_recovery_process()` has the correct pattern (warn + `terminate()` + second join), but this exit path in `autosubmit.py` does not.

---

### 🟡 Moderate Remaining Issues

#### Issue 3: `clean_log_recovery_process()` leaves inconsistent state (L809-841)

After `clean_log_recovery_process()` returns, the Platform object is in a partially-cleaned state:
- `self.cleanup_event` → still the old, **SET** event (not reset)
- `self.log_recovery_process` → still references the dead process object (not set to None)
- `self.recovery_queue` → closed queue (not replaced)
- `self.work_event` → newly created Event (intermediate, will be overwritten by `prepare_process()`)

The normal call sequence (`clean_log_recovery_process()` → `prepare_process()`) handles this because `prepare_process()` overwrites `cleanup_event`, creates a new queue, etc. But the **intermediate state is inconsistent**, meaning:
- If code checks `p.log_recovery_process` between the two calls, it finds a dead process and may make wrong decisions
- If `cleanup_event.is_set()` is checked in this window, it returns `True` (stale), which could cause the new child process to exit immediately
- If process creation fails in `create_new_process()`, `log_recovery_process` is set to `None`, but `cleanup_event` and `recovery_queue` are left in the post-`clean_log_recovery_process()` inconsistent state

**Fix**: Reset `cleanup_event`, `log_recovery_process`, and `recovery_queue` at the end of `clean_log_recovery_process()`:
```python
self.cleanup_event = None
self.log_recovery_process = None
self.recovery_queue = None
```

#### Issue 4: Intermediate `work_event` Event/semaphore wastefully created (L840)

`clean_log_recovery_process()` ends by creating a new `work_event`:
```python
self.work_event = self.ctx.Event()  # ← Creates semaphore
```

Then `prepare_process()` immediately overwrites it with another new one:
```python
self.work_event = self.ctx.Event()  # ← Creates another semaphore
```

The intermediate Event from `clean_log_recovery_process()` is discarded without being explicitly closed. Its underlying OS semaphore is not freed until Python's resource tracker or GC runs — this is exactly the class of semaphore leak the PR is trying to fix. The fix is to set `self.work_event = None` in `clean_log_recovery_process()` (after removing it from `worker_events`) and let `prepare_process()` create the only Event.

#### Issue 5: `update_workers()` not protected by `cls.lock` (L238-241)

`remove_workers()` now correctly uses `with cls.lock:` but `update_workers()` still mutates the shared class-level list without any lock:
```python
@classmethod
def update_workers(cls, event_worker):
    cls.worker_events.append(event_worker)  # ← No lock!
```

If `update_workers()` is called concurrently (e.g., during a platform spawn) while `remove_workers()` is iterating the list, this is an unsynchronized mutation. The lock is already defined and used for removal — it should also cover appends.

#### Issue 6: Missing SIGKILL fallback when SIGTERM fails in `clean_log_recovery_process()` (L828-829)

```python
if self.log_recovery_process.is_alive():
    Log.error("Log recovery process refused to terminate")
    # ← Missing: self.log_recovery_process.kill()
```

When a process ignores SIGTERM and `join(timeout=60)` expires a second time, only an error is logged but the process remains alive forever. `kill()` (SIGKILL) should be called as a last resort to ensure the process is actually terminated.

---

### 🟢 Minor Issues

#### Issue 7: `wait_mandatory_time()` catches wrong exception (L967)

```python
except queue.Empty:
    Log.warning(f"Queue is empty and process log is {process_log} seconds.")
```

`queue.Empty` is raised by `Queue.get()` / `Queue.get_nowait()`, **not** by `Queue.empty()`. The `multiprocessing.Queue.empty()` call could raise `OSError` or other system-level errors under stress. This `except queue.Empty` clause is dead code that will never fire and masks the real potential error. The exception type should be `OSError` or simply `Exception`.

Also, the warning message itself is incorrect — `process_log` is a `bool`, so `f"{process_log} seconds"` prints `"True seconds"` or `"False seconds"`.

#### Issue 8: `_atexit_registered = None` instead of `False` in `__init__` (L151)

```python
self._atexit_registered = None
```

The subsequent check `if not getattr(self, "_atexit_registered", False)` works because `not None` is `True`, but semantically this should be `False` from the start. This also means the `getattr(..., False)` default will never be hit (since `_atexit_registered` is always set in `__init__`), so the `getattr` pattern could simply be `if not self._atexit_registered`.

#### Issue 9: `daemon = True` missing from new `create_new_process()` (L894)

The PR's new version of `create_new_process()` does not set `self.log_recovery_process.daemon = True`. The original code in PR #2668 had `self.log_recovery_process.daemon = True` before `start()`. Without this, if the parent process is killed while the child is running, the child may continue as an orphan. (This is a regression from PR #2668's implementation.)

---

### Summary Table

| # | Location | Issue | Severity | Type |
|---|---|---|---|---|
| 1 | `platform.py:903-904` | Double-start `AssertionError` bug introduced by this PR | 🔴 Critical | Regression / crash |
| 2 | `autosubmit.py:2380-2381` | Process not killed after join timeout at experiment end | 🔴 Critical | Orphan process leak |
| 3 | `platform.py:809-841` | `clean_log_recovery_process()` leaves inconsistent state (stale `cleanup_event`, `log_recovery_process`, `recovery_queue`) | 🟡 Moderate | State inconsistency |
| 4 | `platform.py:840` | Intermediate `work_event` semaphore created then discarded | 🟡 Moderate | Semaphore leak |
| 5 | `platform.py:238-241` | `update_workers()` not under `cls.lock` | 🟡 Moderate | Race condition |
| 6 | `platform.py:828-829` | No SIGKILL when SIGTERM fails in `clean_log_recovery_process()` | 🟡 Moderate | Zombie process |
| 7 | `platform.py:967` | Wrong exception caught in `wait_mandatory_time()` | 🟢 Minor | Dead code / misleading |
| 8 | `platform.py:151` | `_atexit_registered = None` should be `False` | 🟢 Minor | Semantics |
| 9 | `platform.py:894` | `daemon = True` not set on new process | 🟢 Minor | Regression |

VindeeR:
@dbeltrankyl I did some improvements on some old code that I found it was easy to fix that the mypy was complaining about. and I believe the leaks are all being treated now.

dbeltrankyl (changes requested):
Hello @VindeeR 

Thanks for the work! 

I think the implementations are better than we had (except for one inverted condition, I believe), but I've added additional stuff 

Here is the patch that contains comments about what I found

```

Subject: [PATCH] Reviewed changes
---
Index: autosubmit/platforms/platform.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- a/autosubmit/platforms/platform.py	(revision da5fd8af2ef7ad5ceafcc0f853de12dcbac8753c)
+++ b/autosubmit/platforms/platform.py	(date 1777022393065)
@@ -242,7 +242,8 @@
             return
         with cls.lock:
             try:
-                if event_worker in cls.worker_events:
+                # I believe this was inverted
+                if event_worker not in cls.worker_events:
                     cls.worker_events.append(event_worker)
             except ValueError:
                 pass
@@ -786,7 +787,6 @@
         """
         raise NotImplementedError  # pragma: no cover
 
-
     def add_job_to_log_recover(self, job):
         if job.id and int(job.id) != 0:
             self.recovery_queue.put(job)
@@ -851,13 +851,16 @@
                 Log.warning(f"Queue cleanup failed: {e}")
 
         self.log_retrieval_process_active = False
-        if self.work_event is not None:
-            self.remove_workers(self.work_event)
-
-        self.cleanup_event = None
         self.recovery_queue = None
         self.log_recovery_process = None
-        self.work_event = None
+
+        # Reuse existing Event objects instead of destroying them.
+        # Clearing the events is sufficient to give the next child process a clean initial state.
+        if self.cleanup_event is not None:
+            self.cleanup_event.clear()
+        if self.work_event is not None:
+            self.work_event.clear()
+
         self.processed_wrapper_logs = set()
 
     def load_process_info(self, platform):
@@ -892,9 +895,13 @@
         new_platform = self.create_a_new_copy()
         if self.ctx is None:
             self.ctx = self.get_mp_context()
-        self.work_event = self.ctx.Event()
-        self.cleanup_event = self.ctx.Event()
-        Platform.update_workers(self.work_event)
+        # Allocate Events only on the first call.
+        # On subsequent respawn cycles the events are reused (cleared by clean_log_recovery_process)
+        if self.work_event is None:
+            self.work_event = self.ctx.Event()
+            Platform.update_workers(self.work_event)
+        if self.cleanup_event is None:
+            self.cleanup_event = self.ctx.Event()
         self.load_process_info(new_platform)
         if self.recovery_queue is not None:
             self.recovery_queue.close()
@@ -933,8 +940,8 @@
             if ret_pid == 0:  # Process is still running
                 Log.info(f"Process {self.log_recovery_process.pid} is still running.")
                 # Optionally, handle cleanup here (terminate / join)
-            else:
-                Log.result(f"Process {self.log_recovery_process.name} finished with pid {self.log_recovery_process.pid}")
+        else:
+            Log.result(f"Process {self.log_recovery_process.name} finished with pid {self.log_recovery_process.pid}")
 
     def spawn_log_retrieval_process(self, as_conf: Optional['AutosubmitConfig']) -> None:
         """Spawns a process to recover the logs of the jobs that have been completed on this platform.
@@ -967,6 +974,11 @@
                 self.log_recovery_process.terminate()
                 self.log_recovery_process.join(timeout=60)
 
+        if self.work_event is not None:
+            self.remove_workers(self.work_event)
+            self.work_event = None
+        self.cleanup_event = None
+
     def wait_mandatory_time(self, sleep_time: int = 60) -> bool:
         """Waits for the work_event to be set or the cleanup_event to be set for a mandatory time.
```
 

 Can you do a short dummy run climate-dt in hubs and check that we don't have regression zombies after the changes? Use mn5, ps ( mn5 login) and local platforms

If it looks good, we can merge it :)

VindeeR:
I've ran I small dummy experiment as you suggested and it looks alright I believe the leaks have been fixed.

```
================================================================================
                Memory, object and file descriptor by iteration                 
================================================================================
Iteration 1: Memory: 112.57 MiB
Iteration 1: Objects: 128186
Iteration 1: File Descriptors: 10
Iteration 1: Loaded jobs: 0
Iteration 1: Loaded edges: 0
Iteration 2: Memory: 115.07 MiB
Iteration 2: Objects: 130642
Iteration 2: File Descriptors: 11
Iteration 2: Loaded jobs: 8
Iteration 2: Loaded edges: 7
Iteration 2: Opened file descriptor: [fd=9] /esarchive/autosubmit/aawf/tmp/autosubmit.lock
Iteration 3: Memory: 115.82 MiB
Iteration 3: Objects: 131084
Iteration 3: File Descriptors: 16
Iteration 3: Loaded jobs: 8
Iteration 3: Loaded edges: 7
Iteration 3: Opened file descriptor: [fd=10] pipe (read) pipe:[5447608]
Iteration 3: Opened file descriptor: [fd=11] pipe (write) pipe:[5447606]
Iteration 3: Opened file descriptor: [fd=12] pipe (write) pipe:[5447608]
Iteration 3: Opened file descriptor: [fd=13] pipe (read) pipe:[5447609]
Iteration 3: Opened file descriptor: [fd=16] pipe (write) pipe:[5447610]
Iteration 4: Memory: 115.82 MiB
Iteration 4: Objects: 131171
Iteration 4: File Descriptors: 16
Iteration 4: Loaded jobs: 8
Iteration 4: Loaded edges: 7
Iteration 5: Memory: 115.82 MiB
Iteration 5: Objects: 131172
Iteration 5: File Descriptors: 16
Iteration 5: Loaded jobs: 8
Iteration 5: Loaded edges: 7
Iteration 6: Memory: 115.82 MiB
Iteration 6: Objects: 131173
Iteration 6: File Descriptors: 16
Iteration 6: Loaded jobs: 8
Iteration 6: Loaded edges: 7
Iteration 7: Memory: 115.82 MiB
Iteration 7: Objects: 131174
Iteration 7: File Descriptors: 16
Iteration 7: Loaded jobs: 8
Iteration 7: Loaded edges: 7
Iteration 8: Memory: 115.82 MiB
Iteration 8: Objects: 131175
Iteration 8: File Descriptors: 16
Iteration 8: Loaded jobs: 8
Iteration 8: Loaded edges: 7
Iteration 9: Memory: 115.82 MiB
Iteration 9: Objects: 131176
Iteration 9: File Descriptors: 16
Iteration 9: Loaded jobs: 8
Iteration 9: Loaded edges: 7

================================================================================
               Overall Memory, Object and File Descriptor Growth                
================================================================================

MEMORY GROW: 3.25 MiB.
INITIAL MEMORY: 112.57 MiB.
FINAL MEMORY: 115.82 MiB.
OBJECTS GROW: 3019 objects.
FILE DESCRIPTORS GROW: 6 file descriptors.

FINAL OPEN FILE DESCRIPTORS:
  [fd=0] stdin (/dev/pts/4)
  [fd=1] stdout (/dev/pts/4)
  [fd=2] stderr (/dev/pts/4)
  [fd=3] /var/lib/sss/mc/passwd
  [fd=4] unix-socket: (unnamed)
  [fd=5] /esarchive/autosubmit/aawf/tmp/ASLOGS/20260427_094223_run.log
  [fd=6] /esarchive/autosubmit/aawf/tmp/ASLOGS/20260427_094223_run_err.log
  [fd=7] /esarchive/autosubmit/aawf/tmp/ASLOGS/jobs_active_status.log
  [fd=8] /esarchive/autosubmit/aawf/tmp/ASLOGS/jobs_failed_status.log
  [fd=10] pipe (read) pipe:[5447608]
  [fd=11] pipe (write) pipe:[5447606]
  [fd=12] pipe (write) pipe:[5447608]
  [fd=13] pipe (read) pipe:[5447609]
  [fd=16] pipe (write) pipe:[5447610]

```

I'll wait for the CI completion and merge after thanks @dbeltrankyl

PR #2953, opened 2026-04-15
fixed recovery/setstatus issue

### LINKED ISSUE #2894: CSCDESTINCLIMADT-1150: sqlite3.OperationalError: too many SQL variables
### Description

From @youcheng-csc in DestinE JIRA CSCDESTINCLIMADT-1150.

The command `autosubmit setstatus o027 -fs WAITING -t COMPLETED -s -np` for that experiment resulted in this:

```bash
[ERROR] 2026-03-18 18:47:20,309 In Configuration files:
Issues in [Jobs] config file:
[INI] Dependency parameter is invalid, job COMPILE is not configured
[eCode=6013] 
```

and

```bash
sqlite3.OperationalError: too many SQL variables 
```

* Autosubmit version: 4.1.16
* Machine/VM/environment name (if applicable): destine
* Experiment ID (if applicable): o027
* Experiment tasks & log path if applicable): /appl/AS/AUTOSUBMIT_DATA/o027/tmp/ASLOGS/20260318_184715_setstatus_err.log

### Reproducible Example

TODO

### Expected Behaviour

Either set status correctly, or a useful message.

dbeltrankyl:
# Fixes #2894

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/history/database_managers/experiment_history_db_manager.py
+++ b/autosubmit/history/database_managers/experiment_history_db_manager.py
@@ -17,6 +17,7 @@
 
 import os
 import textwrap
+import time
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Generator, Optional, Protocol, cast
@@ -889,14 +890,15 @@ class SqlAlchemyExperimentHistoryDbManager:
         # Local import avoids shadowing any `table` / `column` parameter.
         from sqlalchemy import column as _col, table as _tbl
 
-        tmp = _tbl("_tmp_job_names", _col("job_name"))
+        table_name = f"_tmp_job_names_{time.time_ns()}"
+        tmp = _tbl(table_name, _col("job_name"))
         with self.engine.connect() as conn:
             conn.execute(text(
-                "CREATE TEMPORARY TABLE IF NOT EXISTS _tmp_job_names (job_name TEXT)"
+                f"CREATE TEMPORARY TABLE IF NOT EXISTS {table_name} (job_name TEXT)"
             ))
-            conn.execute(text("DELETE FROM _tmp_job_names"))
+            conn.execute(text(f"DELETE FROM {table_name}"))
             conn.execute(
-                text("INSERT INTO _tmp_job_names (job_name) VALUES (:name)"),
+                text(f"INSERT INTO {table_name} (job_name) VALUES (:name)"),
                 [{"name": n} for n in job_names],
             )
             try:
@@ -905,7 +907,7 @@ class SqlAlchemyExperimentHistoryDbManager:
                 conn.rollback()
                 raise
             finally:
-                conn.execute(text("DROP TABLE IF EXISTS _tmp_job_names"))
+                conn.execute(text(f"DROP TABLE IF EXISTS {table_name}"))
                 conn.commit()
 
     def select_jobs_data(self, table, job_names: list[str]) -> list[tuple[str, Any]]:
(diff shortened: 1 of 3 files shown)


dbeltrankyl:
Warning: The test only covers SQLite, but in 4.2.0, I'll extend it to PostgreSQL

Hello! 

I added two things to fix this... I had some issues reproducing it because the LIMIT depends on the Python and sqlalchemy versions, etc. I set it to 250.000 because it is what Python 3.11 provides... The known limits, 999 or 32xxx, worked with the old version. 


First, changed the approach of the select to use a TEMP-table (taking @LuiggiTenorioK comments into account)

```

Original IN -> crashes depending on the LIMIT
Chunked IN-> O(N / 999), and also for Python <3.11, the limit is obscure
Temp-table JOIN O(N + M × index lookup) ( uses optimized hash join)
```


Then I created a context manager to simplify the code for other selects ( looking at 4.2.0)


Context:

```
_job_names_tmp_table(job_names)
│
├── CREATE TEMPORARY TABLE IF NOT EXISTS _tmp_job_names
├── DELETE FROM _tmp_job_names          
├── INSERT INTO _tmp_job_names ...    
│
│   yield (conn, tmp) 
│
└── finally:
    ├── DROP TABLE IF EXISTS _tmp_job_names
    └── conn.commit()
```

Ready to review! @LuiggiTenorioK or @kinow

autosubmit/history/database_managers/experiment_history_db_manager.py
```diff
@@ -871,15 +872,52 @@ def get_jobs_data_last_row(self, job_names) -> dict[str, Any]:
...
+            conn.execute(text("DELETE FROM _tmp_job_names"))
+            conn.execute(
+                text("INSERT INTO _tmp_job_names (job_name) VALUES (:name)"),
+                [{"name": n} for n in job_names],
+            )
+            try:
+                yield conn, tmp
+            finally:
+                conn.execute(text("DROP TABLE IF EXISTS _tmp_job_names"))
+                conn.commit()
```
LuiggiTenorioK: In the case this function is called many times at the same time, it will populate the same table with all the job names of those calls. It will be better to add some random characters or a timestamp to the name of the table to avoid those collisions. This might be an edge case, but it is quite possible when we have big experiments.

LuiggiTenorioK, line 911: In case the caller of the context manager raises an error, it is risky to just run the commit in the finally block because it could commit changes that are not complete. A better approach is to add a rollback when an exception is raised. In that way, it wipes all operations before dropping the tmp table.


```suggestion
            try:
                yield conn, tmp
            except Exception:
                conn.rollback()
                raise
            finally:
                conn.execute(text("DROP TABLE IF EXISTS _tmp_job_names"))
                conn.commit()
```

This is not urgent since it is called by functions that will do read-only operations for now.

LuiggiTenorioK (commented):
Hi @dbeltrankyl, this looks like a solid solution! Definitely using a temporal table, and using a context manager to ensure dropping the temporal table makes the code cleaner and more efficient. Also, the regression test is nice!

Great job! 💯 

However, I added a few comments for rare cases where this solution could fail.

dbeltrankyl:
Thanks for the feedback @LuiggiTenorioK ! 

committed your suggestion and added time.time_ns() to the table_name

LuiggiTenorioK (approved):

LuiggiTenorioK:
Perfect! Now it looks pretty good :raised_hands:

PR #2969, opened 2026-04-22
Fix ugly and incorrect logs in ASLOGS directory of experiments.

tiggi:
I got tired of reading ASLOGS/ run.log entries like this:

2026-04-20 10:07:34,871 Loading HPC parameters...
2026-04-20 10:07:34,938

38826 of 131101 jobs remaining (10:07)
2026-04-20 10:07:35,063 14 jobs has been  failed (10:07) 2026-04-20 10:07:35,063 Sleep: 10

- Remove the extra double newlines.
- Fix the "jobs has been  " to "jobs have " to make the message legible in English.

No change in logic, just formatting the output.

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
   command line if you want this) #2879
 - Allow to set AS variables as lowercase/uppercase (,,/^^) #2504
 - Slurm tests in CICD are using a new container by @giovtorres #2871
+- Fixed main-loop iteration message handling singular/plural and reducing number of line breaks #2969
 
 ### 4.1.16: Postgres (experimental) support, bug fixes, and enhancements
 
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2088,10 +2088,10 @@ class Autosubmit:
         :return: common parameters for the iteration
         """
         total_jobs = len(job_list.get_job_list())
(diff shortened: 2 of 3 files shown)


dmcclainbsc:
I like the change but I'll note that if exactly 1 job has failed, then it should be "1 job has failed" rather than have.

kinow:
> .

Good point, @dmcclainbsc . I can add some quick code that checks the number and uses has/have. But what if we use "N failed jobs" instead? That may simplify, and avoids having to write code + tests for this :grimacing:

dmcclainbsc:
It would still be incorrect to say "1 failed jobs" but maybe you can say "N failed job(s)"

kinow:
Ah, dang. Let me add that function then. Thanks Damien!

dmcclainbsc:
Just to be clear, it's totally acceptable in English to say "job(s)" (indicating the s may or may not apply) though slightly less clean looking than putting in the work with the conditional.

kinow:
> Just to be clear, it's totally acceptable in English to say "job(s)" (indicating the s may or may not apply) though slightly less clean looking than putting in the work with the conditional.

Yup, but it won't be too difficult to match the singular/plural, and I was going to add a test anyway. Thanks Damien!

tiggi:
+1 For job(s). The old style of always messing up singular/plural was even worse.

kinow (approved):

kinow (approved):

PR #2971, opened 2026-04-22
Refactored part of the filter dependencies, Added previous-N and fixes #29

### LINKED ISSUE #29: delete experiment
In GitLab by @dmanubens on May 21, 2015, 17:16

Delte should test first if the folder can be deleted (no permission denied).
After delete from database.

The following current behaviour is wrong:

``` 
dmanubens@odin~ /cfu/autosubmit~$ autosubmit delete test0004
Do you want to delete test0004 ? [y/n]
y
Removing experiment directory...
[WARNING] Can not delete experiment folder: [Errno 13] Permission denied: '/cfu/autosubmit/test0004/conf/jobs_test0004.conf'
Deleting experiment from database...
Experiment test0004 deleted

```

### LINKED ISSUE #2958: Split-dependencies in different split sizes
### Description

Hi my beloved AS developers! I am encountering an issue while trying to set a dependency structure under splits that looks like this:

```
DN_1_1  -- DN_1_2 --   DN_1_3  -- DN_1_4        (...)       DN_1_31
              |                       |                        |
           OPA_1_1       --       OPA_1_2         (...) OPA_1_16
```
(from original issue https://gitlab.earth.bsc.es/digital-twins/de_340-2/workflow/-/issues/1292 )

* Autosubmit version: 4.1.16
* Machine/VM/environment name (if applicable): does not apply
* Experiment ID (if applicable): does not apply
  * AS create 

very much related with this previous issue https://github.com/BSC-ES/autosubmit/issues/2631

### Reproducible Example

joblist I am using:

```
JOBS:
  LOCAL_SETUP:
    FILE: templates/local_setup.sh
    PLATFORM: LOCAL
    RUNNING: once
    CHECK: on_submission
    NOTIFY_ON:
    NOTIFY_ON:
    - FAILED
    DEPENDENCIES: {}
    ADDITIONAL_FILES: []
  SYNCHRONIZE:
    FILE: templates/synchronize.sh
    PLATFORM: LOCAL
    DEPENDENCIES:
      LOCAL_SETUP: {}
    RUNNING: once
    CHECK: on_submission
    NOTIFY_ON:
    - FAILED
    ADDITIONAL_FILES: []
  REMOTE_SETUP:
    FILE: templates/remote_setup/remote_setup.sh
    PLATFORM: MARENOSTRUM5-login
    DEPENDENCIES:
      SYNCHRONIZE: {}
    RUNNING: once
    WALLCLOCK: 02:00
    CHECK: on_submission
    NOTIFY_ON:
    - FAILED
    ADDITIONAL_FILES: []
  DN:
    FILE: templates/dn.sh
    DEPENDENCIES:
      REMOTE_SETUP:
      DN:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: previous
      DN-1:
    RUNNING: chunk
    WALLCLOCK: 02:00
    PLATFORM: MARENOSTRUM5-readdatabridge
    SPLITS: auto
    TOTALJOBS: 1
    CHECK: on_submission
    RETRIALS: 1
    NOTIFY_ON:
    - FAILED
    ADDITIONAL_FILES:
    - conf/applications/request/request_ENERGY_INDICATORS.yml
    - conf/applications/request/request_ENERGY_OFFSHORE.yml
    - conf/applications/request/request_HYDROLAND.yml
    - conf/applications/request/request_HYDROMET.yml
    - conf/applications/request/request_WILDFIRES_FWI.yml
    - conf/applications/request/request_WILDFIRES_WISE.yml
    - conf/applications/request/request_OBSALL.yml
    - conf/applications/request/request_DATA.yml
    SPLITSIZE: 1
  OPA_ENERGY_INDICATORS:
    DEPENDENCIES:
      DN:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: '[1:auto]*\2'
      OPA_ENERGY_INDICATORS:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: previous
      OPA_ENERGY_INDICATORS-1: {}
    FILE: templates/opa.sh
    PLATFORM: MARENOSTRUM5
    PARTITION: '%CURRENT_APP_PARTITION%'
    RUNNING: chunk
    NODES: 1
    PROCESSORS: 1
    THREADS: '%CURRENT_OPA_PROCESSES%'
    CHECK: on_submission
    CUSTOM_DIRECTIVES: '%CURRENT_OPA_CUSTOM_DIRECTIVES%'
    EXCLUSIVE: '%CURRENT_OPA_EXCLUSIVE%'
    RETRIALS: 1
    WALLCLOCK: 00:20
    NOTIFY_ON:
    - FAILED
    SPLITS: auto
    SPLITSIZE: 2
    ADDITIONAL_FILES:
    - conf/applications/request/request_ENERGY_INDICATORS.yml

```

The graph I am getting is:

Image

that looks good but at some point (split 15) the dependencies get broken. The behaviour repeats for each chunk.

Image

### Expected Behaviour

Dependencies to follow the behaviour that it already shows in the first half of the chunk for the whole chunk, as described here:

```
DN_1_1  -- DN_1_2 --   DN_1_3  -- DN_1_4        (...)       DN_1_31
              |                       |                        |
           OPA_1_1       --       OPA_1_2     --      (...) OPA_1_16
```

Thanks!

dbeltrankyl:
Fixes #2958 cc: @franra9 

Refactors the code touched and  adds -N  to the previous keyword, which was discussed here https://github.com/BSC-ES/autosubmit/discussions/2808 because I remembered it and the function was the same one that handles #2958 cc: @lauherrero  

**Check List**

- [x] Documentation updated.


dbeltrankyl:
Added also regression tests

dbeltrankyl:
Docs added, regression test missed the ref_workflow

I'm not sure that the docs are showing the figures, but it seems not to be related to this PR changes

Ready to review cc: @kinow , @ntorqulu ( at the end, this doesn't touch the same code you were touching about the splits, but if you want to review it, feel free)

franra9:
Thanks @dbeltrankyl ! I can not wait to try it! :)

test/unit/test_dependencies.py
```diff
@@ -762,6 +763,81 @@ def test_job_dict_get_jobs_filtered(mocker, joblist):
...
+])
+def test_get_jobs_filtered_grouped_splits(parent_splits, child_splits, slice_size, as_conf):
+    """Test grouped split dependency mapping (*\\N) for N-to-1, 1-to-1, and 1-to-N ratios.
+
+    Verifies that every child split resolves to the correct parent split(s) regardless
+    of the parent-to-child split ratio and slice size.  The SPLITS_TO filter is set to
+    the minimal form ``1*\\<slice_size>`` so the only information used is the slice size
+    encoded in the ``*\\N`` suffix; the actual index computation must be purely arithmetic.
+    """
+    from datetime import datetime as dt
```
kinow: datetime is already imported in this file. Maybe we can use that one instead of renaming the import here?

test/unit/test_dependencies.py
```diff
@@ -3,6 +3,7 @@
 from datetime import datetime
 from mock.mock import MagicMock
 from networkx import DiGraph
+from datetime import datetime as dt
```
kinow: Datetime is already imported three lines above. Maybe we don't need this one aliasing the import (I think datetime is not normally renamed as numpy, xarray, etc.)

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+        :param jobs_aux: Accumulator for unresolved nested dicts.
+        :return: Updated ``jobs_aux``.
+        """
+        for key in keys:
+            value = jobs.get(key, None)
+            if value is None:
+                continue
+            if type(value) is list:
+                final_jobs_list.extend(value)
+            elif type(value) is Job:
```
kinow: Is `Job` a valid type? Codecov shows this line as not covered, but looking at the only place where this private function gets called, I think it only sends the `.keys()` from a dict (likely strings?), and the `natural_key` object from another function, which I think might be a list or dict? If so, we could also improve the type to be something like `dict[str, Optional[list, dict]]`, which would help mypy to prevent issues with this function in the future when we have mypy fixed in most of the code base.

Image

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+        :param natural_key: The job's own key at this level (date, member, or chunk).
+        :param use_all_keys: When True and no filter is set, iterate every key (e.g. ``running==once``).
+        :param final_jobs_list: Accumulator for resolved Job objects; mutated in place.
+        :return: The merged nested dict for the next traversal level (empty when this level is terminal).
+        """
+        jobs_aux: dict = {}
+        if filter_value:
+            filter_lower = filter_value.lower()
+            if "none" in filter_lower:
+                return {}
```
kinow, line 287: I think we are missing a test with the filter "none", by looking at codecov.

Image

kinow (commented):
I am still reviewing the job dict file, @dbeltrankyl but you can have a look at my partial review for now. I have two meetings now, and a new oper issue in destine but I will try to finish reviewing it today.

kinow:
(Oh, there are some Git conflicts too)

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+            elif "all" in filter_lower:
+                return self._collect_by_keys(jobs, jobs.keys(), final_jobs_list, jobs_aux)
+            else:
+                return self._collect_by_keys(jobs, explicit_keys, final_jobs_list, jobs_aux)
+        else:
+            if use_all_keys:
+                return self._collect_by_keys(jobs, jobs.keys(), final_jobs_list, jobs_aux)
+            elif natural_key is not None and jobs.get(natural_key, None):
+                return self._collect_by_keys(jobs, [natural_key], final_jobs_list, jobs_aux)
```
kinow, line 296: Also not tested, and I couldn't find it in the diff (i.e. I think this is new code, not existing code?). If it's really new, and it's not too difficult to include a test for this it'd be good, I think.

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+                        if type(result_special) is list:
+                            result_special = [result_special[-1]] if result_special else []
+            else:  # 1-to-1 mapping (*)
+                split_index = job.split - 1
+                result_special = [final_jobs_list[split_index]] if split_index < len(final_jobs_list) else []
+
+        if type(result_special) is not list:
+            result_special = [result_special]
+        if type(result_normal) is not list:
+            result_normal = [result_normal]
```
kinow: Are these two `if` above possible? Reading the code above, `result_special` and `result_normal` are either the default empty list, or lists created with `var = [ ... ]`, or they are the given `final_jobs_list` arg. I think mypy will alert us if the `final_jobs_list` is being passed as anything that's not a list. As we are not using it widely, we can add a check in the beginning of the function to ensure it's always a list, and then we can drop these two if's at the end, I think.

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+        if "*" in one_to_one_splits and not skip_one_to_one:
+            if not final_jobs_list:
+                return list(set(result_normal))
+            easier_to_filter = "," + one_to_one_splits + ","
+            slice_matches = re.findall(r"\\[0-9]+", easier_to_filter)
+            if slice_matches:  # N-to-1 or 1-to-N grouped mapping (*\N)
+                split_slice = int(slice_matches[0].split("\\")[1])
+                is_n_to_1 = int(job.splits) <= int(final_jobs_list[0].splits)
+                if is_n_to_1:
+                    split_index = (job.split - 1) * split_slice
```
ntorqulu: Is it possible to have a negative split index reached this stage?
Currently in the jobs yaml configuration file you can define `SPLITS_FROM` and `SPLITS_TO` without defining the number of splits `SPLITS`.
So a job can have `job.split` set to -1 (no splits) and obtain a negative `split_index`, failing silently or with an error difficult to understand.
I don't know if it's useful to add a safeguard for this case, in which SPLITS is not defined and `SPLITS_TO` uses the star mapping.

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+            ]
+
+        result_special: list = []
+        if "*" in one_to_one_splits and not skip_one_to_one:
+            if not final_jobs_list:
+                return list(set(result_normal))
+            easier_to_filter = "," + one_to_one_splits + ","
+            slice_matches = re.findall(r"\\[0-9]+", easier_to_filter)
+            if slice_matches:  # N-to-1 or 1-to-N grouped mapping (*\N)
+                split_slice = int(slice_matches[0].split("\\")[1])
```
ntorqulu: Is it necessary to add a safeguard when `split_slice` is 0?
E.g. currently you obtain a critical error that is difficult to understand when executing `SPLITS_TO=1*\0`.

test/unit/test_dependencies.py
```diff
@@ -762,6 +763,81 @@ def test_job_dict_get_jobs_filtered(mocker, joblist):
...
+        },
+        'JOBS': {
+            'PARENT': {
+                'FILE': 'parent.sh', 'RUNNING': 'chunk',
+                'SPLITS': parent_splits, 'WALLCLOCK': '00:10',
+            },
+        },
+    }
+    as_conf.experiment_data["JOBS"] = as_conf.experiment_data["JOBS"]
+    as_conf.experiment_data["JOBS"] = as_conf.experiment_data["JOBS"]
```
ntorqulu: Should be assigned just one time?

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -686,13 +686,72 @@ Example 3: 1-to-N dependency
...
+    :caption: Example showing ``previous`` (``previous-1``) and ``previous-2`` split dependencies.
+
 Job Splits with calendar
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 For jobs running at any level, it may be useful to split each task into different parts based on the calendar.
 This behaviour can be achieved setting the ``SPLITS: auto`` and using the ``%EXPERIMENT.SPLITSIZE%`` and ``%EXPERIMENT.SPLITSIZEUNIT%`` variables.
 
-Example4: Auto split
+Example5: Auto split
```
ntorqulu: Could you keep the same format as with the previous example, with a space in between?

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+                            result_special = [result_special[-1]] if result_special else []
+            else:  # 1-to-1 mapping (*)
+                split_index = job.split - 1
+                result_special = [final_jobs_list[split_index]] if split_index < len(final_jobs_list) else []
+
+        if type(result_special) is not list:
+            result_special = [result_special]
+        if type(result_normal) is not list:
+            result_normal = [result_normal]
+        return list(set(result_normal + result_special))
```
kinow: Can't we simplify a bit `_apply_splits_filter`? It looks like the function returns a set converted into a list. Maybe it could return the `set` directly, or just compute everything using a set instead of two lists and creating the unique list.

It looks to me like if the filter is `all`, then we just use the `final_jobs_list`, and sets a flag skip_one_to_one`. Maybe we can just return immediately at the beginning of the function. Same with the `none` value. When it's enabled, looks like we just use a new list based on `final_jobs_list`. So we can short-circuit which I think simplifies a bit the logic for reading the code, e.g.,

```
def fn(...):
    split_filter = split_filter.lower()

    if "all" in split_filter:
         return final_jobs_list
    if "none" in split_filter:
        return [f for f in final_jobs_list if (f.split or 0) in (-1, 0) and f.name != job.name]

    # continue
```

This also eliminates the need to have the `skip_one_to_one` variable, and keeps the logic more linear, I think. With less or simpler branches/conditionals.

If we move the logic to handle the `*` to its own function (something like `def _resolve_star_splits`), then this function would really only apply splits filter, and could be refactored to be simpler, something like this, I think:

```py
    def _apply_splits_filter(
        self,
        final_jobs_list: list,
        job: "Job",
        split_filter: str,
        filters_to_of_parent: dict,
    ) -> list:
        split_filter = split_filter.lower()

        if "all" in split_filter:
            return final_jobs_list
        if "none" in split_filter:
            return [
                f for f in final_jobs_list
                if (f.split is None or f.split in (-1, 0)) and f.name != job.name
            ]

        results = set()
        parts = [p.strip() for p in split_filter.split(",")]

        for part in parts:
            # 1-to-1 or N-M mapping with *
            if '*' in part:
                # TODO: move the logic for star-splitting to a function, and return a list/set.
                results.update(self._resolve_start_splits(final_jobs_list, job, part, filters_to_of_parent))
            # previous-N
            elif 'previous' in part:
                match = re.search(r"previous(?:-(\d+))?", part)
                steps = int(match.group(1)) if match and match.group(1) else 1
                results.update(
                    f for f in final_jobs_list
                    if job.split and f.split == job.split - steps and f.name != job.name
                )
            else:
                # Natural splits
                val = str(job.split) if part == "natural" else part
                results.update(
                    f for f in final_jobs_list
                    if (f.name != job.name) and (str(f.split) == val or (f.split or 0) <= 0)
                )

        return list(results)
```

ntorqulu (commented):
Hi @dbeltrankyl! I've reviewed it focusing on some edge cases I found while trying to create experiments with splits.

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+
+        #  splits
+        split_filter = filters_to.get("SPLITS_TO", None)
+        if final_jobs_list and split_filter:
+            final_jobs_list = self._apply_splits_filter(
+                final_jobs_list, job, split_filter, filters_to_of_parent
+            )
+
         if type(final_jobs_list) is not list:
             return [final_jobs_list]
```
kinow: I think here we are quite sure `final_jobs_list` should be a list. Is this still needed? I think the only place where it's recreated without an `[ ]` is when it's set to the result of `self._apply_splits_filter`, which we know it's a `list`.

autosubmit/job/job_dict.py
```diff
@@ -233,232 +233,242 @@ def update_jobs_filtered(self, current_jobs, next_level_jobs):
...
+                )
+                natural_member_key = natural_member.upper() if natural_member else None
+                jobs = self._filter_level(
+                    jobs, member_filter, explicit_members, natural_member_key,
+                    use_all_keys=(job.running == "once" or not job.member),
+                    final_jobs_list=final_jobs_list,
+                )
+
+        # chunk
         if len(jobs) > 0:
```
kinow: This if len(jobs) > 0 is repeated three times. Within each of `if`, we perform one operation, handling jobs filtered at different levels. Wouldn't be better to move this to one or multiple functions, and simplify this function `get_jobs_filtered`? It might be easier to test too.

We can enforce the type of variables too, as the three times after checking the length of `jobs` we check the type of `jobs` and convert it to dict if it's of type list. That can be done once, or we can use the function signature + mypy to enforce a certain type (or an if right at the beginning of the function until we get mypy working as expected in the CICD).

dbeltrankyl:
Hello @kinow , @ntorqulu 

Sorry for the delay

Thanks for the feedback. I think all of them were valid concerns

I've applied the changes based on the feedback. If the pipeline passes, I think it may be ready to merge

dbeltrankyl:
Hello @kinow , @ntorqulu 

All feedback has been addressed the day before my week off... I see that there are no objections since then, so I believe this is ready to merge?. The latest commit only resolves the changelog conflict.

CHANGELOG.md
```diff
@@ -55,6 +58,8 @@
 - Added batch canceling to the `setstatus` command #2919
 - verbosity adjusted #2976
 - Implement filters for `recovery` command #2923
+- Enchanced previous to accept previous-N #
```
kinow: Missing issue number? Or extra # that can be removed?

autosubmit/job/job_dict.py
kinow: Has this one been fixed, @dbeltrankyl ?

kinow (commented):
@dbeltrankyl I had a look, and everything looks in order. Only the tests seem to have missed a part of the new code? Maybe the GH Actions pipeline running as I write this will update that number?

CHANGELOG.md
dbeltrankyl: Right

dbeltrankyl:
Added more tests waiting for the coverage results...

dbeltrankyl:
I guess it was temporal but it is the first time encountering this error in the CI/CD: 


`e = HTTPError('404 Client Error: Not Found for url: http+docker://localhost/v1.48/containers/create')`

kinow:
> I guess it was temporal but it is the first time encountering this error in the CI/CD:
> 
> `e = HTTPError('404 Client Error: Not Found for url: http+docker://localhost/v1.48/containers/create')`

Haven't had that on GH Actions, but when I was trying to rework the tests, and trigger **a lot** of pytest + Docker containers, I hit some 404 errors with the Docker client. Trying again worked, so I never investigated the logs in my system to see if there was anything useful. Good to know it happened, @dbeltrankyl , I'll keep an eye if I see this in one of my PRs. Thanks!

kinow:
> Added more tests waiting for the coverage results...

We don't need 100% for everything in this PR, but at least the new code in that if/else looks useful to have a test or two. Thanks Dani!

dbeltrankyl:
Only 1 partially missed, so I guess it is good enough?

kinow:
> Only 1 partially missed, so I guess it is good enough?

I think so,

image

I think it must be missing when the `job` in the dict is not `None`, and not a `list` nor a `dict`. Is that allowed? If not we can raise a warning or log? As I think when that happens we just keep iterating, and eventually return `jobs_aux`.

dbeltrankyl:
I think it is not possible to have a different value than dict or list in the current code so I've changed it to if else

kinow (approved):
I think you will get all the if/else in the new function covered, @dbeltrankyl . Leaving it approved here for when the pipeline succeeds. Thanks!

PR #2979, opened 2026-04-23
Fix estimated cores in autosubmit stats

### LINKED ISSUE #2978: Fix autosubmit stats to use processing elements instead of requested processors
### Description

As reported in Mattermost about the GUI/API (https://github.com/BSC-ES/autosubmit-api/issues/281), the `autosubmit stats` command doesn't calculate the `PROCESSORS` based on other settings like `NODES` and `PROCESSORS_PER_NODE`.

We have to update it to use the values that are available in the YAML files. At least for the current run.

* Autosubmit version: 4.1.16

LuiggiTenorioK:
Closes #2978

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Removed invalid message about the template script showing the error with a difference of ~5 lines #2962
 - Fix queuing time for the first vertical job
 - Fixes an issue with SUSPENDED parents being ignored when STATUS: RUNNING is set in the dependencies #2960
+- Fixed `autosubmit stats` command that was not considering parameters defined in the platform, like NODES, TASKS, etc. #2978
 
 **Enhancements:**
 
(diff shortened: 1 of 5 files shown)


kinow (approved):
The fix looks simple, and I found nothing that appears to be wrong in the updated code. Thanks for adding the test as well! +1

LuiggiTenorioK:
> The fix looks simple, and I found nothing that appears to be wrong in the updated code. Thanks for adding the test as well! +1

Yes, it is pretty simple. Everything else was tested, but the right parameters (NODES, PROCESSORS_PER_NODE, etc) never arrived to the statistics module to calculate the cores :sweat_smile:

kinow:
Good catch! Feel free to merge whenever you want it then. Thanks Luiggi

Issue #1286, opened 2024-04-30
DDBB migration framework setup


kinow:
In GitLab by @LuiggiTenorioK on Apr 30, 2024, 11:15

Eventually, we have to manage DDBB migrations to enable a maintainable way to change the DDBB schema. This will be important after we complete the Postgres support #1285.

Because we are going to use SQLAlchemy to deal with the multiple DDBBs, we might use this tool for managing them https://alembic.sqlalchemy.org/en/latest/

@kinow

kinow:
Never implemented Alembic in a project, the only tools I used were Flyway, Liquibase, Spring Java tools, and custom migration (just a table `migrations` with `v1`, `v2`, etc., and then a tool that looks for `migrations/v1.sql`, and compares with the latest/current migration applied...)

kinow:
And an example of the custom way Cylc does DB migrations when upgrading the Cylc version (different than what I remembered!):

https://github.com/cylc/cylc-flow/blob/f41602137f5f251c4fdb90d7f34223ac9790b6bd/cylc/flow/workflow_db_mgr.py#L727-L759

dbeltrankyl:
Hello @kinow 

From #3114 issue.

Since that PR will be big (I plan to do it in grouped commits to be easier to review) and you and Luiggi mentioned two possibilities ( custom vs using alembic), what I'm doing is to prepare the db structure to be tool agnostic  and adding a TODO:

 TODO(#1286): a migration tool should eventually take over this: 



Shortly, the custom runner is simple, but every migration is manual. Alembic would solve the writing migrations, but needs extra setup to work with our per-experiment databases.

Personally, I would go for alembic because autogenerates migration scripts by diffing the models against the DB, orders them, and handles SQLite ALTER via batch mode but in another PR

kinow:
Hi @dbeltrankyl 

> Personally, I would go for alembic because autogenerates migration scripts by diffing the models against the DB, orders them, and handles SQLite ALTER via batch mode but in another PR

Yeah me too, but I think you are correct here too,

> Alembic would solve the writing migrations, but needs extra setup to work with our per-experiment databases.

That per-experiment database is a bit of a pain to maintain and work around it. At some point we will have to address that... not sure if that will be fixed first, or the migrations with a tool, etc.

[...]

Issue #2957, opened 2026-04-16
Use enums instead of strings/ints


kinow:
We have several parts of our code that are still using strings/ints for set of values (chunkunit for instance). It'd be simpler and less error-prone to use enums instead.

kinow:
Re-opening as we added chunk unit enums. There are heaps of other data that needs to be moved too. I'm doing ExecutionMode & PlatformType in #3086 

But I think having a central enum type may backfire. It's common to have `project.package.exceptions` and `project.package.enums` in small code base. In medium/large code bases, it's normally preferably to have the data modelled/designed where it makes more sense.

This is useful, for instance, if we need to break it into modules, or refactor. As you don't have to modify `project.package.enums`, but instead you modify something like `autosubmit.jobs.enums` or something like that.

Then, if you have an important pull request changing a lot of files (e.g., the 4.2.0 pull request we have now, the old performance optimization we had some years ago, or one of the large refactoring PRs we have sometimes), the chances of having Git conflicts, or changes blocked by another PR, are reduced because you don't have a central large file with all exceptions and enums.

If you ever need to move things out (e.g., we had a config-parser in the past, with all config related code) it's easier too, as things are within its own package), etc.

kinow:
Re-opening as there are still parts of the code where enums are needed 👍

rajeshsuraj:
I think this got closed again due to me updating the my issues roadmap. my bad

kinow:
I imagined that, @rajeshsuraj :-) Not a problem. Thanks again for working on this!!!! 🙇

ntorqulu:
`get_job_status` has an `if/elif` set of chains that could be substituted by:
```
def get_job_status(status: str) -> int | None:
    """Convert a status key to its integer code. Return None if status is unknown."""
    return Status.KEY_TO_VALUE.get(status.upper())
```
Taking into account that `Status` already has this information in the mapping.

Or better, make it a `Status` classmethod and drop the function:
```
@classmethod
def from_key(cls, key: str) -> int | None:
    """Convert a status key to its integer code. Return None if status is unknown."""
    return cls.KEY_TO_VALUE.get(key.upper())
```

This avoid having to manually update the constant and dicts.

kinow:
Just pasting a comment from @ntorqulu here, about the function `get_job_status` (was `_get_status` in `autosubmit.py` prior to #3171 )

[...]

Issue #2641, opened 2025-10-09
AS master is not creating the experiment_status table


kinow:
### Description

Found by @LuiggiTenorioK while writing the migration script for the database. It's created by the API at the moment, but if it doesn't exist then Autosubmit does not update the experiment status (so it's probably better that Autosubmit creates since it uses it to write data).

* Autosubmit version: master only, after the postgres branch merge
* Machine/VM/environment name (if applicable): climatedt
* Experiment ID (if applicable): NA

ntorqulu:
Things to consider for this issue:
- `install` should create `autosubmit.db` and `as_times.db` if they don't exist. Currently it only creates `autosubmit.db` and it leaves out table `description`. It should initialize both with all the tables in the db.
- It should not raise an error if they don't exists, I would log just a warning to prevent from exiting early. E.g. if `autosubmit.db` exists but `as_times.db` does not exist, log a warning that `autosbumit.db` is already created and create `as_times.db`.

dbeltrankyl:
I'm not sure if this issue is similar to the fix in #3175

dbeltrankyl:
Ah, no, this table is in the general db, not the experiment one, nvm then

PR #2980, opened 2026-04-24
Refactor experiment status traceability

### LINKED ISSUE #2049: Write keep alive signal in the DDBB
If I am not wrong, the API currently checks if an experiment is running by looking at the modification date of the PKL file

I think that we should make Autosubmit write a "keep alive" timestamp every 5 minutes in the DDBB either. This way the API does not need to have access to the filesystem, something that will give us more flexibility in the future

@dbeltrankyl @LuiggiTenorioK @kinow

### LINKED ISSUE #2137: Extend experiment possible statuses to determine ARCHIVE, or DELETED state
After merging https://github.com/BSC-ES/autosubmit-api/pull/165, it raised a question of how to determine if an experiment is archived or deleted.

So, additionally to what has been proposed in #2049, we can also extend the `experiment_status` table to have the status `ARCHIVED` or `DELETED` (and not only `RUNNING` or `NOT RUNNING`).

This will give better feedback to the user which has been already requested on https://github.com/BSC-ES/autosubmit-gui/issues/225.

ntorqulu:
Closes #2049, #2137
Related to https://github.com/BSC-ES/autosubmit/issues/2957: `RunningStatus` is now an enum

**Things done**
- Extended available experiment status in `experiment_status` table. Possible experiment status are: `NOT_RUNNING`, `RUNNING`, `ARCHIVED` and `DELETED`.
- Experiments created (`expid` and `create`) are marked as `NOT_RUNNING`. Implementation left open if we want to add more states in the future.
- Keep traceability of deleted experiments in both `autosubmit.db` and `as_times.db`. Once the experiment is deleted, it's kept in the database with status `DELETED` (tombstone).
- Added column `last_heartbeat` to table `experiment_status`. Each experiment creates a daemon thread (autosubmit-heartbeat-{expid}) that updates the `last_heartbeat` timestamp every 2 min while the experiment is running.
- Enforce unique indx on `name` in `experiment_status` table, kept most recent one.

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2177,7 +2177,7 @@ class Autosubmit:
         """
         Autosubmit.exit = False
         status_tracker = ExperimentStatus(expid)
-        experiment_status = Optional[str] = None
+        experiment_status: Optional[str] = None
         # Start profiling if the flag has been used
         if profile is not None:
             from .profiler.profiler import Profiler
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -48,6 +48,7 @@ from pyparsing import nestedExpr
 from ruamel.yaml import YAML
 
 import autosubmit.helpers.autosubmit_helper as AutosubmitHelper
+from autosubmit.helpers.processes import process_id
 import autosubmit.statistics.utils as StatisticsUtils
 from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.config.configcommon import AutosubmitConfig
@@ -4137,6 +4138,9 @@ class Autosubmit:
         :return: ``True`` if the experiment has been successfully archived. ``False`` otherwise.
         :rtype: bool
         """
+        if process_id(expid) is not None:
+            raise AutosubmitCritical("Ensure no processes are running in the experiment directory", 7076)
+        
         exp_folder = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(expid)
 
         if not noclean:
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1918,13 +1918,6 @@ class Autosubmit:
                 # This error is important
             except Exception:
                 pass
-        try:
-            ExperimentStatus(expid).set_as_running()
-        except Exception as e:
-            # Connection to status database ec_earth.db can fail.
-            # API worker will fix the status.
-            Log.debug(f"Autosubmit couldn't set your experiment as running on the autosubmit times database: "
-                      f"{os.path.join(BasicConfig.DB_DIR, BasicConfig.AS_TIMES_DB)}. Exception: {str(e)}", 7003)
         return exp_history
 
     @staticmethod
@@ -2183,6 +2176,8 @@ class Autosubmit:
 
         """
         Autosubmit.exit = False
+        status_tracker = ExperimentStatus(expid)
+        experiment_status = Optional[str] = None
         # Start profiling if the flag has been used
         if profile is not None:
             from .profiler.profiler import Profiler
@@ -2216,6 +2211,15 @@ class Autosubmit:
 
                 as_conf_config = as_conf.experiment_data.get('CONFIG', {})
                 git_operational_check_enabled = as_conf_config.get('GIT_OPERATIONAL_CHECK_ENABLED', True)
+
+                # set experiment as running in the experiment_status database
+                try:
+                    status_tracker.set_as_running()
+                except Exception as e:
+                    # Connection to status database ec_earth.db can fail.
+                    # API worker will fix the status.
+                    Log.debug(f"Autosubmit couldn't set your experiment as running on the autosubmit times database: "
+                            f"{os.path.join(BasicConfig.DB_DIR, BasicConfig.AS_TIMES_DB)}. Exception: {str(e)}", 7003)
 
                 if git_operational_check_enabled:
                     Log.debug('Checking for dirty local Git repository')
@@ -2256,13 +2260,16 @@ class Autosubmit:
 
                 while job_list.get_active():
                     try:
+                        # TODO: heartbeat should be here
                         if profile is not None:
                             Autosubmit.exit = profiler.iteration_checkpoint(loaded_jobs, loaded_edges)
 
                         if Autosubmit.exit:
                             Autosubmit.check_logs_status(job_list, as_conf, new_run=False)
                             if job_list.get_failed():
+                                experiment_status = "FAILED"
                                 return 1
+                            experiment_status = "PAUSED" # TODO: not sure about this one
                             return 0
                         Autosubmit.refresh_log_recovery_process(platforms_to_test, as_conf)
                         for job in job_list.get_ready():
@@ -2456,8 +2463,10 @@ class Autosubmit:
                     p.close_connection()
                 if len(job_list.get_failed()) > 0:
                     Log.info("Some jobs have failed and reached maximum retrials")
+                    experiment_status = "FAILED"
                 else:
                     Log.result("Run successful")
+                    experiment_status = "COMPLETED"
                     if profile is not None:
                         profiler.iteration_checkpoint(loaded_jobs, loaded_edges)
                     # Updating finish time for job data header
@@ -2473,12 +2482,26 @@ class Autosubmit:
                 else:
                     Log.info("ROCRATE not present in experiment YAML configuration. No RO-Crate archive created.")
         except BaseLockException:
+            experiment_status = "FAILED"
             raise
         except AutosubmitCritical:
+            experiment_status = "FAILED"
             raise
         except BaseException:
+            experiment_status = "FAILED"
             raise
         finally:
+            try:
+                # TODO: this should be an enum
+                if experiment_status == "COMPLETED":
+                    status_tracker.set_as_completed()
+                elif experiment_status == "PAUSED":
+                    status_tracker.set_as_paused()
+                elif experiment_status == "FAILED":
+                    status_tracker.set_as_failed()
+            except Exception as e:
+                Log.warning(f"Autosubmit couldn't update the final experiment status for {expid}: {str(e)}", 7003)
+            
             if profile is not None:
                 profiler.stop()
 
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1487,7 +1487,7 @@ class Autosubmit:
         except Exception:
             Log.warning(f"Could not update experiment details for {exp_id}. Omitting this step.")
 
-        ExperimentStatus(exp_id).set_as_not_running()
+        # ExperimentStatus(exp_id).set_as_not_running()
 
         Log.result(f"Experiment {exp_id} created")
         return exp_id
@@ -4496,7 +4496,7 @@ class Autosubmit:
                     Log.warning(
                         "Remember to MODIFY the MODEL config files!")
                     
-                    ExperimentStatus(expid).set_as_not_running()
+                    # ExperimentStatus(expid).set_as_not_running()
                     fh.flush()
                     os.fsync(fh.fileno())
                     if detail:
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -4205,6 +4205,8 @@ class Autosubmit:
                         "Can not remove or rename experiments folder", 7012, str(e))
 
         Log.result("Experiment archived successfully")
+        with suppress(Exception):
+            ExperimentStatus(expid).set_as_archived() # TODO: unsure if we should archive the experiment even with exceptions in the process
         return True
 
     @staticmethod
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1487,6 +1487,8 @@ class Autosubmit:
         except Exception:
             Log.warning(f"Could not update experiment details for {exp_id}. Omitting this step.")
 
+        ExperimentStatus(exp_id).set_as_not_running()
+
         Log.result(f"Experiment {exp_id} created")
         return exp_id
 
@@ -4493,6 +4495,8 @@ class Autosubmit:
                     Log.result("\nJob list created successfully")
                     Log.warning(
                         "Remember to MODIFY the MODEL config files!")
+                    
+                    ExperimentStatus(expid).set_as_not_running()
                     fh.flush()
                     os.fsync(fh.fileno())
                     if detail:
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1488,7 +1488,7 @@ class Autosubmit:
         except Exception:
             Log.warning(f"Could not update experiment details for {exp_id}. Omitting this step.")
 
-        # ExperimentStatus(exp_id).set_as_not_running()
+        ExperimentStatus(exp_id).set_as_not_running()
 
         Log.result(f"Experiment {exp_id} created")
         return exp_id
@@ -4500,7 +4500,7 @@ class Autosubmit:
                     Log.warning(
                         "Remember to MODIFY the MODEL config files!")
                     
-                    # ExperimentStatus(expid).set_as_not_running()
+                    ExperimentStatus(expid).set_as_not_running()
                     fh.flush()
                     os.fsync(fh.fileno())
                     if detail:
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2494,11 +2494,11 @@ class Autosubmit:
             try:
                 # TODO: this should be an enum
                 if experiment_status == "COMPLETED":
-                    status_tracker.set_as_completed()
+                    status_tracker.set_as_not_running()
                 elif experiment_status == "PAUSED":
-                    status_tracker.set_as_paused()
+                    status_tracker.set_as_not_running()
                 elif experiment_status == "FAILED":
-                    status_tracker.set_as_failed()
+                    status_tracker.set_as_not_running()
             except Exception as e:
                 Log.warning(f"Autosubmit couldn't update the final experiment status for {expid}: {str(e)}", 7003)
             
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2269,7 +2269,7 @@ class Autosubmit:
                             if job_list.get_failed():
                                 experiment_status = "FAILED"
                                 return 1
-                            experiment_status = "PAUSED" # TODO: not sure about this one
+                            experiment_status = "PAUSED"
                             return 0
                         Autosubmit.refresh_log_recovery_process(platforms_to_test, as_conf)
                         for job in job_list.get_ready():
@@ -2482,7 +2482,8 @@ class Autosubmit:
                 else:
                     Log.info("ROCRATE not present in experiment YAML configuration. No RO-Crate archive created.")
         except BaseLockException:
-            experiment_status = "FAILED"
+            # Multiple instances of autosubmit running the same experiment or previous instance didn't release the lock file
+            # In both cases, we don't want to overwrite the status of the experiment to avoid errors with the API and GUI
             raise
         except AutosubmitCritical:
             experiment_status = "FAILED"
@@ -2492,7 +2493,6 @@ class Autosubmit:
             raise
         finally:
             try:
-                # TODO: this should be an enum
                 if experiment_status == "COMPLETED":
                     status_tracker.set_as_not_running()
                 elif experiment_status == "PAUSED":
@@ -4205,8 +4205,7 @@ class Autosubmit:
                         "Can not remove or rename experiments folder", 7012, str(e))
 
         Log.result("Experiment archived successfully")
-        with suppress(Exception):
-            ExperimentStatus(expid).set_as_archived() # TODO: unsure if we should archive the experiment even with exceptions in the process
+        ExperimentStatus(expid).set_as_archived()
         return True
 
     @staticmethod
@@ -4263,6 +4262,7 @@ class Autosubmit:
             return False
 
         Log.info("Unpacking finished")
+        ExperimentStatus(experiment_id).set_as_not_running()
 
         try:
             archive_path.unlink()
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1488,7 +1488,7 @@ class Autosubmit:
         except Exception:
             Log.warning(f"Could not update experiment details for {exp_id}. Omitting this step.")
 
-        ExperimentStatus(exp_id).set_as_not_running()
+        # ExperimentStatus(exp_id).set_as_not_running()
 
         Log.result(f"Experiment {exp_id} created")
         return exp_id
@@ -4500,7 +4500,7 @@ class Autosubmit:
                     Log.warning(
                         "Remember to MODIFY the MODEL config files!")
                     
-                    ExperimentStatus(expid).set_as_not_running()
+                    # ExperimentStatus(expid).set_as_not_running()
                     fh.flush()
                     os.fsync(fh.fileno())
                     if detail:
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -2263,7 +2263,7 @@ class Autosubmit:
 
                 while job_list.get_active():
                     try:
-                        # TODO: heartbeat should be here
+                        status_tracker.update_heartbeat()
                         if profile is not None:
                             Autosubmit.exit = profiler.iteration_checkpoint(loaded_jobs, loaded_edges)
 
--- a/autosubmit/database/db_common.py
+++ b/autosubmit/database/db_common.py
@@ -286,11 +286,11 @@ def last_name_used(test=False, operational=False, evaluation=False):
 
 def delete_experiment(experiment_id):
     """
-    Removes experiment from database. Anti-lock version.
+    Marks experiment as deleted in metadata tables. Anti-lock version.
 
     :param experiment_id: experiment identifier
     :type experiment_id: str
-    :return: True if delete is successful
+    :return: True if operation is successful
     :rtype: bool
     """
     fn = _delete_experiment
@@ -554,11 +554,12 @@ def _last_name_used(test=False, operational=False, evaluation=False):
 
 def _delete_experiment(experiment_id):
     """
-    Removes experiment from database
+    Marks an experiment as deleted in status metadata while preserving
+    the experiment row to avoid reusing assigned experiment IDs.
 
     :param experiment_id: experiment identifier
     :type experiment_id: str
-    :return: True if delete is successful
+    :return: True if operation is successful
     :rtype: bool
     """
     check_db()
@@ -569,9 +570,7 @@ def _delete_experiment(experiment_id):
         (conn, cursor) = open_conn()
     except DbException as e:
         raise AutosubmitCritical("Could not establish a connection to database", 7001, str(e))
-    cursor.execute('DELETE FROM experiment '
-                   'WHERE name=:name', {'name': experiment_id})
-    
+    # Keep experiment rows as tombstones to avoid reusing experiment IDs.
     # keep the experiment trace in experiment_status for deleted experiments
     try:
         cursor.execute('UPDATE experiment_status SET status=:status, seconds_diff=:seconds_diff, modified=CURRENT_TIMESTAMP '
@@ -584,9 +583,7 @@ def _delete_experiment(experiment_id):
     except Exception:
         pass
 
-    row = cursor.fetchone()
-    if row is None:
-        Log.debug(f'The experiment {experiment_id} has been deleted!!!')
+    Log.debug(f'The experiment {experiment_id} has been marked as deleted.')
     close_conn(conn, cursor)
     return True
 
@@ -834,13 +831,6 @@ def _delete_experiment_sqlalchemy(experiment_id: str) -> bool:
         return True
 
     with _get_sqlalchemy_conn() as conn:
-        # Delete from experiment table
-        query = delete(tables.ExperimentTable).where(
-            tables.ExperimentTable.c.name == experiment_id  # type: ignore
-        )
-        result = conn.execute(query)
-        conn.commit()
-
         # Drop schema
         conn.execute(text(f'DROP SCHEMA IF EXISTS "{experiment_id}" CASCADE'))
         conn.commit()
@@ -859,8 +849,7 @@ def _delete_experiment_sqlalchemy(experiment_id: str) -> bool:
         except Exception as e:
             Log.debug(f"The experiment {experiment_id} has no status: {str(e)}")
 
-        if cast(int, result.rowcount) > 0:
-            Log.debug(f"The experiment {experiment_id} has been deleted!!!")
+        Log.debug(f"The experiment {experiment_id} has been marked as deleted.")
         return True
 
 
--- a/autosubmit/database/db_common.py
+++ b/autosubmit/database/db_common.py
@@ -571,6 +571,19 @@ def _delete_experiment(experiment_id):
         raise AutosubmitCritical("Could not establish a connection to database", 7001, str(e))
     cursor.execute('DELETE FROM experiment '
                    'WHERE name=:name', {'name': experiment_id})
+    
+    # keep the experiment trace in experiment_status for deleted experiments
+    try:
+        cursor.execute('UPDATE experiment_status SET status=:status, seconds_diff=:seconds_diff, modified=CURRENT_TIMESTAMP '
+                       'WHERE name=:name',
+                       {
+                           'status': 'DELETED',
+                           'seconds_diff': 0,
+                           'name': experiment_id
+                       })
+    except Exception:
+        pass
+
     row = cursor.fetchone()
     if row is None:
         Log.debug(f'The experiment {experiment_id} has been deleted!!!')
@@ -832,10 +845,14 @@ def _delete_experiment_sqlalchemy(experiment_id: str) -> bool:
         conn.execute(text(f'DROP SCHEMA IF EXISTS "{experiment_id}" CASCADE'))
         conn.commit()
 
-        # Delete from experiment_status table
+        # Keep the experiment trace in experiment_status for deleted experiments
         try:
-            query = delete(tables.ExperimentStatusTable).where(
+            query = update(tables.ExperimentStatusTable).where(
                 tables.ExperimentStatusTable.c.name == experiment_id  # type: ignore
+            ).values(
+                status="DELETED",
+                seconds_diff=0,
+                modified=func.current_timestamp()
             )
             conn.execute(query)
             conn.commit()
--- a/autosubmit/database/db_common.py
+++ b/autosubmit/database/db_common.py
@@ -22,7 +22,7 @@ import os
 import sqlite3
 from contextlib import suppress
 from pathlib import Path
-from typing import List, Optional, cast
+from typing import List, Optional
 
 from autosubmit.config.basicconfig import BasicConfig
 from sqlalchemy import delete, select, Connection, insert, text, update, func
--- a/autosubmit/database/tables.py
+++ b/autosubmit/database/tables.py
@@ -68,6 +68,7 @@ ExperimentStatusTable = Table(
     Column("status", Text, nullable=False),
     Column("seconds_diff", Integer, nullable=False),
     Column("modified", Text, nullable=False),
+    Column("last_heartbeat", Text, nullable=True),
 )
 """Stores the status of the experiments."""
 
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -148,7 +148,7 @@ def _delete_experiment(expid: str, force: bool) -> None:
     # Try to delete the experiment details
     try:
         ExperimentDetails(expid).delete_details()
-        ExperimentStatus(expid).set_as_not_running() # TODO: unsure if we should set the experiment as deleted even with exceptions in the process
+        ExperimentStatus(expid).set_as_deleted() # TODO: unsure if we should set the experiment as deleted even with exceptions in the process
     except Exception as e:
         Log.warning(f'Failed to delete DB details for experiment {expid}: {str(e)}')
         raise
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -148,7 +148,7 @@ def _delete_experiment(expid: str, force: bool) -> None:
     # Try to delete the experiment details
     try:
         ExperimentDetails(expid).delete_details()
-        ExperimentStatus(expid).set_as_deleted() # TODO: unsure if we should set the experiment as deleted even with exceptions in the process
+        ExperimentStatus(expid).set_as_not_running() # TODO: unsure if we should set the experiment as deleted even with exceptions in the process
     except Exception as e:
         Log.warning(f'Failed to delete DB details for experiment {expid}: {str(e)}')
         raise
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -146,9 +146,9 @@ def _delete_experiment(expid: str, force: bool) -> None:
     Log.info(f'Deleting experiment {expid}')
 
     try:
-        _delete_expid(expid, force)
-        ExperimentDetails(expid).delete_details()
         ExperimentStatus(expid).set_as_deleted()
+        ExperimentDetails(expid).delete_details()
+        _delete_expid(expid, force)
         Log.info(f'Experiment {expid} has been deleted')
     except Exception as e:
         raise AutosubmitCritical("Seems that something went wrong, please check the trace", 7012, str(e))
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -148,10 +148,8 @@ def _delete_experiment(expid: str, force: bool) -> None:
 
     try:
         _delete_expid(expid, force)
+        ExperimentDetails(expid).delete_details()
         ExperimentStatus(expid).set_as_deleted()
-        # Details cleanup should not fail deletion after files are removed and status updated. Supress exception
-        with suppress(Exception):
-            ExperimentDetails(expid).delete_details()
         Log.info(f'Experiment {expid} has been deleted')
     except Exception as e:
         raise AutosubmitCritical("Seems that something went wrong, please check the trace", 7012, str(e))
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -29,6 +29,7 @@ from autosubmit.database import db_common
 from autosubmit.experiment.detail_updater import ExperimentDetails
 from autosubmit.helpers.processes import process_id
 from autosubmit.helpers.utils import user_yes_no_query
+from autosubmit.history.experiment_status import ExperimentStatus
 from autosubmit.log.log import Log, AutosubmitCritical, AutosubmitError
 
 __all__ = [
@@ -147,6 +148,7 @@ def _delete_experiment(expid: str, force: bool) -> None:
     # Try to delete the experiment details
     try:
         ExperimentDetails(expid).delete_details()
+        ExperimentStatus(expid).set_as_deleted() # TODO: unsure if we should set the experiment as deleted even with exceptions in the process
     except Exception as e:
         Log.warning(f'Failed to delete DB details for experiment {expid}: {str(e)}')
         raise
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -20,6 +20,7 @@
 import os
 import pwd
 import string
+from contextlib import suppress
 from pathlib import Path
 from shutil import rmtree
 from typing import Optional
@@ -145,23 +146,19 @@ def _delete_experiment(expid: str, force: bool) -> None:
 
     Log.info(f'Deleting experiment {expid}')
 
-    # Try to delete the experiment details
-    try:
-        ExperimentDetails(expid).delete_details()
-        ExperimentStatus(expid).set_as_deleted() # TODO: unsure if we should set the experiment as deleted even with exceptions in the process
-    except Exception as e:
-        Log.warning(f'Failed to delete DB details for experiment {expid}: {str(e)}')
-        raise
-
     try:
         _delete_expid(expid, force)
+        ExperimentStatus(expid).set_as_deleted()
+        # Details cleanup should not fail deletion after files are removed and status updated. Supress exception
+        with suppress(Exception):
+            ExperimentDetails(expid).delete_details()
         Log.info(f'Experiment {expid} has been deleted')
     except Exception as e:
         raise AutosubmitCritical("Seems that something went wrong, please check the trace", 7012, str(e))
 
 
 def _delete_expid(expid_delete: str, force: bool = False) -> None:
-    """Removes an experiment from the path and database.
+    """Removes an experiment from disk and marks it as deleted in status metadata.
     If the current user is eadmin and the -f flag has been sent, it deletes regardless of experiment owner.
 
     :param expid_delete: Identifier of the experiment to delete.
@@ -207,7 +204,7 @@ def _delete_expid(expid_delete: str, force: bool = False) -> None:
                 f"Current user is not the owner of the experiment. {expid_delete} cannot be deleted!", 7012)
 
     message_parts = [
-        f"The {expid_delete} experiment was removed from the local disk and from the database.",
+        f"The {expid_delete} experiment was removed from local disk and marked as deleted in metadata.",
         "Note that this action does not delete any data written by the experiment.",
         "Complete list of files/directories deleted:",
         ""
@@ -246,13 +243,6 @@ def _perform_deletion(experiment_path: Path, structure_db_path: Path, job_data_d
 
     is_sqlite = BasicConfig.DATABASE_BACKEND == 'sqlite'
 
-    Log.info(f"Deleting experiment from {BasicConfig.DATABASE_BACKEND} database...")
-    try:
-        db_common.delete_experiment(expid_delete)
-        Log.result(f"Experiment {expid_delete} deleted from database")
-    except Exception as e:
-        error_message.append(f"Cannot delete experiment entry: {e}")
-
     Log.info("Removing experiment directory...")
     try:
         rmtree(experiment_path)
@@ -272,6 +262,15 @@ def _perform_deletion(experiment_path: Path, structure_db_path: Path, job_data_d
         db_path.unlink(missing_ok=True)
         sql_path.unlink(missing_ok=True)
         Log.info(f"Experiment {expid_delete} job_data db deleted")
+
+    # Mark experiment status as deleted only after cleanup successful
+    if not error_message:
+        Log.info(f"Updating experiment status in {BasicConfig.DATABASE_BACKEND} database...")
+        try:
+            db_common.delete_experiment(expid_delete)
+            Log.result(f"Experiment {expid_delete} marked as deleted in database")
+        except Exception as e:
+            error_message.append(f"Cannot update experiment metadata: {e}")
 
     return "\n".join(error_message)
 
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -20,7 +20,6 @@
 import os
 import pwd
 import string
-from contextlib import suppress
 from pathlib import Path
 from shutil import rmtree
 from typing import Optional
--- a/autosubmit/history/database_managers/database_models.py
+++ b/autosubmit/history/database_managers/database_models.py
@@ -38,6 +38,8 @@ MaxCounter = collections.namedtuple('MaxCounter', ['maxcounter'])
 class RunningStatus:
   RUNNING = "RUNNING"
   NOT_RUNNING = "NOT RUNNING"
+  ARCHIVED = "ARCHIVED"
+  DELETED = "DELETED"
 
 
 class RowType:
--- a/autosubmit/history/database_managers/database_models.py
+++ b/autosubmit/history/database_managers/database_models.py
@@ -27,7 +27,7 @@ ExperimentRunRow = collections.namedtuple('ExperimentRunRow', [
                                            'run_id', 'created', 'modified', 'start', 'finish', 'chunk_unit', 'chunk_size', 'completed', 'total', 'failed', 'queuing', 'running', 'submitted', 'suspended', 'metadata'])
 
 ExperimentStatusRow = collections.namedtuple(
-    'ExperimentStatusRow', ['exp_id', 'name', 'status', 'seconds_diff', 'modified'])
+  'ExperimentStatusRow', ['exp_id', 'name', 'status', 'seconds_diff', 'modified', 'last_heartbeat'])
 
 ExperimentRow = collections.namedtuple('ExperimentRow', ["id", "name", "autosubmit_version", "description"])
 
--- a/autosubmit/history/database_managers/experiment_status_db_manager.py
+++ b/autosubmit/history/database_managers/experiment_status_db_manager.py
@@ -154,9 +154,12 @@ class ExperimentStatusDbManager(DatabaseManager):
         self.create_exp_status(exp_row.id, expid, status)
 
     def update_heartbeat(self, expid: str) -> None:
-        # dummy function
-        pass
-
+        now = HUtils.get_current_datetime()
+        statement = ''' UPDATE experiment_status SET last_heartbeat = ?, modified = ?
+        WHERE name = ?'''
+        arguments = (now, now, expid)
+        self.execute_statement_with_arguments_on_dbfile(
+            self._as_times_file_path, statement, arguments)
 
 class ExperimentStatusDatabaseManager(Protocol):
 
--- a/autosubmit/history/database_managers/experiment_status_db_manager.py
+++ b/autosubmit/history/database_managers/experiment_status_db_manager.py
@@ -260,7 +260,6 @@ class SqlAlchemyExperimentStatusDbManager:
                 status=status,
                 seconds_diff=0,
                 modified=now,
-                # Update last_heartbeat only if status is RUNNING, otherwise keep the previous value
                 last_heartbeat=now if status == Models.RunningStatus.RUNNING else ExperimentStatusTable.c.last_heartbeat  # type: ignore
             )
         )
--- a/autosubmit/history/database_managers/experiment_status_db_manager.py
+++ b/autosubmit/history/database_managers/experiment_status_db_manager.py
@@ -252,15 +252,13 @@ class SqlAlchemyExperimentStatusDbManager:
         return row_count
 
     def update_exp_status(self, expid: str, status="RUNNING") -> None:
-        now = HUtils.get_current_datetime()
         query = (
             update(ExperimentStatusTable).
             where(ExperimentStatusTable.c.name == expid).  # type: ignore
             values(
                 status=status,
                 seconds_diff=0,
-                modified=now,
-                last_heartbeat=now if status == Models.RunningStatus.RUNNING else None
+                modified=HUtils.get_current_datetime()
             )
         )
         with self.engine.connect() as conn:
--- a/autosubmit/history/database_managers/experiment_status_db_manager.py
+++ b/autosubmit/history/database_managers/experiment_status_db_manager.py
@@ -92,7 +92,6 @@ class ExperimentStatusDbManager(DatabaseManager):
         experiment_row = self.get_experiment_row_by_expid(expid)
         return self.get_experiment_status_row_by_exp_id(experiment_row.id)
 
-    # TODO: it's fetching by name and not by exp_id column?
     def get_experiment_row_by_expid(self, expid: str) -> Models.ExperimentRow:
         """Get the experiment from ecearth.db by expid as Models.ExperimentRow."""
         statement = self.get_built_select_statement("experiment", "name=?")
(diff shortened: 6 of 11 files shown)


ntorqulu:
### PROBLEMS/OBSERVATIONS:

## 1. `experiment_status.db` allows repeated experiment ids as rows

What I mean by that: you can have multiple experiements with the same expid in the db. It should be enforced uniqueness of the `name` column too.

## 2. AS already creates the experiment_status db. This step does not depend on the api.

So, I will be investigating more and add some tests for this case, but #2641 is solved

## 3. Enforce single source of truth

I'll enforce that only AS backend writes into the `experiment_status.db`. API should not be able to write into the db, only read the info stored

ntorqulu:
Which should be the possible states for an experiment?
Current RunningStatus:
- RUNNING
- NOT_RUNNING

Should we track when:
- an experiment is paused -> not running or paused?
- an experiment has failed -> not running or failed?
- experiment archived and experiment deleted -> yes based on specifications with one state for each

ntorqulu:
@LuiggiTenorioK I've a few questions about how to handle the statuses in the `experient_status` database.

I'm thinking about future maintainability of the databases (both `experiment_status` and `experiment` tables) and I'm not sure if it's correct to store the DELETED experiments... Why the GUI needs to show the deleted experiments? Is not possible that if a experiment is deleted it's trace is completely lost? 
I understand that they want to visualize the archived ones, but why would they want to visualize the deleted ones?

kinow:
I can answer on the deleted part, @ntorqulu . Not sure about the part of showing the experiment, but we want to keep the register of a deleted experiment so we know it was deleted, and so that we cannot re-use the expid.

Keeping deleted entries in a DB is sometimes called a tombstone entry -- https://en.wikipedia.org/wiki/Tombstone_(data_store). We can purge deleted entries after some time if needed, but it could be useful for traceability.

At the moment, the main problem we have with the deleted experiments is this one: https://github.com/BSC-ES/autosubmit/issues/1072

ntorqulu:
Great, thanks for the explanation @kinow! It makes sense to delete them in batch once the current_time - modified_time is greater than a specified threshold if necessary.

About the issue with the deleted experiments, yes, I've been facing the same problem. If you:
- expid & create & run an experiment
- delete the experiment -> then this experiment is deleted from the `experiment` table but not from the `experiment_status` table.
- expid & create & run -> it will assign the same expid that was previously deleted, because it's no longer present in the `experiment` table. And will then create a duplication of expid in the `experiment_status` table because in this table the experiment row was not deleted...

For now, I'll enforce uniqueness in expid column and do not delete experiment rows in neither table. The only thing that will be updated is the status of the experiment to DELETED. For the old ones, I'll keep the newest row with the same expid.

LuiggiTenorioK:
@ntorqulu is just as Bruno mentioned above. A logical delete is almost always a must if we want to keep traceability, especially if we don't want to reuse expids.

In the GUI, it is also nice to know if the experiment has been deleted to keep the user informed (probably with a message like `this experiment has been deleted on 20XX-XX-XX at XX:XX`.

Also, take into account that Autosubmit is not the only one that modifies the `experiment_status` data. The API does it as well every 5 minutes, meaning that your changes to change the status to `DELETED` will be `quickly` overwritten by the API, which will set it as `NOT RUNNING`. So, in every case, we have to modify the API to better orchestrate the status management.

ntorqulu:
Yes, I've this in mind! I was just checking first the backend because it's what I'm currently more familiar.
The idea would be that the API only overwrites statuses for RUNNING experiments when the heartbeat is outdated (should define what outdated means, 15 min maybe? I'm not sure about this).

How do you want to handle the API modifications @LuiggiTenorioK? I open an issue for you to handle it or I assign it to myself to continue the refactor in the API?

LuiggiTenorioK:
> The idea would be that the API only overwrites statuses for RUNNING experiments when the heartbeat is outdated (should define what outdated means, 15 min maybe? I'm not sure about this).

Filtering by status seems like a good idea. I'll probably include the `NOT RUNNING` as well to do a sanity check in case Autosubmit failed to set it up as `RUNNING`.

> How do you want to handle the API modifications @LuiggiTenorioK? I open an issue for you to handle it or I assign it to myself to continue the refactor in the API?

First, I think it's better if we have an issue to discuss the modifications to be made in the API. I'll create it and tag you to continue the thread there.

ntorqulu:
Okay! I'll take a look @LuiggiTenorioK .  Thanks!
About the state diagram of an experiment, I've a few doubts about the granularity of the detail for the states of the experiments and if we should track also experiments that are created but not run yet.

experiment_status

1. Currently `RUNNING` experiments can be archived, but `RUNNING` experiments cannot be deleted. Is it correct? Should only `NOT_RUNNING` experiments be archived?

2. When creating the exeperiment (`EXPID` and `CREATE`), the experiment is not stored in the `experiment_status` table. Only when the experiment has been archived, deleted or run it appears as a new entry. Should we enforce the existance in the table once the experiment is created?

3. `NOT_RUNNING` is a very broad status. It captures paused experiments, failed ones and just the ones that have finished the execution. Is it correct to keep it as `NOT_RUNNING`?

LuiggiTenorioK:
> 1. Currently `RUNNING` experiments can be archived, but `RUNNING` experiments cannot be deleted. Is it correct? Should only `NOT_RUNNING` experiments be archived?

`RUNNING` experiments shouldn't be archived. This could potentially cause experiments to crash and generate corrupted data.

> 2. When creating the exeperiment (`EXPID` and `CREATE`), the experiment is not stored in the `experiment_status` table. Only when the experiment has been archived, deleted or run it appears as a new entry. Should we enforce the existance in the table once the experiment is created?

To keep data consistent, we should

> 3. `NOT_RUNNING` is a very broad status. It captures paused experiments, failed ones and just the ones that have finished the execution. Is it correct to keep it as `NOT_RUNNING`?

It is not incorrect to have it like that.

`RUNNING` and `NOT_RUNNING` experiments can have failed jobs, so splitting from there will diversify all states because of mutual exclusivity. IMO, experiments with failed jobs are more an observation than a state.

Then, "paused" and "completed" (or better said, "stopped" (because there is no pause command, just stop or kill the process) and "finished" (because there might be pending tasks to do, like restart failed jobs)) with an additional "created" state could be valid states to split the `NOT_RUNNING` state. However, I think there is enough complexity in this issue to add those. I suggest keeping the same 4 states for now, while leaving the solution extensible for this change in the future.

kinow:
+1 to what @LuiggiTenorioK said. I think @ntorqulu 's state diagram comes from the code. I don't think the `archive` command checks for the lock file.

ntorqulu:
Yes, the current state diagram is from the code, I'll update it with @LuiggiTenorioK's suggestions before changing anything in the code. This way we can have it as reference.
- commands `expid` and `create` -> set experiment as `NOT RUNNING`
- from RUNNING cannot transition to ARCHIVE
- from NOT RUNNING it can transition to ARCHIVE and to DELETE
- status STOPPED, FAILED, COMPLETED will be mapped to NOT RUNNING for now, taking into account maintainability and scalability for future states

ntorqulu:
Untitled Diagram drawio

ntorqulu:
QUESTION: What to do with old experiments and its `last_heartbeat` value. Leave as NULL or prefill with modified time?

LuiggiTenorioK:
> QUESTION: What to do with old experiments and its `last_heartbeat` value. Leave as NULL or prefill with modified time?

`last_heartbeat` won't always be equal to `modified` time? I think `last_heartbeat` might be redundant in this case.

ntorqulu:
From what I'm understanding, `modified` is more like a timestamp for the table, that keeps track of when the last modifications were done in the row (change of status, change of the last_heartbeat etc...).
And `last_heartbeat` acts like an i'm-alive signal while the experiment is running, updating its time while there are active jobs in the experiment.
For instance, if an experiment transition from `NOT RUNNING` to `ARCHIVED` currently `status` and  `modified` columns get updated, but not `last_heartbeat`. `last_heartbeat` keeps the last value it had while the `status` was `RUNNING`.
If we use `modified` as the single column to store the heartbeat too, we will lose the timestamp of the last run when the experiment transitions to `NOT RUNNING`.

ntorqulu:
Also is it usefull to have the timestamp when the experiment was created?

LuiggiTenorioK:
> From what I'm understanding, `modified` is more like a timestamp for the table, that keeps track of when the last modifications were done in the row (change of status, change of the last_heartbeat etc...). And `last_heartbeat` acts like an i'm-alive signal while the experiment is running, updating its time while there are active jobs in the experiment. For instance, if an experiment transition from `NOT RUNNING` to `ARCHIVED` currently `status` and `modified` columns get updated, but not `last_heartbeat`. `last_heartbeat` keeps the last value it had while the `status` was `RUNNING`.

Ok, it makes sense like that. This heartbeat will serve to determine if the experiment stopped abruptly and didn't achieve the change in status to `NOT_RUNNING`.

> If we use `modified` as the single column to store the heartbeat too, we will lose the timestamp of the last run when the experiment transitions to `NOT RUNNING`.

That timestamp should also be available in the `experiment_run` table. Do you think that `modified` timestamp in the `experiment_run` table could work as the heartbeat, or is it better to keep it separated?

> Also is it usefull to have the timestamp when the experiment was created?

Yes, this is another issue, because we currently use the creation time of the dir as the experiment creation, and we don't want to rely on the file system. More likely, we can have it as another column in the `experiment` table.

ntorqulu:
> Ok, it makes sense like that. This heartbeat will serve to determine if the experiment stopped abruptly and didn't achieve the change in status to `NOT_RUNNING`.

Yes, this way we can filter even more the experiments the API can modify. For now I'll leave `last_heartbeat` as NULL, only updated when the experiment is RUNNING. Not pre-filled with `modified` for old experiments.
This way we don't introduce false data, the API will handle that.

> That timestamp should also be available in the experiment_run table. Do you think that modified timestamp in the experiment_run table could work as the heartbeat, or is it better to keep it separated?

I've been debugging where `modified` is updated for this table in `run_experiment` method.
Around line 2324 it calls 
```
exp_history = Autosubmit.process_historical_data_iteration(job_list, job_changes_tracker,
                                                                                       expid)
```
The problem is that only when there are job changes the `modified` column gets updated. So we are in the same situation as before, `modified` acts more as a time-of-edit of the row rather than a keep-alive signal of the experiment.
If it's necessary to have a keep-alive signal in table `experiment_run`, I would add the `last_heartbeat` there too.
Could happen that an experiment is `RUNNING` but it's job statuses remain the same during a long period of time. The `last_heartbeat` would keep being updated (because the job is running), but the modified time would be fixed for this period.

> Yes, this is another issue, because we currently use the creation time of the dir as the experiment creation, and we don't want to rely on the file system. More likely, we can have it as another column in the experiment table.

Is it ok to leave it for another pr? Or better implement it in this one too?

LuiggiTenorioK:
> Is it ok to leave it for another pr? Or better implement it in this one too?

It seems unrelated to this PR. I would leave it for another one.

ntorqulu:
> > Is it ok to leave it for another pr? Or better implement it in this one too?
> 
> It seems unrelated to this PR. I would leave it for another one.

Okay! I'll open a new issue to not forget this.

ntorqulu:
QUESTION: 
1. Currently `autosubmit install` creates only database `autosubmit.db` with  tables `experiment` and `db_version`.
I'm  not sure why:
- table `details` from `autosubmit.db` is not created
- database `as_times.db` with table `experiment_status` is not initialized
I would initialize them all together with the install command. 

2. With both databases empty, when executing `create`:

- `autosubmit create <EXPID with a configuration folder stored>` -> Incorrect. Does not raise error and does not insert the experiment into `autosubmit.db` and neither in `as_times.db`. I'm not sure if it's better to raise an error or insert the experiment in the database in this case. Probably insert it to be more robust to fallbacks, because the configuration is correctly stored in its experiment folder.

`autosubmit create <EXPID not created yet>` -> raise error. Correct.

3. With both databases empty, when executing expid:

- currently it autofills the `experiment` table with consecutive experiment names, with consecutive id's and all of them with `description = No description` and `autosubmit_version = 3.14.0` until it reaches the first expid that does not have a configuration folder assigned (the first one, not the last one!). For this one it correctly assigns the description given in the command and autosubmit version 4.1.17. 
 Incorrect. I'm not sure about this one. Iterate through all the configuration folders to insert again the stored experiments and assign always the highest expid to a new experiment (not the first not assigned yet!)?

image

LuiggiTenorioK:
> QUESTION:
> 
> 1. Currently `autosubmit install` creates only database `autosubmit.db` with  tables `experiment` and `db_version`.
>    I'm  not sure why:
> 
> * table `details` from `autosubmit.db` is not created


The `details` table was only maintained by the API before, and it was also its responsibility to create it. However, now that Autosubmit also makes changes in that table, it should create the table as well. I think we just forgot to add that create statement in the implementation.

> * database `as_times.db` with table `experiment_status` is not initialized

Same as the `details` table, but take into account that the API machine and the Autosubmit one can be different Linux users. We should chmod it to allow write for the group of users.

>   I would initialize them all together with the install command.

Agree +1


> 2. With both databases empty, when executing `create`:
> 
> * `autosubmit create <EXPID with a configuration folder stored>` -> Incorrect. Does not raise error and does not insert the experiment into `autosubmit.db` and neither in `as_times.db`. I'm not sure if it's better to raise an error or insert the experiment in the database in this case. Probably insert it to be more robust to fallbacks, because the configuration is correctly stored in its experiment folder.
> 
> `autosubmit create <EXPID not created yet>` -> raise error. Correct.

Currently, neither table is essential for running an experiment. So, probably it is ok to keep it like that because it is worse if the user cannot run the experiment than having that traceability.


> 3. With both databases empty, when executing expid:
> 
> * currently it autofills the `experiment` table with consecutive experiment names, with consecutive id's and all of them with `description = No description` and `autosubmit_version = 3.14.0` until it reaches the first expid that does not have a configuration folder assigned (the first one, not the last one!). For this one it correctly assigns the description given in the command and autosubmit version 4.1.17.
>   Incorrect. I'm not sure about this one. Iterate through all the configuration folders to insert again the stored experiments and assign always the highest expid to a new experiment (not the first not assigned yet!)?

I saw that happening before, and it is annoying. For me, it shouldn't do that. However, I am not 100% sure if removing that process can affect other things; maybe @kinow knows better.

ntorqulu:
> Same as the `details` table, but take into account that the API machine and the Autosubmit one can be different Linux users. We should chmod it to allow write for the group of users.

Okay thanks! I was not thinking about this. I see that install is not applying `chmod` in `autosubmit.db` neither. I'll apply it to both (autosubmit.db and as_times.db).

> Currently, neither table is essential for running an experiment. So, probably it is ok to keep it like that because it is worse if the user cannot run the experiment than having that traceability.

Mmm okay. I was thinking that when user executes `create`, before inserting the status into the experiment_status table:
- check if the experiment exists in autosubmit.db
- if it does not exist (db empty but config files exists) automatically insert it into autosubmit.db?
- if registration into autosubmit.db fails, log warning but do not fail, continue with the execution

But I don't know if it's in the scope of this pr. It's an edge case when the exp has config files but is not registered into autosubmit.db.

LuiggiTenorioK:
> Mmm okay. I was thinking that when user executes `create`, before inserting the status into the experiment_status table:
> 
> * check if the experiment exists in autosubmit.db
> * if it does not exist (db empty but config files exists) automatically insert it into autosubmit.db?
> * if registration into autosubmit.db fails, log warning but do not fail, continue with the execution
> 
> But I don't know if it's in the scope of this pr. It's an edge case when the exp has config files but is not registered into autosubmit.db.

Yes, I think it goes a little bit outside the scope. It is an edge case of another issue. In this PR, it is better to assume there is a correct common setup.

Also, the creation of the tables and permissions can be handled in a different PR. There is also an existing separated issue related to that: https://github.com/BSC-ES/autosubmit/issues/2641

ntorqulu:
I'll try to just focus on the experiment status table traceability and heartbeat update for this pr as it's growing too much. I'll add more detail on what's missing for the creation of the tables with `install` and the edge case with `create` in issue #2641, todo after this pr.

ntorqulu:
I'm content with the current implementation of the exp status traceability. I'll move to the API now. Should I open the pr to review or wait for any branch to be merged into master first to integrate the changes here before revision?

LuiggiTenorioK:
> I'm content with the current implementation of the exp status traceability. I'll move to the API now. Should I open the pr to review or wait for any branch to be merged into master first to integrate the changes here before revision?

Great! No need to wait. However, it would probably be better to be merged after 4.1.17, but we will see.

autosubmit/database/db_common.py
```diff
@@ -821,29 +830,25 @@ def _delete_experiment_sqlalchemy(experiment_id: str) -> bool:
...
+                modified=func.current_timestamp()
             )
             conn.execute(query)
             conn.commit()
         except Exception as e:
             Log.debug(f"The experiment {experiment_id} has no status: {str(e)}")
 
-        if cast(int, result.rowcount) > 0:
-            Log.debug(f"The experiment {experiment_id} has been deleted!!!")
+        Log.debug(f"The experiment {experiment_id} has been marked as deleted.")
```
LuiggiTenorioK: You can still use the `result.rowcount` to verify that the experiment status row has been updated successfully. For that, you need to save the result with `result = conn. execute ()` after building the update query.

autosubmit/autosubmit.py
```diff
@@ -2473,12 +2493,27 @@ def run_experiment(expid: str, start_time: Optional[str] = None, start_after: Op
...
             raise
         finally:
+            heartbeat_monitor.stop(timeout=10)
+            try:
+                if experiment_status == "COMPLETED":
+                    status_tracker.set_as_not_running()
+                elif experiment_status == "PAUSED":
+                    status_tracker.set_as_not_running()
+                elif experiment_status == "FAILED":
+                    status_tracker.set_as_not_running()
```
LuiggiTenorioK: This looks like a fancy way to write a TODO for the upcoming split of the `NOT RUNNING` statuses 😆

Probably, a comment explaining that will be nice, to not confuse the developers with a `experiment_status in [...]` to not have many redundant ifs.

ntorqulu:
Will be opened and reviewed after 4.2.0 branch #2619 is merged into master.

autosubmit/database/db_common.py
```diff
@@ -569,11 +570,19 @@ def _delete_experiment(experiment_id):
         (conn, cursor) = open_conn()
```
LuiggiTenorioK: `_delete_experiment` doesn't delete the experiment in the not SQLAlchemy version, because `open_conn` opens the `autosubmit.db` file and not the `as_times.db`.

You can check that by changing the `pass` to `raise` below after executing the query.

The reason that `autosubmit delete` still changes the status is because in the `_delete_experiment` of `experiment_common.py`, you call `ExperimentStatus(expid).set_as_deleted()` before calling this function and the exception is lost by the pass statement I already mentioned.

autosubmit/experiment/experiment_common.py
```diff
@@ -144,22 +145,17 @@ def _delete_experiment(expid: str, force: bool) -> None:
...
-    # Try to delete the experiment details
     try:
+        ExperimentStatus(expid).set_as_deleted()
         ExperimentDetails(expid).delete_details()
-    except Exception as e:
-        Log.warning(f'Failed to delete DB details for experiment {expid}: {str(e)}')
-        raise
-
-    try:
         _delete_expid(expid, force)
```
LuiggiTenorioK: Double deletion here from `ExperimentStatus(expid).set_as_deleted()` and `_delete_expid(expid, force)`

_related to another comment in this review_

autosubmit/history/database_managers/experiment_status_db_manager.py
LuiggiTenorioK: To be consistent with the not SQLAlchemy version of this, the index creation and validation could be included here.

LuiggiTenorioK (changes requested):
Great job @ntorqulu! The code is clean and pretty maintainable.

There are only some comments mainly in the deletion process, but nothing really hard to fix overall.

autosubmit/database/db_common.py
ntorqulu: Okay, thanks for the review @LuiggiTenorioK ! 
About the `rowcount` method:
- Why the cast to integer is needed? Checking the implementation, it should always return an integer no?
- Should we check cases `>0` and `0`? I see that there's also the possibility to return `-1`, but only when DBAPI does not support rowcount (I think we're safe here because we're using PostgreSQL via psycopg2)
- Checking again this part, I think the exception should be logged as a warning because it's a real failure, not just a missing row.
E..g
```
    with _get_sqlalchemy_conn() as conn:
        # Drop schema
        conn.execute(text(f'DROP SCHEMA IF EXISTS "{experiment_id}" CASCADE'))
        conn.commit()

        # Keep the experiment trace in experiment_status for deleted experiments
        try:
            query = update(tables.ExperimentStatusTable).where(
                tables.ExperimentStatusTable.c.name == experiment_id  # type: ignore
            ).values(
                status="DELETED",
                seconds_diff=0,
                modified=func.current_timestamp()
            )
            result = conn.execute(query)
            conn.commit()
            if result.rowcount > 0:
                Log.debug(f"The experiment {experiment_id} has been marked as deleted.")
            else:
                Log.debug(f"The experiment {experiment_id} was not found in experiment_status table to be marked as deleted.")
        except Exception as e:
            Log.warning(f"Failed to mark experiment {experiment_id} as deleted: {str(e)}")

        return True
```

autosubmit/autosubmit.py
ntorqulu: Yep, I implemented it with multiple status at the beginning and did not change it. Sure, I'll change for a TODO and only use status `NOT RUNNING` to prevent incorporating code that currently does not make sense. Also, I'll change the string-status for the use of enums `Model.RunningStatus` in `run_experiment`. I think it's better this way.

autosubmit/database/db_common.py
LuiggiTenorioK: > * Why the cast to integer is needed? Checking the implementation, it should always return an integer no?

I am also not sure why the cast. As you said, it should always return an int.

> * Should we check cases `>0` and `0`? I see that there's also the possibility to return `-1`, but only when DBAPI does not support rowcount (I think we're safe here because we're using PostgreSQL via psycopg2)

It should check `>0`, as the successful case. The `-1`, as far as I know, should not happen with SQLite and Postgres.

> * Checking again this part, I think the exception should be logged as a warning because it's a real failure, not just a missing row.

Yes, you are right, at least in the scope of that function. Just make sure that raising it doesn't stop a critical part in the call hierarchy.

autosubmit/database/db_common.py
ntorqulu: True, it did not realize I was trying to update the experiment status two times! Moved it to `db_common.py/_delete_experiment` only.
I was reviewing this part:
```
    try:
        # TODO: not sure if the details should be deleted
        ExperimentDetails(expid).delete_details()
        _delete_expid(expid, force) 
```
Should the `details` from deleted experiments be deleted? WIll the GUI show extra info about deleted experiments such as the user who created it or the creation time?

LuiggiTenorioK: > Should the `details` from deleted experiments be deleted? WIll the GUI show extra info about deleted experiments such as the user who created it or the creation time?

Now that we keep the deleted experiments, I think the details shouldn't be deleted even if the experiment has been deleted or archived. This will let us have the last values before the delete/archive.

Good catch! :100:

autosubmit/history/database_managers/experiment_status_db_manager.py
ntorqulu: I'm realizing that in `test_experiment_status_db_manager` I'm just testing with the SQLite db. I'll use the shared `as_db` fixture to test both cases.

ntorqulu:
Done correcting the errors from the revision!

- db_common.py:
	- used `result.rowcount` to verify that the DELETE status has been updated correctly. Removed previous cast to integer, as this function always return an `int`
	- Notified user of an error in the update via `Log.warning` instead of `Log.debug`
	- Corrected the double update of DELETED status. Now it's only updated inside `_delete_expid`
	- Prevented deletion of the experiment details row. Deleted experiments will keep its details from the `details` table.

- autosubmit.py:
	- removed all reference of unused statuses COMPLETED, PAUSED and FAILED. This will be handled in another pr. Added a `TODO` comment to remember where they should be handled in the future.
	- All references to RUNNING, NOT RUNNING, ARCHIVED and DELETED statuses now use enums instead of hardcoded strings.

- experiment_status_db_manager.py:
	- check that the last_heartbeat column exists, add it if not present, and ensure unique rows by name in the sqlalchemy version, similarly on how it's implemented in the non-sqlalchemy one.

- add more tests overall to increase coverage of the pr

docs/source/database/index.rst
```diff
@@ -38,7 +38,7 @@ Core databases
 +===================+====================================+==========================================================================================================+
 | ``autosubmit.db`` | ``$HOME/autosubmit/autosubmit.db`` | The main database of Autosubmit. The location can be customized in the ``autosubmitrc`` file.            |
 +-------------------+------------------------------------+----------------------------------------------------------------------------------------------------------+
-| ``as_times.db``   | ``$HOME/autosubmit/as_times.db``   | Deprecated API. Used by Autosubmit API with Autosubmit ``3.x``. Kept for backward compatibility for now. |
+| ``as_times.db``   | ``$HOME/autosubmit/as_times.db``   | The secondary database of Autosubmit. Used to keep track of the experiment status.                       |
```
ntorqulu, line 41: `Databases` section of the documentation should be updated. Currently only changed the part related to the ` as_times.db` to avoid making this pr bigger.
But the rst is also referencing some `status.db` and a `test` folder I don't find anywhere.

LuiggiTenorioK: `autosubmit/metadata/test/status.db` used to exist to measure the IO latency of `/esarchive` at BSC. This was one of the first things I removed from the old API since we wanted Autosubmit to be a generalized tool rather than one made for BSC.

autosubmit/database/db_common.py
```diff
@@ -599,27 +599,23 @@ def _last_name_used(test=False, operational=False, evaluation=False):
...
-        raise AutosubmitCritical("Could not establish a connection to database", 7001, str(e))
-    cursor.execute('DELETE FROM experiment '
-                   'WHERE name=:name', {'name': experiment_id})
-    row = cursor.fetchone()
-    if row is None:
-        Log.debug(f'The experiment {experiment_id} has been deleted!!!')
-    close_conn(conn, cursor)
+        ExperimentStatus(experiment_id).set_as_deleted()
+    except Exception as e:
+        raise AutosubmitCritical('Could not mark experiment as deleted', 7001, str(e))
```
pablogoitia: I really liked the suggestion to keep the experiments in the database even if deleted for a traceability matter :) This is common practice in software design because not all the information can be always immediately deleted, for example, because of accountancy reasons

autosubmit/database/db_common.py
```diff
@@ -862,30 +858,14 @@ def _delete_experiment_sqlalchemy(experiment_id: str) -> bool:
...
-                    tables.ExperimentStatusTable.c.name == experiment_id  # type: ignore
-                )
-                conn.execute(query)
-            except Exception as e:
-                Log.debug(f"The experiment {experiment_id} has no status: {str(e)}")
-
-        if cast(int, result.rowcount) > 0:
-            Log.debug(f"The experiment {experiment_id} has been deleted!!!")
-        return True
+    from autosubmit.history.experiment_status import ExperimentStatus
```
pablogoitia: Could this import be moved to the header?

autosubmit/database/db_common.py
```diff
@@ -599,27 +599,23 @@ def _last_name_used(test=False, operational=False, evaluation=False):
...
     :type experiment_id: str
-    :return: True if delete is successful
+    :return: True if operation is successful
     :rtype: bool
     """
-    check_db()
-
     if not _check_experiment_exists(experiment_id, False):  # Reference the no anti-lock version.
         return True
+    from autosubmit.history.experiment_status import ExperimentStatus
```
pablogoitia: Could this import be moved to the header?

autosubmit/database/db_common.py
```diff
@@ -862,30 +858,14 @@ def _delete_experiment_sqlalchemy(experiment_id: str) -> bool:
...
-                Log.debug(f"The experiment {experiment_id} has no status: {str(e)}")
-
-        if cast(int, result.rowcount) > 0:
-            Log.debug(f"The experiment {experiment_id} has been deleted!!!")
-        return True
+    from autosubmit.history.experiment_status import ExperimentStatus
+    try:
+        ExperimentStatus(experiment_id).set_as_deleted()
+    except Exception as e:
+        raise AutosubmitCritical("Could not mark experiment as deleted", 7001, str(e))
```
pablogoitia: Same here :)

autosubmit/autosubmit.py
```diff
@@ -1513,6 +1515,8 @@ def expid(
             Log.warning(f"Could not update experiment details for {exp_id}. Omitting this step.")
             Log.debug(f"Error calling save_update_details: {str(e)}")
 
+        ExperimentStatus(exp_id).set_as_not_running()
```
pablogoitia: I understand that `set_as_not_running` can throw an exception for example if the experiment do not exist.

Line 1855 already wraps `set_as_running` within a try-except clause :)

autosubmit/autosubmit.py
```diff
@@ -2319,8 +2338,10 @@ def run_experiment(expid: str, start_time: str | None = None, start_after: str |
                     p.close_connection()
                 if len(job_list.get_failed_from_db()) > 0:
                     Log.info("Some jobs have failed and reached maximum retrials")
+                    experiment_status = Models.RunningStatus.NOT_RUNNING
                 else:
                     Log.result("Run successful")
+                    experiment_status = Models.RunningStatus.NOT_RUNNING
```
pablogoitia: `experiment_status = Models.RunningStatus.NOT_RUNNING` seems to be redundant. It can be moved outside the conditional because it's always executed anyway.

autosubmit/autosubmit.py
```diff
@@ -2334,12 +2355,27 @@ def run_experiment(expid: str, start_time: str | None = None, start_after: str |
                 else:
                     Log.info("ROCRATE not present in experiment YAML configuration. No RO-Crate archive created.")
         except BaseLockException:
+            # Multiple instances of autosubmit running the same experiment 
+            # or previous instance didn't release the lock file
+            # In both cases, we don't want to overwrite the status 
+            # of the experiment to avoid errors with the API and GUI
             raise
```
pablogoitia: Well spotted! :)

autosubmit/autosubmit.py
```diff
@@ -3789,6 +3824,7 @@ def archive(expid, noclean=True, uncompress=True, rocrate=False):
                         "Can not remove or rename experiments folder", 7012, str(e))
 
         Log.result("Experiment archived successfully")
+        ExperimentStatus(expid).set_as_archived()
```
pablogoitia: Same as in my previous comment. I think it's better to encapsulate within a try-except

autosubmit/autosubmit.py
```diff
@@ -3845,6 +3881,7 @@ def unarchive(experiment_id, uncompressed=True, rocrate=False):
             return False
 
         Log.info("Unpacking finished")
+        ExperimentStatus(experiment_id).set_as_not_running()
```
pablogoitia: Same as in my previous comment. I think it's better to encapsulate within a try-except

autosubmit/autosubmit.py
```diff
@@ -5384,7 +5424,7 @@ def stop(expids: str, force=False, all_expids=False, force_all=False, cancel=Fal
         :param force_yes: force yes answer to prompts
         :type force_yes: bool
         """
-        from autosubmit.helpers.processes import process_id, retrieve_expids
+        from autosubmit.helpers.processes import retrieve_expids
         from autosubmit.job.job_utils import cancel_jobs
```
pablogoitia: Is there any specific reason these imports are within the function scope or could be moved to the header? @kinow 

From my POV it's ok if we keep it as is because `autosubmit.py` will vanish from the face of the earth anyway

autosubmit/history/database_managers/database_manager.py
```diff
@@ -55,6 +55,8 @@ def get_connection(self, path):
     def _create_database_file(self, path):
         # type : (str) -> None
         """ creates a database files with full permissions """
+        # FIXME: umask is never restored. Every file 
+        # created after this will also lose umask filtering (too permissive).
         os.umask(0)
```
pablogoitia, line 58: Maybe it might be worth opening a separate issue to check it out later. Great catch!

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -187,50 +293,72 @@ def get_experiment_status_row_by_exp_id(self, exp_id: int) -> Models.ExperimentS
...
                 seconds_diff=0,
-                modified=HUtils.get_current_datetime()
+                modified=now,
+                last_heartbeat=now if status == Models.RunningStatus.RUNNING else ExperimentStatusTable.c.last_heartbeat  # type: ignore
             )
         )
         with self.engine.connect() as conn, conn.begin():
             conn.execute(query)
 
+    def set_exp_status(self, expid:str, status:str) -> None:
```
pablogoitia: I was a little confused to see that this function is duplicated, I think then I understood 😅 This is my first time delving into the persistence layer of AS. Is this redundancy (and others I've observed in this file) done to ensure compatibility between DB types?

pablogoitia (changes requested):
Hi, @ntorqulu. Thanks for such a clever solution for the experiment status traceability! It will be so helpful to improve the accuracy in the experiment monitoring in the GUI side. I left few comments requesting/suggesting very small changes, and also a couple questions.

autosubmit/database/db_common.py
ntorqulu: Yes! And will be really useful when implementing the issue related to avoid repeating existing expid: https://github.com/BSC-ES/autosubmit/issues/1072

autosubmit/autosubmit.py
ntorqulu: Yes, I implemented it thinking about other possible status and didn't remember to revert it back to RUNNING/NOT_RUNNING/ARCHIVED/DELETED. Changing it!

autosubmit/history/database_managers/experiment_status_db_manager.py
ntorqulu: Yes, this is the ongoing problem of having two diff dbs backends + legacy code for the sqlite db.
Currently we're trying to normalize it all under the SQLAlchemy, but it's a work in progress.

test/integration/conftest.py
```diff
@@ -540,12 +541,23 @@ def as_db(request: "FixtureRequest", autosubmit: Autosubmit, tmp_path: "LocalPat
...
     # TODO: check which functions call as_db twice or if this is used in
     #  combination other fixture that calls autosubmit.install)
     with suppress(AutosubmitCritical):
         autosubmit.install()
 
-    return backend
+    yield backend
+
+    # drop the per-test database so they do not accumulate in the
+    # Postgres container
```
kinow: 💪

autosubmit/autosubmit.py
```diff
@@ -2334,12 +2358,27 @@ def run_experiment(expid: str, start_time: str | None = None, start_after: str |
...
+            experiment_status = Models.RunningStatus.NOT_RUNNING
             raise
         except BaseException:
+            experiment_status = Models.RunningStatus.NOT_RUNNING
             raise
         finally:
+            heartbeat_monitor.stop(timeout=10)
+            try:
+                # TODO: Separate NOT_RUNNING into FAILED, COMPLETED and PAUSED statuses.
+                # Currently it's all treated as NOT_RUNNING for simplicity
```
kinow: 👍

autosubmit/autosubmit.py
```diff
@@ -1849,14 +1856,13 @@ def get_historical_database(expid, job_list, as_conf):
...
             Log.warning(f"Couldn't access the historical database for experiment {expid}")
-
         try:
             ExperimentStatus(expid).set_as_running()
         except Exception as e:
             # Connection to status database ec_earth.db can fail.
             # API worker will fix the status.
             Log.debug(f"Autosubmit couldn't set your experiment as running on the autosubmit times database: "
-                      f"{os.path.join(BasicConfig.DB_DIR, BasicConfig.AS_TIMES_DB)}. Exception: {str(e)}", 7003)
+                      f"{os.path.join(BasicConfig.DB_DIR, BasicConfig.AS_TIMES_DB)}. Exception: {str(e)}", 7003)
```
kinow: Extra space added at the end of the line?

autosubmit/autosubmit.py
```diff
@@ -2132,6 +2141,20 @@ def run_experiment(expid: str, start_time: str | None = None, start_after: str |
                 as_conf_config = as_conf.experiment_data.get('CONFIG', {})
                 git_operational_check_enabled = as_conf_config.get('GIT_OPERATIONAL_CHECK_ENABLED', True)
 
+                try:
+                    status_tracker.set_as_running()
+                except Exception as e:
+                    Log.debug(
```
kinow: Maybe log that as warning? I guess sysadmins will likely be watching for warnings like this (I mean, for messages like this, as a warning, not skimming through many debug lines). Unless this is expected to be a very common issue?

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -89,53 +140,342 @@ def test_experiment_status_db_manager(tmp_path: 'LocalPath', as_db: str, use_sql
...
+def test_update_heartbeat_stores_last_heartbeat(tmp_path: "LocalPath", as_db: str, autosubmit_exp, mocker):
+    """Test that update_heartbeat() stores the last heartbeat timestamp in the database."""
+    database_manager, experiments = _create_experiment_status_db_manager_and_rows(
+        as_db=as_db,
+        tmp_path=tmp_path,
+        expids=["a000"],
+        autosubmit_exp=autosubmit_exp,
+    )
+
+    # Ensure heartbeat timestamps are deterministic, ordering stable across diff runs
```
kinow: 👏

kinow (commented):
Haven't reviewed it. I was just waiting for something to run, so I started reading the changes and marking files as reviewed here 😬 If you are confident after Pablo's review, no objection if this is merged. Otherwise, I can have a look and send a review later if you'd like too 👍 

Now I'll go back to splitting autosubmit.py (which I can rebase and fix after this one is merged too, no worries :) )

Thanks!

autosubmit/autosubmit.py
pablogoitia: Oh, I see. Thanks :)

autosubmit/autosubmit.py
kinow: @pablogoitia some time ago there was a huge PR to improve performance and fix some memory leaks. Analysing the memory, every command was always consuming a certain amount of memory, plus the rest of the memory used for the command (and the leak).

Even simple commands like `autosubmit describe` or `clean` started with high memory usage.

It was due to the imports in `autosubmit.py`. The top of the file had imports required for all commands. `autosubmit stats`, for example, requires (I think) matplotlibs nd networkx. Two of the largest dependencies we have. I think that was increasing ~30-60 MB mem because of imports from other commands?

That's why we moved imports to within functions.

Then, as time passed, imports were added again to `autosubmit.py` :rofl: Because it is pretty much impossible to maintain Autosubmit with the single `autosubmit.py`, all the roles/code it accumulates, and the imports of all commands.

Another good reason to really nuke it away from the face of earth! :boom: :laughing:

autosubmit/autosubmit.py
ntorqulu: Yep, deleted the extra spaces + changed it from `os.path` to `Path`

autosubmit/autosubmit.py
ntorqulu: Yes, it shouldn't be a common error and it only logs during the experiment run (one time) so it's safe to log it as a warning instead of logging it at debug level. Changing it!
Also, now that you've pointed it out, I see some places (like this one) where the log is followed by an error code :sweat_smile: 
I'll get rid of the error code for this two logs in `autosubmit.py`:
```
Log.warning(
                        f"Autosubmit couldn't set your experiment as running on the autosubmit times database: "
                        f"{Path(BasicConfig.DB_DIR) / BasicConfig.AS_TIMES_DB}. Exception: {str(e)}",
                        7003,
                    )
```
```
except Exception as e:
                Log.warning(f"Autosubmit couldn't update the final experiment status for {expid}: {str(e)}", 7003)
```

autosubmit/history/database_managers/database_manager.py
ntorqulu: Created it in the internal repo, as any file created after without chmod set is world-writable: https://stackoverflow.com/questions/12116121/php-umask0-what-is-the-purpose 

Which can be dangerous...
I've to investigate more on this, not sure which are the implications. This is why I've moved it to the internal repo.

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -101,38 +120,74 @@ def get_experiment_status_row_by_exp_id(self, exp_id: int) -> Models.ExperimentS
         return Models.ExperimentStatusRow(*current_rows[0])
 
     def create_exp_status(self, exp_id: int, expid: str, status: str) -> int:
-        """Create experiment status."""
-        statement = ''' INSERT INTO experiment_status(exp_id, name,
-        status, seconds_diff, modified) VALUES(?,?,?,?,?) '''
+        """Insert a new experiment status row in the database.
+        
+        Raises IntegrityError if a row with the same exp_id already exists.
```
pablogoitia: > Raises IntegrityError if a row with the same **exp_id** already exists.

And/or "name", in this case, no?

autosubmit/history/database_managers/experiment_status_db_manager.py
pablogoitia, line 1: I spent a while reviewing this module and I haven't found any kind of inconsistency and everything looked good to me. Since this is my first time taking a look into the persistence part of AS it would be nice if another one could double-check it. Anyway, if not, I think the tests can also be a reliable reviewer in this case :)

pablogoitia (approved):
All files reviewed! I think that, since you have successfully rebased the code and all the tests passed, it can be merged, @ntorqulu. I am approving the changes, but please, check the comments I have left.

autosubmit/autosubmit.py
pablogoitia: Ohhh, I see. My first thought was just a matter of coding practice 😅 I think that the split of the imports in your PR will improve the performance a bit again then 👀 Thanks for the explanation @kinow!

autosubmit/autosubmit.py
kinow: Thanks!

autosubmit/history/database_managers/experiment_status_db_manager.py
ntorqulu: True!

ntorqulu:
@kinow, when you have time, could you do a quick review of this pr?
I think it's ready to merge, but I would like a double check it as it involves db operations.
It can be merged before or after the `autosubmit.py` split pr, whatever you prefer.

kinow:
@ntorqulu I am waiting for CICD to finish in the #3171 , and came to have a look at the PRs. I can review this one tomorrow, but looking at your changes, I think I can rebase and fix conflicts in under 10-20 minutes. So probably easier if yours is merged first (assuming we don't find anything to be fixed tomorrow). Or we do the other way too, and I can help you fixing conflicts here if needed. Cheers

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -33,24 +36,72 @@
...
+        with sqlite3.connect(autosubmit_db_path) as conn:
+            cursor = conn.cursor()
+            cursor.execute(
+                """
+                CREATE TABLE experiment (
+                    id INTEGER PRIMARY KEY,
+                    name TEXT NOT NULL,
+                    description TEXT NOT NULL,
+                    autosubmit_version TEXT
+                )
```
kinow: Maybe the test should do the same as the code, instead of writing the DDL here.

The `install` command calls `db_common.py`, which reads the `autosubmit.sql` to create the DB. You can call the `db_common#create_db`  function passing the same sql file. Having the extra stuff in that file shouldn't matter. And that way, if one changes `create_db`, they can find other places where this is used and update...

For the DB + path, there should be something in conftext that may be reused here to create DB + initial DDL, maybe. Or maybe we could have a fixture just to create an empty DB 🤔 for later I guess.

test/integration/test_expid.py
```diff
@@ -928,12 +928,12 @@ def test_delete_experiment_not_owner(mocker, tmp_path, autosubmit_exp, autosubmi
     assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/data").iterdir())
     assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/logs").iterdir())
     assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/structures").iterdir())
-    # Consult if the expid is not in the database
-    db_path = Path(BasicConfig.DB_PATH)
+    # Consult if the expid is still in the database (tombstone)
+    db_path = Path(f"{run_dir}/tests.db")
```
kinow: Why use the f-string with run_dir, and `tests.db` here, instead of the `DB_PATH` constant that was used before? I think both point to the same location?

Image

test/integration/test_expid.py
```diff
@@ -928,12 +928,12 @@ def test_delete_experiment_not_owner(mocker, tmp_path, autosubmit_exp, autosubmi
...
     with sqlite3.connect(db_path) as conn:
         cursor = conn.cursor()
         cursor.execute(f"SELECT name FROM experiment WHERE name='{as_exp.expid}'")
-        assert cursor.fetchone() is None
+        assert cursor.fetchone() is not None
```
kinow: Maybe we can drop the cursor, and close the conn (not only the transaction).

```
# from contextlib import closing
    db_path = Path(BasicConfig.DB_PATH)
    with closing(sqlite3.connect(db_path)) as conn:
        row = conn.execute(
            f"SELECT name FROM experiment WHERE name='{as_exp.expid}'"
        ).fetchone()
    assert row is not None
```

But not important.

test/integration/test_expid.py
```diff
@@ -882,12 +882,12 @@ def test_delete_experiment(mocker, tmp_path, autosubmit_exp, autosubmit: Autosub
     assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/data").iterdir())
     assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/logs").iterdir())
     assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/structures").iterdir())
-    # Consult if the expid is not in the database
-    db_path = Path(BasicConfig.DB_PATH)
+    # Consult if the expid is still in the database (tombstone)
+    db_path = Path(f"{run_dir}/tests.db")
```
kinow: Same question as in another test. Why use the run dir + tests.db, instead of `DB_PATH`? I think `DB_PATH` is configured somewhere in the `conftest.py` fixtures. Changing that one may break this test, I think...

test/integration/conftest.py
```diff
@@ -540,12 +541,23 @@ def as_db(request: "FixtureRequest", autosubmit: Autosubmit, tmp_path: "LocalPat
         raise ValueError(f'Unsupported database backend: {backend}')
 
     BasicConfig.read()
+    if backend == "postgres":
+        PostgreSQLEngineSingleton.reset()
```
kinow: Argh. I think we can get rid of this singleton completely. For another issue. 👍

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -89,53 +140,342 @@ def test_experiment_status_db_manager(tmp_path: 'LocalPath', as_db: str, use_sql
...
+            cursor = conn.cursor()
+            cursor.execute("""
+                CREATE TABLE experiment_status (
+                    exp_id INTEGER PRIMARY KEY,
+                    name TEXT NOT NULL,
+                    status TEXT NOT NULL,
+                    seconds_diff INTEGER NOT NULL,
+                    modified TEXT NOT NULL
+                )
+                """)
```
kinow: Probably better to use the same mechanism AS/conftest use to create the DB. To avoid issues in the tests if/when we change that.

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -89,53 +140,342 @@ def test_experiment_status_db_manager(tmp_path: 'LocalPath', as_db: str, use_sql
...
+            conn.commit()
+
+        # Initialize the manager
```
kinow: s/z/s 😬 so we keep our comments/codebase closer to British or Euro English.

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -33,24 +36,72 @@
...
+                CREATE TABLE experiment (
+                    id INTEGER PRIMARY KEY,
+                    name TEXT NOT NULL,
+                    description TEXT NOT NULL,
+                    autosubmit_version TEXT
+                )
+                """
+            )
+            cursor.executemany(
+                "INSERT INTO experiment (id, name, description, autosubmit_version) VALUES (?, ?, ?, ?)",
```
kinow: And this one maybe we can use `db_common`'s `_save_experiment`? Although you control the ID here, so maybe we can't use that function directly? Or maybe we could rely on `autosubmit_exp` for both SQLite and PG?

test/integration/experiment/test_experiment_common.py
```diff
@@ -37,21 +39,56 @@ def test_delete_experiment_cancelled(autosubmit_exp, mocker):
...
 def test_delete_experiment_failed(autosubmit_exp, mocker):
+    """Test that if the experiment deletion fails, the experiment is not deleted."""
     exp = autosubmit_exp(experiment_data={})
     mocker.patch('autosubmit.experiment.experiment_common._delete_experiment', side_effect=ValueError)
 
     assert not delete_experiment(exp.expid, force=True)
 
 
 def test_delete_experiment_that_is_running(autosubmit_exp, mocker):
+    """Test that if the experiment is running, the experiment is not deleted."""
```
kinow: 🙇 thank you!

autosubmit/database/db_common.py
```diff
@@ -597,29 +596,24 @@ def _last_name_used(test=False, operational=False, evaluation=False):
...
-        (conn, cursor) = open_conn()
-    except DbException as e:
-        raise AutosubmitCritical("Could not establish a connection to database", 7001, str(e))
-    cursor.execute('DELETE FROM experiment '
-                   'WHERE name=:name', {'name': experiment_id})
-    row = cursor.fetchone()
-    if row is None:
-        Log.debug(f'The experiment {experiment_id} has been deleted!!!')
-    close_conn(conn, cursor)
+        ExperimentStatus(experiment_id).set_as_deleted()
```
kinow: We are recreating an `ExperimentStatus` for every operation. The object seems to reload the `BasicConfig`, and create the database manager. Maybe we can improve this, somehow, to avoid recreating it in case we have a large simulation and this operation occurs a lot?

autosubmit/database/session.py
```diff
@@ -54,6 +54,14 @@ def get_instance(cls) -> Engine:
                 cls._instance = _resolve_engine(connection_url)
         return cls._instance
 
+    @classmethod
+    def reset(cls) -> None:
+        """Dispose the cached engine if it exists, so a new one is built on next use."""
+        with cls._lock:
+            if cls._instance is not None:
+                cls._instance.dispose()
+                cls._instance = None
```
kinow: For another issue, but I think we could have something like this,

```py
_postgres_engine: Engine | None = None
_postgres_connection_url: str | None = None
_postgres_lock = threading.Lock()


def _get_postgres_engine(connection_url: str) -> Engine:
    global _postgres_engine, _postgres_connection_url

    with _postgres_lock:
            if _postgres_engine is not None:
                _postgres_engine.dispose()

            _postgres_engine = _resolve_engine(connection_url)
            _postgres_connection_url = connection_url

        return _postgres_engine
```

In Python, it's common devs with OOP from Java, JavaScript to create classes and instantiate things. But in reality, modules are also objects, created during module import/first use.

We can use that if we don't need class-hierarchy, OOP, etc...

So here the "singleton" is replaced by a global variable. We lock/create a barrier, then check if it exists and the URL changed. If so, dispose. If it does not exist, create a new. Or... well, just return the engine (le singleton 🇫🇷 😆 ).

I think that may work, and simplify the code (but it's Friday, and I cannot really trust my 🧠 so close to the weekend). I will create a separate issue later.

autosubmit/database/tables.py
```diff
@@ -61,6 +61,7 @@
     Column("status", Text, nullable=False),
     Column("seconds_diff", Integer, nullable=False),
     Column("modified", Text, nullable=False),
+    Column("last_heartbeat", Text, nullable=True),
```
kinow: My sqlalchemy-fu is quite rusty 🙇 

Can't we use `DATETIME`? I think SqlAlchemy supports it for sqlite too, https://docs.sqlalchemy.org/en/21/dialects/sqlite.html#sqlalchemy.dialects.sqlite.DATETIME

Even though in SQLite it will be probably stored as text anyway. Maybe it could be the last_heartbeat in UTC time? 😬 Not sure if this needs to be in UTC, or if for the heartbeat it'd make sense to have the system TZ.

autosubmit/experiment/experiment_common.py
```diff
@@ -265,6 +250,17 @@ def _perform_deletion(experiment_path: Path, structure_db_path: Path, job_data_d
...
+    if not error_message:
+        Log.info(
+            f"Updating experiment status in {BasicConfig.DATABASE_BACKEND} database..."
+        )
+        try:
+            db_common.delete_experiment(expid_delete)
+            Log.result(f"Experiment {expid_delete} marked as deleted in database")
+        except Exception as e:
+            error_message.append(f"Cannot update experiment metadata: {e}")
+
```
kinow: Not related to your change directly, but I think this is a sort of XA/2PC (two-phase commit) problem, @ntorqulu -- https://en.wikipedia.org/wiki/Two-phase_commit_protocol

What AS does at the moment, I think, is remove the folders, and if that succeeds, then open a transaction and delete or mark as deleted (now) in the DB.

But if the DB step fails, then now the files were already removed. Not important as this is already possible to happen. I will test to confirm and open an issue for that later, some day 😬

autosubmit/history/database_managers/database_manager.py
```diff
@@ -107,35 +124,42 @@ def get_from_statement(self, path, statement):
...
+            cursor = conn.cursor()
+            cursor.execute(statement, arguments)
+            lastrow_id = cursor.lastrowid
+            conn.commit()
+            return lastrow_id
+        except Exception:
+            conn.rollback()
+            raise
+        finally:
+            conn.close()
```
kinow: Thank you for improving this!!! 🙇

autosubmit/history/database_managers/database_models.py
```diff
@@ -29,17 +30,23 @@
...
+)
 
 ExperimentRow = collections.namedtuple('ExperimentRow', ["id", "name", "autosubmit_version", "description"])
 
 PragmaVersion = collections.namedtuple('PragmaVersion', ['version'])
 MaxCounter = collections.namedtuple('MaxCounter', ['maxcounter'])
 
 
-class RunningStatus:
+class RunningStatus(str, Enum):
```
kinow: 👏 you can update the PR description with something like "Related to #2957" too. It will not close that, but will link the PR to that issue, so others are aware this PR is doing something related to that... but not important.

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -55,31 +54,51 @@ def __init__(
...
+                modified text NOT NULL,
+                last_heartbeat text
+            );""")
         self.execute_statement_on_dbfile(self._as_times_file_path, create_table_query)
 
-    def set_existing_experiment_status_as_running(self, expid: str) -> None:
-        """ Set the experiment_status row as running. """
-        self.update_exp_status(expid, Models.RunningStatus.RUNNING)
+        # backward compatibility
+        self._add_column_if_missing("last_heartbeat", "text")
```
kinow: It'll be great if someday we can fix #1286 so that these migrations are more clearly separated.

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -55,31 +54,51 @@ def __init__(
...
+        # keep only latest row by name
+        self.execute_statement_on_dbfile(
+            self._as_times_file_path,
+            """DELETE FROM experiment_status
+               WHERE exp_id NOT IN (
+                   SELECT MAX(exp_id)
+                   FROM experiment_status
+                   GROUP BY name
+               );""",
+        )
```
kinow: Is this needed for the API? I **think** not doing this doesn't break anything, but it's better to keep only the last one as the other ones are not used?

kinow: Looking at the as_times.db that I have, @ntorqulu I seem to recall that it was going to be deprecated? At least that's what my memory tells me today 😅 but I might be wrong.

It looks like the original design was to log every experiment status change. Here we will change and keep the last status only. If so, wouldn't it make sense to be a column of `experiment`? It's a ONE-to-ONE relationship between experiment and experiment_status, and in DB normalisation I think that's normally an indicator that something belongs/is part of another table.

I had to google that to remember the name 😆 Boyce-Codd normal form, or normal forms -- https://en.wikipedia.org/wiki/Database_normalization. I remember having to normally go through 1->3NF at least. But I didn't recall there were that many forms.

But trivia apart, I think if this table now will have one row per experiment, then it could be merged?

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -55,31 +54,51 @@ def __init__(
...
+            alter_query = (
+                f"ALTER TABLE experiment_status ADD COLUMN {column_name} {column_type};"
+            )
+            self.execute_statement_on_dbfile(self._as_times_file_path, alter_query)
+
+    def _column_exists(self, path: str, column_name: str) -> bool:
+        """Check whether a column exists in the experiment_status table for SQLite."""
+        query = "PRAGMA table_info(experiment_status);"
+        current_columns = [row[1] for row in self.get_from_statement(path, query)]
+        return column_name in current_columns
```
kinow: As this is Python, maybe we can simplify and [EAFP](https://docs.python.org/2/glossary.html#term-eafp)

```
def _add_column_if_missing(self, column_name: str, column_type: str) -> None:
    try:
        self.execute_statement_on_dbfile(
            self._as_times_file_path,
            f"ALTER TABLE experiment_status ADD COLUMN {column_name} {column_type}",
        )
    except sqlite3.OperationalError as e:
        if "duplicate column name" not in str(e):
            raise
```

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -151,18 +212,64 @@ def __init__(self) -> None:
         self.engine = session.get_engine(
             db_path=Path(BasicConfig.DB_DIR, BasicConfig.AS_TIMES_DB)
         )
+        self._validate_status_database()
+
+    def _validate_status_database(self) -> None:
```
kinow: Ah, looks like sqlalchemy one was not using the `_validate_status_database` like in sqlite. Thanks for matching both.

kinow (commented):
@ntorqulu I finished reviewing 22 files out of the 24. Stopped midway through `experiment_status.py`, and left `autosubmit.py` for when I have had more ☕ and am less tired 😅 

Left a few comments, and some random rambling. Feel free to ignore the latter! 😆 

I will continue next week. Once I'm done with the code review, I will do a few tests with real experiments. Cheers

autosubmit/autosubmit.py
kinow: Sorry, haven't looked at autosubmit.py today to close this one. I think it should be done, will review on the next pass.

docker/Dockerfile
```diff
@@ -126,7 +126,7 @@ USER $MAMBA_USER
 # Install Python
 RUN micromamba install --yes --name base --channel conda-forge \
       git=2.41.0 \
-      python=3.9.16 && \
+      python=3.10.20 && \
```
kinow: Ah! This is where we needed to update it 😆

kinow: If this fixes it, feel free to send a separate PR if you'd like, to fix it as I think other PRs will start failing that same job.

ntorqulu: yes, give me a second; it was just a test to see if it worked

ntorqulu: i've bad news, i'll comment in the `autosubmit.py` pr

kinow: Ooohh. Bearer of bad news, but with excellent reporting on what went wrong. Another good reason for us to automate our release process, https://github.com/BSC-ES/autosubmit/issues/2498 & https://github.com/BSC-ES/autosubmit/issues/2498

Thanks Núria!

kinow (commented):
Sorry for so many git conflicts. @ntorqulu I have two files pending review, autosubmit.py and autosubmit/history/experiment_status.py.

autosubmit/history/experiment_status.py is not in the lsit of conflicts. For autosubmit.py, what I'd do is

    squash commits here
    rebase interactively
    fix other files first (should be easier than autosubmit.py)
    then open autosubmit.py changes in GH UI
    make sure my rebase is now showing autosubmit.py as removed (by them)
    manually apply the changs of autosubmi.py again, to the code moved

get_historical_database is now in autosubmit.history.experiment_history

run_experiment is now experiment/manage, function run

archive is in experiment/manage, function archive

If you rebase, then GH will show which files changed. I believe experiment_status.py will remain pending review for me, but it will show what else I have to check (the changes that were in autosubmit.py). Then, I will finish the review 👍 🙇

ntorqulu:
Don't know exactly what happened but the pr was closed automatically and couldn't open it again.
Moving all the important info and conversations to this new one: https://github.com/BSC-ES/autosubmit/pull/3258

PR #2986, opened 2026-04-29
Enhancement: Detect inefficiencies in SIM jobs with CPMIPS thresholds - Part 4

paparici:
Hi AS team!

This is the fourth part of a large [PR](https://github.com/MissterP/autosubmit/pull/1) with the objective to implement the detection of inefficiencies once the SIM job is completed (or section indicated) using CPMIPS and thresholds. If a threshold is violated, it will send a notification to the user indicated in the config. 

Last PR: #2927 

This fourth part consists in adding the notification functions for the CPMIPS in MailNotifier.

**Check List**

- [ ] Documentation updated.


autosubmit/notifications/notifier.py
```diff
@@ -26,4 +26,13 @@ def notify_status_change(implementation, exp_id, job_name="", prev_status="", st
...
+        implementation.notify_experiment_status(exp_id,notify_to,platform)
+
+    @staticmethod
+    def notify_cpmip_threshold_violations(
+            implementation,
+            exp_id,
+            job_name="",
+            violations=None,
+            notify_to=""):
+        implementation.notify_cpmip_threshold_violations(exp_id, job_name, violations, notify_to)
```
VindeeR: EoF

test/integration/test_mail.py
```diff
@@ -141,6 +153,52 @@ def test_experiment_status(mail_notifier, fake_smtp_server, mock_platform):
...
+    emails = resp.json()["items"]
+    _check_metadata(
+        emails,
+        "CPMIP Threshold Violation detected",
+        expid,
+        'notifier@localhost',
+        to_email)
+
+    expected_body = _expected_cpmip_mail_body(expid, job_name, violations)
+    assert _normalize_mail_text(emails[0]["Content"]["Body"]) == _normalize_mail_text(expected_body)
```
VindeeR: The test looks solid overall, but I have a concern.

Right now, the email is generated twice using the same code, and the test compares those two results. Since both outputs come from the same generator, this doesn’t guarantee determinism you're essentially comparing one generated result against another.

While the test verifies consistency between “sent” and “generated” values, it doesn’t protect against unintended changes in the email structure.

It might be worth adding a deterministic check using a fixed, expected email (one defined manually with a known structure). That way, if the generation logic changes, the test will fail and signal that the output has been modified.

Do you think that approach would make sense?

test/unit/test_mail.py
```diff
@@ -220,3 +220,63 @@ def test_notify_status_change(
...
+        message_arg = mock_smtp.return_value.sendmail.call_args[0][2]
+        assert 'Subject: [Autosubmit] CPMIP Threshold Violation detected for Job1' in message_arg
+        assert 'Autosubmit notification' in message_arg
+        assert '----------------------------------------' in message_arg
+        assert 'Metric: SYPD' in message_arg
+        assert 'Comparison: must be >= effective bound (greater_than)' in message_arg
+        assert 'Configured threshold: 5.0' in message_arg
+        assert 'Accepted error (%): 10' in message_arg
+        assert 'Effective bound: 4.5' in message_arg
+        assert 'Observed value: 3.9' in message_arg
```
VindeeR: I believe it might be simpler to do something like this it is clearer and validates the structure too

```suggestion
        assert ("----------------------------------------\n
            "Metric: SYPD\n
            "----------------------------------------\n
            "Comparison: must be >= effective bound (greater_than)\n
            "Configured threshold: 5.0\n
            "Accepted error (%): 10\n
            "Effective bound: 4.5\n
            "Observed value: 3.9") == message_arg
```

VindeeR (changes requested):
The changes looks really good to mee, just did a few suggestion related to the tests

test/integration/test_mail.py
```diff
@@ -87,6 +88,40 @@ def _check_metadata(
...
+
+def _expected_cpmip_mail_body() -> str:
+    """Return the expected CPMIP threshold violation notification body."""
+    return (
+        "        Autosubmit notification\n\n"
+        "        -------------------------\n\n\n"
+        "        Experiment id:  a000\n\n\n"
+        "        Job name: SIM\n\n\n"
+        "        The following CPMIP metrics violated their configured thresholds:\n\n"
+        "        ----------------------------------------\n"
```
VindeeR: Are those spaces at the beginning expected to be like this? if not, I believe removing the blank spaces at the notifier should solve it

VindeeR (approved):
Changes looking good to me thanks 🙌

Issue #2664, opened 2025-10-21
HPC2020 issue with Autosubmit


eferre1:
Hi @dbeltrankyl @kinow, I'm having an issue with Autosubmit, we've checked a bit with @dbeltrankyl offline but it doesn't seem to be any of the usual mistakes.

### Description

When running an experiment in HPC2020 (ECMWF), using ecaccess as the connection mechanism, it stops without any clear error. It usually transfers some of the files (template and additional) and then stops. Probably during the ecaccess commands, and it might be receiving some output from it that it is not treated properly making the autosubmit process stop without dying (the process is still active in the machine, as well as the ecaccess). But checking manually the ecaccess commands I have not been able to reproduce the issue (it might also be due to too many simultaneous connections, which could be solved by using mput instead of put if that is the issue, but I have no idea).

- Short log:
```
9 of 9 jobs remaining (09:06)
No jobs to check for platform ECMWF-HPC2020
File /esarchive/autosubmit/t0lr/proj/auto-ecearth4/templates/synchronize.tmpl.sh exists
Calculating possible ready jobs for ECMWF-HPC2020
Section:SYNCHRONIZE can submit 20 jobs at this time
[WARNING] Script t0lr_SYNCHRONIZE has some empty variables. An empty value has substituted these variables
```
- ASLOG last lines:
```
9 of 9 jobs remaining (09:06)
2025-10-21 09:06:55,746 Sleep: 10
2025-10-21 09:06:55,746 Number of retrials: 0
2025-10-21 09:06:55,747 Checking jobs for platform=ECMWF-HPC2020
2025-10-21 09:06:55,748 No jobs to check for platform ECMWF-HPC2020
2025-10-21 09:06:55,750 Updating FAILED jobs
2025-10-21 09:06:55,818 Updating WAITING jobs
2025-10-21 09:06:55,819 Update finished
2025-10-21 09:06:55,826 Saving JobList: /esarchive/autosubmit/t0lr/pkl/job_list_t0lr.pkl.tmp
2025-10-21 09:06:55,854 JobList saved in /esarchive/autosubmit/t0lr/pkl/job_list_t0lr.pkl
2025-10-21 09:06:55,857 File /esarchive/autosubmit/t0lr/proj/auto-ecearth4/templates/synchronize.tmpl.sh exists
2025-10-21 09:06:55,859 Number of jobs available: 20
2025-10-21 09:06:55,860 Number of jobs ready: 1
[...]

PR #2987, opened 2026-04-30
2264 added  o txt option to as create

### LINKED ISSUE #2264: Add -o txt option to AS create
Dear all,

Getting the graph in a txt format is sometimes very useful but it is only available at `autosubmit monitor $expid`

Could we have it as well in the `autosubmit create $expid` in the same way?

In this way, the -o flags will be homogeneous along create and monitor, as for now they differ in this option:

### create options:

```
create specified experiment joblist

positional arguments:
  expid                 experiment identifier

optional arguments:
(...)
  -o {pdf,png,ps,svg}, --output {pdf,png,ps,svg}
                        chooses type of output for generated plot
(...)

```

### monitor options:

```
optional arguments:
  -h, --help            show this help message and exit
  -o {pdf,png,ps,svg,txt}, --output {pdf,png,ps,svg,txt}
                        chooses type of output for generated plot
(...)

```

Best,

Francesc

### LINKED ISSUE #2372: Replace print statements by log calls in code, remove print from tests
Print statements in tests are not useful as they are captured and ignored during normal runs in pytest. When there is a failure they may or may not be printed based on user settings, but they normally clutter the output (pytest already tries hard to provide good feedback from fixtures, and assertions).

Our code has some `print` statements too, even though we have `log.log.` and `logging`. We should replace that by proper `Log` calls. Example found while writing tests for #2187:

https://github.com/BSC-ES/autosubmit/blob/b34076442a02546b90b5c53361c7bdb792e3c22f/autosubmit/job/job_list.py#L3170

https://github.com/BSC-ES/autosubmit/blob/b34076442a02546b90b5c53361c7bdb792e3c22f/autosubmit/job/job_list.py#L3208

https://github.com/BSC-ES/autosubmit/blob/b34076442a02546b90b5c53361c7bdb792e3c22f/autosubmit/job/job_list.py#L3326

### LINKED ISSUE #2477: [refactor] Replace variable in list(dict.keys()) by variable in dict
Searching for a variable/value in a dictionary by key can be done simply by

```py
if var in dict:
   ...
```

It will automatically search the variable in the keys of the dictionary. I'm reviewing the code, especially of wrappers, for #2474 , and noticed several parts of the code where we check if the variable exists in the `dict.keys()` (unnecessary), or we wrap the `keys` in a `list` (also unnecessary).

Image

That should make our code a little more concise and maybe give us some microseconds in performance.

rajeshsuraj:
The changes made are

- Added txt to the -o flag choices in autosubmit create to match monitor.
- Made -d an alias for -o txt internally in create so that txt file gets created.
- With the autosubmit create <EXPID> -o txt command does not need the -plt flag to be passed.

Closes #2264 

**Check List**

- [X] Changelog entry included in `CHANGELOG.md` if this is a change that can affect users.
- [X] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 
 **Enhancements:**
 
+- Normalized the flags across create / monitor to have the txt option #2264
 - Fix characters escaped in regexes #2526
 - Fixed RO-Crate validation issues with the latest `rocrate-validator`,
   now Autosubmit crates conform to RO-Crate 1.1, process 0.5, workflow run
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 
 **Enhancements:**
 
-- Normalized the flags across create / monitor to have the txt option #2264
+- Added txt report generation using autosubmit create -o txt, similar to autosubmit monitor -o txt #2264"
 - Fix characters escaped in regexes #2526
 - Fixed RO-Crate validation issues with the latest `rocrate-validator`,
   now Autosubmit crates conform to RO-Crate 1.1, process 0.5, workflow run
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -4435,6 +4435,7 @@ class Autosubmit:
                                 "Couldn't recover the Historical database, AS will continue without it, GUI may be affected")
                     if detail:
                         output = 'txt'
+                    if output == 'txt':
                         noplot = False
                     if not noplot:
                         from .monitor.monitor import Monitor
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -4433,6 +4433,9 @@ class Autosubmit:
                         except Exception:
                             Log.warning(
                                 "Couldn't recover the Historical database, AS will continue without it, GUI may be affected")
+                    if detail:
+                        output = 'txt'
+                        noplot = False
                     if not noplot:
                         from .monitor.monitor import Monitor
                         if group_by:
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -532,7 +532,7 @@ class Autosubmit:
                                    help='hides plot window')
             subparser.add_argument('-d', '--detail', action='store_true',
                                    default=False, help='Show Job List view in terminal')
-            subparser.add_argument('-o', '--output', choices=('pdf', 'png', 'ps', 'svg'),
+            subparser.add_argument('-o', '--output', choices=('pdf', 'png', 'ps', 'svg', 'txt'),
                                    help='chooses type of output for generated plot')  # Default -o value comes from .conf
             subparser.add_argument('-group_by', choices=('date', 'member', 'chunk', 'split', 'automatic'), default=None,
                                    help='Groups the jobs automatically or by date, member, chunk or split')
--- a/docs/source/userguide/modifying_workflow/index.rst
+++ b/docs/source/userguide/modifying_workflow/index.rst
@@ -149,6 +149,8 @@ Additionally, you can have re-run only jobs that won't be included in the defaul
 
 By default, ``create`` does **not** generate plots. To generate plots of the new job list, use the ``-plt`` or ``--plot`` option.
 
+The only exception is ``-txt`` flag that does not require ``-plt`` or ``--plot`` option.
+
 ::
 
     autosubmit create <EXPID>
--- a/test/integration/test_autosubmit.py
+++ b/test/integration/test_autosubmit.py
@@ -24,6 +24,7 @@ from shutil import copy
 from typing import TYPE_CHECKING
 
 import pytest
+import time
 
 from autosubmit.autosubmit import Autosubmit
 from autosubmit.config.basicconfig import BasicConfig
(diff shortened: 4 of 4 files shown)


autosubmit/job/job_list.py
```diff
@@ -320,7 +320,7 @@ def split_by_platform(self):
 
     def _add_all_jobs_edge_info(self, dic_jobs, option="DEPENDENCIES"):
         jobs_data = dic_jobs.experiment_data.get("JOBS", {})
-        sections_gen = (section for section in jobs_data.keys())
+        sections_gen = (section for section in jobs_data)
         for job_section in sections_gen:
```
LuiggiTenorioK, line 324: I think that this change doesn't resolve the goal of #2477. 

By changing `(section for section in jobs_data.keys())` to `(section for section in jobs_data)`, both methods are functionally equal. In both, you will pass through the N elements of `jobs_data` and create a tuple with the keys. Then, in `for job_section in sections_gen:` you will go through the same N elements again.

However, the goal that @kinow mentioned in #2477 was to use the iterator directly instead of building the tuple mentioned above. With that, only one pass through the elements is done.


```suggestion
        for job_section in jobs_data:
```

autosubmit/job/job_list.py
```diff
@@ -401,7 +401,7 @@ def _add_dependencies(
                 self.dependency_map_with_distances[section].remove(section)
 
         # Generate all graph before adding dependencies.
-        for job_section in (section for section in jobs_data.keys()):
+        for job_section in (section for section in jobs_data):
```
LuiggiTenorioK: Unnecessary list of keys creation here

autosubmit/job/job_list.py
```diff
@@ -410,7 +410,7 @@ def _add_dependencies(
                       self.graph.nodes.get(job.name).get("job", None) is None):
                     self.graph.nodes.get(job.name)["job"] = job
 
-        for job_section in (section for section in jobs_data.keys()):
+        for job_section in (section for section in jobs_data):
```
LuiggiTenorioK: Unnecessary list of keys creation here

autosubmit/job/job_list.py
```diff
@@ -998,7 +998,7 @@ def _apply_jobs_edge_info(self, job: Job, dependencies: dict) -> None:
                                             filters_to_apply_by_section[key], parent)
 
     def find_current_section(self, job_section, section, dic_jobs, distance, visited_section):
-        sections = dic_jobs.as_conf.jobs_data[section].get("DEPENDENCIES", {}).keys()
+        sections = dic_jobs.as_conf.jobs_data[section].get("DEPENDENCIES", {})
```
LuiggiTenorioK: In this one, it is better to keep the `.keys()` since it is expected that the `sections` variable is a list of strings and not a dictionary.

autosubmit/job/job_list.py
```diff
@@ -3175,7 +3175,7 @@ def retrieve_packages(BasicConfig, expid, current_jobs=None):
                     else:
                         package_to_symbol[list_packages[i]] = 'hexagon'
             except Exception:
-                print((traceback.format_exc()))
+                Log.warning(traceback.format_exc())
```
LuiggiTenorioK: This traceback is more a debug than a warning message

autosubmit/job/job_list.py
```diff
@@ -3292,7 +3292,7 @@ def retrieve_times(status_code, name, tmp_path, make_exception=False, job_times=
                     finish_time = 0
 
         except Exception:
-            print((traceback.format_exc()))
+            Log.warning(traceback.format_exc())
```
LuiggiTenorioK: This traceback is more a debug than a warning message

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@
 
 **Enhancements:**
 
+- Normalized the flags across create / monitor to have the txt option #2264
+- Replaced redundant `list(dict.keys())` and `.keys()` patterns with direct dictionary iteration and membership checks, and replaced `print` statements with `Log.warning` calls in `job_list.py` #2477 #2372
```
LuiggiTenorioK: Better to separate this line of #2477 #2372 into two

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@
 
 **Enhancements:**
 
+- Normalized the flags across create / monitor to have the txt option #2264
```
LuiggiTenorioK: From the user side, this line is better explained if it says: "Added txt report generation using `autosubmit create -o txt`, similar to `autosubmit monitor -o txt` #2264"

LuiggiTenorioK (changes requested):
## Issue #2664

The changes seem good enough to close it. However, I need to ask you to create a test for it to ensure that the txt file is created or displayed depending on the command executed.

## Issue #2372 

There are some print statements that need to be in another log level (e.g., tracebacks should be in debug level, not warning). Then, there are other print statements (outside the test module and the job_list.py) that should be replaced, like in diagram.py. Please take a closer look.

## Issue #2477

I added multiple comments in the changes related to this issue. The goal was not only to remove the `.key()` method, if applicable (sometimes it is needed, like in one comment below). Also, it should avoid generating a list in "for" loops and using the iterator right away. Like in the previous issue, there are more places where this issue happens and need a closer look.

LuiggiTenorioK:
@rajeshsuraj also I think it is better if you split this PR into three, so we can handle each issue separately and not block the progress.

I suggest doing it before making more changes, as it will be easier to do it now than after.

LuiggiTenorioK (approved):
Seems good! The test is really appreciated here. Good job @rajeshsuraj! 💯

PR #2989, opened 2026-04-30
2477 iterator keys

### LINKED ISSUE #2477: [refactor] Replace variable in list(dict.keys()) by variable in dict

rajeshsuraj:
Closes #2477

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 
 **Enhancements:**
 
+- Replaced redundant `list(dict.keys())` and `.keys()` patterns with direct dictionary iteration and membership checks #2477
 - Replaced `print` statements with `Log.warning` calls in `job_list.py` and `diagram.py` #2372
 - Fix characters escaped in regexes #2526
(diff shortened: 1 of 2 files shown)


LuiggiTenorioK (approved):
The changes in the `job_list.py` seem good! As that file is the one that might cause more problems, this PR should be enough of an improvement.

However, there are more cases in other files along the codebase. A good way to catch more is to search using regex like `in\s.+\.keys\(\)` that searches for `in` and `keys()` statements.

If you want to continue the code cleaning, you can try the method above, but this is good as it is already.

PR #2994, opened 2026-05-05
[Enhancement] - Infinite loops while substituing variables

### LINKED ISSUE #2602: Infinite loops while substituing variables
### Description

https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/merge_requests/978#note_393129

The idea is to detect that a variable is referring itself and cut it by raising an AutosubmitCritical

### Reproducible Example

```yaml
Platforms:
  test:
    FDB_COPY_BIN: "%CURRENT_FDB_COPY_BIN%/fdb-copy"
    FDB_LIST_BIN: "%CURRENT_FDB_LIST_BIN%/fdb-list"
```

Causes an infinite loop, which causes mem issues and can lead to a crash. 

### Expected Behaviour

Raise an AutosubmitCritical

VindeeR:
closes #2602 

Creating a new validation in order to check and return a proper error message when validating dynamic variables avoiding infinite loop when auto referenced

**Check List**

- [ ] Documentation updated.


test/unit/conftest.py
```diff
@@ -378,3 +379,81 @@ def pbs_platform(autosubmit_config, tmp_path):
     aslogs.mkdir(parents=True, exist_ok=True)
     (aslogs / "submit_local.sh").touch()
     return PBSPlatform(expid="a000", name="pytest-pbs", config=as_conf.experiment_data)
+
+
+@pytest.fixture(scope="function")
+def general_data(tmp_path: Path) -> dict[str, Any]:
+    """
+    Provides part of the `experiment_data` dictionary used by the
```
kinow: Comment can be moved to first line of doc, `"""Provides part of...` 😬

test/unit/conftest.py
```diff
@@ -378,3 +379,81 @@ def pbs_platform(autosubmit_config, tmp_path):
...
+    integration tests in `commands`.
```
kinow: experiment_data and commands need to backticks to be formatted as monospace/code in pydocs/rst.

test/unit/conftest.py
```diff
@@ -378,3 +379,81 @@ def pbs_platform(autosubmit_config, tmp_path):
...
+
+    :param tmp_path: Temporary directory fixture from pytest.
+    :type tmp_path: Path
+    :return: A dictionary compatible with AutosubmitConfig.experiment_data
+    :rtype: Dict[str, object]
```
kinow: I think we don't need to duplicate the types in function signatures + pydoc. Here it's also using Dict from typing, whereas the signature has dict from stdlib (probably latter is better to be kept).

autosubmit/config/configcommon.py
```diff
@@ -2021,6 +2020,11 @@ def deep_add_missing_starter_conf(self, experiment_data, starter_conf):
                 experiment_data[key] = starter_conf[key]
             elif isinstance(starter_conf[key], collections.abc.Mapping):
                 experiment_data[key] = self.deep_add_missing_starter_conf(experiment_data[key], starter_conf[key])
+            if isinstance(experiment_data[key], str) and "%" in experiment_data[key]:
+                check_value = experiment_data[key].split("%")[1]
+                special_var = check_value[0] == "^"
+                if key == check_value[8:] or (special_var and key == check_value[9:]):
```
kinow: I was just going through open pull requests, and saw this one assigned to Dani for review. I opened just to know what's this about, and saw this :8. Why is :8 and :9 used here? Maybe that could have a comment somewhere... or is there a chance this was tailored for the test written?

VindeeR: This validation checks whether a `Dynamic Variable` references itself, which would create an infinite loop.

During validation, all variable names are prefixed with `current_`. However, variables starting with the `^` character can break this logic because the special character becomes part of the generated reference name.

```
variable: %variable%
generated reference: current_variable
result: variable -> variable (infinite loop)
```
```
variable: %^variable%
generated reference: current_^variable
result: variable -> variable (infinite loop)
```

In the second case, the `^` symbol is included in the generated reference name, which causes the self-reference validation to fail.

VindeeR:
Thanks for the review @kinow addressed the pointed correction

test/unit/conftest.py
kinow: Still pending these 😬

autosubmit/config/configcommon.py
kinow: Ah, so that's because of the current_, thanks! I think that deserves a comment in the code so others are aware where that's coming from.

test/integration/test_autosubmit.py
```diff
@@ -453,4 +453,14 @@ def test_create_txt_output_writes_status_file(autosubmit_exp):
...
\ No newline at end of file
+    assert len(txt_files_after_detail) == 2, "Expected a second txt file in status/ for -d"
+
+
+def test_infinite_loop_dynamic_variable(autosubmit_exp):
+    with pytest.raises(AutosubmitCritical) as ac:
+        autosubmit_exp(experiment_data={
+            'JOB': {
+                'FDB_COPY_BIN': "%CURRENT_FDB_COPY_BIN%/fdb-copy",
+            }
```
kinow, line 499: Maybe move 'JOB' to under 'JOBS'. I debugged the code and noticed it a custom attribute, and not a real job (just to make sure the config parser is triggering the logic for the jobs in the test).

kinow (commented):
Sorry, I have to start working on something else. Nearly done, I have one thing that I am testing in PyCharm, but I think it may be for another issue. I'll send the comments I had saved here so you can have a look in the meantime, Erick. Ta

autosubmit/config/configcommon.py
```diff
@@ -2075,6 +2074,15 @@ def deep_add_missing_starter_conf(self, experiment_data, starter_conf):
                 experiment_data[key] = starter_conf[key]
             elif isinstance(starter_conf[key], collections.abc.Mapping):
                 experiment_data[key] = self.deep_add_missing_starter_conf(experiment_data[key], starter_conf[key])
+            # This validation checks whether a `Dynamic Variable ` references itself, which would create an infinite loop.
```
kinow: s/`Dynamic Variable `/dynamic variable... or `Dynamic Variable` ? Asking because of the extra space before the last backtick 👍

autosubmit/config/configcommon.py
```diff
@@ -2075,6 +2074,15 @@ def deep_add_missing_starter_conf(self, experiment_data, starter_conf):
...
+            # During validation, all variable names are prefixed with `current_` hence the [8:].
+            # However, special variables start with `^` that can break this logic since
```
kinow, line 2018: since it becomes? I think it's missing the it

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
 - Fixed `CPU per task` for new version of autosubmit #2897
 - Fixed issue overwritting expid config variables with the ones from the github repo #2877
 - Fixed inspect infinite loop when PLATFORMS.TOTALJOBS or PLATFORMS.MAX_WAITING_JOBS is set to 0 #2749
+- Fixed the infinite loop generated by point a dynamic variable at itself by raising an error #2602
```
kinow: I think there's something missing in the point a dynamic variable. Maybe by pointing? Or maybe "by a dynamic variable referencing itself (e.g., var=%CURRENT_var%)"?

test/unit/experiment/test_dynamic_variables.py
```diff
@@ -0,0 +1,57 @@
...
+        },
+        {
+            'TEST_SLURM': {
+                'TESTS': '%"CURRENT_TESTS%/fdb-copy',
+                'FDB_DICT_BIN': 'TEST_FOLDER/%^CURRENT_FDB_DICT_BIN%/fdb-list',
+            },
+        }, {
+            'TEST_SLURM': {
+                'FDB_COPY_BIN': 'TEST/%CURRENT_FDB_COPY_BIN%/fdb-copy',
+                'FDB_LIST_BIN': 'TEST/%CURRENT_FDB_LIST_BIN%/fdb-list',
```
kinow: I think you need to have one variable per test. I modified `FDB_LIST_BIN`, thinking it would break the test, but it didn't. That's because FDB_COPY_BIN still causes the exception to be raised.

kinow: Second question: What about this?

```
            'TEST_SLURM': {
                'fdb_list_bin': 'TEST/%CURRENT_FDB_LIST_BIN%/fdb-list',
            },
```

Is this ever allowed or possible? Right now doing so causes the test to fail (I removed the `FDB_COPY_BIN` as otherwise it would still pass).

kinow: Maybe include a test with the lower case current too? It seems to be detected fine with your current code 👍 

```
                'FDB_LIST_BIN': 'TEST/%current_FDB_LIST_BIN%/fdb-list',
```

kinow: And do we have a test that without the CURRENT_ it doesn't raise the error??

I don't know if that should or not. I guess the variable is always prefixed by CURRENT_? This one doesn't raise anything right now,

```
            'TEST_SLURM': {
                'FDB_LIST_BIN': 'TEST/%FDB_LIST_BIN%/fdb-list',
            },
```

kinow (changes requested):
Ok, took me a while, but I think I managed to reproduce the infinite loop, and almost crash my laptop 😆 

@VindeeR here you go

```
JOBS:
  LOCAL_SETUP:
    FILE: LOCAL_SETUP.sh
    PLATFORM: 'LOCAL'
    SCRIPT: >
      echo "OK!"
      echo "SCRIPT: %current_script%"
```

Last week I mentioned I was going to create a bug about lowercase/uppercase in AS. Even with the validation that @ntorqulu started working on, I think the current situation is a bit of a mess (cc @LuiggiTenorioK). This bug is similar to what I found in platforms, but here both `Current_scripT`, or this version all in lower case still causes the infinite loop. Internally we probably should default to lower-case + enums, unless we need to keep the same value used by the user. I don't know.

Anyhow, there's more to do in this PR, @VindeeR. Besides fixing this corner case, check my comments in the unit tests. I think we can add a few more cases 👍 But great job so far chasing this bug!!!

kinow:
Tested with `% CURRENT_SCRIPT%` (space), `"\x25CURRENT_SCRIPT\x25"` & `echo "SCRIPT: %25CURRENT_SCRIPT%25"` (encoding the percentage). The upper/lower case was the only that triggered it.

autosubmit/config/configcommon.py
VindeeR: the space was a type, good catch, thanks :raised_hands: 
`dynamic variable` should do it, I like adding the upper case on those, but it'll be clear not using it

test/unit/experiment/test_dynamic_variables.py
VindeeR: > I think you need to have one variable per test. I modified `FDB_LIST_BIN`, thinking it would break the test, but it didn't. That's because FDB_COPY_BIN still causes the exception to be raised.

This is the expected behavior, if I'm not mistaken, as long as there's at least one infinite loop it should always raise the error, in the test you select there are 2 variable self referencing

The only scenario that would still be happening would be the cross reference, which I'm not sure how should be handled.
```
{
    'TEST_SLURM': {
        'FDB_COPY_BIN': 'TEST/%CURRENT_FDB_LIST_BIN%/fdb-copy',
        'FDB_LIST_BIN': 'TEST/%CURRENT_FDB_COPY_BIN%/fdb-list',
    },
}
```



> Is this ever allowed or possible? Right now doing so causes the test to fail (I removed the FDB_COPY_BIN as otherwise it would still pass).

That would never happen because AS always generate those variables as uppercase, hence why it is not treated in the code, although, if this behavior was ever to change it would fail, but many other things would fail too since other parts of the code are case sensitive, despite not being a consistent thing throughout the project.



> Maybe include a test with the lower case current too? It seems to be detected fine with your current code

currently the `current` case sensitivity is ignored for the treatment, as long as the word has that specific amount of characters, `current` just indicates is the one being considered in the code for the substitutions. AS always generate the new file with variables as `CURRENT_` to which I decided to simply ignore to make the code easier to handle.

> And do we have a test that without the CURRENT_ it doesn't raise the error??
> 
> I don't know if that should or not. I guess the variable is always prefixed by CURRENT_? This one doesn't raise anything right now,
> 
> ```
>             'TEST_SLURM': {
>                 'FDB_LIST_BIN': 'TEST/%FDB_LIST_BIN%/fdb-list',
>             },
> ```

It doesn't raise an error because only the `CURRENT_` variable will be loaded, which makes it so that all the variable that does not contains it to be discarded it will never get to that piece of code

VindeeR: This was a pretty lengthy answer let me know if anything was confusing and I'll try my best to clarify

VindeeR:
> Tested with `% CURRENT_SCRIPT%` (space), `"\x25CURRENT_SCRIPT\x25"` & `echo "SCRIPT: %25CURRENT_SCRIPT%25"` (encoding the percentage). The upper/lower case was the only that triggered it.

Correct me if I'm wrong, but I believe that when configuring the AS experiment they wouldn't use the dynamic variable as `%CURRENT_SCRIPT%` but as `%SCRIPT%` and after generating the files AS would substitute the variable for `%CURRENT_SCRIPT%` which would make them standard uppercase always.

test/unit/experiment/test_dynamic_variables.py
kinow: Yup, I think that's all correct. But I think I wasn't clear. What I mean is that all tests have two strings, testing that both fail. In reality, only one will fail, causing the issue.

It might be worth confirming both that a single string causes the failure, as well as that two or more strings cause the error.

The test could also be updated to support positive & negative. Cases where the strings are valid (i.e., not self-referencing) and invalid. But I think you have that somewhere else. Just having more cases like a single string that's invalid... two strings that are invalid... a single string with that caret^ char, two strings, one with that char, two strings, both with that char, and so it goes.

kinow:
> > Tested with `% CURRENT_SCRIPT%` (space), `"\x25CURRENT_SCRIPT\x25"` & `echo "SCRIPT: %25CURRENT_SCRIPT%25"` (encoding the percentage). The upper/lower case was the only that triggered it.
> 
> Correct me if I'm wrong, but I believe that when configuring the AS experiment they wouldn't use the dynamic variable as `%CURRENT_SCRIPT%` but as `%SCRIPT%` and after generating the files AS would substitute the variable for `%CURRENT_SCRIPT%` which would make them standard uppercase always.

Do you mean in the user template scripts? They will have some variables used directly, and others with CURRENT_.

e.g.,

`CURRENT_ARCH` in https://github.com/DestinE-Climate-DT/Workflow/blob/e9f7ad7ec127eeda152b55d2fa259c03dbd6c086/templates/application.sh#L66

and

`HPC_` vars, https://github.com/DestinE-Climate-DT/Workflow/blob/e9f7ad7ec127eeda152b55d2fa259c03dbd6c086/templates/application.sh#L99-L100

The script also uses locally created variables like `SCRIPTDIR`, that use user configuration. Looking at the DestinE config I think you will be able to tell what users used, and come up with some test scenarios for the PR :+1: 

If you mean other YAML files, I believe they can use either. If it's under the Job in the user YAML job configuration, the CURRENT_ could be used (I believe).

VindeeR:
I wasn't aware they could use the variables `current_`, as I thought this was used only by AS after generating them, but I can see that the users could be able to use them if they want, it does feels not proper, why would it be used the variables in its `current_` format instead of declared by the user format, I`ll try to think of a way to handle this.

.benchmarks/artifacts/performance-metrics-by-iteration-4.2.0.csv
kinow, line 1: @VindeeR did you intend to commit these? If you run the tests locally, you get that `:memory:` file, and performance artifacts. I think both should be commented out/disabled by default (the latter we can probably disable via pytest.ini + markers). It annoys a bit, but I haven't had time to look into that yet. For now I think you can just remove these 2 benchmark files.

CHANGELOG.md
```diff
@@ -84,6 +84,7 @@ This release also includes several bug fixes and enhancements to improve the ove
 - Fixed inspect infinite loop when PLATFORMS.TOTALJOBS or PLATFORMS.MAX_WAITING_JOBS is set to 0 #2749
 - Fixed wrappers on Lumi platform #3059 
 - Fixed HPC2020 platform #3059
+- Fixed the infinite loop generated by pointing a dynamic variable referencing itself (e.g., var=%CURRENT_var%) raising an error #2602
```
kinow: 💪 👏 good changelog entry w/ an example!

test/unit/conftest.py
```diff
@@ -425,3 +426,88 @@ def fake_job_list(mocker) -> JobList:
     job_list = JobList('a000', as_conf, YAMLParserFactory())
     job_list.dbmanager = mocker.MagicMock()
     return job_list
+
+
+@pytest.fixture(scope="function")
+def general_data(tmp_path: Path) -> dict[str, Any]:
```
kinow, line 420: I think this looks useful, but right now it's a general fixture, used in a single test (all xdist sessions/tests still load it when pytest is launched).

Do you plan to make other tests use this fixture, @VindeeR ? If so, maybe create an issue so it's not forgotten/duplicated.

autosubmit/config/configcommon.py
```diff
@@ -2470,9 +2473,11 @@ def set_safetysleeptime(self, sleep_time: int):
         :param sleep_time: value to set
         :type sleep_time: int
         """
-        content = open(self._conf_parser_file).read()
-        content = content.replace(re.search('SAFETYSLEEPTIME:.*', content).group(0), "SAFETYSLEEPTIME: %d" % sleep_time)
-        open(self._conf_parser_file, 'w').write(content)
+        with open(self._conf_parser_file) as content_in_file:
```
kinow: 👌 👏

autosubmit/config/configcommon.py
```diff
@@ -2017,10 +2009,19 @@ def deep_add_missing_starter_conf(self, experiment_data, starter_conf):
...
+            if key not in experiment_data:
                 experiment_data[key] = starter_conf[key]
             elif isinstance(starter_conf[key], collections.abc.Mapping):
                 experiment_data[key] = self.deep_add_missing_starter_conf(experiment_data[key], starter_conf[key])
+            # This validation checks whether a `dynamic variable` references itself, which would create an infinite loop.
+            # During validation, all variable names are prefixed with `current_` hence the [8:].
+            # However, special variables start with `^` that can break this logic since
+            # it becomes part of the generated reference name.
+            if isinstance(experiment_data[key], str) and "%" in experiment_data[key]:
+                check_value = experiment_data[key].split("%")[1]
```
kinow: We can avoid accessing the dictionary twice, by moving it to a variable just before the `if`. Not super important here because it's accessed only twice.

```
bla = experiment_data[key]
if isinstance(bla, str) and "%" in bla": ...

```

autosubmit/config/configcommon.py
```diff
@@ -2017,10 +2009,19 @@ def deep_add_missing_starter_conf(self, experiment_data, starter_conf):
...
+                special_var = check_value[0] == "^"
```
kinow: Can the `experiment_data[key]` be a single `%`?

```
kinow@ranma:~$ python3
Python 3.14.4 (main, Jun 18 2026, 14:25:02) [GCC 15.2.0] on linux
>>> checked_value = "%".split("%")[1]
>>> checked_value
''
>>> special_var = checked_value[0] == "^"
  File "<python-input-3>", line 1, in <module>
    special_var = checked_value[0] == "^"
                  ~~~~~~~~~~~~~^^^
IndexError: string index out of range
>>> 

```

autosubmit/config/configcommon.py
kinow: I think sometimes the `.split("%")` can give different results, depending on the user input.

```
>>> "foo%bar%baz".split("%")
['foo', 'bar', 'baz']
>>> "%bar%baz".split("%")
['', 'bar', 'baz']
>>> "bar%baz".split("%")
['bar', 'baz']

```

autosubmit/config/configcommon.py
```diff
@@ -2017,10 +2009,19 @@ def deep_add_missing_starter_conf(self, experiment_data, starter_conf):
...
+                if key.upper() == check_value[8:].upper() or (special_var and key.upper() == check_value[9:].upper()):
```
kinow: Normally it's `.lower` or `.casefold`. But `upper` should work too (the casefold is useful if we have german/turkish data/etc... that have some special chars -- pt/sp/english are all safe, as far as I know).

autosubmit/config/configcommon.py
```diff
@@ -2017,10 +2009,19 @@ def deep_add_missing_starter_conf(self, experiment_data, starter_conf):
...
+                    raise AutosubmitCritical(f"Dynamic variables {key} causing infinite recursion during evaluation")
```
kinow: s/variables/variable? I think `key` will be a single variable. Maybe we could print a bit more of context, like the variable and the definition (like `KEY=%CURRENT_KEY%` or so), if we have that in the context here.

autosubmit/config/configcommon.py
```diff
@@ -1996,15 +1989,14 @@ def quick_deep_diff(self, current_data, last_run_data, changed=False):
...
-                else:
-                    if key not in last_run_data.keys() or str(last_run_data[key]).lower() != str(val).lower():
-                        changed = True
-                        break
+                elif key not in last_run_data or str(last_run_data[key]).lower() != str(val).lower():
+                    changed = True
+                    break
         except Exception:
             changed = True
         return changed
```
kinow, line 2007: Not part of your change, but I think there's no good reason to have a `changed` variable at all in this function.

Right now:

```py
        if not current_data:
            return changed
        if changed:
            return True
        try:
            for key, val in current_data.items():
                if isinstance(val, collections.abc.Mapping):
                    if not last_run_data or key not in last_run_data:
                        changed = True
                        break
                    else:
                        changed = self.quick_deep_diff(last_run_data[key], val, changed)
                elif key not in last_run_data or str(last_run_data[key]).lower() != str(val).lower():
                    changed = True
                    break
        except Exception:
            changed = True
        return changed
```

There's no reason to not return immediately.

```py
    # A function like this deserves good docstrings, and/or code comments.
    # I'd remove changed from function signature (why would changed==True, call the function, and then return True again????0 and then:

    if not current_data:
        return False

    try:
        for key, val in current_data.items():
            if key not in last_run_data:
                return True

            previous_val = last_run_data[key]

            if isinstance(val, collections.abc.Mapping):
                if self.quick_deep_diff(previous_val, val):
                    return True
            elif str(previous_val).lower() != str(val).lower():
                return True

        return False

    except Exception:
        # TODO: This is bad and could hide real bugs....
        return True

```

I'd need to write a unit test to compare the results before/after, but I think the results through the boolean logic should match...

Your change above looks harmless, but we just need to take care to check if there are tests... let's leave discussing about fixing this to another issue/later👍

:memory:
kinow, line 1: This file was probably accidentally added when running the tests.

kinow (commented):
A few questions, some test data that needs to be removed, I think. I tested the tests without the code fix and they fail, which indicates the code is really fixing the bug (or at least the tests are testing 😬 ) Thanks Erick!

kinow:
Also, merging conflicts :+1:

.benchmarks/artifacts/performance-metrics-by-iteration-4.2.0.csv
VindeeR: I'll add them to the .ignore file they were not intended to be here

test/unit/conftest.py
VindeeR: I think it is useful, and can be used for other tests, I will create an issue to evaluate where it makes sense to use it and substitute it.

Thanks :raised_hands:

.benchmarks/artifacts/performance-metrics-by-iteration-4.2.0.csv
dbeltrankyl: Maybe you ran all the tests with all the markers locally, and the file was generated? 

The benchmark stuff was merged somewhat incompletely due to the 4.2.0 merge in master, #3162 this branch is finishing that work

test/unit/conftest.py
kinow: Thanks!

autosubmit/config/configcommon.py
VindeeR: Since this is checking dynamic variables that are defined by `%<variable>%` I believe it shouldn't be but the user could very well make a typo which should fail, maybe creating a more explicit and clear error message would be good

autosubmit/config/configcommon.py
VindeeR: The last one you commented is not a `dynamic variable` so it should substitute anything AFAIK.

So the idea would be get the firs element in between the %% which would be the second element of the split

autosubmit/config/configcommon.py
VindeeR: So I'll change to `.casefold()` just in case since we have users from different languages

autosubmit/config/configcommon.py
VindeeR: I believe the key shouldn't be repeated right? but I'll add the `experiment_data[key]` value for context, I believe printing the `experiment_data` might make things too verbose and confusing

test/integration/test_autosubmit.py
```diff
@@ -680,3 +680,16 @@ def test_set_status_with_detail(autosubmit_exp):
     )
 
     assert result is True
+
+
+
```
kinow, line 685: three blank lines, I think it's supposed to be 2? 😬

autosubmit/config/configcommon.py
```diff
@@ -2022,10 +2014,23 @@ def deep_add_missing_starter_conf(self, experiment_data, starter_conf):
...
+            # it becomes part of the generated reference name.
+            key_experiment_data = experiment_data[key]
+            if isinstance(key_experiment_data, str) and "%" in key_experiment_data:
+                try:
+                    check_value = key_experiment_data.split("%")[1]
+                    special_var = check_value[0] == "^"
+                    if key.casefold() == check_value[8:].casefold() or (special_var and key.casefold() == check_value[9:].casefold()):
+                        raise AutosubmitCritical(f"Dynamic variables {key} causing infinite recursion during evaluation of the element {key_experiment_data}")
+                except IndexError:
+                    raise AutosubmitCritical(f"The following key: {key} Has a problem with it's value {key_experiment_data}")
```
kinow, line 2033: s/it's/its

And I think "has" and not "Has".

Maybe something like

```py
raise AutosubmitCritical(
    f"Invalid value for key '{key}': {key_experiment_data}"
)
```

kinow (commented):
Code's looking good. Left the last few comments I think. Will run a couple of tests now, and if nothing pops up, and you can look at these latest review comments, then I think it'll be ready to merge soon!

kinow (commented):
@VindeeR maybe I'm doing something wrong. Or maybe you could have a look if you can reproduce this.

``` bash
$ autosubmit expid -H local -d "bla"
```

Edit <expid>/conf/jobs.yml, and set it to the following:

```yaml
JOBS:
  A:
    SCRIPT: |
      echo "That is OK %TEST.TEST_NAME%"
    RUNNING: once
    WALLCLOCK: 00:05
TEST:
  NAME: "BRUNO"
  TEST_NAME: "Hi %TEST.TEST_NAME%"
```

Maybe, first replace the `SCRIPT` to `echo "That is OK %TEST.NAME%"`.

Then `autosubmit create <Expid> && autosubmit refresh <expid> && autosubmit inspect <Expid>`. Then, open `<expid>/tmp/<expid>_A.cmd`.

Near the bottom, you should have

```bash
################
# Autosubmit job
################
echo "That is OK BRUNO"

###################
# Autosubmit tailer
###################
```

Now change it back to 

```yaml
SCRIPT: |
      echo "That is OK %TEST.TEST_NAME%"
```

I CTRL+C my process after a few seconds, as it looked like it was walking towards an ugly recursion error, maybe due to a bug replacing variables.

VindeeR:
> @VindeeR maybe I'm doing something wrong. Or maybe you could have a look if you can reproduce this.
> 
> ```shell
> $ autosubmit expid -H local -d "bla"
> ```
> 
> Edit /conf/jobs.yml, and set it to the following:
> 
> ```yaml
> JOBS:
>   A:
>     SCRIPT: |
>       echo "That is OK %TEST.TEST_NAME%"
>     RUNNING: once
>     WALLCLOCK: 00:05
> TEST:
>   NAME: "BRUNO"
>   TEST_NAME: "Hi %TEST.TEST_NAME%"
> ```
> 
> Maybe, first replace the `SCRIPT` to `echo "That is OK %TEST.NAME%"`.
> 
> Then `autosubmit create <Expid> && autosubmit refresh <expid> && autosubmit inspect <Expid>`. Then, open `<expid>/tmp/<expid>_A.cmd`.
> 
> Near the bottom, you should have
> 
> ```shell
> ################
> # Autosubmit job
> ################
> echo "That is OK BRUNO"
> 
> ###################
> # Autosubmit tailer
> ###################
> ```
> 
> Now change it back to
> 
> ```yaml
> SCRIPT: |
>       echo "That is OK %TEST.TEST_NAME%"
> ```
> 
> I CTRL+C my process after a few seconds, as it looked like it was walking towards an ugly recursion error, maybe due to a bug replacing variables.

Thanks for catching it I'll try this scenery and hopefully get anything that might be not working

VindeeR:
> @VindeeR maybe I'm doing something wrong. Or maybe you could have a look if you can reproduce this.
> 
> ```shell
> $ autosubmit expid -H local -d "bla"
> ```
> 
> Edit /conf/jobs.yml, and set it to the following:
> 
> ```yaml
> JOBS:
>   A:
>     SCRIPT: |
>       echo "That is OK %TEST.TEST_NAME%"
>     RUNNING: once
>     WALLCLOCK: 00:05
> TEST:
>   NAME: "BRUNO"
>   TEST_NAME: "Hi %TEST.TEST_NAME%"
> ```
> 
> Maybe, first replace the `SCRIPT` to `echo "That is OK %TEST.NAME%"`.
> 
> Then `autosubmit create <Expid> && autosubmit refresh <expid> && autosubmit inspect <Expid>`. Then, open `<expid>/tmp/<expid>_A.cmd`.
> 
> Near the bottom, you should have
> 
> ```shell
> ################
> # Autosubmit job
> ################
> echo "That is OK BRUNO"
> 
> ###################
> # Autosubmit tailer
> ###################
> ```
> 
> Now change it back to
> 
> ```yaml
> SCRIPT: |
>       echo "That is OK %TEST.TEST_NAME%"
> ```
> 
> I CTRL+C my process after a few seconds, as it looked like it was walking towards an ugly recursion error, maybe due to a bug replacing variables.

Okay so after further investigation I believe that it is not an issue, since we can't predict how deep that recursion can be and how many of it can be done at the same file what will happen is, it'll run for an X amount of times, where X will be the length of the dynamic variable, this way stopping the recursion once the X reaches 0 and cancelling the build of the experiment.

What I believe can be done to make it clear is pop an error telling which is the variable in question, but the behavior you mentioned is already treated.

kinow:
> Okay so after further investigation I believe that it is not an issue, since we can't predict how deep that recursion can be and how many of it can be done at the same file what will happen is, it'll run for an X amount of times, where X will be the length of the dynamic variable, this way stopping the recursion once the X reaches 0 and cancelling the build of the experiment.

So what happens if a user has a variable self-referencing? Wouldn't that cause an infinite loop while substituting variables, and crash the user workflow?

VindeeR:
> > Okay so after further investigation I believe that it is not an issue, since we can't predict how deep that recursion can be and how many of it can be done at the same file what will happen is, it'll run for an X amount of times, where X will be the length of the dynamic variable, this way stopping the recursion once the X reaches 0 and cancelling the build of the experiment.
> 
> So what happens if a user has a variable self-referencing? Wouldn't that cause an infinite loop while substituting variables, and crash the user workflow?

yes, but it was designed a stop after `len(dynamic_variable)` lets say that we have a depth of 20 in recursion until it reach a value and then substituting everything, but if the depth is deeper than `len(dynamic_variable)` it will stop regardless if you have a value

kinow:
> > > Okay so after further investigation I believe that it is not an issue, since we can't predict how deep that recursion can be and how many of it can be done at the same file what will happen is, it'll run for an X amount of times, where X will be the length of the dynamic variable, this way stopping the recursion once the X reaches 0 and cancelling the build of the experiment.
> > 
> > 
> > So what happens if a user has a variable self-referencing? Wouldn't that cause an infinite loop while substituting variables, and crash the user workflow?
> 
> yes, but it was designed a stop after `len(dynamic_variable)` lets say that we have a depth of 20 in recursion until it reach a value and then substituting everything, but if the depth is deeper than `len(dynamic_variable)` it will stop regardless if you have a value

But is that the best we can do? Can't we identify it will lead to a recursion, and warn the user? In the example I sent, I think there's nothing we can do about it, whether the depth limit is 1, 20, or 100, it will eventually crash. And a user setting a recursive value like that might prefer to be warned they accidentally added a circular dependency, that they have to fix instead?

VindeeR:
> > > > Okay so after further investigation I believe that it is not an issue, since we can't predict how deep that recursion can be and how many of it can be done at the same file what will happen is, it'll run for an X amount of times, where X will be the length of the dynamic variable, this way stopping the recursion once the X reaches 0 and cancelling the build of the experiment.
> > > 
> > > 
> > > So what happens if a user has a variable self-referencing? Wouldn't that cause an infinite loop while substituting variables, and crash the user workflow?
> > 
> > 
> > yes, but it was designed a stop after `len(dynamic_variable)` lets say that we have a depth of 20 in recursion until it reach a value and then substituting everything, but if the depth is deeper than `len(dynamic_variable)` it will stop regardless if you have a value
> 
> But is that the best we can do? Can't we identify it will lead to a recursion, and warn the user? In the example I sent, I think there's nothing we can do about it, whether the depth limit is 1, 20, or 100, it will eventually crash. And a user setting a recursive value like that might prefer to be warned they accidentally added a circular dependency, that they have to fix instead?

Maybe we can keep track of the elements that were used to substitute and add them to an array, and if the element is already there we throw an error with the list of recursive elements? 

What do you have any other ideas to solve it that would create little to no impact to other developers?

VindeeR:
I still have a lot of cleaning to do in this code before being ready for proper review :melting_face:

VindeeR:
I did a bunch of manual tests and it seems to be working, the only part I'm a bit unsure is related to the `CURRENT_` variables, I believe It was properly tested, but I'm not extremely confident if there are other scenarios that I could be missing.

@kinow, I think it should be fine to be reviewed.

autosubmit/config/configcommon.py
```diff
@@ -288,8 +289,9 @@ def set_new_user(self, section: str, new_user: str) -> None:
                 'USER:.*', content_to_mod).group(0)[1:], "USER: " + new_user)
             content_to_mod = content_to_mod.replace(re.search(
                 'USER_TO:.*', content_to_mod).group(0)[1:], "USER_TO: " + old_user)
-        open(self._platforms_parser_file, 'w').write(content)
-        open(self._platforms_parser_file, 'a').write(content_to_mod)
+        with open(self._platforms_parser_file, 'w') as file:
+            file.write(content)
+            file.write(content_to_mod)
```
kinow, line 295: 💥 good catch!

autosubmit/config/configcommon.py
```diff
@@ -471,9 +473,8 @@ def is_section_in_any_wrapper(self, section: str) -> bool:
         :return: True if the section is found in at least one wrapper, False otherwise.
         """
         for wrapper_data in self.experiment_data.get("WRAPPERS", {}).values():
-            if isinstance(wrapper_data, dict):
-                if section in wrapper_data.get("JOBS_IN_WRAPPER", []):
-                    return True
+            if isinstance(wrapper_data, dict) and section in wrapper_data.get("JOBS_IN_WRAPPER", []):
+                return True
         return False
```
kinow, line 479: If we are simplifying here, then maybe remove the if altogether?

```python
def is_section_in_any_wrapper(self, section: str) -> bool:
    """Return ``True`` if the section appears in any wrapper's ``JOBS_IN_WRAPPER``.

    :param section: The job section name to look up.
    :return: True if the section is found in at least one wrapper, False otherwise.
    """
    return any(
        isinstance(wrapper_data, dict)
        and section in wrapper_data.get("JOBS_IN_WRAPPER", [])
        for wrapper_data in self.experiment_data.get("WRAPPERS", {}).values()
    )
```

I think it does the same, but it's not important for this change, and not sure if the code has good tests for this refactoring.

autosubmit/config/configcommon.py
kinow: Ping here...

kinow (commented):
Quick review during meeting, @VindeeR .

This is not a placeholder, so it shouldn't crash! 😆 

Image

```bash
(venv) kinow@ranma:~/Development/python/workspace/autosubmit$ autosubmit create a00i && autosubmit inspect a00i
Autosubmit v4.2.0

Command traceability:
  Command: /home/kinow/Development/python/workspace/autosubmit/venv/bin/autosubmit create a00i
  Host: ranma
  User: kinow
  PID: 212805
  Python: 3.10.20
  Platform: Linux-7.0.0-31-generic-x86_64-with-glibc2.43
  Timezone: CEST (UTC+02:00)
  Working directory: /home/kinow/Development/python/workspace/autosubmit

  File "/home/kinow/Development/python/workspace/autosubmit/venv/bin/autosubmit", line 6, in <module>
    sys.exit(main())
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/scripts/autosubmit.py", line 320, in main
    ret = _autosubmit(argv[1:])
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/scripts/autosubmit.py", line 304, in _autosubmit
    return execute_cmd(entry_point, *command_args)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/scripts/_entry_points.py", line 107, in execute_cmd
    return entry_point.load()(*args)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/scripts/_cli_function.py", line 147, in _cli_function
    initialise_command(subcommand, opts)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/scripts/_initialise.py", line 103, in initialise_command
    as_conf.reload(force_load=True)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/config/configcommon.py", line 1804, in reload
    self.deep_add_missing_starter_conf(self.experiment_data, starter_conf)
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/config/configcommon.py", line 2018, in deep_add_missing_starter_conf
    experiment_data[key] = self.deep_add_missing_starter_conf(experiment_data[key], starter_conf[key])
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/config/configcommon.py", line 2018, in deep_add_missing_starter_conf
    experiment_data[key] = self.deep_add_missing_starter_conf(experiment_data[key], starter_conf[key])
  File "/home/kinow/Development/python/workspace/autosubmit/autosubmit/config/configcommon.py", line 2029, in deep_add_missing_starter_conf
    raise AutosubmitCritical(f"Dynamic variables {key} causing infinite recursion during evaluation of the element {key_experiment_data}")
autosubmit.log.log.AutosubmitCritical: Dynamic variables ANOTHER_QUEUE causing infinite recursion during evaluation of the element %CURRENT_ANOTHER_QUEUE
Error code: 7000
```

```yaml
PLATFORMS:
  MARENOSTRUM5:
    TYPE: slurm
    HOST: glogin1.bsc.es, glogin2.bsc.es
    PROJECT: bsc32
    USER: bruno
    SPECIAL_QUEUE: ultra
    ANOTHER_QUEUE: "%CURRENT_ANOTHER_QUEUE"
    QUEUE: gp_debug
    SCRATCH_DIR: /gpfs/scratch
    ADD_PROJECT_TO_HOST: false
    MAX_WALLCLOCK: 48:00
    TEMP_DIR: ''
```

VindeeR:
> ```yaml
> PLATFORMS:
>   MARENOSTRUM5:
>     TYPE: slurm
>     HOST: glogin1.bsc.es, glogin2.bsc.es
>     PROJECT: bsc32
>     USER: bruno
>     SPECIAL_QUEUE: ultra
>     ANOTHER_QUEUE: "%CURRENT_ANOTHER_QUEUE"
>     QUEUE: gp_debug
>     SCRATCH_DIR: /gpfs/scratch
>     ADD_PROJECT_TO_HOST: false
>     MAX_WALLCLOCK: 48:00
>     TEMP_DIR: ''
> ```

Based on the description of the [PR](https://github.com/BSC-ES/autosubmit/issues/2602), I believe this behavior is correct

autosubmit/config/configcommon.py
```diff
@@ -1841,8 +1829,6 @@ def reload(self, force_load=False, only_experiment_data=False):
 
     def set_default_parameters(self) -> None:
         """Sets the default parameters for the experiment."""
-        self.default_parameters: dict = {'d': '%d%', 'd_': '%d_%', 'Y': '%Y%', 'Y_': '%Y_%', 'M': '%M%', 'M_': '%M_%',
-                                         'm': '%m%', 'm_': '%m_%'}
```
kinow, line 1845: Note to self: test if Eric's variables can still be used when passing the date format pattern in configuration files.

PR #2998, opened 2026-05-07
2640 create autosubmit directories when calling autosubmit install not configure

### LINKED ISSUE #2640: Create Autosubmit directories when calling autosubmit install, not configure
### Problem

At the moment running autosubmit configure it modifies your autosubmitrc, and creates folders for you. But if you want to create folders using an existing autosubmitrc then it becomes a problem

### Proposed Solution

Move it to autosubmit install, as suggested by @dbeltrankyl today during the coding sprint.

rajeshsuraj:
closes #2640 

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -3607,15 +3607,6 @@ class Autosubmit:
                     parser.write(config_file)
                     Log.result(f"Configuration file written successfully: \n\t{rc_path}")
 
-                    dirs = [local_root_path, global_logs_path, structures_path, historicdb_path, historiclog_path]
-
-                    for dir in dirs:
-                        create_me = Path(dir)
-                        create_me.mkdir(parents=True, exist_ok=True)
-                        create_me.chmod(0o775)
-
-                    sep = '\n\t'
-                    Log.result(sep.join(['Directories configured successfully:'] + [str(d) for d in dirs]))
                 except (IOError) as e:
                     raise AutosubmitCritical(f"Can not write config file: {e.message}", 7012)
                 except (OSError) as e:
@@ -3629,6 +3620,21 @@ class Autosubmit:
     @staticmethod
     def install():
         """Creates a new database instance for autosubmit at the configured path."""
+        dirs = [
+            BasicConfig.LOCAL_ROOT_DIR,
+            BasicConfig.GLOBAL_LOG_DIR,
+            BasicConfig.STRUCTURES_DIR,
+            BasicConfig.JOBDATA_DIR,
+            BasicConfig.HISTORICAL_LOG_DIR,
+        ]
+        for dir in dirs:
+            create_me = Path(dir)
+            create_me.mkdir(parents=True, exist_ok=True)
+            create_me.chmod(0o775)
+
+        sep = '\n\t'
+        Log.result(sep.join(['Directories configured successfully:'] + [str(d) for d in dirs]))
+        
         if BasicConfig.DATABASE_BACKEND == 'sqlite':
             if not os.path.exists(BasicConfig.DB_PATH):
                 Log.info("Creating autosubmit database...")
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -3634,6 +3634,7 @@ class Autosubmit:
             BasicConfig.STRUCTURES_DIR,
             BasicConfig.JOBDATA_DIR,
             BasicConfig.HISTORICAL_LOG_DIR,
+            BasicConfig.DB_DIR,
         ]
         for dir in dirs:
             create_me = Path(dir)
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -3607,6 +3607,7 @@ class Autosubmit:
 
                     paths_info = [
                         f"Local root path: {local_root_path}",
+                        f"Database path: {database_path}",
                         f"Global logs path: {global_logs_path}",
                         f"Structures path: {structures_path}",
                         f"Historic DB path: {historicdb_path}",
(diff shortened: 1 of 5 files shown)


rajeshsuraj:
Also for the issue , i am updating the changelog but within the documentation where should i include this change ?

LuiggiTenorioK:
I think you have to review the "Installation" page in the ReadTheDocs site: https://autosubmit.readthedocs.io/en/master/installation/index.html

Then, modify it (`docs/source/installation/index.rst`), so it is clear that the process should be `autosubmit configure` -> `autosubmit install`, and in the install is where the directories are created. Also, it seems important to mention that from version `<=4.1.16` this was different.

rajeshsuraj:
Okay, will review and make changes under installation within the docs.

autosubmit/autosubmit.py
```diff
@@ -3629,6 +3620,21 @@ def configure(
...
+            BasicConfig.JOBDATA_DIR,
+            BasicConfig.HISTORICAL_LOG_DIR,
+        ]
+        for dir in dirs:
+            create_me = Path(dir)
+            create_me.mkdir(parents=True, exist_ok=True)
+            create_me.chmod(0o775)
+
+        sep = '\n\t'
+        Log.result(sep.join(['Directories configured successfully:'] + [str(d) for d in dirs]))
```
LuiggiTenorioK: It will be better to log that the directories have been **created** and configured successfully.

autosubmit/autosubmit.py
```diff
@@ -3607,15 +3607,6 @@ def configure(
...
 
-                    dirs = [local_root_path, global_logs_path, structures_path, historicdb_path, historiclog_path]
-
-                    for dir in dirs:
-                        create_me = Path(dir)
-                        create_me.mkdir(parents=True, exist_ok=True)
-                        create_me.chmod(0o775)
-
-                    sep = '\n\t'
-                    Log.result(sep.join(['Directories configured successfully:'] + [str(d) for d in dirs]))
```
LuiggiTenorioK, line 3618: Here the logging should remain, but with a different message. Something like:

```
Directories added to the configuration file:
- Local root path: /home/blablabla
...
...
- Historical logs path: /blablabla
```

autosubmit/autosubmit.py
LuiggiTenorioK: Here is a directory creation that has not been moved

LuiggiTenorioK: Here is a directory creation that has not been moved

docs/source/installation/index.rst
```diff
@@ -194,15 +194,21 @@ Manually generate an ``autosubmitrc`` file in one of these locations, which is t
 Quick Installation - Non-shared database (user level)
 ------------------------------------------------------
 
-After the package installation, you have to configure at least the database and path for Autosubmit.
```
LuiggiTenorioK, line 197: The `you have to configure at least the database and path for Autosubmit.` should remain somewhere else (probably after the code block), because you have always to give the root path, database path, and database filename in `autosubmit configure`

test/unit/test_configure.py
```diff
@@ -145,3 +145,33 @@ def test_configure_advanced(mocker, tmp_path, suffix: str, autosubmit) -> None:
...
+        platforms_conf_path=None,
+        jobs_conf_path=None,
+        smtp_hostname=None,
+        mail_from=None,
+        machine=False,
+        local=False,
+    )
+
+    assert (tmp_path / ".autosubmitrc").exists()
+    assert not (lr_path / "logs").exists()
```
LuiggiTenorioK: From what was stated in another comment of this review.

```suggestion
    assert not lr_path.exists()
    assert not (lr_path / "logs").exists()
```

LuiggiTenorioK (changes requested):
The PR overall seems good @rajeshsuraj. It has good tests and documentation.

However, there are some extra changes to make to keep both commands fully consistent.

autosubmit/autosubmit.py
```diff
@@ -3629,6 +3628,21 @@ def configure(
     @staticmethod
     def install():
         """Creates a new database instance for autosubmit at the configured path."""
+        dirs = [
+            BasicConfig.LOCAL_ROOT_DIR,
+            BasicConfig.GLOBAL_LOG_DIR,
+            BasicConfig.STRUCTURES_DIR,
+            BasicConfig.JOBDATA_DIR,
+            BasicConfig.HISTORICAL_LOG_DIR,
+        ]
```
LuiggiTenorioK, line 3639: In the last commit, the database path dir creation was removed. However, it is not created in the new install.

So, when you do, for example, `autosubmit configure -lr /home/ltenorio/as_test_2640 -db /home/ltenorio/as_test_2640_db 
-dbf autosubmit.db` where the local root path is different from the db path, the `autosubmit install` will fail because the db path has not been created.

LuiggiTenorioK (changes requested):

autosubmit/autosubmit.py
rajeshsuraj: yeah my bad , i have added the db creation in install now.

autosubmit/autosubmit.py
```diff
@@ -3607,15 +3605,16 @@ def configure(
                     parser.write(config_file)
                     Log.result(f"Configuration file written successfully: \n\t{rc_path}")
 
-                    dirs = [local_root_path, global_logs_path, structures_path, historicdb_path, historiclog_path]
+                    paths_info = [
+                        f"Local root path: {local_root_path}",
+                        f"Global logs path: {global_logs_path}",
+                        f"Structures path: {structures_path}",
+                        f"Historic DB path: {historicdb_path}",
+                        f"Historic logs path: {historiclog_path}",
```
LuiggiTenorioK, line 3614: It will be good to also log that the database path has been added to the configuration.

LuiggiTenorioK (approved):
Good! Now, everything seems in place and as intended.

Just one small comment left, but after that, it can be merged without any problem.

Issue #1117, opened 2023-08-19
[enhancement] Add type hints


kinow:
This is related to #876, but it should help us to simplify and improve our code regardless of whether we use `mypyc` or not.

Someone tweeted this today, from "Types and Programming Languages" (Pierce, 2002):

![image](https://earth.bsc.es/gitlab/es/autosubmit/uploads/69ef09139829f09b17e859ee4f029259/image.png)

This pretty much summarizes my experience so far after seeing types being added to Python projects. (Also true for JavaScript projects that adopted types in JS or TS.)

The profiler code was annotated for types, and we can start adding more types little by little, as we work on different parts of the code.

VindeeR:
I just ran mypy and I got the following results

```
Found 631 errors in 95 files (checked 329 source files)

Mypy Type Check Coverage Summary
================================

Script: index

+--------------------------------------------------------------------------------+-------------------+-----------+
| Module                                                                         | Imprecision       | Lines     |
+--------------------------------------------------------------------------------+-------------------+-----------+
| autosubmit                                                                     |   0.00% imprecise |    21 LOC |
| autosubmit.autosubmit                                                          |  32.83% imprecise |  5884 LOC |
| autosubmit.config                                                              |   0.00% imprecise |    18 LOC |
| autosubmit.config.basicconfig                                                  |  49.00% imprecise |   251 LOC |
| autosubmit.config.configcommon                                                 |  42.09% imprecise |  2884 LOC |
| autosubmit.config.yamlparser                                                   |  20.59% imprecise |    34 LOC |
| autosubmit.database                                                            |   0.00% imprecise |    18 LOC |
| autosubmit.database.data                                                       |   0.00% imprecise |    18 LOC |
| autosubmit.database.db_common                                                  |  35.32% imprecise |   872 LOC |
| autosubmit.database.db_manager                                                 |  23.21% imprecise |   112 LOC |
| autosubmit.database.db_structure                                               |  19.63% imprecise |   107 LOC |
| autosubmit.database.session                                                    |   0.00% imprecise |    34 LOC |
| autosubmit.database.tables                                                     |   2.65% imprecise |   264 LOC |
| autosubmit.experiment                                                          |   0.00% imprecise |     0 LOC |
| autosubmit.experiment.detail_updater                                           |  20.30% imprecise |   330 LOC |
| autosubmit.experiment.experiment_common                                        |  19.63% imprecise |   428 LOC |
[...]

PR #2999, opened 2026-05-08
[Enhancement] - Fixing mypy

VindeeR:
Handling issues reported at the #1117

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/platforms/ecplatform.py
+++ b/autosubmit/platforms/ecplatform.py
@@ -690,8 +690,8 @@ class EcPlatform(ParamikoPlatform):
         :raises AutosubmitError: For transient errors.
         :raises AutosubmitCritical: For permanent errors.
         """
-        out = self._ssh_output or ""
-        err = self._ssh_output_err or ""
+        out = self._ssh_output
+        err = self._ssh_output_err
 
         # Fast-exit: stdout matches a known ecaccess success pattern.
         if any(pat.search(out) for pat in _EC_EXPECTED_OUTPUT):
--- a/autosubmit/platforms/paramiko_platform.py
+++ b/autosubmit/platforms/paramiko_platform.py
@@ -148,7 +148,7 @@ class ParamikoPlatform(Platform):
         self.job_status: Optional[dict[str, list]] = None
         self._ssh: Optional[paramiko.SSHClient] = None
         self._ssh_config = None
-        self._ssh_output = None
+        self._ssh_output = ""
         self._host_config: Optional[dict] = None
         self._host_config_id = None
         self.submit_cmd = ""
@@ -167,13 +167,13 @@ class ParamikoPlatform(Platform):
 
         self.remove_log_files_on_transfer = False
         if self.config:
-            platform_config: dict = self.config.get("PLATFORMS", {}).get(
+            platform_config = self.config.get("PLATFORMS", {}).get(
                 self.name.upper(), {}
             )
             self.remove_log_files_on_transfer = platform_config.get(
                 "REMOVE_LOG_FILES_ON_TRANSFER", False
             )
-        self._uses_local_api: bool = False
+        self._uses_local_api = False
         # Pre-submission snapshot used by get_submitted_jobs_by_name to exclude
         # stale processes from previous runs on process-based platforms.
         self._pre_submission_pids: dict[str, set[int]] = {}
@@ -201,7 +201,7 @@ class ParamikoPlatform(Platform):
         self.connected = False
         self._ssh = None
         self._ssh_config = None
-        self._ssh_output = None
+        self._ssh_output = ""
         self._host_config = None
         self._host_config_id = None
         self._ftpChannel = None
@@ -1508,8 +1508,6 @@ class ParamikoPlatform(Platform):
         :return: output from last command
         :rtype: str
         """
-        if self._ssh_output is None or not self._ssh_output:
-            self._ssh_output = ""
         return self._ssh_output
 
     def get_ssh_output_err(self):
--- a/autosubmit/platforms/platform.py
+++ b/autosubmit/platforms/platform.py
@@ -23,23 +23,25 @@ import traceback
 from contextlib import suppress
 from multiprocessing.queues import Queue
 from multiprocessing.synchronize import Event
+
 # noinspection PyProtectedMember
 from os import _exit  # type: ignore
 from pathlib import Path
-from typing import Any, Optional, Union, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Optional, Union
 
 import setproctitle
 
 from autosubmit.helpers.parameters import autosubmit_parameter
 from autosubmit.job.job_common import Status
-from autosubmit.log.log import Log
+from autosubmit.log.log import AutosubmitCritical, Log
 
 if TYPE_CHECKING:
+    from multiprocessing.process import BaseProcess
+
     from autosubmit.config.configcommon import AutosubmitConfig
-    from autosubmit.job.job_packages import JobPackageBase
     from autosubmit.job.job import Job
     from autosubmit.job.job_list import JobList
-    from multiprocessing.process import BaseProcess
+    from autosubmit.job.job_packages import JobPackageBase
 
 
 def _init_logs_log_process(as_conf: 'AutosubmitConfig', platform_name: str) -> None:
@@ -98,7 +100,7 @@ class CopyQueue(Queue):
     A queue that copies the object gathered.
     """
 
-    def __init__(self, maxsize: int = -1, block: bool = True, timeout: float = None, ctx: Any = None) -> None:
+    def __init__(self, maxsize: int = -1, block: bool = True, timeout: Optional[float] = None, ctx: Any = None) -> None:
         """Initializes the Queue.
 
         :param maxsize: Maximum size of the queue. Defaults to -1 (infinite size).
@@ -166,8 +168,8 @@ class Platform:
         self._partition = None
         self.ec_queue = "hpc"
         self.processors_per_node = None
-        self.scratch_free_space = None
-        self.custom_directives = None
+        self.scratch_free_space: Optional[str] = None
+        self.custom_directives: Optional[dict] = None
         self._host = ''
         self._user = ''
         self._project = ''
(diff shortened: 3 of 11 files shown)


autosubmit/platforms/paramiko_platform.py
```diff
@@ -141,13 +141,13 @@ def __init__(self, expid: str, name: str, config: dict, auth_password: Optional[
         """
         Platform.__init__(self, expid, name, config, auth_password=auth_password)
         self._proxy: Optional[ProxyCommand] = None
-        self._ssh_output_err = ""
+        self._ssh_output_err: str = ""
```
kinow: Do we **have to** type this? Typing things that are already explicit sounds a bit tedious I think? Like `a = 42`. Even in Java we don't type the type anymore, and just `val a = 42`.

VindeeR: I added it but later on I solved the issue, so we can remove it, but I didn't know it wasn't needed.

I will adjust the future fixes to only add the type when necessary thanks :raised_hands:

kinow: :pray: thank you!!!

VindeeR:
I just ran the same mypy command on master and got:
`Found 315 errors in 12 files (checked 12 source files)`

This branch is fixing 97 mypy related items, I believe it makes more sense to merge this branch as is, and then move to a different set of files in other to keep it fairly small changes.

wdyt? @kinow

report0800/index.txt
kinow, line 1: Do we need to add this file???

test/unit/platforms/test_ecplatform.py
```diff
@@ -241,7 +241,6 @@ def _send_command(command: str, **_) -> bool:
 @pytest.mark.parametrize("ssh_output,ssh_output_err", [
     # Empty or whitespace output – no error possible.
     ("", ""),
-    (None, None),
```
kinow, line 285: Was it needed to remove this test case? I think we can assume mypy is checking for that, but it should be OK to be extra careful having tests for some features, no?

autosubmit/platforms/ecplatform.py
```diff
@@ -458,8 +458,8 @@ def _check_for_unrecoverable_errors(self) -> None:
         :raises AutosubmitError: For transient errors.
         :raises AutosubmitCritical: For permanent errors.
         """
-        out = self._ssh_output or ""
-        err = self._ssh_output_err or ""
+        out: str = self._ssh_output
+        err: str = self._ssh_output_err
```
kinow: The ssh_output & err attributes of the ParamikoPlatform are already strings. In theory we don't need the types here?

```
$ mypy autosubmit/platforms/ecplatform.py --ignore-missing-imports --follow-imports=skip
autosubmit/platforms/ecplatform.py:107: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
autosubmit/platforms/ecplatform.py:208: error: Argument 1 to "decode" of "bytes" has incompatible type "Optional[str]"; expected "str"  [arg-type]
autosubmit/platforms/ecplatform.py:235: error: Argument 1 to "decode" of "bytes" has incompatible type "Optional[str]"; expected "str"  [arg-type]
Found 2 errors in 1 file (checked 1 source file)
```

I get the same with and without this change. Since the parent type is a string and not an optional string, we can keep the removal of `or ""`, I think.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -147,7 +147,7 @@ def __init__(self, expid: str, name: str, config: dict, auth_password: Optional[
         self.job_status: Optional[dict[str, list]] = None
         self._ssh: Optional[paramiko.SSHClient] = None
         self._ssh_config = None
-        self._ssh_output = None
+        self._ssh_output = ""
         self._host_config: Optional[dict] = None
```
kinow, line 152: Here we define that the ssh output will always be a string. That means we have to review everywhere where it was still treated as optional. e.g., `get_ssh_output` in this same class.

Image

We only need to check for that before write/set the value.

autosubmit/platforms/platform.py
```diff
@@ -154,7 +154,7 @@ def __init__(self, expid: str, name: str, config: dict, auth_password: Optional[
         self.connected = False
         self.expid: str = expid
         self._name: str = name
-        self.config = config
+        self.config: dict = config
```
kinow: `config` is a dict. I guess mypy is able to infer the type. Hovering the variable on IDE's should display the correct type, and tools that produce documentation should also already have the type available. I think this dict here is redundant and may not be needed?

autosubmit/platforms/platform.py
```diff
@@ -232,7 +234,7 @@ def __init__(self, expid: str, name: str, config: dict, auth_password: Optional[
             self.compression_level = platform_config.get("COMPRESSION_LEVEL", 9)
 
         self.log_queue_size = log_queue_size
-        self.remote_log_dir = None
+        self.remote_log_dir: str = ""
```
kinow, line 239: This is already implicit, I think we don't need : str = ""

autosubmit/platforms/platform.py
```diff
@@ -749,7 +752,7 @@ def get_files_path(self) -> str:
         :rtype: str
         """
         if self.type == "local":
-            path = Path(self.root_dir) / self.config.get("LOCAL_TMP_DIR") / f'LOG_{self.expid}'
+            path = Path(self.root_dir).joinpath(self.config.get("LOCAL_TMP_DIR", "") + f'/LOG_{self.expid}')
```
kinow: Probably easier to use only the constructor instead of constructor + function calls. Something like Path(self.root_dir, self.config....).

autosubmit/platforms/platform.py
```diff
@@ -994,6 +997,9 @@ def wait_mandatory_time(self, sleep_time: int = 60) -> bool:
         for _ in range(0, sleep_time, 5):
             time.sleep(5)
             try:
+                if self.work_event is None or self.recovery_queue is None or self.cleanup_event is  None:
+                    raise AutosubmitCritical("An issue occurred at the Mandatory Wait time in which one or more"
+                                             "of the values are  is None")
```
kinow: I haven't reviewed if this check is needed or not. I only read the text, but it's quite odd, with mixed upper/lower case, and the end has an are/is None with extra blank spaces.

autosubmit/platforms/platform.py
```diff
@@ -1026,6 +1033,8 @@ def recover_job_log(self, identifier: str, jobs_pending_to_process: set[Any],
         :param as_conf: The Autosubmit configuration object containing experiment data.
         :return: Updated set of jobs pending to process.
         """
+        if self.recovery_queue is None:
+            raise AutosubmitCritical("An issue occurred at the Recovery queue is None")
```
kinow: I think we need to be more careful with error messages. I don't think we need to write Recovery instead of recovery here. At least, I think we normally don't always write it this way (if we choose to use it here, we must make sure we are consistent). And instead of None, it might be better to indicate to users "something is missing". Maybe that will help those not familiar with Python (and we always end up searching the code for that text in the end 😅 so for AS devs should be OK).

autosubmit/autosubmit.py
```diff
@@ -3234,7 +3234,7 @@ def check(experiment_id: str) -> bool:
             as_conf.check_conf_files(False)
             as_conf.get_project_type()
             submitter = ParamikoSubmitter(as_conf=as_conf)
-            if not submitter.platforms:
+            if len(submitter.platforms) == 0:
```
kinow: Is this needed? The `len` is semantically right, I think. But in Python it's common to just check `if not submitter.platforms`. And the `if a`, `if not a`, is normally a bit faster too -- https://pytutorial.com/check-if-python-dictionary-is-empty/

report0800/index.txt
VindeeR: oops

test/unit/platforms/test_ecplatform.py
VindeeR: I removed since this test will be unrealistic to happen, but I can reintroduce the test in a different test, since it needs a new error handling to be done

autosubmit/platforms/platform.py
VindeeR: if I remove this one it raises to 258 errors :grimacing: from the 222 we currently have

autosubmit/platforms/platform.py
VindeeR: If there's no `None` treatment in here the mypy complain, but yeah the message is poorly worded

autosubmit/job/job_list.py
```diff
@@ -3337,9 +3337,9 @@ def retrieve_times(status_code, name, tmp_path, make_exception=False, job_times=
         energy = 0
         seconds_queued = 0
         seconds_running = 0
-        submit_time = datetime.timedelta()
-        start_time = datetime.timedelta()
-        finish_time = datetime.timedelta()
+        submit_time: datetime.timedelta = datetime.timedelta()
+        start_time: datetime.timedelta = datetime.timedelta()
+        finish_time: datetime.timedelta = datetime.timedelta()
```
kinow: I think we don't need to declare types here as they are already implied from the assignment.

kinow (commented):
@VindeeR I saw one place where I thought the types wouldn't be need. But I think the rest looks good. Do you think we can maybe merge it without closing the mypy issue? As this already improves the platform code, maybe we could do a final review and merge it this week?

VindeeR:
> @VindeeR I saw one place where I thought the types wouldn't be need. But I think the rest looks good. Do you think we can maybe merge it without closing the mypy issue? As this already improves the platform code, maybe we could do a final review and merge it this week?

I believe it won't be closing but I'll keep an eye, I'll add your suggestion and merge this part of the mypy issue

test/unit/platforms/test_ecplatform.py
kinow: > will be unrealistic to happen

Because there's already code that prevents that from happening, or because it's100% certain that can never ever happen in our code (even if a bug somewhere adds a `None`?)?

PR #3005, opened 2026-05-12
removed redundancy from autosubmit report flag all

### LINKED ISSUE #1043: Redundancy in the autosubmit report output
Hello @dbeltrankyl and @kinow,

### Autosubmit Version

`master` 9592607b231bb0e6e5a0f68ebf97d343dbc6f041

### Expid affected(If applicable)

NA

### Which task has issues? Where is the log(If applicable)

NA

### Summary

Running `autosubmit report -all ...`, the output YAML has some redundant information. Each job appears to have information about the other jobs that are already in this report.

image

image

If you look at the screenshot above, `SYNC_TO_REMOTE` has a `JOBS.SYNC_TO_REMOTE`, and also to other jobs.

### Steps to reproduce

`autosubmit report -all $expid`

Redundant information in the report (for a 7-8 tasks workflow, I am getting a report with 1205 lines. Out of these lines, 746 have the token "JOBS". Fixing this issue might greatly simplify the report output.

```bash
$ cat a008_parameter_list_20230517-151204.txt  | grep JOBS | wc -l
746

```

No redundant information.

NA

### Any other relevant information(if applicable)

()

rajeshsuraj:
Closes #1043 

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/test/integration/commands/test_run_command.py
+++ b/test/integration/commands/test_run_command.py
@@ -343,3 +343,61 @@ def test_run_command_logs_warning(
 
     with pytest.warns(UserWarning, match=warning_log_message):
         exp.autosubmit.run_command(args=args)
+
+
+@pytest.mark.parametrize(
+    'command,expected',
+    [(
+        ['autosubmit', 'report', '{expid}', '-all'],
+        {
+            'required_top_level': ('HPCARCH',),
+            'required_section_keys': ('FILE', 'RUNNING'),
+        },
+    )],
+    ids=['report -all'],
+)
+def test_run_report_command(
+        command: list[str],
+        expected: dict,
+        autosubmit_exp: 'AutosubmitExperimentFixture',
+        mocker: 'MockerFixture',
+        get_next_expid: Callable[[], str]):
+    """Validate `autosubmit report -all` output (issue #1043).
+
+    Checks that the parameter list contains the expected content and is free
+    of the two redundancy patterns the issue describes: global keys duplicated
+    under every JOBS.<section>.* prefix, and sibling jobs re-nested under each
+    section (JOBS.<X>.JOBS.<Y>.*).
+
+    """
+    exp, args, command = set_up_test(get_next_expid(), command, autosubmit_exp, mocker)
+    assert exp.autosubmit.run_command(args=args)
+
+    report = next(Path(exp.tmp_dir).glob(f"{exp.expid}_parameter_list_*.txt"))
+    keys = {line.split('=', 1)[0] for line in report.read_text().splitlines() if '=' in line}
+
+    for k in expected['required_top_level']:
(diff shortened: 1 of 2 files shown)


VindeeR:
I notice your jobs were running for over 10 minutes, even the unit that's, I'm rerunning the tests just to see if this is some sort of intermittent issue or not.

All the tests should get the green ticks always, so if something was broken we would need to investigate it.

VindeeR:
It would be good to always add test to the changes you do in the code, in the case of this change I would recommend adding a test to the
`test/integration/commands/test_run_command.py`

Here is a snippet to help you getting started on creating the test, don't hesitate in asking me if anything feels confusing, and I did this code in here on github, so there might have some problems on it.

```
@pytest.mark.parametrize(
    'command,output',
    [(
        ['autosubmit', 'report', '-all', '{expid}'],
        {
        #dict representing the expected output
        }
    )],
    ids=[
        'report'
    ]
)
def test_run_report_command(
        command: list[str],
        output: dict
        autosubmit_exp: 'AutosubmitExperimentFixture',
        mocker: 'MockerFixture',
        get_next_expid: Callable[[], str]):
    """This test will be use to validate if the report being generated have the expected output
    """
    exp, args, command = set_up_test(get_next_expid(), command, autosubmit_exp, mocker)
    assert exp.autosubmit.run_command(args=args) # this will fail in case the run raises an error
     
    yaml = YAML(typ='rt')
    for report_file in Path(exp.tmp_dir).iterdir(): #This is incomplete, here you need to find where the report is being stored
        with open(report_file, 'r+') as file:
            yaml_data = yaml.load(file)
            
            ##Validate report with the output created above
```

test/integration/commands/test_run_command.py
```diff
@@ -343,3 +343,61 @@ def test_run_command_logs_warning(
...
+    top_level_namespaces = {
+        k.split('.', 1)[0] for k in keys
+        if '.' in k and not k.startswith('JOBS.')
+    }
+    for section in sections:
+        for ns in top_level_namespaces:
+            duplicated = [k for k in keys if k.startswith(f'JOBS.{section}.{ns}.')]
+            assert not duplicated, \
+                f"Global namespace {ns!r} duplicated under JOBS.{section}.* " \
+                f"(issue #1043): {duplicated[:3]}"
```
VindeeR: Don't forget to always leave a blank line `EoF`

VindeeR (changes requested):
I did some manual investigation, and I believe everything is fine, it looks like all parameters were exported into the report correctly.

Thanks for the bug QoL improvement 🙌 

There just a small fix I requested, but after that PR can be merged.

VindeeR (approved):
Thanks, you can merge it now 🙌

PR #3012, opened 2026-05-14
optimization of autosubmit report

### LINKED ISSUE #3008: Improve autosubmit report templating and add docs for report and describe
### Problem

The template-rendering block in autosubmit report -t has rough edges (regex quirks, hardcoded .txt output, no feedback on unmatched placeholders), and neither report nor describe has user-facing documentation.

### Proposed Solution

- Clean up the template-rendering block in report so substitution behaves predictably and surfaces unmatched tokens.
- Add a starter template users can copy and adapt.
- Document autosubmit report (both -all and -t) and autosubmit describe.

rajeshsuraj:
Closes #3008

**Check List**

- [x] Documentation updated.


docs/source/userguide/monitor_and_check/index.rst
```diff
@@ -565,72 +565,232 @@ The location where user can put this stats is in the file:
...
 
-    autosubmit report EXPID
+    autosubmit report EXPID -all
 
 Or combined as follows:
+
 ::
 
-    autosubmit report EXPID -t "absolute_file_path"
+    autosubmit report EXPID -all -t "absolute_file_path"
```
VindeeR: This is not a working command

It should be either `-t` or `-all`, but not both
```suggestion
    autosubmit report <EXPID> -t "absolute_file_path"
```

VindeeR (changes requested):
The changes looks good thanks for improving the quality of the documentation 🙌

I have just one comment, and I would like you to add a test for this too.
It should be fairly similar to the previous one you created for the other branch

I ended up having a doubt as I was going through the commit but I'm not exactly sure how the scenarios below would or should be treated

`%%EXPERIMENT.CHUNKSIZEUNIT%%`
`%^EXPERIMENT.CHUNKSIZE%`
`% EXPERIMENT.CHUNKSIZE %`

docs/source/userguide/monitor_and_check/index.rst
rajeshsuraj: I tested it out and it works because either of the conditionals for those flags are on same level . If thats not intended i will remove it.

VindeeR: correct me if I'm wrong but, It would be misleading to say it works, because the `template` would superpose the `-all` and just print the template and all the variables in all would never be added.

rajeshsuraj:
> The changes looks good thanks for improving the quality of the documentation 🙌
> 
> I have just one comment, and I would like you to add a test for this too. It should be fairly similar to the previous one you created for the other branch
> 
> I ended up having a doubt as I was going through the commit but I'm not exactly sure how the scenarios below would or should be treated
> 
> `%%EXPERIMENT.CHUNKSIZEUNIT%%` `%^EXPERIMENT.CHUNKSIZE%` `% EXPERIMENT.CHUNKSIZE %`

I just tested the edge cases and these are the results.

- %%EXPERIMENT.CHUNKSIZEUNIT%% → renders as the literal text %EXPERIMENT.CHUNKSIZEUNIT%. The double %% is the escape for "leave this alone."
- %^EXPERIMENT.CHUNKSIZE% → renders as -. The ^ is treated as part of the key name, that key doesn't exist, so it falls through to the unknown-key fallback (same as a typo).
- % EXPERIMENT.CHUNKSIZE % → renders unchanged. The spaces around the key mean it isn't recognised as a placeholder at all, so it's just left as-is in the output.

VindeeR:
> > The changes looks good thanks for improving the quality of the documentation 🙌
> > I have just one comment, and I would like you to add a test for this too. It should be fairly similar to the previous one you created for the other branch
> > I ended up having a doubt as I was going through the commit but I'm not exactly sure how the scenarios below would or should be treated
> > `%%EXPERIMENT.CHUNKSIZEUNIT%%` `%^EXPERIMENT.CHUNKSIZE%` `% EXPERIMENT.CHUNKSIZE %`
> 
> I just tested the edge cases and these are the results.
> 
>     * %%EXPERIMENT.CHUNKSIZEUNIT%% → renders as the literal text %EXPERIMENT.CHUNKSIZEUNIT%. The double %% is the escape for "leave this alone."
> 
>     * %^EXPERIMENT.CHUNKSIZE% → renders as -. The ^ is treated as part of the key name, that key doesn't exist, so it falls through to the unknown-key fallback (same as a typo).
> 
>     * % EXPERIMENT.CHUNKSIZE % → renders unchanged. The spaces around the key mean it isn't recognised as a placeholder at all, so it's just left as-is in the output.

Perfect I believe it to be the expected behavior.
The one I was worried the most would be the one with `^` since we have the special dynamic variables

docs/source/userguide/monitor_and_check/index.rst
rajeshsuraj: I think you are right it would be misleading cause definitely the user think using all plus template gives everything or something like that. Will remove that to avoid confusion.

docs/source/userguide/monitor_and_check/index.rst
```diff
@@ -565,72 +565,226 @@ The location where user can put this stats is in the file:
 .. _report:
 
 How to extract information about the experiment parameters
-----------------------------------------------------------
+------------------------------------------------------------
 
-This procedure allows you to extract the experiment variables that you want.
+The ``autosubmit report`` command extracts the parameters and resolved values
+of an experiment. It has two modes, and they can be used together:
```
VindeeR:
```suggestion
The ``autosubmit report`` command extracts the parameters and resolved values
of an experiment. It has two modes, each generating their own file, and they can be used together:
```

VindeeR (approved):
One small change, that I believe clarifies a bit the explanation. let me know what you think.

Thanks for the great improvement in the documentation and the upgrade on the code. 🙌

PR #3016, opened 2026-05-15
Fix intermittent failures with mail integration tests

kinow:
Fixing an intermittent failure. The tests ran in parallel, using a module-scoped Docker container. Each test tried to delete the messages from the inbox before running, and every test used a temporary directory with the expid a000.

I used the `autosubmit_exp` everywhere, and made the fixtures function-scoped. That may launch more containers, but the mail containers are small/fast, so that should be OK. A compromise so we don't have intermittent failures.

**Check List**

- [ ] Documentation updated.


kinow:
Cherry-picked commit passed on my other PR. The last job has been running for 10 minutes.... so I may have to kick it a few times...

VindeeR (approved):
The changes looks great, thanks for fixing the test and the fixture.

I'll rerun the test just for peace of mind, but reading the code the use of a specific expid should be fixed now.

Thanks for the hard work 🙌

kinow:
Thanks @VindeeR ! Good idea to rerun the tests. One integration test failed due to timeout. I will address that in another PR. Merging...

PR #3018, opened 2026-05-18
Enhancement: Detect inefficiencies in SIM jobs with CPMIPS thresholds - Part 5 (Final)

paparici:
Hi AS team!

This is the fith part of a large [PR](https://github.com/MissterP/autosubmit/pull/1) with the objective to implement the detection of inefficiencies once the SIM job is completed (or section indicated) using CPMIPS and thresholds. If a threshold is violated, it will send a notification to the user indicated in the config. 

Last PR: #2986

This fith and final part consists in adding the integration with the AS main loop and the manager for handling the computation.

**Check List**

- [X] Changelog entry included in `CHANGELOG.md` if this is a change that can affect users.
- [x] Documentation updated.


paparici:
Hi! Do you have any recommendation on where to add documentation about this feature?

kinow:
> Hi! Do you have any recommendation on where to add documentation about this feature?

I'm not following the development around CPMIPS, notifications/thresholds/etc., but that seems important enough to maybe have its own section. For now it could be a new page, I think? Later we can move it around, and Sphinx can handle redirects (there's an extension... reredirects, I think?).

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 - Improving the handling of leakages that were happening when dealing with queues and processes #2900
 - verbosity adjusted #2976
 - Implement filters for `recovery` command #2923
+- Add detection of inneficiencies for key CPMIP metrics.
```
VindeeR: Please add the number of the issue related to this development

autosubmit/notifications/cpmip_notifier.py
```diff
@@ -0,0 +1,152 @@
...
+        if start_time <= 0 or end_time <= 0:
+            return None
+
+        data = CPMIPMetricsData(
+            start_time=start_time,
+            end_time=end_time,
+            run_years=_simulated_years(job, as_conf),
+            ncpus=_processing_elements(job),
+        )
+        return CPMIPEvaluation(data=data, thresholds=dict(thresholds))
```
VindeeR: Isn't `thresholds` already a dict? Why would you need the conversion here?

autosubmit/notifications/cpmip_notifier.py
```diff
@@ -0,0 +1,152 @@
...
+        Notifier.notify_cpmip_threshold_violations(
+            MailNotifier(BasicConfig),
+            expid,
+            job.name,
+            violations,
+            as_conf.experiment_data["MAIL"]["TO"],
+        )
+
+
+def _processing_elements(job) -> Optional[int]:
```
VindeeR: this name looks to be mismatching the behavior of the function maybe something like `_estimating_cpus`, `_calculate_number_cpus`, or `_get_used_cpus`

test/integration/test_cpmip_metrics.py
```diff
@@ -0,0 +1,217 @@
...
+    runtime_hours = (end_time - start_time) / 3600.0
+    sypd = SIM_SIMULATED_YEARS * 24.0 / runtime_hours
+    chsy = SIM_PROCESSORS * runtime_hours / SIM_SIMULATED_YEARS
+    core_hours = SIM_PROCESSORS * runtime_hours
+    return sypd, chsy, core_hours
+
+
+def _get_messages(api_base: str) -> list[dict[str, Any]]:
+    response = requests.get(f"{api_base}/api/v2/messages")
+    return response.json()["items"]
```
VindeeR: maybe the things related to the container can be imported directly from the `utils`

test/integration/test_cpmip_metrics.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+@pytest.fixture
+def configured_mail(monkeypatch, fake_smtp_server) -> tuple[int, str]:
+    """Point ``BasicConfig`` at the fake SMTP server and reset its inbox."""
+    smtp_port, api_base = fake_smtp_server
+    monkeypatch.setattr(BasicConfig, "MAIL_FROM", "notifier@localhost")
+    monkeypatch.setattr(BasicConfig, "SMTP_SERVER", f"127.0.0.1:{smtp_port}")
+    requests.delete(f"{api_base}/api/v1/messages")
+    return smtp_port, api_base
```
VindeeR: is it actually needed, wouldn't it work by just using the `fake_smtp_server` fixture?

[Fixture](https://github.com/BSC-ES/autosubmit/blob/197bd377b445a495cd87742ea6e43b19c8546567/test/integration/test_mail.py#L53-L61)

[Example of it being used](https://github.com/BSC-ES/autosubmit/blob/197bd377b445a495cd87742ea6e43b19c8546567/test/integration/test_mail.py#L100-L121)

test/integration/test_cpmip_metrics.py
```diff
@@ -0,0 +1,217 @@
...
+        "violation_emits_notification",
+        "no_violation",
+        "violation_but_notifications_off",
+        "no_violation_and_notifications_off",
+    ],
+)
+def test_cpmip_notification(autosubmit_exp, configured_mail, slurm_server,
+                             thresholds, notifications, expected_inbox_count):
+    """Cover every (violation, notifications-enabled) combination in one place."""
+    _, api_base = configured_mail
```
VindeeR: I believe using the fixture as I mentioned above should be enough here, no need to create a new fixture.

Let me know if I'm overlooking something.

VindeeR (changes requested):
I have some questions but everything looks really well done thanks for the tests too 🙌

CHANGELOG.md
paparici: There is no issue. Do you want to create one?

test/integration/test_cpmip_metrics.py
paparici: In this case, configured_mail is necessary because, sets the parameters of mail for the AS run. Inside the function the fixture is used.

CHANGELOG.md
VindeeR: hmmm... not sure why we didn't had one so far, my bad for overlooking it, it would be great to have one :pray:

paparici: I mean, I think there is the old one. Let me check

test/integration/test_cpmip_metrics.py
VindeeR: Can't it be set here inside the test, maybe even before getting the values form the fixture?

I'm wondering that because if I got it right you're just changing the `BasicConfig`, and in this way we could avoid code repetition.

I just found this other fixture too `create_mail_notifier`, that uses the same parameters you use, I believe that that it would work.

paparici: Trueee. However: 

- I created a function because, if in the future more tests are created, they can reuse that.
- I don't need the MailNotifier class that create_mail_notifier returns.

CHANGELOG.md
paparici: Well, didn't find one xD. I will create it once everything is finished.

test/integration/test_cpmip_metrics.py
VindeeR: Yeah, that sounds fair, so the last thing I want will be to keep all the fixture related to the mail in the same place for the sake of facilitating future development and avoiding the possibility of duplicating code.

I haven't tested it, but I believe if you do it like the example below, you can avoid `monkeypatch` too, which is good, the least we mock the better so we can replicate the usual behavior.

```
@pytest.fixture
def configured_mail( fake_smtp_server, autosubmit_experiment) -> tuple[int, str]:  # noqa: F811
    """Point ``BasicConfig`` at the fake SMTP server.

    Inbox is fresh for free since ``fake_smtp_server`` is function-scoped.
    """
    smtp_port, api_base = fake_smtp_server
    basic_config = autosubmit_experiment.as_conf.basic_config
    basic_config.MAIL_FROM = 'notifier@localhost'
    basic_config.SMTP_SERVER = f'127.0.0.1:{smtp_port}'
    return smtp_port, api_base
```

You might have to make a change on the initialization of the fixture let me know if you need help with that.

paparici: Agree to moving the fixture. However.

- I think the suggested change without mokeypatch could cause a leak of the state of the BasicConfig class. If a future test depends on the default values, depending on when the test is run, that changes of the values could affect in the future test.  Even if the current values are used across the tests. The mokeypatch provides the rollback for the next test.

VindeeR: Well in this case, we have the other fixture not setting the parameters, so it shouldn't be a problem, we would have both fixtures with different states and the developer can choose which best apply to his scenario.

paparici:
Added in documentation:

- How to configure performance notifications after section of how to activate e-mail.
- Performance metrics explanation

paparici:
Created issue: #3022

docs/source/userguide/configure/index.rst
```diff
@@ -247,6 +247,119 @@ Example:
                 - FAILED
                 - COMPLETED
 
+.. _cpmip-notifications-config:
```
VindeeR, line 250: What exactly is this directive? Do you need to have it set in here?

docs/source/userguide/configure/index.rst
```diff
@@ -247,6 +247,119 @@ Example:
...
+
+How to configure CPMIP threshold notifications
+----------------------------------------------
+
+Autosubmit can send an email when one or more
+:doc:`CPMIP performance metrics </userguide/performance/index>`
+fall outside a target range. The feature reuses the same ``MAIL.NOTIFICATIONS``
+and ``MAIL.TO`` settings as job-status notifications, so make sure those are
+enabled first (see the previous section).
+
```
VindeeR:
```suggestion
Autosubmit can send email alerts when one or more :doc:`CPMIP performance metrics </userguide/performance/index>` 
fall outside the configured target range. This feature uses the same ``MAIL.NOTIFICATIONS`` and ``MAIL.TO`` 
settings as job-status notifications, so make sure those are enabled first (see the previous section).

```

docs/source/userguide/configure/index.rst
```diff
@@ -247,6 +247,119 @@ Example:
...
+CPMIP metrics are designed for the **simulation job** (commonly named
+``SIM``), which runs per chunk and exposes the calendar metadata (``CHUNK``,
+``CHUNKSIZE``, ``CHUNKSIZEUNIT``) required to compute SYPD and CHSY. Declare
+the ``CPMIP_THRESHOLDS`` block under that job in ``jobs_<EXPID>.yml``. Each
+entry maps a metric name (``SYPD``, ``CHSY``, ``CORE_HOURS``) to its target
```
VindeeR:
```suggestion
entry maps a metric named (``SYPD``, ``CHSY``, ``CORE_HOURS``) to its target
```

docs/source/userguide/configure/index.rst
```diff
@@ -247,6 +247,119 @@ Example:
...
+                    COMPARISON: less_than    # CHSY must be <= THRESHOLD
+                    "%_ACCEPTED_ERROR": 5
+                CORE_HOURS:
+                    THRESHOLD: 1000
+                    COMPARISON: less_than    # CORE_HOURS must be <= THRESHOLD
+                    "%_ACCEPTED_ERROR": 5
+
+The keys are:
+
+* ``THRESHOLD``: numeric target the metric is compared against. Must be > 0.
```
VindeeR:
```suggestion
* ``THRESHOLD``: The numeric target value used for comparison against the metric. This value must be > 0.
```

docs/source/userguide/configure/index.rst
```diff
@@ -247,6 +247,119 @@ Example:
...
+* ``COMPARISON``: ``greater_than`` (metric must be ≥ threshold) or
+  ``less_than`` (metric must be ≤ threshold).
+* ``%_ACCEPTED_ERROR``: percentage slack applied to the threshold before
+  flagging a violation. For ``greater_than`` the effective bound is
+  ``THRESHOLD * (1 - error/100)``; for ``less_than``,
+  ``THRESHOLD * (1 + error/100)``. Use ``0`` for a strict check.
```
VindeeR:
```suggestion
* ``%_ACCEPTED_ERROR``: Percentage slack applied to the threshold before flagging a violation. 
For ``greater_than``, the effective lower bound is ``THRESHOLD * (1 - error / 100)``; 
for ``less_than``, the effective upper bound is ``THRESHOLD * (1 + error / 100)``. 
Set to 0 to enforce a strict threshold check.
```

docs/source/userguide/configure/index.rst
```diff
@@ -247,6 +247,119 @@ Example:
...
+
+When the job completes, Autosubmit recomputes the configured metrics, applies
+the bound, and — if any metric is outside it — sends a single
```
VindeeR:
```suggestion
the bound, and — if any of the metrics are outside of it — sends a single
```

docs/source/userguide/configure/index.rst
```diff
@@ -247,6 +247,119 @@ Example:
...
+threshold, effective bound, and observed value.
+
+Example notification
+~~~~~~~~~~~~~~~~~~~~
+
+With the configuration above, suppose the simulation completes with
+``SYPD = 3.9``, ``CHSY = 55000``, and ``CORE_HOURS = 1280``. All three fall
+outside their effective bounds (SYPD ``≥ 4.5`` after the 10% slack;
+CHSY ``≤ 52500`` after the 5% slack; CORE_HOURS ``≤ 1050`` after the 5%
+slack), so Autosubmit sends a single email looking like this:
```
VindeeR:
```suggestion
With the configuration above, suppose the simulation completes with ``SYPD = 3.9``, ``CHSY = 55000``, and ``CORE_HOURS = 1280``. 
All three values fall outside their effective bounds (SYPD ``≥ 4.5`` after the 10% slack;
CHSY ``≤ 52500`` after the 5% slack; CORE_HOURS ``≤ 1050`` 
after the 5% slack), so Autosubmit sends a single email notification similar to the following:
```

docs/source/userguide/configure/index.rst
```diff
@@ -247,6 +247,119 @@ Example:
...
+    Configured threshold: 5.0
+    Accepted error (%): 10.0
+    Effective bound: 4.5
+    Observed value: 3.9
+
+    INFO: This message was auto generated by Autosubmit,
+    remember that you can disable these messages on Autosubmit config file.
+
+Metrics that pass their thresholds are omitted from the body, and no email is
+sent at all when every configured metric is within its effective bound.
```
VindeeR:
```suggestion
Metrics that remain within their effective bounds are omitted from the email body.
If all configured metrics satisfy their thresholds, no notification email is sent.
```

docs/source/userguide/performance/index.rst
```diff
@@ -0,0 +1,70 @@
+###########
+Performance
+###########
+
+Autosubmit ships with built-in support for **CPMIP** (Coupled Model Intercomparison
+Project) performance metrics, a community-standard set of indicators originally
+proposed for climate models. The metrics give a quick, comparable read on how
+efficiently a simulation runs on a given platform, making them a useful first
+signal when looking for inefficiencies — undersized partitions, oversubscribed
+nodes, I/O bottlenecks, or regressions after a code change.
```
VindeeR:
```suggestion
Autosubmit includes built-in support for **CPMIP** (Coupled Model Intercomparison Project)
performance metrics, a community-standard set of indicators originally proposed for climate models. 
These metrics provide a quick and comparable view of how efficiently a simulation runs on a given platform, making them a useful first indicator 
when investigating for inefficiencies — undersized partitions, oversubscribed nodes, I/O bottlenecks, or regressions after a code change.
```

docs/source/userguide/performance/index.rst
```diff
@@ -0,0 +1,70 @@
...
+
+CPMIP metrics target the **simulation job** of an experiment — the job that
+advances model time chunk by chunk (commonly named ``SIM``). SYPD and CHSY are
+defined in terms of simulated years, so they rely on the chunk calendar
+(``CHUNK``, ``CHUNKSIZE``, ``CHUNKSIZEUNIT``) that only chunk-based simulation
+jobs carry; CORE_HOURS can still be computed on any job.
```
VindeeR, line 14: This part feels repetitive since you already spoke about it previously

docs/source/userguide/performance/index.rst
```diff
@@ -0,0 +1,70 @@
...
+
+CPMIP metrics are not computed automatically: today they are evaluated only
+as part of the notification feature. Autosubmit computes a job's CPMIP
+metrics when, and only when, both conditions hold:
```
VindeeR:
```suggestion
CPMIP metrics are not computed automatically: today they are evaluated only
as part of the notification workflow. 
A job’s CPMIP metrics are computed only when both of the following conditions are met:
```

docs/source/userguide/performance/index.rst
```diff
@@ -0,0 +1,70 @@
...
+* **Low SYPD** → the job is slow relative to wall-clock. Useful for catching
+  regressions after a code change, a new compiler, or a different
+  platform/partition. Compare against a known-good SYPD baseline for the same
+  model on the same platform.
+* **High CHSY** → the job is *expensive per simulated year*. A run can have an
+  acceptable SYPD but still burn excessive core-hours if scaled past the
+  strong-scaling sweet spot. CHSY is the canonical metric for catching that.
+* **High CORE_HOURS** → the job is consuming more allocation than expected.
+  Useful as a budget guard against runaway runs (e.g., a chunk that should
+  take 1 000 core-hours but takes 10 000 because of a wait/retry loop).
```
VindeeR: Idk about you defining it for the user, I believe each user might see a situation on its own way, So setting yourself which is considered high or low use, might be a bit too much no? especially when you start to think about different platforms, with different consumption/uses

VindeeR (changes requested):
I have a few suggestion on your wording feel free to change it however you see fit, just try to be a bit more clear and precise, at times, I feel the wording can be fairly vague or confusing.

Outside this pointers, this is a great documentation explaining and  broadening everything about this new feature, thanks for the hard work 🙌

docs/source/userguide/configure/index.rst
paparici: It is an anchor for referencing from the performance page

docs/source/userguide/performance/index.rst
paparici: Where do you think is better, in the config section or in the performance section?

docs/source/userguide/performance/index.rst
paparici: Agree, I just wanted to add an example. I will delete it. Better to leave the interpretation to the user.

docs/source/userguide/performance/index.rst
VindeeR: Leave it here at the performance IMHO, It explains how the metrics work, and ultimately how they are related/used to the performance.

Someone that wants to understand the metrics would look into performance, and someone looking into how to configure it would fall into performance too.

At least I believe it would be like this.

VindeeR (approved):
Waiting for the CI to finish so it can be merged 👍 
Thanks for the hard-work on finishing this new feature

PR #3021, opened 2026-05-18
Refactor describe to enumerate from the database and support archived experiments

### LINKED ISSUE #1110: The describe command should ignore directories that are not valid
Hello @dbeltrankyl,

### Autosubmit Version

Latest from Git, `master`

### Summary

I had archived an experiment, which produced the file `~/autosubmit/2023/a002.tar.gz`. After that, when I tried `autosubmit -lc DEBUG describe` I noticed the following in the logs

```bash
(autosubmit4) kinow@ranma:~$ autosubmit -lc DEBUG describe
/home/kinow/mambaforge/envs/autosubmit4/bin/autosubmit:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('autosubmit==4.0.85')
Autosubmit is running with 4.0.87
Describing a000
Owner: kinow
Created: 2023-07-12 20:06:33.129013
Model: https://github.com/kinow/auto-mhm-test-domains.git
Branch: rocrate
HPC: LOCAL
Description: Autosubmit mHM test domains
Describing a002
Owner: kinow
Created: 2023-08-11 09:21:37.931250
Model: Not Found
Branch: master
HPC: LOCAL
Description: test
[WARNING] Could not describe the following experiments:
['logs', '9999', '2023'][eCode=3000]
```

I created a directory `blabla`, and another directory `9999`. As seen above, the `blabla` directory is ignored (probably the code filters for 4-characters long), but other directories like `2023` and `9999` are treated as experiment directories and raise the warning above.

### Steps to reproduce

See above.

Warning, and possibly wasted resources in scanning unnecessary directories.

I think we have at least a couple of ways to fix this.

1. Filter for 4 characters and also for valid expid, something like `^[a-z][a-z0-9]+$`
2. and/or consider what's in the database, and either ignore the rest or raise a warning for every directory that is a valid expid, but is not in the DB or that's in the DB but it's missing a directory (do not describe it)

@rohanali1798 & @mcastril  found some experiments in the ClimateDT VM that existed in the DB but users had `rm -rf <EXPID>`. Implementing 2 could assist with that, perhaps.

(Paste any relevant logs - use code blocks (```) to format console output, logs, and code, as
it's very hard to read otherwise.)

### Any other relevant information(if applicable)

()

### LINKED ISSUE #2717: Check archived experiments
### Problem

Archived experiments not describable

### Proposed Solution

Check that commands like `autosubmit describe $expid` work on compressed experiments

(suggestion from @dbeltrankyl )

rajeshsuraj:
Closes #1110 
Closes #2717 

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -3448,7 +3448,7 @@ class Autosubmit:
                     Log.warning(f"Experiment '{expid}' not found in the database, skipping")
 
         for experiment_id in experiments_ids:
-            exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, experiment_id)
+            exp_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(experiment_id)
             try:
                 try:
                     # Preferred source of truth: the on-disk config files.
@@ -3494,18 +3494,18 @@ class Autosubmit:
                     branch = snapshot["branch"]
                     hpc = snapshot["hpc"]
                     description = get_experiment_description(experiment_id)
-                    description = description if description else ""
+                    description = description[0][0] if description else ""
                     Log.info(f"Experiment '{experiment_id}' files not found; "
                             f"it may have been archived. Showing the last "
                             f"stored snapshot.")
-                Log.result("Describing {0}", experiment_id)
-                Log.result("Owner: {0}", user)
-                Log.result("Location: {0}", exp_path)
-                Log.result("Created: {0}", created)
-                Log.result("Model: {0}", model)
-                Log.result("Branch: {0}", branch)
-                Log.result("HPC: {0}", hpc)
-                Log.result("Description: {0}", description)
+                Log.result(f"Describing {experiment_id}")
+                Log.result(f"Owner: {user}")
+                Log.result(f"Location: {exp_path}")
+                Log.result(f"Created: {created}")
+                Log.result(f"Model: {model}")
+                Log.result(f"Branch: {branch}")
+                Log.result(f"HPC: {hpc}")
+                Log.result(f"Description: {description}")
             except Exception:
                 not_described_experiments.append(experiment_id)
         if len(not_described_experiments) > 0:
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -3427,8 +3427,7 @@ class Autosubmit:
 
         user = created = model = branch = hpc = ""
         not_described_experiments = []
-
-        db_expids = get_experiment_expids()
+        experiments_ids_not_in_db = []
 
         if ',' in input_experiment_list:
             requested = [e.strip().lower() for e in input_experiment_list.split(',') if e.strip()]
@@ -3438,17 +3437,25 @@ class Autosubmit:
             requested = [e.strip().lower() for e in input_experiment_list.split(' ') if e.strip()]
 
         if requested is None:
-            experiments_ids = sorted(db_expids)
+            experiments_ids = sorted(get_experiment_expids())
         else:
+            found = get_experiment_expids(expids=requested)
             experiments_ids = []
-            for expid in requested:
-                if expid in db_expids:
-                    experiments_ids.append(expid)
-                else:
-                    Log.warning(f"Experiment '{expid}' not found in the database, skipping")
+            for e in requested:
+                (experiments_ids if e in found else experiments_ids_not_in_db).append(e)
+
+        if experiments_ids_not_in_db:
+            Log.warning(f"Experiments not found in the database, skipping: {experiments_ids_not_in_db}")
 
         for experiment_id in experiments_ids:
             exp_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(experiment_id)
+            if exp_path.is_dir():
+                try:
+                    folder_owner = pwd.getpwuid(exp_path.stat().st_uid).pw_name
+                    if folder_owner != get_from_user:
+                        continue
+                except Exception:
+                    pass
             try:
                 try:
                     # Preferred source of truth: the on-disk config files.
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -3428,7 +3428,6 @@ class Autosubmit:
         user = created = model = branch = hpc = ""
         not_described_experiments = []
 
-        # --- 1. Build the candidate list from the database ---
         db_expids = get_experiment_expids()
 
         if ',' in input_experiment_list:
@@ -3448,7 +3447,6 @@ class Autosubmit:
                 else:
                     Log.warning(f"Experiment '{expid}' not found in the database, skipping")
 
-        # --- 2. Describe each experiment ---
         for experiment_id in experiments_ids:
             exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, experiment_id)
             try:
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -54,7 +54,7 @@ from autosubmit.config.configcommon import AutosubmitConfig
 from autosubmit.config.yamlparser import YAMLParserFactory
 from autosubmit.database.db_common import (
     create_db, get_experiment_description, get_autosubmit_version, check_experiment_exists,
-    update_experiment_description_version
+    update_experiment_description_version, get_experiment_expids
 )
 from autosubmit.database.db_structure import get_structure
 from autosubmit.experiment.detail_updater import ExperimentDetails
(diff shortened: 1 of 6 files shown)


autosubmit/autosubmit.py
```diff
@@ -3426,6 +3429,9 @@ def describe(input_experiment_list="*", get_from_user=""):
             try:
```
VindeeR, line 3439: I believe that this try block can be removed, once they won't try to access the folder/file that doesn't exist anymore the exception should happen if I'm not mistaken

autosubmit/autosubmit.py
```diff
@@ -3426,6 +3429,9 @@ def describe(input_experiment_list="*", get_from_user=""):
...
                 experiment_id = experiment_id.strip(" ")
                 exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, experiment_id)
+                if not expid_re.match(experiment_id) or not Path(exp_path, "conf").is_dir():
+                    Log.warning(f"'{experiment_id}' is not a valid experiment id, skipping")
+                    continue
```
VindeeR:
```suggestion
                if not expid_re.match(experiment_id) or not Path(exp_path, "conf").is_dir():
                    not_described_experiments.append(experiment_id)
                    Log.warning(f"'{experiment_id}' is not a valid experiment id, skipping")
                    continue
```

VindeeR (changes requested):
I would like you to create tests validating the change in behavior and try to poke a variety of scenarios even forcing wrong ones, to see if the behavior is as expected

autosubmit/autosubmit.py
rajeshsuraj: I think we can keep the try and move the validation on top because the validator checks the exp at face level so its a good filter but the try is needed in places where the valid exp can have issues then the exception can highlight it for that.

autosubmit/autosubmit.py
```diff
@@ -3434,10 +3426,13 @@ def describe(input_experiment_list="*", get_from_user=""):
         else:
             experiments_ids = experiments_ids.split(' ')
         for experiment_id in experiments_ids:
+            experiment_id = experiment_id.strip(" ")
+            exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, experiment_id)
+            if not expid_re.match(experiment_id) or not Path(exp_path, "conf").is_dir():
+                Log.warning(f"'{experiment_id}' is not a valid experiment id, skipping")
+                not_described_experiments.append(experiment_id)
+                continue
```
VindeeR:
```suggestion
            experiment_id = experiment_id.strip(" ")
            exp_path = Path(BasicConfig.LOCAL_ROOT_DIR, experiment_id)
            if not expid_re.match(experiment_id) or not exp_path.joinpath("conf").is_dir():
                Log.warning(f"'{experiment_id}' is not a valid experiment id, skipping")
                not_described_experiments.append(experiment_id)
                continue
```

VindeeR:
As discussed on slack and using the [documentation](https://autosubmit.readthedocs.io/en/master/userguide/expids.html) as basis and discussed with @kinow.
we agreed that the best solution, and a way to actually fix it, is to check on the db first, validating the folder name in the db, because otherwise anything you put on the folder will assume is a valid project.

The only thing I would point out is that you might have to implement 2 different validation depending on the DB, since the 4.2 will use a different DB, maybe you can ask for @LuiggiTenorioK help on what we already have, and/or if anything need to be implemented

LuiggiTenorioK:
Summarizing what I discussed with @rajeshsuraj, the idea is now to check directly from the DB to get all the experiments, rather than relying on the dirs in the root folder. The process should follow something like:

1. Get experiments from the DB applying the filters
2. For each experiment:
  2.1. Get the details directly from the file system
  2.2. If it is not possible to get them, try the last snapshot  from the `details` table

You can use the classes in `autosubmit/experiment/detail_updater.py` for that, and even to make the code cleaner, if possible. Just take into account that the `ExperimentDetails` class purpose was to get the properties from the file system and save snapshots, but not to retrieve the snapshots to get the properties (which is still something that you are free to implement).

VindeeR:
Thanks for the time and the help @LuiggiTenorioK :raised_hands:

autosubmit/autosubmit.py
```diff
@@ -3413,69 +3422,90 @@ def describe(input_experiment_list="*", get_from_user=""):
...
-            experiments_ids = experiments_ids.split(' ')
+            experiments_ids = []
+            for expid in requested:
+                if expid in db_expids:
+                    experiments_ids.append(expid)
+                else:
+                    Log.warning(f"Experiment '{expid}' not found in the database, skipping")
+
         for experiment_id in experiments_ids:
+            exp_path = os.path.join(BasicConfig.LOCAL_ROOT_DIR, experiment_id)
```
VindeeR: we want to convert all the `os.path` to `Path`, since the library `Path` is newer is more convenient.

[os.path vs path](https://stackoverflow.com/questions/67112343/pathlib-path-vs-os-path-join-in-python)
[Paths in Python](https://www.pythonsnacks.com/p/paths-in-python-comparing-os-path-and-pathlib) recommended reading

```suggestion
            exp_path = Path(BasicConfig.LOCAL_ROOT_DIR).joinpath(experiment_id)
```

autosubmit/autosubmit.py
```diff
@@ -3413,69 +3422,90 @@ def describe(input_experiment_list="*", get_from_user=""):
...
+                            f"stored snapshot.")
                 Log.result("Describing {0}", experiment_id)
                 Log.result("Owner: {0}", user)
                 Log.result("Location: {0}", exp_path)
                 Log.result("Created: {0}", created)
                 Log.result("Model: {0}", model)
                 Log.result("Branch: {0}", branch)
                 Log.result("HPC: {0}", hpc)
-                Log.result("Description: {0}", description[0][0])
+                Log.result("Description: {0}", description)
```
VindeeR: we have been transitioning to f-strings too since they are more readable

[here's a short explanation on them](https://stackoverflow.com/questions/43123408/f-strings-vs-str-format)

```suggestion
                Log.result(f"Describing {experiment_id}", )
                Log.result(f"Owner: {user}")
                Log.result(f"Location: {exp_path}")
                Log.result(f"Created: {created}")
                Log.result(f"Model: {model}")
                Log.result(f"Branch: {branch}")
                Log.result(f"HPC: {hpc}")
                Log.result(f"Description: {description}")
```

VindeeR (changes requested):
I have just a few things that I would like changes, but I have a big doubt, and maybe @LuiggiTenorioK might be able to help me with that better.

So as I was reading through the changes, I realized that maybe we could filter by owner on the DB instead of a loop after getting all experiments, this would definitely improve the quality and speed of the code whenever describe is used.

Another thing I had in mind is, the way the code works is trying to bring anything on the db and filtering it by the id of the experiments found in the pointed folder (folder name).

So I was wondering wouldn't it be better to filter the experiments by concatenating that to the DB query much like what was suggested for the owner.

I might be shortsighted, but I do think it would be better and easier to generate the query for that, than getting all experiments from the db and filtering on the code, since the DB already is thought to process the data in a reliable and faster way.

autosubmit/autosubmit.py
rajeshsuraj: Sorry missed this change last week , will change from os.path to Path.

LuiggiTenorioK (changes requested):
> So as I was reading through the changes, I realized that maybe we could filter by owner on the DB instead of a loop after getting all experiments, this would definitely improve the quality and speed of the code whenever describe is used.

That's a valid concern. However, since the details table is a snapshot and probably doesn't reflect the real situation, it is better to filter it then in a loop rather than inside the query.

I still don't review this PR deeply, but I think that I saw that this filter by owner is not included in the changes @rajeshsuraj 



> Another thing I had in mind is, the way the code works is trying to bring anything on the db and filtering it by the id of the experiments found in the pointed folder (folder name).
> 
> So I was wondering wouldn't it be better to filter the experiments by concatenating that to the DB query much like what was suggested for the owner.
> 
> I might be shortsighted, but I do think it would be better and easier to generate the query for that, than getting all experiments from the db and filtering on the code, since the DB already is thought to process the data in a reliable and faster way.

That is true. It is more efficient, especially when there are many experiments created.

@rajeshsuraj, you might need to add a `FROM name in ( ... )` to the query, and beware that SQLite has a limit of 999 parameters, so chunking it is needed. I think Dani made that recently in some place.

@rajeshsuraj I suggest resolving the two comments above before moving to the next review round.

autosubmit/experiment/detail_updater.py
```diff
@@ -269,6 +269,16 @@ def save_update_details(self):
...
+    
+    def get_details(self) -> Union[Dict[str, Any], None]:
+        """
+        Retrieve the last stored snapshot of the experiment's details
+        from the database.
+        """
+        exp_id = getattr(self, "exp_id", None)
+        if exp_id is None:
+            exp_id = get_experiment_id(self.expid)
+        return self._details_repo.get_details(exp_id)
```
VindeeR, line 281: is this piece of code being used? an d what is the purpose of it, I'm a bit confused.

autosubmit/autosubmit.py
```diff
@@ -3413,69 +3422,90 @@ def describe(input_experiment_list="*", get_from_user=""):
...
+        if requested is None:
+            experiments_ids = sorted(db_expids)
         else:
-            experiments_ids = experiments_ids.split(' ')
+            experiments_ids = []
+            for expid in requested:
+                if expid in db_expids:
+                    experiments_ids.append(expid)
+                else:
+                    Log.warning(f"Experiment '{expid}' not found in the database, skipping")
```
VindeeR: I believe it might be better to have a different variable that stores all the selected `expids` that are not in the db and print this log just once the loop is finished

Something like:

```suggestion
                else:
                    experiments_ids_not_in_db.append(expid)
            if experiments_ids_not_in_db is not None:
                Log.warning(f"Experiment '{experiments_ids_not_in_db}' not found in the database, skipping")
```

test/integration/test_describe.py
```diff
@@ -26,108 +27,119 @@
...
+    # describe reads the database; autosubmit_exp creates it on first call.
+    autosubmit_exp(experiment_data=_experiment_data())
 
-    The ``expid_len`` defines how many expids we will provide to the command.
-    If the value is ``None``, then we will use a non-existent expid.
-    If the value is ``0``, then we will not provide any expid (i.e., empty).
+    input_list = ''
+    exps = []
+    if expid_count is None:
+        input_list = 'zzzz'  # Valid format, not in the database.
```
VindeeR: I just want to double check if this comment is right

test/integration/test_describe.py
```diff
@@ -26,108 +27,119 @@
...
-        }
-    )
+    exp = autosubmit_exp(experiment_data=_experiment_data(hpcarch='TEST_SLURM'))
 
     mocker.patch('sys.argv', ['autosubmit', '-lc', 'ERROR', '-lf', 'WARNING', 'describe', exp.expid])
     _, args = autosubmit.parse_args()
     output = autosubmit.run_command(args=args)
 
-    assert getuser() == output[0]
+    assert getuser() == output[0]
```
VindeeR: EoF

test/unit/test_expid.py
```diff
@@ -79,8 +79,7 @@ def test_expid(mocker, copy_id, expected, tmp_path, autosubmit_config, monkeypat
 
     with expected:
         expid = Autosubmit.expid("Test", copy_id=copy_id)
-        experiment = Autosubmit.describe(expid)
+        # The `describe` call was removed here due to describe restructuring.
```
VindeeR:
```suggestion
        # The `describe` call was removed here due to describe now needing the db, which conflicts with mock.
```

VindeeR (commented):
I want to reread the changes before approving, but it looks good so far, thanks for the hard work

It would be great if the headers had at least a summary of the feature and problem they are fixing, this way during meeting it can be easier to track, understand and remember what it is about.

autosubmit/experiment/detail_updater.py
rajeshsuraj: yes, this is used in autosubmit.py under describe command to get the details from the database if the physical files don't exist.

dbeltrankyl:
Hello, how is this feature going? I see that you requested a PR review to @LuiggiTenorioK 

Checking because I believe we need to have this for the 4.1.17, but it can be added later to the dev module as it is kinda isolated command

rajeshsuraj:
> Hello, how is this feature going? I see that you requested a PR review to @LuiggiTenorioK
> 
> Checking because I believe we need to have this for the 4.1.17, but it can be added later to the dev module as it is kinda isolated command

I was finished with the feature just waiting for the review from @LuiggiTenorioK.

LuiggiTenorioK:
> Hello, how is this feature going? I see that you requested a PR review to @LuiggiTenorioK

I was reviewing it today. It seems good, there are only really minor issues, but they are not really important.

I'll approve it rn.

LuiggiTenorioK (approved):

Issue #2951, opened 2026-04-15
`list index out of range` in squeue verification after wrapper submission causes infinite resubmission loop


MuhammadShafeeque:
### Autosubmit Version

v4.1.16.1

### Workflow release used / branch

v6.1.1

### Expid affected(If applicable)

o02f

### Summary

After wrapper submission, the squeue -o %A,%.50j verification step (I think) triggers `list index out of range`, causing `LUMI submission failed [eCode=6015]`. This marks all jobs in the cycle as FAILED, triggering a resubmission loop.

logs: experiment **o02f** logs at `/appl/AS/AUTOSUBMIT_DATA/o02f/tmp/ASLOGS/`

Critically, when this happens with APP/OPA wrappers, other parallel jobs (CLEAN, FIX_CONSTANT_VARIABLES, etc.) are also re-submitted in every cycle, leading to hundreds of duplicate SLURM jobs!

## Related
Gitlab issue: https://gitlab.earth.bsc.es/digital-twins/de_340-2/workflow/-/issues/1562
IFS-FESOM end-to-end testing for E-26.1: https://jira.eduuni.fi/browse/CSCDESTINCLIMADT-1075

@dbeltrankyl @kinow @mandresm

dbeltrankyl:
I would recommend disabling the wrappers in LUMI until this is investigated

kinow:
Thank you for creating the issue here, @MuhammadShafeeque ! 🙇

TracyMcBean:
Could this also affect the runs on MN5? @dbeltrankyl ?

dbeltrankyl:
Hello @TracyMcBean 

I'm not sure, as I haven't had time to investigate yet, but as far as I know, all the reports are coming from experiments launched in Lumi. Maybe workflow developers can confirm it cc : ( @franra9 , @dmcclainbsc, or @fmacchia?)

dmcclainbsc:
I have never really worked with wrappers but I was wondering if it was related to Cheng's issue, which is on MN5: https://jira.eduuni.fi/browse/CSCDESTINCLIMADT-1152

ialsina:
As @MuhammadShafeeque, I created a separate issue, #3038, which is a duplicate, at least the `list index out of range` part. That occurred in an end-to-end in LUMI, **without wrappers**. Just to point out that, while the infinite resubmission loop is a wrapper-related issue, the `list index out of range` bug might block the TS for the workflow release v6.2.0.

[...]

PR #3024, opened 2026-05-20
Fix slurm ssh banner parsing

MuhammadShafeeque:
**Check List**

- [ ] Documentation updated.

TESTS:

t0qr
t0qy (with wrappers)

linked to #2951 
@kinow

### DIFF
--- a/autosubmit/platforms/paramiko_platform.py
+++ b/autosubmit/platforms/paramiko_platform.py
@@ -1179,10 +1179,13 @@ class ParamikoPlatform(Platform):
             retries -= 1
             sleep(2)
         if retries >= 0:
-            # get id last line
-            job_ids_names = job_id_name.split('\n')[1:-1]
-            # get all ids by job-name
-            job_ids = [job_id.split(',')[0] for job_id in job_ids_names]
+            # Only keep lines where the first comma-delimited field is a numeric job ID,
+            # skipping the squeue header and any SSH banner lines in stdout.
+            job_ids = [
+                line.split(',')[0].strip()
+                for line in job_id_name.splitlines()
+                if line.split(',')[0].strip().isdigit()
+            ]
         return job_ids
 
     def get_check_job_cmd(self, job_id: str) -> str:
--- a/autosubmit/platforms/slurmplatform.py
+++ b/autosubmit/platforms/slurmplatform.py
@@ -286,7 +286,7 @@ class SlurmPlatform(ParamikoPlatform):
         :return: Command to look for a job in the queue.
         :rtype: str
         """
-        return f'squeue -o %A,%j -n {job_name}'
+        return f'squeue -o %A,%.50j -n {job_name}'
 
     def get_job_energy_cmd(self, job_id: str) -> str:
         """Generates a command to get data from a job
--- a/autosubmit/platforms/slurmplatform.py
+++ b/autosubmit/platforms/slurmplatform.py
@@ -214,7 +214,11 @@ class SlurmPlatform(ParamikoPlatform):
             if x11:
                 return int(output.splitlines()[0])
             for line in output.splitlines():
-                jobs_id.append(int(line.split(' ')[3]))
+                m = re.search(r'Submitted batch job (\d+)', line, re.IGNORECASE)
+                if m:
+                    jobs_id.append(int(m.group(1)))
+            if not jobs_id:
+                raise AutosubmitCritical("Submission failed. No job ID found in sbatch output", 7014)
             return jobs_id
         except IndexError as exc:
             raise AutosubmitCritical("Submission failed. There are issues on your config file", 7014) from exc
@@ -477,7 +481,13 @@ class SlurmPlatform(ParamikoPlatform):
 
         for pattern in critical_patterns:
             if pattern in err_lower:
+                # Filter SSH banner/module-load lines; keep only Slurm-relevant stderr lines
+                slurm_lines = [
+                    line for line in err.splitlines()
+                    if re.search(r'(?:sbatch|salloc|srun|sinfo|slurm)\b', line, re.IGNORECASE)
+                ]
+                clean_err = "\n".join(slurm_lines) if slurm_lines else err
                 raise AutosubmitCritical(
-                    f"Permanent Slurm error: {err}",
+                    f"Permanent Slurm error: {clean_err}",
                     7014,
                 )
--- a/autosubmit/platforms/slurmplatform.py
+++ b/autosubmit/platforms/slurmplatform.py
@@ -142,9 +142,9 @@ class SlurmPlatform(ParamikoPlatform):
             self.scratch, self.project_dir, self.user, self.expid)
         self.remote_log_dir = os.path.join(self.root_dir, "LOG_" + self.expid)
         self.cancel_cmd = "scancel"
-        self._submit_cmd = f'sbatch --no-requeue -D {self.remote_log_dir} '
+        self._submit_cmd = f'sbatch --parsable --no-requeue -D {self.remote_log_dir} '
         self._submit_command_name = "sbatch"
-        self._submit_hold_cmd = f'sbatch -H -D {self.remote_log_dir} '
+        self._submit_hold_cmd = f'sbatch --parsable -H -D {self.remote_log_dir} '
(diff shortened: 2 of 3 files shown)


MuhammadShafeeque:
This PR does not close #2951

autosubmit/platforms/slurmplatform.py
```diff
@@ -216,7 +216,11 @@ def get_submitted_job_id(self, output: str, x11: bool = False) -> list[str]:
             if x11:
                 return int(output.splitlines()[0])
             for line in output.splitlines():
-                jobs_id.append(int(line.split(' ')[3]))
+                m = re.search(r'Submitted batch job (\d+)', line, re.IGNORECASE)
+                if m:
+                    jobs_id.append(int(m.group(1)))
+            if not jobs_id:
+                raise AutosubmitCritical("Submission failed. No job ID found in sbatch output", 7014)
```
kinow, line 222: This sounds like something that should have been done before 👍 . Maybe we could have a pure-python version, avoiding regex?

Something like 

```python
parts = line.split(' ')
if len(parts) >= 4 and parts[3].isdigit():
    jobs_id.append(int(parts[3]))
if not jobs_id:
   raise...
```

Or even better, @MuhammadShafeeque , a tip I learned today from @manuel-g-castro , what if we change the `._submit_cmd` in the Slurm platform from

```
self._submit_cmd = f'sbatch --no-requeue -D {self.remote_log_dir} '
```

To

```
self._submit_cmd = f'sbatch --parsable --no-requeue -D {self.remote_log_dir} '
```

With `--parsable`, the output becomes this:

```
$ sbatch --parsable --no-requeue --chdir=/home/bsc/bsc032841 job.sh
41318847
```

So here we would only have to check for isdigit without having to worry in case Slurm changes the text used in the successful submission message?

The --parsable has been available in sbatch since around release 14.03, from 2014, https://support.schedmd.com/show_bug.cgi?id=594, so it should be fairly safe to use.

autosubmit/platforms/paramiko_platform.py
```diff
@@ -1120,10 +1120,13 @@ def get_job_id_by_job_name(self, job_name, retries=2):
...
-            job_ids_names = job_id_name.split('\n')[1:-1]
-            # get all ids by job-name
-            job_ids = [job_id.split(',')[0] for job_id in job_ids_names]
+            # Only keep lines where the first comma-delimited field is a numeric job ID,
+            # skipping the squeue header and any SSH banner lines in stdout.
+            job_ids = [
+                line.split(',')[0].strip()
+                for line in job_id_name.splitlines()
+                if line.split(',')[0].strip().isdigit()
+            ]
```
kinow, line 1188: Why did we have to modify this part, @MuhammadShafeeque ? I do think this version testing for digit/numeric is safer, but just wondering if you had a case where it returned something that was not a digit?

autosubmit/platforms/slurmplatform.py
```diff
@@ -514,7 +518,13 @@ def _check_for_unrecoverable_errors(self) -> None:
...
             if pattern in err_lower:
+                # Filter SSH banner/module-load lines; keep only Slurm-relevant stderr lines
+                slurm_lines = [
+                    line for line in err.splitlines()
+                    if re.search(r'(?:sbatch|salloc|srun|sinfo|slurm)\b', line, re.IGNORECASE)
+                ]
+                clean_err = "\n".join(slurm_lines) if slurm_lines else err
                 raise AutosubmitCritical(
-                    f"Permanent Slurm error: {err}",
+                    f"Permanent Slurm error: {clean_err}",
```
kinow, line 492: @MuhammadShafeeque I couldn't figure a way to get the code to reach this part. Do you recall what was the server response that made you add this?

If I understand correctly, here you are changing the code so that it considers an error only messages that contain the tokens above (sbatch, salloc, etc.). But I am not sure if that's the best way to filter other possible errors.

For instance, `squeue` is not considered above, even though that command is used by the Slurm platform and `send_command` (which calls `_check_for_unrecoverable_errors`, this function). The job cancellation command is also not included in the list, `scancel`.

In `__check_running_jobs`, the `send_command` function is also used by `_check_running_jobs` which sends a variety of Linux commands. This code predates me, so I am not sure if there are no side effects from making this part of `critical_patterns` to filter only Slurm lines, but from the function name and docs, I'm +1 to make it Slurm specific. We may just have to add scancel,squeue... these are the only two that I think are missing.

kinow (commented):
We have a Slurm fixture, that launches a Slurm Docker container, @MuhammadShafeeque . I can help you with the tests (the more important the PR, the more important it's for us to cover it with as many/as good tests as we can). Let me just have a quick lunch and I will try to send an example that should be able to cover some of the changes, or be used as example.

kinow:
Sorry, got sidetracked copying an experiment for debugging the issue.

autosubmit/platforms/slurmplatform.py
MuhammadShafeeque: Thank you very much @kinow for the review and suggestions.

Good point, thank you! I've taken your `--parsable` suggestion — `sbatch` now outputs only `<jobid>` or `<jobid>;<cluster>`, so parsing is a simple `isdigit()` check with no regex needed at all.

In experiment `o02f` on LUMI (log: `20260413_234431_run.log`), the old `line.split(' ')[3]` raised `IndexError: list index out of range` 35+ times over ~50 minutes because LUMI's SSH session injected MOTD banner lines before sbatch's stdout. The repeated failures ultimately exhausted the QoS submit-job limit and killed the run. Similar behaviour in `o02e` on MN5 was obsered.

autosubmit/platforms/paramiko_platform.py
MuhammadShafeeque: Same root cause as the banner issue above. On LUMI/MN5, SSH banner/MOTD lines appear in stdout before the actual `squeue` output. The old `split('\n')[1:-1]` skipped only the header/trailing-newline, so a banner line in the middle would be treated as a job ID and fail. The `.isdigit()` filter handles this regardless of where banner lines appear.

I also could not reproduce this case with test scripts without running an actual experiment. This addition is acting as safty buffer, but with other changes in place this change may not add any value, and it can be reverted. 

I am testing this PR with an expid t0qr

autosubmit/platforms/slurmplatform.py
MuhammadShafeeque: Good catch! @kinow . Thanks.
 `squeue` and `scancel` are now added to the filter regex. The motivation is the same banner issue: LUMI/MN5 append MOTD text to stderr, making the `AutosubmitCritical` message extremely noisy. The filter strips non-Slurm lines so only the actionable error is shown. The `else err` fallback preserves the full message if no Slurm-specific lines are present.

We can rename `Clean_error` if you prefer another name.

autosubmit/platforms/slurmplatform.py
```diff
@@ -291,7 +296,7 @@ def get_job_id_by_job_name_cmd(self, job_name: str) -> str:
         :return: Command to look for a job in the queue.
         :rtype: str
         """
-        return f'squeue -o %A,%.50j -n {job_name}'
+        return f'squeue -o %A,%j -n {job_name}'
```
MuhammadShafeeque: Testing this change in t0qr

kinow (commented):
@MuhammadShafeeque sorry the delay to come back to this issue. I've finished a few tasks, and now I finally have plenty of time to review & fix issues here. Can you rebased it, please? There were some failures in the GH Actions pipeline, but maybe rebasing will fix or reduce the failures. Thanks!

MuhammadShafeeque:
> @MuhammadShafeeque sorry the delay to come back to this issue. I've finished a few tasks, and now I finally have plenty of time to review & fix issues here. Can you rebased it, please? There were some failures in the GH Actions pipeline, but maybe rebasing will fix or reduce the failures. Thanks!

Hi @kinow 
Done ✅

kinow:
Thank you @MuhammadShafeeque ! I'll start reviewing it tomorrow or Friday :muscle:

kinow:
We had an operational issue last afternoon, which I am still working on with Cheng. I'm updating some issues related to the issue, and creating a couple more. I'm afraid this review is now postponed for next week, sorry Muhammad.

PR #3029, opened 2026-05-22
Refactor `inspect` command to allow execution with multiple filters

### LINKED ISSUE #3025: Refactor `inspect` command to allow filter combinations as `setstatus` and `recovery`
Similarly to issue #3020, `inspect` command uses if/elif/else statements for applying filters (-fl, -fs, -ft, -fc). I would propose normalize this command too to allow:
- disable mutually exclusive filtering, allow multiple filters in the same command execution
- call same utils function `apply_job_filter()`

This way all the commands that use filters (setstatus, recovery, monitor and inspect) will follow the same filter code structure.
I think there are no more commands that use filters.

ntorqulu:
Closes #3025 

As implemented previously for setstatus and recovery and monitor, inspect now allows being executed with multiple filters at the same time (-ft, -fs, -fc, -fl) and follows the same filter code structure.

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 - Implement filters for `recovery` command #2923
 - Add detection of inneficiencies for key CPMIP metrics #3022
 - Allow multiple filters in the `monitor` command #3020
+- Allow multiple filters in the `inspect` command #3029
 
 ### 4.1.16: Postgres (experimental) support, bug fixes, and enhancements
 
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -455,22 +455,17 @@ class Autosubmit:
             subparser.add_argument(
                 '-q', '--quick', action="store_true", help='Only checks one job per each section')
 
-            group.add_argument('-fs', '--filter_status', type=str,
-                               choices=('Any', 'READY', 'COMPLETED',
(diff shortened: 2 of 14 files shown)


ntorqulu:
What I've seen so far that could be changed but is not in the scope of this pr:
---
- `inspect` command first checks if there's a lock file (experiment is being accessed). In this case, it only evaluates the unsubmitted jobs.
- if exp is not locked and inspect is executed with command --force, all jobs are selected to inspect
- **else**, then filter the jobs by the filters by chunk, status, section or list. So, you cannot combine force and filters in the same command. I think is better first filter the jobs and then check the lock and force flags.
---
- The following warning is always displayed:
`[WARNING] Generating the auxiliary job_list used for the -cw flag.`
even when I don't activate the `-cw` flag. This can confuse users... Should be changed in method `generate_scripts_andor_wrappers`
---
- The lock file is not deleted by default, currently I've to delete it manually. I've opened an issue for that.

ntorqulu:
About the integration tests for the `inspect` command, at the end I've decided to create a new file `test_inspect` because I found `test_inspect_local.py` to narrowed to check the cmd files generated.
I'm not sure about that.

VindeeR (approved):
This is beautifully done, thanks for the improvement in the tests, and reducing the amount of repetitive code on the fixtures.

The reading on the filter are much more clear and easy to follow too.

Keep up the good work 🙌

Issue #1345, opened 2024-07-08
Consistency in Autosubmit command-line


kinow:
Reported by a user, `autosubmit` (sub)commands follow a pattern of `autosubmit` + verb. Like autosubmit+run, autosubmit+stop, autosubmit+check. Except for recovery, which is a noun, instead of recover. We could review all the commands at some point, and deprecate recovery making an alias recover (less cognitive burden on users, as it's easier to remember as+verb for all commands).

kinow:
In GitLab by @sebastianbeyer on Jul 10, 2024, 22:55

Also short and long names should be consistent. In this example a long option `-all` only has a single `-`. Probably a special case, but it is still confusing.

```
usage: autosubmit report [-all] [-t] [-fp] [-p] expid
    expid                                Experiment identifier
    -t, --template <path_to_template>    Allows to select a set of parameters to be extracted
    -fp, --show_all_parameters           All parameters will be extracted to a different file
    -fp, --folder_path                   By default, all parameters will be put into experiment tmp folder
    -p, --placeholders                   disable the replacement by - if the variable doesn't exist
```

kinow:
Agreed. We should probably stick with conventions from unix tools and others tools like git/jq/argparse/etc., as most users will be already familiar with that.

This is a good guideline https://clig.dev/#arguments-and-flags (cc @ainagaya as I think she's working on something similar in the workflow, with `--` vs. `-`).

Thanks @sebastianbeyer !

kinow:
In GitLab by @ainagaya on Jul 11, 2024, 09:47

I opened this issue some time ago for exactly the same that Sebastian is reporting above https://earth.bsc.es/gitlab/es/autosubmit/-/issues/1237 :)

Issue #1227, opened 2024-02-12
[enhancement] Add auto-completion to our command-line tool


kinow:
Hi @dbeltrankyl, this comes from a comment from @mcastril on Slack.

We can improve Autosubmit user experience by providing auto-completion in the command line. So that users can type, for example, `autosubmit`, hit `TAB`, and that should bring `run`, `expid`, `setstatus`, etc.

Then, if they select `autosubmit run`, this could also try to auto-complete with experiment ID's, or options.

kinow:
(I thought we had an issue for that, but couldn't find after searching on GitLab).

kinow:
And here's the file from Cylc, in case it helps: https://github.com/cylc/cylc-flow/blob/master/cylc/flow/etc/cylc-completion.bash

rajeshsuraj:
Hi @dbeltrankyl @kinow , I've assigned this to myself. 
My initial plan is to use argcomplete, which derives completions directly from the existing argparse definitions, plus a few custom completers for dynamic things (experiment IDs from the experiments directory, job names from the pickle for setstatus, status enums). 
This is much less infrastructure than the Cylc completion-server approach , what do you guys think ?

kinow:
We have another issue to split `autosubmit.py`, where there will be some custom decorators involved to move that argparse from that huge class into separate files. So the whole argparse configuration may change soon-ish (I will tackle that issue in July/August). So there's a small chance that whatever is done now could stop working. Maybe it'd be better to pick another issue for now IMO, @rajeshsuraj.

rajeshsuraj:
Okay will pick up a new issue then

PR #3035, opened 2026-05-27
Updated delete lock file to actually delete

### LINKED ISSUE #3033: `autosubmit.lock` is not deleted in `autosubmit create` & `autosubmit run`
### Description

Function `delete_lock_file` does not delete file `autosubmit/<EXPID>/tmp/autosubmit.lock` after executing commands `autosubmit create <EXPID>` or `autosubmit run <EXPID>`.
Function docstring specifies that the `base_path` is by default the experiment `tmp` directory, but in runtime this path is empty.
The lock file is not found and therefore is never deleted. 

* Autosubmit version: master

### Reproducible Example

Just run `autosubmit create <EXPID>` in local and check file `autosubmit/a05l/tmp/autosubmit.lock`.
After the execution of the command, it is not deleted.

### Expected Behaviour

File should be deleted after execution is completed.

### LINKED ISSUE #3055: `archive` and `delete` commands should acquire the lock file
### Description

Currently the following commands correctly acquire the lock file: `run`, `create`, `recovery`, `setstatus`, `pklfix`.
But there are two other commands that perform filesystem operations and do not acquire the lock:
- `archive`: deletes the experiment directory after compressing it. Problems when executing this command + `run` or `create`.
- `delete`: deletes the experiment folder. Same problem.

### Reproducible Example

1. Execute `autosubmit run <EXPID>`
2. Run `autosubmit archive <EXPID>` in parallel
3. The archive is executed without blocking

### Expected Behaviour

`archive` and `delete` should acquire autosubmit.lock file before proceeding.

rajeshsuraj:
Closes #3033 
Closes #3055 

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/scripts/_args.py
+++ b/autosubmit/scripts/_args.py
@@ -19,12 +19,10 @@
 
 import traceback
 from argparse import ArgumentParser, Namespace, RawDescriptionHelpFormatter
-from contextlib import suppress
 from dataclasses import dataclass
 from difflib import get_close_matches
 from enum import Enum
 from inspect import cleandoc
-from pathlib import Path
 from re import split, sub
 from typing import Any, TypeVar, cast, get_type_hints
 
@@ -323,27 +321,14 @@ def create_argparse_parser(docs: str | None, *, add_help=True) -> ArgumentParser
     )
 
 
-def _delete_lock_file(
-    base_path: str = Log.file_path, lock_file: str = "autosubmit.lock"
-) -> None:
-    """Delete the lock file if it exists. Suppresses permission errors raised.
-
-    :param base_path: Base path to locate the lock file. Defaults to the experiment ``tmp`` directory.
-    :param lock_file: The name of the lock file. Defaults to ``autosubmit.lock``.
-    :return: None
-    """
-    with suppress(PermissionError):
-        Path(base_path, lock_file).unlink(missing_ok=True)
-
-
 def exit_from_error(e: BaseException) -> int:
     """Called by ``Autosubmit`` when an exception is raised during a command execution.
 
     Prints the exception in ``CRITICAL`` if it is an ``AutosubmitCritical`` or an
     ``AutosubmitError`` exception, including any trace attached to the exception.
 
-    Exceptions raised by ``porta-locker` library print a message informing the user
-    about the locked experiment. Other exceptions raised cause the lock to be deleted.
+    Exceptions raised by the ``portalocker`` library print a message informing the user
+    that another process holds the experiment lock.
 
     :param e: The exception being raised.
     """
@@ -359,11 +344,9 @@ def exit_from_error(e: BaseException) -> int:
 
     if isinstance(e, BaseLockException):
         Log.warning(
-            "Another Autosubmit instance using the experiment\n. Stop other Autosubmit instances that are "
-            "using the experiment or delete autosubmit.lock file located on the /tmp folder."
+            "Another Autosubmit process is using this experiment. Wait for it to finish, "
+            "or stop it, before running this command again."
         )
-    else:
-        _delete_lock_file()
 
     if is_autosubmit_error:
         as_error: AutosubmitError | AutosubmitCritical = cast(
--- a/autosubmit/scripts/_cli_function.py
+++ b/autosubmit/scripts/_cli_function.py
@@ -19,9 +19,12 @@
 
 from argparse import ArgumentParser
 from collections.abc import Callable
+from contextlib import AbstractContextManager, nullcontext
 from functools import wraps
+from portalocker import TemporaryFileLock
 from typing import TYPE_CHECKING
 
+from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.log import setup_log_files
 from autosubmit.scripts._args import (
     CommandGroup,
@@ -44,12 +47,33 @@ if TYPE_CHECKING:
 __all__ = ["cli_function"]
 
 
+def _experiment_lock(expid: str) -> AbstractContextManager:
+    """Return the lock that stops two commands from modifying one experiment at once.
+
+    ``TemporaryFileLock`` deletes ``autosubmit.lock`` on release, including when the
+    command raises. From portalocker 4.0 the file is unlinked while the lock is still
+    held, so no other process can end up locking a stale copy of it.
+
+    ``timeout=1`` with ``fail_when_locked=False`` keeps the previous behaviour: retry
+    for one second, then raise ``AlreadyLocked``, which ``exit_from_error`` reports.
+
+    :param expid: The experiment identifier.
+    :return: A context manager that holds the experiment lock.
+    """
+    return TemporaryFileLock(
+        str(BasicConfig.expid_lock_file(expid)),
+        timeout=1,
+        fail_when_locked=False
+    )
+
+
 def cli_function(
     *,
     args_parser: Callable[[], ArgumentParser],
     options_type: type["AutosubmitOptions"],
     validators: "Validator[OptionsT] | list[Validator[OptionsT]] | None" = None,
     group: "CommandGroup" = CommandGroup.GENERAL,
+    lock: bool = False,
 ) -> Callable:
     """Decorator for Autosubmit CLI functions.
 
@@ -60,12 +84,14 @@ def cli_function(
     * parsing command-line arguments;
     * validating options;
     * handling command exceptions;
+    * holding the experiment lock while the command runs, if ``lock`` is set;
     * converting command return values into process exit codes.
 
     :param args_parser: The argparse parser for the sub-command.
     :param options_type: The type of the options for the sub-command.
     :param validators: A validation function for the sub-command.
     :param group: CLI command group. Default is general.
+    :param lock: Whether the sub-command holds ``<EXPID>/tmp/autosubmit.lock`` while it runs.
     :raises ValueError: If ``options_type`` or ``args_parser`` are not provided.
     """
     if options_type is None:
@@ -169,7 +195,8 @@ def cli_function(
 
                     opts._profiler = profiler
 
-                return_value = func(opts, **kwargs)
+                with _experiment_lock(opts.expid) if lock else nullcontext():
+                    return_value = func(opts, **kwargs)
 
                 return normalise_return_value(return_value)
             except KeyboardInterrupt:
--- a/autosubmit/scripts/archive.py
+++ b/autosubmit/scripts/archive.py
@@ -70,6 +70,7 @@ def args_parser() -> ArgumentParser:
     group=CommandGroup.EXPERIMENT,
     options_type=ArchiveOptions,
     validators=validate_expid,
+    lock=True,
 )
 def main(opts: ArchiveOptions) -> int | bool | None:
     from autosubmit.experiment.manage import archive
--- a/autosubmit/scripts/autosubmit.py
+++ b/autosubmit/scripts/autosubmit.py
@@ -33,32 +33,17 @@ from autosubmit.log.log import Log, AutosubmitCritical, AutosubmitError  # noqa:
 # Extend this set if a new lock-acquiring command is added.
 LOCK_OWNING_COMMANDS = frozenset({'run', 'create', 'recovery', 'setstatus', 'pklfix'})
 
-def delete_lock_file(base_path: Optional[Union[str, Path]] = None, lock_file: str = 'autosubmit.lock') -> None:
-    """Delete lock file if it exists. Suppresses permission errors raised.
+def delete_lock_file(base_path: Union[str, Path], lock_file: str = 'autosubmit.lock') -> None:
+    """Delete the autosubmit lock file at ``base_path/lock_file`` if it exists.
 
-    :param base_path: Base path to locate the lock file. Defaults to the experiment ``tmp`` directory.
-    :type base_path: Optional[Union[str, Path]]
-    :param lock_file: The name of the lock file. Defaults to ``autosubmit.lock``.
+    :param base_path: Directory containing the lock file.
+    :type base_path: Union[str, Path]
+    :param lock_file: The lock file name. Defaults to ``autosubmit.lock``.
     :type lock_file: str
     :return: None
     """
-    if base_path is None:
-        base_path = Log.file_path
     with suppress(PermissionError):
         Path(base_path, lock_file).unlink(missing_ok=True)
-
-
-def _get_lock_path(args: Optional[argparse.Namespace]) -> Optional[Path]:
-    """Return the experiment ``tmp`` directory for the expid in ``args``, or ``None`` if no expid.
-
-    :param args: Parsed CLI arguments.
-    :type args: Optional[argparse.Namespace]
-    :return: Path to the experiment ``tmp`` directory, or ``None``.
-    :rtype: Optional[Path]
-    """
-    if args and getattr(args, 'expid', None):
-        return BasicConfig.expid_tmp_dir(args.expid)
-    return None
 
 
 def _owns_lock(args: Optional[argparse.Namespace]) -> bool:
@@ -135,7 +120,7 @@ def main():
         if args:
             return_value = Autosubmit.run_command(args)
         if _owns_lock(args):
-            delete_lock_file(_get_lock_path(args))
+            delete_lock_file(BasicConfig.expid_tmp_dir(args.expid))
     except BaseException as e:
         command = "<no command provided>"
         expid = "<no expid provided>"
@@ -151,7 +136,8 @@ def main():
                     version = f"{as_conf.experiment_data.get('CONFIG', {}).get('AUTOSUBMIT_VERSION', 'unknown')}"
         Log.error(f"Arguments provided: {str(args)}")
         Log.error(f"This is the experiment: {expid} which had an issue with the command: {command} and it is currently using the Autosubmit Version: {version}.")
-        return_value = exit_from_error(e, _get_lock_path(args) if _owns_lock(args) else None)
+        lock_path = BasicConfig.expid_tmp_dir(args.expid) if _owns_lock(args) else None
+        return_value = exit_from_error(e, lock_path)
     # TODO: we need to define whether the function called here will return an int or bool
     if type(return_value) is bool:
         return_value = 0 if return_value else 1
--- a/autosubmit/scripts/autosubmit.py
+++ b/autosubmit/scripts/autosubmit.py
@@ -98,8 +98,7 @@ def exit_from_error(e: BaseException, lock_path: Optional[Union[str, Path]] = No
     elif lock_path is not None:
         delete_lock_file(lock_path)
 
-    if is_autosubmit_error:
-        e: Union[AutosubmitError, AutosubmitCritical] = e
+    if isinstance(e, (AutosubmitError, AutosubmitCritical)):
         if e.trace:
             Log.critical(f"Trace: {str(e.trace)}")
         Log.critical(f"{e.message} [eCode={e.code}]")
--- a/autosubmit/scripts/autosubmit.py
+++ b/autosubmit/scripts/autosubmit.py
@@ -20,46 +20,66 @@
 import argparse
 import traceback
 from contextlib import suppress
-from os import _exit  #noqa
-# noinspection PyProtectedMember
 from pathlib import Path
 from typing import Optional, Union
 
 from portalocker.exceptions import BaseLockException
 
 from autosubmit.autosubmit import Autosubmit  # noqa: E402
+from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.config.configcommon import AutosubmitConfig  # noqa: E402
 from autosubmit.log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
 
+# Extend this set if a new lock-acquiring command is added.
+LOCK_OWNING_COMMANDS = frozenset({'run', 'create', 'recovery', 'setstatus', 'pklfix'})
 
-def delete_lock_file(base_path: str = Log.file_path, lock_file: str = 'autosubmit.lock') -> None:
+def delete_lock_file(base_path: Optional[Union[str, Path]] = None, lock_file: str = 'autosubmit.lock') -> None:
     """Delete lock file if it exists. Suppresses permission errors raised.
 
     :param base_path: Base path to locate the lock file. Defaults to the experiment ``tmp`` directory.
-    :type base_path: str
+    :type base_path: Optional[Union[str, Path]]
     :param lock_file: The name of the lock file. Defaults to ``autosubmit.lock``.
     :type lock_file: str
     :return: None
     """
+    if base_path is None:
+        base_path = Log.file_path
     with suppress(PermissionError):
         Path(base_path, lock_file).unlink(missing_ok=True)
 
 
-def exit_from_error(e: BaseException) -> int:
+def _get_lock_path(args: Optional[argparse.Namespace]) -> Optional[Path]:
+    if args and getattr(args, 'expid', None):
+        return BasicConfig.expid_tmp_dir(args.expid)
+    return None
+
+
+def _owns_lock(args: Optional[argparse.Namespace]) -> bool:
+    """Return True if the command in args is one that acquires the autosubmit lock."""
+    return bool(args and getattr(args, 'command', None) in LOCK_OWNING_COMMANDS)
+
+
+def exit_from_error(e: BaseException, lock_path: Optional[Union[str, Path]] = None) -> int:
     """Called by ``Autosubmit`` when an exception is raised during a command execution.
 
     Prints the exception in ``CRITICAL`` if is it an ``AutosubmitCritical`` or an
     ``AutosubmitError`` exception, including any trace attached to the exception.
 
-    Exceptions raised by ``porta-locker` library print a message informing the user
-    about the locked experiment. Other exceptions raised cause the lock to be deleted.
+    Exceptions raised by the ``portalocker`` library print a message informing the
+    user about the locked experiment. For non-lock errors, the lock file is removed
+    only when ``lock_path`` is provided (i.e. only when the failing command was
+    a lock-owning command).
 
-    After printing the exception, this function calls ``os._exit(1)``, which will
-    forcefully exit the executable running.
+    Returns the resulting error code so the caller can use it as the process exit status.
 
     :param e: The exception being raised.
     :type e: BaseException
-    :return: None
+    :param lock_path: Path to the experiment tmp directory holding the lock file.
+        ``None`` when the failing command does not own the lock; in that case no
+        deletion is attempted.
+    :type lock_path: Optional[Union[str, Path]]
+    :return: The error code corresponding to the exception.
+    :rtype: int
     """
     err_code = 1
     trace = traceback.format_exc()
@@ -72,10 +92,11 @@ def exit_from_error(e: BaseException) -> int:
     is_autosubmit_error = isinstance(e, (AutosubmitCritical, AutosubmitError))
 
     if isinstance(e, BaseLockException):
-        Log.warning('Another Autosubmit instance using the experiment\n. Stop other Autosubmit instances that are '
-                    'using the experiment or delete autosubmit.lock file located on the /tmp folder.')
-    else:
-        delete_lock_file()
+        lock_file = Path(lock_path, 'autosubmit.lock') if lock_path else 'autosubmit.lock'
+        Log.warning(f'Another Autosubmit instance is using the experiment.\n Stop the other instance(s) '
+                    f'or delete the lock file located at: {lock_file}')
+    elif lock_path is not None:
+        delete_lock_file(lock_path)
 
     if is_autosubmit_error:
         e: Union[AutosubmitError, AutosubmitCritical] = e
@@ -101,16 +122,16 @@ def main():
         return_value, args = Autosubmit.parse_args()
         if args:
             return_value = Autosubmit.run_command(args)
-        delete_lock_file()
+        if _owns_lock(args):
+            delete_lock_file(_get_lock_path(args))
     except BaseException as e:
-        delete_lock_file()
         command = "<no command provided>"
         expid = "<no expid provided>"
         version = "<no version found>"
         if args:
-            if 'command' in args and args.command:
+            if getattr(args, 'command', None):
                 command = f"<{args.command}>"
-            if 'expid' in args and args.expid:
+            if getattr(args, 'expid', None):
                 expid = f"<{args.expid}>"
                 with suppress(BaseException):
                     as_conf = AutosubmitConfig(args.expid)
@@ -118,7 +139,7 @@ def main():
                     version = f"{as_conf.experiment_data.get('CONFIG', {}).get('AUTOSUBMIT_VERSION', 'unknown')}"
         Log.error(f"Arguments provided: {str(args)}")
         Log.error(f"This is the experiment: {expid} which had an issue with the command: {command} and it is currently using the Autosubmit Version: {version}.")
-        return_value = exit_from_error(e)
+        return_value = exit_from_error(e, _get_lock_path(args) if _owns_lock(args) else None)
     # TODO: we need to define whether the function called here will return an int or bool
     if type(return_value) is bool:
         return_value = 0 if return_value else 1
--- a/autosubmit/scripts/autosubmit.py
+++ b/autosubmit/scripts/autosubmit.py
@@ -49,13 +49,26 @@ def delete_lock_file(base_path: Optional[Union[str, Path]] = None, lock_file: st
 
 
 def _get_lock_path(args: Optional[argparse.Namespace]) -> Optional[Path]:
+    """Return the experiment ``tmp`` directory for the expid in ``args``, or ``None`` if no expid.
+
+    :param args: Parsed CLI arguments.
+    :type args: Optional[argparse.Namespace]
+    :return: Path to the experiment ``tmp`` directory, or ``None``.
+    :rtype: Optional[Path]
+    """
     if args and getattr(args, 'expid', None):
         return BasicConfig.expid_tmp_dir(args.expid)
     return None
 
 
 def _owns_lock(args: Optional[argparse.Namespace]) -> bool:
-    """Return True if the command in args is one that acquires the autosubmit lock."""
+    """Return ``True`` if the command in ``args`` acquires the autosubmit lock.
+
+    :param args: Parsed CLI arguments.
+    :type args: Optional[argparse.Namespace]
+    :return: ``True`` if the command is in ``LOCK_OWNING_COMMANDS``.
+    :rtype: bool
+    """
     return bool(args and getattr(args, 'command', None) in LOCK_OWNING_COMMANDS)
 
 
--- a/autosubmit/scripts/delete.py
+++ b/autosubmit/scripts/delete.py
@@ -79,6 +79,7 @@ def args_parser() -> ArgumentParser:
     group=CommandGroup.EXPERIMENT,
     options_type=DeleteOptions,
     validators=validate_expid,
+    lock=True,
 )
 def main(opts: DeleteOptions) -> int | bool | None:
     from autosubmit.experiment.manage import delete_experiment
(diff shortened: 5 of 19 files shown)


autosubmit/scripts/autosubmit.py
```diff
@@ -28,11 +28,12 @@
 from portalocker.exceptions import BaseLockException
 
 from autosubmit.autosubmit import Autosubmit  # noqa: E402
+from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.config.configcommon import AutosubmitConfig  # noqa: E402
 from autosubmit.log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
```
ntorqulu, line 32: `from os import _exit` imported but never called in the file

autosubmit/scripts/autosubmit.py
```diff
@@ -28,11 +28,12 @@
 from portalocker.exceptions import BaseLockException
```
ntorqulu, line 28: Function `exit_from_error` docstring says _exit is called, but it's not. Better correct this part to avoid confussions.

autosubmit/scripts/autosubmit.py
```diff
@@ -94,16 +97,22 @@ def exit_from_error(e: BaseException) -> int:
...
+
 # noinspection PyProtectedMember
 def main():
     args: Optional[argparse.Namespace] = None
     try:
         return_value, args = Autosubmit.parse_args()
         if args:
             return_value = Autosubmit.run_command(args)
-        delete_lock_file()
+        delete_lock_file(_get_lock_path(args))
```
ntorqulu: Lock could be deleted before being acquired. The delete is unconditional, there's no guarantee the lock was acquired here.

autosubmit/scripts/autosubmit.py
```diff
@@ -94,16 +97,22 @@ def exit_from_error(e: BaseException) -> int:
...
     except BaseException as e:
-        delete_lock_file()
+        delete_lock_file(_get_lock_path(args))
```
ntorqulu: Here there could be a potential double-deletion with mismatched paths.
In the first `delete_lock_file` (line 115)

- `except BaseException` catches all exceptions, including BaseLockException. It should not run for BaseLockException. BaseLockException is an exception raised when a process fails to get a lock file that is already held by another one. Imagine that:
1. process A executes run_command
2. process B already hold the lock
3. portalocker raises BaseLockException
4. processA deletes the lock even though the process never owned it.
So process A is deleting a lock file that belongs to process B, which is still running. Then imagine process C starts executing the same experiment. It could do it because there's no lock  protection now.

- And then in this same path, line 130 calls `exit_from_error', that calls again `delete_lock_file()` this time with another filepath.

autosubmit/scripts/autosubmit.py
```diff
@@ -94,16 +97,22 @@ def exit_from_error(e: BaseException) -> int:
...
         command = "<no command provided>"
         expid = "<no expid provided>"
         version = "<no version found>"
```
ntorqulu, line 109: Maybe it's outside of the scope, but I would access `args` attributes always the same.
In line 101:
```
if args and getattr(args, 'expid', None):
```
And in lines 120, 122:
```
if 'command' in args and args.command:
if 'expid' in args and args.expid:
```
I would prefer the first option, as it's more robust.

ntorqulu (commented):
Hi @rajeshsuraj, great job with the lock file handling! There's a few comments focused on when should the lock be deleted and which process should be the ones that delete it.

rajeshsuraj:
From your comments i have made some major changes , the unconditional delete in the except block has been removed, so there's no double deletion as only the success path in main() deletes, and the failure path is delegated to exit_from_error. The only failure where a process could reach the delete without having acquired the lock is BaseLockException (raised when another instance already holds it), and that's precisely the case the isinstance(e, BaseLockException) guard excludes and it takes the Log.warning branch instead. Any other failure means this process did acquire the lock, so deleting it there is correct. Net result: no path where a process deletes a lock it didn't acquire.

autosubmit/scripts/autosubmit.py
```diff
@@ -94,32 +98,36 @@ def exit_from_error(e: BaseException) -> int:
     return err_code
 
 
+def _get_lock_path(args: Optional[argparse.Namespace]) -> Optional[Path]:
+    if args and getattr(args, 'expid', None):
+        return Path(BasicConfig.LOCAL_ROOT_DIR) / args.expid / BasicConfig.LOCAL_TMP_DIR
+    return None
+
+
```
VindeeR: we already have the `expid_tmp_dir` function inside `BasicConfig`, you would only need to pass the `expid` as parameter so no need for this funcion

```suggestion
```

autosubmit/scripts/autosubmit.py
```diff
@@ -94,32 +98,36 @@ def exit_from_error(e: BaseException) -> int:
...
 # noinspection PyProtectedMember
 def main():
     args: Optional[argparse.Namespace] = None
     try:
         return_value, args = Autosubmit.parse_args()
         if args:
             return_value = Autosubmit.run_command(args)
-        delete_lock_file()
+        delete_lock_file(_get_lock_path(args))
```
VindeeR:
```suggestion
        path_to_lock = BasicConfig.expid_tmp_dir(args.expid) if args and getattr(args, 'expid', None) else None
        delete_lock_file(path_to_lock)
```

autosubmit/scripts/autosubmit.py
```diff
@@ -94,32 +98,36 @@ def exit_from_error(e: BaseException) -> int:
...
                 expid = f"<{args.expid}>"
                 with suppress(BaseException):
                     as_conf = AutosubmitConfig(args.expid)
                     as_conf.reload()
                     version = f"{as_conf.experiment_data.get('CONFIG', {}).get('AUTOSUBMIT_VERSION', 'unknown')}"
         Log.error(f"Arguments provided: {str(args)}")
         Log.error(f"This is the experiment: {expid} which had an issue with the command: {command} and it is currently using the Autosubmit Version: {version}.")
-        return_value = exit_from_error(e)
-    # TODO: we need to define whether the function called here will return an int or bool
+        return_value = exit_from_error(e, _get_lock_path(args))
```
VindeeR:
```suggestion
        return_value = exit_from_error(e, path_to_lock)
```

autosubmit/scripts/autosubmit.py
```diff
@@ -72,10 +76,10 @@ def exit_from_error(e: BaseException) -> int:
     is_autosubmit_error = isinstance(e, (AutosubmitCritical, AutosubmitError))
 
     if isinstance(e, BaseLockException):
-        Log.warning('Another Autosubmit instance using the experiment\n. Stop other Autosubmit instances that are '
-                    'using the experiment or delete autosubmit.lock file located on the /tmp folder.')
+        Log.warning('Another Autosubmit instance using the experiment.\n Stop other Autosubmit instances that are '
+                    'using the experiment or delete autosubmit.lock file located on the <expid>/tmp folder.')
```
VindeeR: maybe we can pass the `expid` here as a parameter

VindeeR (changes requested):
I just have a few comments on you changes, but good job on them 🙌 

On top of those changes, it would be great if the headers had at least a summary of the feature and problem they are fixing, this way during meeting it can be easier to track, understand and remember what it is about.

Thanks for the good work 😄

autosubmit/scripts/autosubmit.py
rajeshsuraj: as discussed the function is required here because it also guards the no-expid commands (configure, readme, etc.) where args.expid is absent and returns None in that case.

autosubmit/scripts/autosubmit.py
```diff
@@ -94,32 +99,36 @@ def exit_from_error(e: BaseException) -> int:
...
+
 # noinspection PyProtectedMember
 def main():
     args: Optional[argparse.Namespace] = None
     try:
         return_value, args = Autosubmit.parse_args()
         if args:
             return_value = Autosubmit.run_command(args)
-        delete_lock_file()
+        delete_lock_file(_get_lock_path(args))
```
ntorqulu: Hi @rajeshsuraj! I still have the same error. 
Steps to reproduce it:
1. execute `autosubmit run <EXPID>` -> this will create the lock file
2. while the experiment is running, execute a command that does not require the lock. e.g. `autosubmit inspect <EXPID>`. This will delete the lock while the previous process is running. Should not happen. Only the process that creates the lock should be the responsible of releasing it.
3. while the experiment is running, execute `autosubmit run <EXPID> -> this will create another lock file without blocking me from running the experiment while the other process is still running the same experiment...

autosubmit/scripts/autosubmit.py
```diff
@@ -94,32 +99,36 @@ def exit_from_error(e: BaseException) -> int:
...
+            if getattr(args, 'expid', None):
                 expid = f"<{args.expid}>"
                 with suppress(BaseException):
                     as_conf = AutosubmitConfig(args.expid)
                     as_conf.reload()
                     version = f"{as_conf.experiment_data.get('CONFIG', {}).get('AUTOSUBMIT_VERSION', 'unknown')}"
         Log.error(f"Arguments provided: {str(args)}")
         Log.error(f"This is the experiment: {expid} which had an issue with the command: {command} and it is currently using the Autosubmit Version: {version}.")
-        return_value = exit_from_error(e)
-    # TODO: we need to define whether the function called here will return an int or bool
```
ntorqulu, line 122: Better keep the TODO comment if it has not been solved.

autosubmit/scripts/autosubmit.py
VindeeR: In this case I believe it would be better to only call it when there is an expid and if there's not you use None.

I'll give an example on the other piece of code

autosubmit/scripts/autosubmit.py
VindeeR: I changed my suggestion here to something I believe should work for what you commented

autosubmit/scripts/autosubmit.py
VindeeR: I tried to run the steps as suggested and I got the following message

Inspect

```
Starting to generate cmd scripts
[WARNING] There is a .lock file and not -f, generating only all unsubmitted cmd scripts
[WARNING] Generating the auxiliary job_list used for the -cw flag.
[WARNING] Aux Job_list was generated successfully
Calculating possible ready jobs for TEST_SLURM
```

second run

```
portalocker.exceptions.LockException: [Errno 11] Resource temporarily unavailable                                                                                                                                          
                                                                                                                                                                                                                           
[WARNING] Another Autosubmit instance is using the experiment.
 Stop the other instance(s) or delete the lock file located at: /home/edoria/autosubmit/a00c/tmp/autosubmit.lock                                                                                                           
```

this message happens both on this branch and on master, maybe there's something else or I'm executing too fast.
I\m not sure how to replicate it.

VindeeR (changes requested):
Made a few more pointers, and I would like you to try replicate the behavior @ntorqulu manage, if that is happening I agree with her where it should be fixed.

autosubmit/scripts/autosubmit.py
ntorqulu: Yes, the problem is in the order and which commands are executed. The problem we're trying to avoid is having two processes that require a lock file run at the same time. This should not be possible, but currently can be done executing the following commands:
Initially you should have two terminals open in order to execute commands in parallel. We will call them terminal_A and terminal_B.
First we'll execute a command that generates a lock file, for instance `autosubmit run`. This process will create a lock file that in theory only himself should be able to release.

1. In terminal_A execute:
`autosubmit run <EXPID>`

While the experiment is running, we will execute in terminal_B another command that does not require a lock, for instance `autosubmit inspect`. The problem is that currently this process do not create a lock (correct) but it releases the previous lock created by `autosubmit run` that is running in terminal_A (incorrect).

2. In terminal_B execute:
`autosubmit inspect <EXPID>`
while in terminal_A autosubmit run is still running. This will delete the lock file created by the first process...

Now we have a process in terminal_A that should require a lock file but that does not have one. We are free to execute again `autosubmit run` without anything blocking us, which can cause a lot of errors (rewriting files from old executions, problems with concurrency, etc...).
While the first process with `autosubmit run <EXPID>` is running in terminal_A.

3. In terminal_B execute:
`autosubmit run <EXPID>`

We have the same experiment being ran by two different process at the same time... This should not happen. Lock files should only be released by the process that creates the lock.

ntorqulu:
Hi @rajeshsuraj, I think you made a push without `--force-with-lease` so you have brought all the changes from `master`. As it's more difficult to review, would be possible to just keep your changes in this branch? 
Thanks!

rajeshsuraj:
Sorry @ntorqulu , Actually i thinking to drop all previous changes start from a new because the current changes i made dont align with whats intended from the discussion from @dbeltrankyl

ntorqulu:
Sure, no problem! Then better delete the current branch and unlink it from the issue to avoid confusion.

rajeshsuraj:
> Sure, no problem! Then better delete the current branch and unlink it from the issue to avoid confusion.

I am using the same branch but dropped all the previous commits that's what i meant in the previous comment.

autosubmit/scripts/autosubmit.py
```diff
@@ -20,46 +20,66 @@
...
 
 from portalocker.exceptions import BaseLockException
 
 from autosubmit.autosubmit import Autosubmit  # noqa: E402
+from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.config.configcommon import AutosubmitConfig  # noqa: E402
 from autosubmit.log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
 
+# Extend this set if a new lock-acquiring command is added.
+LOCK_OWNING_COMMANDS = frozenset({'run', 'create', 'recovery', 'setstatus', 'pklfix'})
```
ntorqulu, line 34: Question: delete should be a lock owning command?

autosubmit/scripts/autosubmit.py
```diff
@@ -20,46 +20,66 @@
...
     :return: None
     """
+    if base_path is None:
+        base_path = Log.file_path
     with suppress(PermissionError):
         Path(base_path, lock_file).unlink(missing_ok=True)
 
 
-def exit_from_error(e: BaseException) -> int:
+def _get_lock_path(args: Optional[argparse.Namespace]) -> Optional[Path]:
```
ntorqulu: Could you add docstrings to functions `_get_lock_path` and `_own_lock` with the used format?

autosubmit/scripts/autosubmit.py
rajeshsuraj: Screenshot from 2026-06-09 15-36-25

So in essence yes , delete should be a lock owning command with archive and unarchive but their sections of code do not have the lock within them at the moment.

ntorqulu: I see, thanks! As it is outside the scope of this pr, I would open an issue for the `archive`, `unarchive` and `delete` commands then, to use the lock too. I'm not sure if there's any already opened.

rajeshsuraj: Ideally that should be the case, but i am waiting for autosubmit to be broken and when the core cli commands are refactored we can add the locks to the commands required but for the current issue i think the pr covers it well. what do you think ?

autosubmit/scripts/autosubmit.py
```diff
@@ -20,46 +20,79 @@
...
+    """Return the experiment ``tmp`` directory for the expid in ``args``, or ``None`` if no expid.
+
+    :param args: Parsed CLI arguments.
+    :type args: Optional[argparse.Namespace]
+    :return: Path to the experiment ``tmp`` directory, or ``None``.
+    :rtype: Optional[Path]
+    """
+    if args and getattr(args, 'expid', None):
+        return BasicConfig.expid_tmp_dir(args.expid)
+    return None
```
ntorqulu: Is there a case where `_get_lock_path` is called with a command that does not have `expid`? Could be easier to just return the `BasicConfig.expid_tmp_dir` always, as it is safeguarded by `_owns_lock`?
I saw this because this line and:
```
if base_path is None:
base_path  = Log.file_path
```
are not covered with tests, I'm not sure if it's possible to reach there.

ntorqulu: And `Log.file_path` is filled in `_init_logs`, wiith the content of  `BasicConfig.expid_tmp_dir`. I would always assign the `BasicConfig.expid_tmp_dir` as it's filled before.

rajeshsuraj: i think that piece of code was assumed in the intial versions of the code where we give path later it evolved to building it and passing it so, now it makes sense that it is unreachable now , good observation the code can be cleaned up here.

autosubmit/scripts/autosubmit.py
ntorqulu: I think so, but I would write it down somewhere that these commands need the lock file too. Either in a new issue or in an existing one (if it exists) about this refactor. This way we don't forget in the future.

autosubmit/scripts/autosubmit.py
```diff
@@ -20,46 +20,64 @@
...
 
 from portalocker.exceptions import BaseLockException
 
 from autosubmit.autosubmit import Autosubmit  # noqa: E402
+from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.config.configcommon import AutosubmitConfig  # noqa: E402
 from autosubmit.log.log import Log, AutosubmitCritical, AutosubmitError  # noqa: E402
 
+# Extend this set if a new lock-acquiring command is added.
+LOCK_OWNING_COMMANDS = frozenset({'run', 'create', 'recovery', 'setstatus', 'pklfix'})
```
ntorqulu, line 34: Hi @rajeshsuraj! Yesterday I was thinking about this list... I think we should not introduce a manually-maintained list here, it could cause compatibility issues.
What I mean by that:
We will have to modify commands `archive`, `unarchive`, `delete` in `autosubmit.py`. This list creates a dependency with future changes, it will break if a new lock-acquiring command is added without updating.
I would avoid using lists for the lock-acquiring commands to avoid maintainability issues.

What I would do (to discuss) is:
- remove `LOCK_OWNING_COMMANDS`
- remove `_owns_lock`
- replace them by `_lock_file_exists`, which checks if `autosubmit.lock` is present on the dir after `run_command` returns. This would eliminate the dependency introduced by the list.

What do you think?

autosubmit/scripts/autosubmit.py
```diff
@@ -101,24 +119,25 @@ def main():
         return_value, args = Autosubmit.parse_args()
         if args:
             return_value = Autosubmit.run_command(args)
-        delete_lock_file()
+        if _owns_lock(args):
```
ntorqulu, line 122: If _lock_file_exists

autosubmit/scripts/autosubmit.py
```diff
@@ -101,24 +119,25 @@ def main():
...
+            if getattr(args, 'expid', None):
                 expid = f"<{args.expid}>"
                 with suppress(BaseException):
                     as_conf = AutosubmitConfig(args.expid)
                     as_conf.reload()
                     version = f"{as_conf.experiment_data.get('CONFIG', {}).get('AUTOSUBMIT_VERSION', 'unknown')}"
         Log.error(f"Arguments provided: {str(args)}")
         Log.error(f"This is the experiment: {expid} which had an issue with the command: {command} and it is currently using the Autosubmit Version: {version}.")
-        return_value = exit_from_error(e)
+        lock_path = BasicConfig.expid_tmp_dir(args.expid) if _owns_lock(args) else None
```
ntorqulu, line 139: if _lock_file_exists

ntorqulu (commented):
Hi @rajeshsuraj, I'm not sure about keeping a list with the lock-acquiring commands. Maybe is better to check if the lock file exists? What do you think?

autosubmit/scripts/autosubmit.py
rajeshsuraj: the problem with this approach is that if we check lock_file_exists then consider this scenario that run expid on terminal A also in terminal B we run expid inspect the nature of the code when inspect finishes will delete the lockfile and then in terminal B you run expid concurrently along with the terminal A now we are back at the original bug that we faced, the issue is the with check lock_file can not distinguish which process is holding it

ntorqulu: True! Still I think we should not have the list hardcoded.
Maybe we can try with `TemporaryFileLock`. It automatically deletes the lock file after it unlocks.
https://github.com/wolph/portalocker/issues/107

rajeshsuraj: I agree that keeping a hard coded list is not ideal . TemporaryFileLock can be alternative but the underlying issue that the lock architecture is only present for 4 commands and other commands lack it and the existing solution is already a work around. I think what should be done is to rework the other lock unaware commands and the 4 lock aware commands to be consistent and remove the lockfile deletion from scripts/autosubmit.py ? what do you think ?

ntorqulu: For this pr I would:
- Every command that currently acquires the lock file, do it with `TemporaryFileLock` (the ones in the hardcoded list: run, create, recovery, setstatus, pklfix)
- For the commands that should also acquire the lock file but they are currently not doing it, we'll handle it in another pr (see issue #3055), and we'll follow the same `TemporaryFIleLock` implementation as here. This way this pr will not handle 2 diff things, just the correctly release + delete of the lock file for the current commands that use it.
- In `scripts/autosubmit.py` only handle the stale lock recovery.

rajeshsuraj: Okay , will move forward with this

kinow: Just about this one,

> I think we should not introduce a manually-maintained list here, it could cause compatibility issues.

@ntorqulu +1. 

My idea was to first split the enormous `autosubmit.py` file (which I have been trying for 4 years 😅 ). Then, I'll move the commands to `autosubmit.scripts` (e.g., `autosubmit.scripts.create`, `autosubmit.scripts.set_status`, etc.).

One script py module per Autosubmit subcommand. And every script file sharing an API built on top of argparse + decorators.

The idea of the decorators is to avoid these `if command_uses_lock`, `if command_is_expidless`, `if command_takes_filters`, etc., and instead have something like (pseudocode, not tested):

```py
# file: autosubmit.scripts.create.py

def parse_args():
    # The argparse parameters used by this subcommand
    ...

@as_command(
  expidless=False,
  lock=True
)
def run():
    # the normal code...
    ...
```

A class structure is probably not needed (can be structured + validated using protocols for instance). That'd make it easier to work on issues like #1345, #1227, and maybe even validation (the validation could be built into that decorator).

When that's done, knowing what's using a lock or not would be more declarative and less imperative, like here. Not sure if we can come up with another solution until that's fixed. But just to give y'all some context.

ntorqulu: Great, thanks for the explanation @kinow! I like the idea of the decorators for the lock and the validation. I'll keep it in mind while refactoring the validation.
As there are commands that validate the config yaml files and raise errors, other that just validate them and log warnings and others that directly do not validate the files I would include a:
```
@as_command(
    validate="offline", # if command needs to validate the config files; options: False, "offline", "runtime"
```
But I'll keep this discussion in the validation pr.


About the lock file not being cleaned out, do you think is better to wait for the decorators and the refactor of the commands or implement a temporary solution first?

rajeshsuraj: thanks @kinow @ntorqulu for the comments, so for now i would put this pr on hold.

kinow: Not a problem. Thanks Suraj!

dbeltrankyl:
@rajeshsuraj is this one stall? Should we bump it to 4.2.1? Or will it be resumed soon? thanks!

rajeshsuraj:
Hi @dbeltrankyl , this will be resumed soon as i was waiting for bruno's PR #3171  to be merged.

autosubmit/scripts/_cli_function.py
```diff
@@ -44,12 +47,33 @@
...
+    for one second, then raise ``AlreadyLocked``, which ``exit_from_error`` reports.
+
+    :param expid: The experiment identifier.
+    :return: A context manager that holds the experiment lock.
+    """
+    return TemporaryFileLock(
+        str(BasicConfig.expid_lock_file(expid)),
+        timeout=1,
+        fail_when_locked=False
+    )
```
kinow, line 67: This `_cli_function.py` module was intentionally kept "lean", containing only the logic for the decorator. We should -- if/when possible -- keep it like this. So ideally the lock file would be created elsewhere.

autosubmit/scripts/_cli_function.py
```diff
@@ -44,12 +47,33 @@
...
+
+
 def cli_function(
     *,
     args_parser: Callable[[], ArgumentParser],
     options_type: type["AutosubmitOptions"],
     validators: "Validator[OptionsT] | list[Validator[OptionsT]] | None" = None,
     group: "CommandGroup" = CommandGroup.GENERAL,
+    lock: bool = False,
```
kinow, line 76: This should go to the options, like the options that define if a command supports multiple expids, or allows multiple users from accessing the experiments. This way we only receive options here, without having to keep adding new parameters to `@cli_function`.

rajeshsuraj: I have added it to Defaultoptions and removed it from cli_function and made a separate file under experiment/lock.py

autosubmit/scripts/_cli_function.py
rajeshsuraj: changed the style in which the lockfile was assigned from imperative to declarative the commands or to be commands just need to set acquires_lock and the lock is set to the command and removed either on success or failure.

autosubmit/experiment/lock.py
```diff
@@ -0,0 +1,46 @@
...
+
+def experiment_lock(expid: str) -> AbstractContextManager:
+    """Return the lock that stops two commands from modifying one experiment at once.
+
+    ``TemporaryFileLock`` deletes ``autosubmit.lock`` on release, including when the
+    command raises. From portalocker 4.0 the file is unlinked while the lock is still
+    held, so no other process can end up locking a stale copy of it.
+
+    ``timeout=1`` with ``fail_when_locked=False`` retries for one second, then raises
+    ``AlreadyLocked``, which ``exit_from_error`` reports to the user.
```
kinow, line 37: 👍 

Maybe worth adding that if the process is abruptly killed, the file is not deleted.

autosubmit/experiment/lock.py
```diff
@@ -0,0 +1,46 @@
...
+
+    :param expid: The experiment identifier.
+    :return: A context manager that holds the experiment lock.
+    """
+    return TemporaryFileLock(
+        str(BasicConfig.expid_lock_file(expid)),
+        timeout=1,
+        fail_when_locked=False,
+    )
```
kinow: Missing EOL at the end... useful/needed for some old linux utilities (every line is a bunch of chars, and a line termination like \n), but also a bother when an editor automatically includes, but another dev's editor doesn't, and you end up with unwanted changes in some files in GitHub pull requests etc 👍

autosubmit/scripts/_args.py
```diff
@@ -72,6 +70,8 @@ class DefaultOptions(AutosubmitOptions):
     """Whether or not the sub-command accepts multiple expids. Default is no."""
     accepts_other_users = False
     """Whether or not the sub-command accepts other users. Default is no."""
+    acquires_lock = False
+    """Whether or not the sub-command holds the experiment lock while it runs. Default is no."""
```
kinow, line 74: 👏

autosubmit/scripts/archive.py
```diff
@@ -36,6 +36,8 @@
 class ArchiveOptions(ExpidOptions):
     """Options for the archive command."""
 
+    acquires_lock = True
+
```
kinow: Nitpick, but I think in other subcommands the arguments are declared at the top... and the extra options like acquires_lock are at the end, separated 😬 e.g., https://github.com/BSC-ES/autosubmit/blob/9b1d85f5e243dd7f0c08d68a192203b8411570ae/autosubmit/scripts/stop.py#L61-L70

:memory:
kinow, line 1: This file must not be committed. It's generated accidentally by some unit tests, I believe. That's being fixed in other(s) pull requests, that apparently have not been merged yet 😅 Sorry. Please delete it here.

test/unit/scripts/test_entry_points.py
```diff
@@ -287,3 +287,14 @@ def test_entry_points_have_short_names():
             # first string after the three double-quotes in a docstring) must be the
             # sub-command short name.
             assert doc.title == cmd
+
+
+_LOCKED_COMMANDS = {"create", "recovery", "run", "setstatus","archive","delete"}
```
kinow: That can be done dynamically?

test/integration/test_utils/misc.py
```diff
@@ -61,6 +61,9 @@ def wait_locker(file_lock: Path, expect_locked: bool, timeout: int, interval=0.0
                                f"({timeout}s timeout, {elapsed:.2f}s elapsed)")
 
         if not file_lock.exists():
+            if not expect_locked:
+                return
+            sleep(interval)
```
kinow, line 66: I was going to ask why the sleep here, but looking at the code, I think we forgot to use `interval` before? 😅 thanks for fixing that.

test/integration/scripts/test_experiment_lock.py
```diff
@@ -0,0 +1,73 @@
...
+
+    assert seen == [True]
+    assert not _lock_file(exp).exists()
+
+
+def test_lock_removed_when_command_raises(autosubmit_exp, mocker):
+    exp = autosubmit_exp()
+    mocker.patch(
+        "autosubmit.experiment.manage.create",
+        side_effect=AutosubmitCritical("boom", 7000),
```
kinow, line 57: 💥

autosubmit/workflow/manage.py
kinow, line 1: I will check this with the IDE later. So far I'm only using the GitHub UI.

Am I understanding correct here that `workflow/manage.py` included logic to handle the lock, so you only removed that code from this file, and the changes here are just for indentation?

kinow (commented):
Looking great, going in the right direction. Thanks @rajeshsuraj !

:memory:
rajeshsuraj: sorry i didn't check , will delete it

autosubmit/workflow/manage.py
rajeshsuraj: yes , alll the manage files were indentation changes with the with Lock as fh and fh.flush removed.

rajeshsuraj: Also when reviewing you can add ?w=1 this to pr url and avoid the indentation changes like this https://github.com/BSC-ES/autosubmit/pull/3035/changes?w=1

:memory:
kinow: Not your fault :) thanks for deleting it!

autosubmit/scripts/archive.py
rajeshsuraj: no issues will move subcommand arguments to match the styling everywhere else.

test/unit/scripts/test_entry_points.py
rajeshsuraj: when i was writing tests i just added that quickly to test lock declarations , now in the latest changed it to be dynamic.

autosubmit/scripts/archive.py
```diff
@@ -35,11 +35,13 @@
 
 class ArchiveOptions(ExpidOptions):
     """Options for the archive command."""
-
+
```
kinow: Extra blank spaces! 😬

autosubmit/scripts/delete.py
```diff
@@ -43,7 +43,7 @@ class DeleteOptions(AutosubmitOptions):
 
     (OOP class hierarchy here is not too important as it is internal code.)
     """
-
+
```
kinow: Extra blank spaces.... 😬 😬

autosubmit/scripts/recovery.py
```diff
@@ -57,7 +57,7 @@
 
 class RecoveryOptions(ExpidOptions):
     """Options for the recovery command."""
-
+
```
kinow: 😬 😬 😬

autosubmit/scripts/setstatus.py
```diff
@@ -55,7 +55,7 @@
 
 class SetstatusOptions(ExpidOptions):
     """Options for the setstatus command."""
-
+
```
kinow: 😬 😬 😬 😬

autosubmit/workflow/manage.py
```diff
@@ -605,294 +604,270 @@ def run(
...
-            except AutosubmitCritical as e:  # Critical errors can't be recovered. Failed configuration or autosubmit error
-                raise AutosubmitCritical(e.message, e.code, e.trace)
-
-        Log.result("No more jobs to run.")
-        # search hint - finished run
-        Log.info("Waiting for all logs to be updated")
-        for p in platforms_to_test:
-            p.clean_log_recovery_process()
-        _process_historical_data_iteration(job_list, job_changes_tracker, expid)
-
```
kinow, line 873: Any reason for deleting the blank lines that were here before? (Hey, thanks for the `?w=1` trick!)

test/unit/scripts/test_entry_points.py
```diff
@@ -287,3 +287,11 @@ def test_entry_points_have_short_names():
             # first string after the three double-quotes in a docstring) must be the
             # sub-command short name.
             assert doc.title == cmd
+
+
+@pytest.mark.parametrize("cmd", sorted(get_commands()))
+def test_command_lock_declaration(cmd):
+    """"Every sub-command declares whether it holds the experiment lock."""
+    options_type = get_commands()[cmd].load().options_type
+    assert isinstance(getattr(options_type, "acquires_lock", False), bool)
```
kinow: >     """"Every sub-command declares whether it holds the experiment lock."""

I think the test is not testing what it's saying here.

If you do a `getattr` with default `False`, then there's no way an `Option` without `acquires_lock` is not a bool?

I think it's testing every command **that** declares whether it holds or not, does so using a variable that's a boolean. But I think testing what the original docstring text says would be better.

autosubmit/workflow/manage.py
rajeshsuraj: It was accidental when i was multiline selecting with alt and shift in vscode it also deletes the selector at the beginning of the line hence effectively removing the line. will clean this up

kinow: Perfect!

test/unit/scripts/test_entry_points.py
rajeshsuraj: Ah i understand the test becomes a closed loop , i have reverted to my original test

test/unit/scripts/test_entry_points.py
```diff
@@ -287,3 +289,11 @@ def test_entry_points_have_short_names():
             # first string after the three double-quotes in a docstring) must be the
             # sub-command short name.
             assert doc.title == cmd
+
+
+@pytest.mark.parametrize("cmd", sorted(get_commands()))
+def test_command_lock_declaration(cmd):
+    """Only commands that change the experiment state hold the experiment lock."""
+    options_type = get_commands()[cmd].load().options_type
+    assert getattr(options_type, "acquires_lock", False) == (cmd in _LOCKED_COMMANDS)
```
kinow, line 298: This test still irks me for some reason... I can see that the `get_commands` is a bit expensive, so we could probably move it to a fixture.

But it's the hard-coded list of locked commands that's making me want to find an alternative for that... let me think about it. But I think I've submitted all my other comments and you fixed everything. So we should be able to merge it tomorrow or Friday tops.

Thanks Suraj!!!

kinow: I think I was overthinking. This looks like the best compromise. Thanks for trying things out!!!

kinow (approved):
+1!

Just remember to rebase, squash, merge.

Also... before merging, please add a changelog so users are informed that this is now fixed, and the lock file will be deleted (something short, with the link to the PR or issue -- there are a few examples to follow).

Thank you Suraj!!!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ several bug fixes and enhancements to improve the overall user experience.
 
 **Bug fixes:**
 
+- Fix `autosubmit.lock` not being deleted after `create` and `run`; `archive` and `delete` now acquire it too #3033 #3055
```
kinow, line 11: ❤️ !

kinow (approved):
Thanks!

PR #3037, opened 2026-05-28
[Enhancement] - args in command log

### LINKED ISSUE #2795: include useful information such as as version + expid + command used + args in every command log
From @dmcclainbsc here https://github.com/BSC-ES/autosubmit/discussions/2794

### Problem

We do not have a convention/standardized way to log the arguments used. Upon errors we dump the whole argparse data, but that's not always available.

Operators/workflow devs might need this info error or not, and it shouldn't be hard for us to include it.

### Proposed Solution

Something similar to 

```
Autosubmit version: 4.1.16.dev123
Expid: a008
Command: /full/path/to/autosubmit
Subcommand: run
Args: -rom fc 0 a008
User: dm...
...
...
...
```

right near the top of each command log.

VindeeR:
Closes #2795 

Adding the descriptor of all variables within a command, used or not, to a log with the goal of facilitating the debugging

**Check List**

- [x] Documentation updated.


autosubmit/autosubmit.py
```diff
@@ -912,6 +912,9 @@ def run_command(args: argparse.Namespace):
 
         if hasattr(args, 'expid'):
             expid = args.expid
+
+        describe_command_details(args)
```
ntorqulu: Move call after `_init_logs`, to write the input commands into the `/tmp/ASLOG/` files. Currently it only writes the commands in `autosubmit/logs`.

autosubmit/helpers/utils.py
```diff
@@ -329,3 +327,18 @@ def user_yes_no_query(question: str) -> bool:
             sys.stdout.write('Please respond with \'y\' or \'n\'.\n')
         except Exception as e:
             raise AutosubmitCritical("No input detected, the experiment will not be erased.", 7011, str(e))
+
+def describe_command_details(args) -> None:
```
ntorqulu, line 517: To discuss, why do we need to create another `details.log` file outside the experiment folder? Why we don't store the commands executed into the `<EXPID/tmp/ASLOGS` folder?
I think that if the purpose is quickly retrieve commands executed, having them stored by experiment is easier for devs to find.

autosubmit/helpers/utils.py
```diff
@@ -329,3 +327,18 @@ def user_yes_no_query(question: str) -> bool:
...
+        str(
+            Path(
+                BasicConfig.GLOBAL_LOG_DIR,
+                args.command + "_details.log",
+            )
+        ),
+        "out",
+        4020,
+    )
+    Log.info("Full range of command descriptors will be printed bellow")
```
ntorqulu: Typo: below.
Is it necessary to add this line? Which should be the header standardized structure?
I think we should define a template to use for easy-search.
Currently:
```
2026-05-28 17:23:06,074 VERSION: False
2026-05-28 17:23:06,074 LOGFILE: DEBUG
2026-05-28 17:23:06,074 LOGCONSOLE: WARNING
2026-05-28 17:23:06,074 COMMAND: create
2026-05-28 17:23:06,074 EXPID: a05r
2026-05-28 17:23:06,074 NOPLOT: True
2026-05-28 17:23:06,074 HIDE: False
2026-05-28 17:23:06,074 DETAIL: False
2026-05-28 17:23:06,074 OUTPUT: None
2026-05-28 17:23:06,074 GROUP_BY: None
2026-05-28 17:23:06,074 EXPAND: None
2026-05-28 17:23:06,074 EXPAND_STATUS: None
2026-05-28 17:23:06,074 NOTRANSITIVE: False
2026-05-28 17:23:06,074 CHECK_WRAPPER: False
2026-05-28 17:23:06,074 UPDATE_VERSION: False
2026-05-28 17:23:06,074 PROFILE: False
2026-05-28 17:23:06,074 FORCE: False
```
I'm not sure if this is the easiest way to facilitate quick-retrieval.
What I mean by that: In the discussion they where talking about if a user had executed the `-rom` flag.
What a dev would like to do is quickly filter among the experiment logs for this flag, `-rom`.
But the current implementation logs the internal argument names, not the user flags. For instance, `NOPLOT` is known for users as `-np`. They don't know how AS encodes them internally.

The template proposed in the discussion seems like a good approach:
```
Autosubmit version: 4.1.16.dev123
Expid: a008
Command: /full/path/to/autosubmit
Subcommand: run
Args: -rom fc 0 a008
User: dm...
```
Where in args you can concatenate all the flags used for easy-search. I don't know if more information should be dumped.

ntorqulu (commented):
Hi @VindeeR! I like the idea, but I'm not sure about what should be logged. What do you think?

autosubmit/helpers/utils.py
VindeeR: Not all commands will be used with an `expid`, which might lead to a crash.

So my idea was to put the logs on the lowest level possible, the `autosubmit` folder, and there's a file generated on that path already for commands that have no `expid`.

So for the logs I believe that a standard directory for the same information might be better than scattering around in different places, still it is easily changed for the `expid` for the commands that have it and generate here whenever there's no `expid`.

VindeeR:
> Hi @VindeeR! I like the idea, but I'm not sure about what should be logged. What do you think?

Based on what I understood by the issue, the example @kinow gave, I believe we should be logging all the information inside the `args`, that is generated after processing the command at `parse_args`, but there might be more things.

Do you have any idea if I'm missing something, worst case we can wait for Bruno?

autosubmit/helpers/utils.py
VindeeR: I understood that was supposed to print everything good to know they had a template in mind
The problem I faced is not always filled which could crash, or some commands don't have some of this values.
The first line I used in order to make it easy to search, but it can be changed

autosubmit/helpers/utils.py
LuiggiTenorioK: I am also not sure about why we need to create an additional file for each command executed.

I think the goal of the issue was to improve the current command log with a header, and each command already has a saved log somewhere. So, that new header should already be in a log, without the need to create a new file.

Also, creating more files will go against our goal to reduce them, which is something Climate DT maintainers are asking us to do heavily.

autosubmit/helpers/utils.py
LuiggiTenorioK: Agree that the current format is noisy.

Probably, an improvement will be to print it at once instead of one by one, so we don't get the timestamp for each line. Another suggestion could be to not include parameters that are not set by the user at INFO log level, so we can easily assume that they are defaults. Those can be at DEBUG level instead.

Finally, I find it really useful to have the `Command`, `Subcommand`, and `Args` fields like in the example header. You can easily have them by using `sys.argv`.

VindeeR:
thanks for the feedback @ntorqulu @LuiggiTenorioK  I haven't implemented the test yet, but I changed to add the specific values mentioned, and the use the same file as suggested by you :raised_hands:

VindeeR:
@ntorqulu @LuiggiTenorioK changes and tests applied :raised_hands:

autosubmit/helpers/utils.py
```diff
@@ -329,3 +327,26 @@ def user_yes_no_query(question: str) -> bool:
...
             raise AutosubmitCritical("No input detected, the experiment will not be erased.", 7011, str(e))
+
+def describe_command_details(args) -> None:
+    keys = ['FLAGS', 'EXPID', 'SUBCOMMAND', 'COMMAND']
+    flags = ''
+    descriptor = '\n'
+    for val in sys.argv:
+        descriptor += f'{keys.pop()} : {val}\n'
+        if keys[-1] == "FLAGS":
+            break
```
LuiggiTenorioK: This brings the wrong result when flags are before the expid. For example, if I do `autosubmit monitor -cw a003`, then this is printed:

```
COMMAND : /home/ltenorio/miniconda3/envs/autosubmit_dev/bin/autosubmit
SUBCOMMAND : monitor
EXPID : -cw
FLAGS : a003 
USER: ltenorio
```

I think for expid you can just use `args.expid`, and not show that if none.

Probably the current header we want is too hard to get. I think this header could facilitate your work:

```bash 
CLI_PATH: /home/ltenorio/miniconda3/envs/autosubmit_dev/bin/autosubmit
COMMAND: autosubmit monitor -cw a0003  # You can append the autosubmit in the beginning and remove the full path to be more readable
EXPID: a003
USER: ltenorio
```

This will keep it simple and achieve most of the value Bruno was looking for. Wdyt?

autosubmit/helpers/utils.py
```diff
@@ -465,3 +465,26 @@ def _parse_stat_file(path: Path) -> tuple[int, int]:
...
+    flags += f'{keys[0]} : '
+    for val in sys.argv[3:]:
+        flags += f'{val} '
+    descriptor += flags+'\n'
```
LuiggiTenorioK: Still getting the wrong output with the EXPID and FLAGS when running:

```sh
$ autosubmit monitor -cw a003

...
COMMAND : /home/ltenorio/miniconda3/envs/autosubmit_dev/bin/autosubmit
SUBCOMMAND : monitor
EXPID : -cw
FLAGS : a003 
USER: ltenorio
...
```

LuiggiTenorioK (changes requested):

autosubmit/helpers/utils.py
VindeeR: :facepalm: sorry I forgot to adjust this one

autosubmit/helpers/utils.py
```diff
@@ -465,3 +465,25 @@ def _parse_stat_file(path: Path) -> tuple[int, int]:
...
+
+def describe_command_details(args) -> None:
+    descriptor = '\n'
+    command = ''
+    for val in sys.argv:
+        if 'autosubmit' in val:
+            descriptor += f'CLI_PATH : {val}\n'
+            command += 'autosubmit '
+        else:
+            command += val+' '
```
LuiggiTenorioK: Probably it will be better to evaluate if there is an "autosubmit" substring just in the first value of `sys.argv` (`sys.argv[0]`). It could be that another value can have the  "autosubmit" substring as well, and it will be replaced.

For example:

```
$ autosubmit expid -d "some autosubmit experiment"

CLI_PATH : /home/ltenorio/miniconda3/envs/autosubmit_dev/bin/autosubmit
CLI_PATH : some autosubmit experiment
COMMAND : autosubmit expid -d autosubmit 

```

LuiggiTenorioK (changes requested):

autosubmit/helpers/utils.py
LuiggiTenorioK: Also, you will lose the quotes in `"some autosubmit experiment"`

LuiggiTenorioK: This is my best attempt to solve it in 5 minutes :sweat_smile: 

```suggestion
    cli_path = ""
    cli_args = []
    if "autosubmit" in sys.argv[0]:
        cli_path = sys.argv[0]
        cli_args = ["autosubmit"] + sys.argv[1:]
    else:
        cli_args = sys.argv

    command = ' '.join(shlex.quote(arg) for arg in cli_args)

    if cli_path:
        descriptor = f'CLI_PATH : {cli_path}\n'

```

VindeeR: changes applied thanks for the improvement, it is very tricky to validate those logs.

autosubmit/helpers/utils.py
```diff
@@ -464,4 +516,36 @@ def _parse_stat_file(path: Path) -> tuple[int, int]:
...
+
+
+def describe_command_details(args) -> None:
+    descriptor = "\n"
+    if "autosubmit" in sys.argv[0]:
+        descriptor += f"CLI_PATH : {sys.argv[0]}\n"
+        cli_args = ["autosubmit"] + sys.argv[1:]
+    else:
+        cli_args = sys.argv
+    command = " ".join(arg for arg in cli_args)
```
LuiggiTenorioK:
```suggestion
    command = " ".join(shlex.quote(arg) for arg in cli_args)
```

I still think it is worth having `shlex.quote` to handle quoted args, like:

```bash
# Without shlex.quote
$ autosubmit updatedescrip aa6g "some test autosubmit"
COMMAND : autosubmit updatedescrip aa6g some test autosubmit

# With shlex.quote
$ autosubmit updatedescrip aa6g "some test autosubmit"
COMMAND : autosubmit updatedescrip aa6g 'some test autosubmit'
```

kinow (commented):
Do we need to upper case everything?

The autosubmit message & version is more "calm". The rest of the log already has entries that are not upper-case, and this way users won't have to `grep -i` if they want to search:

```
(venv) kinow@ranma:~/Development/python/workspace/autosubmit$ autosubmit describe
Autosubmit is running with 4.1.17

CLI_PATH : /home/kinow/Development/python/workspace/autosubmit/venv/bin/autosubmit
COMMAND : autosubmit describe

YAML configuration loaded:
  /home/kinow/autosubmit/a000/proj/git_project/conf/bootstrap
  /home/kinow/autosubmit/a000/proj/git_project/conf/bootstrap/include.yml
  /home/kinow/autosubmit/a000/proj/git_project/conf
  /home/kinow/autosubmit/a000/proj/git_project/conf/experiment.yml
  /home/kinow/autosubmit/a000/proj/git_project/conf/jobs.yml
  /home/kinow/autosubmit/a000/proj/git_project/conf/mhm.yml
  /home/kinow/autosubmit/a000/proj/git_project/conf/platforms.yml
  /home/kinow/autosubmit/a000/proj/git_project/conf/rocrate.yaml
  /home/kinow/.config/autosubmit/platforms.yml
  /home/kinow/autosubmit/a000/conf/as_misc.yml
Describing a000
Owner: kinow
Location: /home/kinow/autosubmit/a000
```

I think it'd be more helpful to have something like

```
Autosubmit version: 4.1.17
Python: 3.9.10
Executable: /home/kinow/Development/python/workspace/autosubmit/venv/bin/autosubmit
Command: autosubmit describe
Args:

YAML configuraiton loaded:
  ...
```

If I run this again with `-u kinow`,

```
(venv) kinow@ranma:~/Development/python/workspace/autosubmit$ autosubmit describe -u kinow
Autosubmit is running with 4.1.17

CLI_PATH : /home/kinow/Development/python/workspace/autosubmit/venv/bin/autosubmit
COMMAND : autosubmit describe -u kinow
```

That's helpful, but I have to separate command & args. If you give me both separated, besides being easier to parse/read it, it's also `grep`'able.

And the Python version is free, I think, probably harmless to add, but could be helpful for us/users when troubleshooting things.

autosubmit/helpers/utils.py
```diff
@@ -464,4 +517,36 @@ def _parse_stat_file(path: Path) -> tuple[int, int]:
...
+        if values
+        else (0, 0)
+    )
+
+
+def describe_command_details(args) -> None:
+    descriptor = "\n"
+    if "autosubmit" in sys.argv[0]:
+        descriptor += f"CLI_PATH : {sys.argv[0]}\n"
+        cli_args = ["autosubmit"] + sys.argv[1:]
```
kinow: I haven't done this in a while, but there's a canonical way to retrieve the command, IIRC.

Test with a symbolic link, and I think the logic here might break?

```bash
(venv) kinow@ranma:~/Development/python/workspace/autosubmit$ which autosubmit
/home/kinow/Development/python/workspace/autosubmit/venv/bin/autosubmit
(venv) kinow@ranma:~/Development/python/workspace/autosubmit$ ln -s /home/kinow/Development/python/workspace/autosubmit/venv/bin/autosubmit as
(venv) kinow@ranma:~/Development/python/workspace/autosubmit$ ./as describe
Autosubmit is running with 4.1.17

COMMAND : ./as describe

YAML configuration loaded:

```

The cli path is gone (BTW, CLI , cli probably doesn't make sense to all users -- e.g., sysadmin, operator, manager-- I think executable is probably more suitable for any user).

Bash shell has a way to figure out the path of the script, and I think you also need to consider cases where `realpath -P` or something is used. I don't recall how that's done in Python, but there's a way with stdlib if I recall well.

autosubmit/helpers/utils.py
```diff
@@ -464,4 +517,36 @@ def _parse_stat_file(path: Path) -> tuple[int, int]:
...
+    else:
+        cli_args = sys.argv
+    command = " ".join(shlex.quote(arg) for arg in cli_args)
```
kinow: Huh, we lost the values passed by the user? I think we need to think another way to print this. shlex is safe, but it's not round-robin.

e.g., see the difference in what I gave, and what your command is now printing.

```bash
(venv) kinow@ranma:~/Development/python/workspace/autosubmit$ autosubmit describe -u "'kinow'"
Autosubmit is running with 4.1.17

CLI_PATH : /home/kinow/Development/python/workspace/autosubmit/venv/bin/autosubmit
COMMAND : autosubmit describe -u ''"'"'kinow'"'"''
```

This is harmless at first, until a user uses this to automate something, or grep, and then for some crazy reason they complain this is breaking and we have to think in a way to give a more predictable result. We see this happening in DestinE quite frequently, where things we didn't think would happen... well, suddenly happen and we have to fix 😅 

I don't know a way to quickly solve it here, but if I were doing it, I'd take a look at the place where the raw data is modified. It should be probably near where argparse is parsing the command... maybe we can just keep the raw data (if argparse doesn't already do that).

autosubmit/helpers/utils.py
```diff
@@ -464,4 +517,36 @@ def _parse_stat_file(path: Path) -> tuple[int, int]:
...
+            current_owner = pwd.getpwuid(current_owner_id).pw_name
+        except (TypeError, KeyError) as e:
+            Log.warning(
+                f"Current owner of experiment {args.expid} could not be retrieved. "
+                f"The owner is no longer in the system database: {str(e)}"
+            )
+        user_descriptor = (
+            current_owner if current_owner is not None else current_owner_id
+        )
+        descriptor += f"USER: {user_descriptor}"
```
kinow: All other args have a space after. This one doesn't. I think it's better to keep the colon together with the word like here. We are more used to "Autosubmit: 4..1.17" rather than "Autosubmit : 4.1.17".

And I think this is not really USER/user. I would understand this as the user running the command.

Here, I think what we have is the owner?

And concatenating strings like this is normally bad & slow in Python (which is already slower than C, etc.). Better to use a list + join in the `Log.info` -- e.g., https://medium.com/@yvan.siyou/stop-using-string-concatenation-in-python-the-superior-efficiency-of-join-explained-eae77d6c1595 & https://realpython.com/python-string-concatenation/ etc

Image

We have string concatenation with `+` in a lot of our code and we need to fix it. Eventually we will do it, I think.

kinow:
@VindeeR great job! I had a look at half of the code, stopped before looking at `autosubmit.py`. Have a look at the comments and see what you think.

If you gather the metadata about the command & experiment to display in a sort of dictionary, we can also have something like this pseudo-code,

```
command_metadata = _get_command_metadata(args, as_conf, ...?)

width = max(len(key) for key in items)

for key, value in items.items():
    print(f"{key:<{width}} : {value}")
```

That'd print things like

```
version    : 4.1.17
python    : 3.9.0
etc...
```

Or we can keep things

```
version: 4.1.17
python: 3.9.0
etc...
```

Right now it's not consistent as I think one (USER) at least is using `"USER:", while the rest is "CLI :".

And personally I'm against the upper cases. Our configuration is a mix of lower & upper case, which causes some bugs. I'll create one now for platform where mixing the keys causes inconsistencies that are not always clear to users as AS doesn't inform what went wrong/where.

I believe AS UPPER_CASE_VARIABLES are probably coming from old Fortran that used upper case, even though it was not case sensitive (I believe it was more of a convention). No modern language has adopted that, which I think shows that besides reducing the risk of bugs, users can adapt & use lower case as well.

test/unit/test_autosubmit.py
```diff
@@ -519,6 +550,61 @@ def test_submit_ready_jobs_raises_on_missing_template(mocker):
...
+        (
+            ["autosubmit", "updatedescrip", "a000", "'description'"],
+            "command: autosubmit updatedescrip\nargs: expid=a000 description='description'",
+        ),
+    ],
+)
+def test_describe_command_details(mocker, command, command_log):
+    """
+    Test the utils.describe_command_details to ensure that command is being printed correctly.
+    """
```
kinow: The comment can be in a single line `"""Test the ..."""`, like in the other functions... 😬

test/unit/test_profiler.py
```diff
@@ -76,57 +74,3 @@ def test_memory_profiling_loop(profiler):
...
-
-    status, args = Autosubmit.parse_args()
-
-    assert status == 0
-    assert args is not None
-
-    with pytest.raises(AutosubmitCritical) as exc_info:
-        Autosubmit.run_command(args)
-
-    assert exc_info.value.code == 7012
```
kinow, line 132: Why did we have to delete these tests?

autosubmit/autosubmit.py
```diff
@@ -848,8 +848,8 @@ def parse_args() -> tuple[int, Optional[argparse.Namespace]]:
                                    help='Mode. Options are c(cat), t(tail). Default is c(cat).')
             subparser.add_argument('-i', '--inspect', default=False, action='store_true',
                                    help='Read job files generated by the inspect subcommand.')
-            subparser.add_argument('ID', metavar='ID',
-                                   help='An ID of a Workflow (eg a000) or a Job (eg a000_20220401_fc0_1_1_APPLICATION).')
+            subparser.add_argument('expid', metavar='expid',
+                                   help='The experiment ID (eg a000) or a Job (eg a000_20220401_fc0_1_1_APPLICATION).')
```
kinow, line 853: Hmmm... I feel like I did that intentionally. I have a vague memory that using `expid` here triggered a check somewhere for expidless/expidful commands, and that didn't work with `cat-log`. Maybe it's safer to leave this improvement for another issue/PR? Feel free to create an issue for this if you'd like. Looks like I never added a comment here :/

autosubmit/helpers/utils.py
```diff
@@ -351,34 +393,47 @@ def build_and_connect_platform(platform_name: str, as_conf: 'AutosubmitConfig',
...
+        plat.temp_dir = platform_config.get("TEMP_DIR", "")
+        plat.root_dir = str(
+            Path(plat.scratch)
+            / (
+                plat.project
+                if hasattr(plat, "project") and plat.project
+                else section_project
+            )
+            / (plat.user if hasattr(plat, "user") and plat.user else "")
+            / expid
```
kinow: Maybe this part needs a quick/small refactoring... We can possibly simplify here by creating the path before we call `str(path)` here?

kinow (commented):
@VindeeR what about something like this?

```diff
diff --git a/autosubmit/autosubmit.py b/autosubmit/autosubmit.py
index 75436045..ef4ca006 100644
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -901,8 +901,8 @@ class Autosubmit:
             expid = args.expid
 
         if args.command != "configure" and args.command != "install":
-            Autosubmit._init_logs(args, args.logconsole, args.logfile, expid)
             describe_command_details(args)
+            Autosubmit._init_logs(args, args.logconsole, args.logfile, expid)
 
         if args.command == 'run':
             if args.trace and args.profile is None:
@@ -1224,8 +1224,6 @@ class Autosubmit:
             Log.info("Locale C.utf8 is not found, using 'C' as fallback")
             locale.setlocale(locale.LC_ALL, 'C')
 
-        Log.info(f"Autosubmit is running with {Autosubmit.autosubmit_version}")
-
     @staticmethod
     def _check_ownership_and_set_last_command(as_conf, expid, command):
         if command not in ["monitor", "describe", "delete", "report", "stats", "dbfix"]:
diff --git a/autosubmit/experiment/experiment_common.py b/autosubmit/experiment/experiment_common.py
index 8f3f580d..12796d36 100644
--- a/autosubmit/experiment/experiment_common.py
+++ b/autosubmit/experiment/experiment_common.py
@@ -414,7 +414,7 @@ def check_ownership(expid: str, raise_error=False) -> tuple[bool, bool, Optional
     try:
         eadmin = current_user_id == pwd.getpwnam(admin_user).pw_uid
     except Exception as e:
-        Log.info(f"Autosubmit admin user: {admin_user} is not set: {str(e)}")
+        Log.debug(f"Autosubmit admin user: {admin_user} is not set: {str(e)}")
     current_owner_id = Path(BasicConfig.LOCAL_ROOT_DIR, expid).stat().st_uid
     try:
         current_owner = pwd.getpwuid(current_owner_id).pw_name
diff --git a/autosubmit/helpers/utils.py b/autosubmit/helpers/utils.py
index 33c20f79..16a811d1 100644
--- a/autosubmit/helpers/utils.py
+++ b/autosubmit/helpers/utils.py
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
-
+import argparse
 import os
 import pwd
 import re
@@ -23,6 +23,8 @@ from collections import defaultdict
 from contextlib import suppress
 from itertools import zip_longest
 from pathlib import Path
+from platform import platform, system
+from shlex import join
 from typing import TYPE_CHECKING, Iterable, Optional, Union
 
 from autosubmit.config.basicconfig import BasicConfig
@@ -525,48 +527,42 @@ def _parse_stat_file(path: Path) -> tuple[int, int]:
     )
 
 
-def describe_command_details(args) -> None:
-    try:
-        descriptor = "\n"
-        descriptor = "".join(f"{descriptor}executable: {sys.argv[0]}\n")
-        descriptor = "".join(f"{descriptor}command: autosubmit {sys.argv[1]}\n")
-        args_print = ""
-        for key, value in args.__dict__.items():
-            if value is None or value == "" or not value:
-                continue
-            if key in [
-                "version",
-                "logfile",
-                "logconsole",
-                "command",
-                "advanced",
-                "database_backend",
-                "database_conn_url",
-                "databasepath",
-                "databasefilename",
-                "localrootpath",
-                "platformsconfpath",
-                "jobsconfpath",
-                "smtphostname",
-            ]:
-                continue
-            args_print = "".join(f"{args_print} {key}={value}")
-
-        descriptor = "".join(f"{descriptor}args:{args_print}\n")
-        if hasattr(args, "expid") and args.expid and args.expid != "*":
-            descriptor = "".join(f"{descriptor}expid: {args.expid}\n")
-            current_owner_id = Path(BasicConfig.LOCAL_ROOT_DIR, args.expid).stat().st_uid
-            try:
-                current_owner = pwd.getpwuid(current_owner_id).pw_name
-            except (TypeError, KeyError) as e:
-                Log.warning(
-                    f"Current owner of experiment {args.expid} could not be retrieved. "
-                    f"The owner is no longer in the system database: {str(e)}"
-                )
-            user_descriptor = (
-                current_owner if current_owner is not None else current_owner_id
-            )
-            descriptor = "".join(f"{descriptor}user: {user_descriptor}\n")
-        Log.info(f"{descriptor}")
-    except Exception as e:
-        Log.error(f"An error occurred as the command Log tried to be generated {e}")
+def get_os_name() -> str:
+    if system() == "Linux":
+        try:
+            for line in Path("/etc/os-release").read_text().splitlines():
+                if line.startswith("PRETTY_NAME="):
+                    return line.split("=", 1)[1].strip().strip('"')
+        except FileNotFoundError:
+            pass
+
+    return platform()
+
+
+def describe_command_details(args: argparse.Namespace) -> None:
+    import socket
+
+    from autosubmit.autosubmit import Autosubmit
+    kv: dict[str, Union[str, dict]] = {
+        "python": sys.version,
+        "system": get_os_name(),
+        "hostname": socket.gethostname(),
+        "argv": str(sys.argv),
+        "executable": sys.argv[0],
+        "subcommand": sys.argv[1],
+        "args": join(sys.argv[1:]),
+        "argparse": args.__dict__,
+    }
+    if "expid" in args and args.expid and args.expid != "*":
+        from autosubmit.experiment.experiment_common import check_ownership
+        expid = args.expid
+        kv["expid"] = expid
+        _, _, current_owner = check_ownership(expid=expid, raise_error=True)
+        if current_owner:
+            kv["owner"] = current_owner
+    width = max(len(key) for key in kv)
+
+    Log.info(f"Autosubmit {Autosubmit.autosubmit_version}")
+    for k, v in kv.items():
+        Log.info(f"  {k:<{width}}  {v}")
+    Log.info("")
\ No newline at end of file
```

Have a look at the code and see what you think. I played trying a few different commands, and I think the output is now always the first to appear. It shows the complete info about AS + other related environment settings, and includes information for users but also some that are helpful to us (like the whole dump of argv).

kinow:
@VindeeR this one might be superseded by #3171 😬 there we have easier access yo args in a central place, and it is easier to choose to do it before or after other actions.

VindeeR:
> @VindeeR this one might be superseded by #3171 😬 there we have easier access yo args in a central place, and it is easier to choose to do it before or after other actions.

So I'll be closing this PR is anything here something you could apply to your branch? or should I delete the branch as well?

kinow:
I've deleted the branch but I can still see the changes here. Will re-use what you were doing. Thanks Erick!

Issue #871, opened 2022-09-22
[enhancement] Validate Autosubmit configuration


kinow:
Hi,

Yesterday after a meeting @mcastril and I were chatting about Autosubmit and other projects at BSC, and at some point we discussed validating YAML files.

I mentioned that in [cwltool](https://github.com/common-workflow-language/cwltool), CWL workflows are written in YAML, and the validation (also code generation, HTML documentation, and more) is handled by a library they created, called [Schema Salad](https://github.com/common-workflow-language/schema_salad/).

This library was created after they tried to use JSON-LD but found features missing from the specification (I believe they discussed it directly with the specification working group).

Schema Salad is able to, given a schema written in JSON or YAML, validate that an input YAML is valid or not, based on the schema structure, types, and attributes.

Maybe something that could be useful for Autosubmit 4 too?

-Bruno

kinow:
For example, given my current AS4 platforms file:

```yaml
Platforms:
  local:
    TYPE: ps
    HOST: localhost
    PROJECT: test
    USER: kinow
    QUEUE: debug
    SCRATCH_DIR: '/tmp/'
    ADD_PROJECT_TO_HOST: False
    MAX_WALLCLOCK: '48:00'
    TEMP_DIR: '/tmp/'
  MaReNoStRuM4:
    TYPE: slurm
    HOST: mn1.bsc.es
    PROJECT: bsc32
    USER: bsc32070
    QUEUE: debug
    SCRATCH_DIR: /gpfs/scratch
    ADD_PROJECT_TO_HOST: False
    MAX_WALLCLOCK: '48:00'
    TEMP_DIR: ''
  marenostrum_archive:
    TYPE: ps
    HOST: dt02.bsc.es
    PROJECT: bsc32
    USER: bsc32070
    SCRATCH_DIR: /gpfs/scratch
    ADD_PROJECT_TO_HOST: False
    TEST_SUITE: False

  transfer_node:
    TYPE: ps
    HOST: dt01.bsc.es
    PROJECT: bsc32
    USER: bsc32070
    ADD_PROJECT_TO_HOST: false
    SCRATCH_DIR: /gpfs/scratch

  transfer_node_bscearth000:
    TYPE: ps
    HOST: bscearth000
    USER: dbeltran
    PROJECT: Earth
    ADD_PROJECT_TO_HOST: false
    QUEUE: serial
    SCRATCH_DIR: /esarchive/scratch

  bscearth000:
    TYPE: ps
[...]

PR #3046, opened 2026-06-04
Raise `AutosubmitCritical` when `TOTALJOBS = 0`

### LINKED ISSUE #2749: Infinite loop in INSPECT calculating possible ready jobs for machines with 4.1.16 when TOTALJOBS is 0 for a machine
### Description

Hi @dbeltrankyl @kinow @manuel-g-castro ,

This is a very specific case that I found copying an operational experiment from Cheng (can't find him here) to improve the performance of the applications workflow.

I was having issues with the end-to-end experiments in DestinE with 4.1.16. When inspecting, I was getting an infinite loop when calculating READY jobs, both with --quick and without. After trying multiple things and multiple configs, **I found that the issue was that there was a TOTALJOBS: 0 defined for a platform (MARENOSTRUM5-TRANSFER), and this was causing an infinite loop in the inspect command**:

```
Calculating possible ready jobs for MARENOSTRUM5-TRANSFER
Calculating possible ready jobs for MARENOSTRUM5
Calculating possible ready jobs for MARENOSTRUM5-LOGIN
Calculating possible ready jobs for MARENOSTRUM5-TRANSFER
Calculating possible ready jobs for MARENOSTRUM5
Calculating possible ready jobs for MARENOSTRUM5-LOGIN
Calculating possible ready jobs for MARENOSTRUM5-TRANSFER
Calculating possible ready jobs for MARENOSTRUM5
Calculating possible ready jobs for MARENOSTRUM5-LOGIN
Terminated (by me)
```

I think that we should not allow running with that configuration, so I would expect an exception.

* Autosubmit version: 4.1.16-dev-6c6f666
* Machine/VM/environment name (if applicable): climateDT-VM
* Experiment ID (if applicable): a31o
  * Infinite loop I believe: `/appl/AS/AUTOSUBMIT_DATA/a31o/tmp/ASLOGS/20251217_113940_inspect.log`
  * Fixed: `/appl/AS/AUTOSUBMIT_DATA/a31w/tmp/ASLOGS/20251217_115011_inspect.log`

### Reproducible Example

a31o:

main.yml

```
RUN:
     
  SUITE: e-suite
  CYCLE: E-25.1
  WORKFLOW: end-to-end
  ENVIRONMENT: intel
  PROCESSOR_UNIT: cpu
  TYPE: operational
 
MODEL:
  NAME: ifs-fesom
  SIMULATION: ifs-fesom-storylines-historical-tco1279
  GRID_ATM: tco1279l137
  PATH: /gpfs/projects/ehpc123/E-25.1/models/ifs-fesom/DE_CY48R1.0_climateDT_20250521bis
  INPUTS: "/gpfs/scratch/ehpc01/input_data/climatemodels/ifs-fesom/inproot"

CONFIGURATION:
  FESOM_PERTURB: "True" 
  CONTAINER_DIR: /gpfs/projects/ehpc123/E-25.1/containers
  ADDITIONAL_JOBS:
    TRANSFER: 'True'
    BACKUP: 'False'
    WIPE: 'False'
    MEMORY_CHECKER: 'False'
    SIZE_CHECK: 'False'
    DQC: 'True'
    CLEAN: 'False'
    AQUA: 'True'

APP:
  ENERGY_INDICATORS: "True"
  ENERGY_OFFSHORE: "True"
  WILDFIRES_FWI: "True"
  WILDFIRES_WISE: "True"
  HYDROLAND: "True"
  HYDROMET: "True"
  OBSALL: "False"
  DATA: "False"

HYDROLAND:
  INIT_FILES: "/gpfs/projects/ehpc01/applications/hydroland_vdev"

HYDROMET:   
   IN_DATA_VERSION: '1.3'
   GRID: "0.1/0.1"  

ENERGY_INDICATORS:
   GRID: "0.1/0.1"

AQUA:
  CONTAINER_VERSION: "0.19.7"
  START_DATE: 20170101
  FIXER_NAME: "climatedt-phase2-production"
  BUCKET: "aqua-web"
  REPOSITORY: "DestinE-Climate-DT/aqua-web"
  AWS_PROFILE: "development"
  EXPERIMENT_NAME: "o01v-story-2017-historical"

PLATFORMS:
  MARENOSTRUM5:
    PROJECT: ehpc123
    DATABRIDGE_FDB_HOME: ''
    PARTITION: ehpc123
    HPCARCH_short: "MN5"
    CATALOG_NAME: "climatedt-o25.1"
    APP_PARTITION: gp_ehpc
  MARENOSTRUM5-login:
    PROJECT: ehpc123
    HPCARCH_short: "MN5"
    CATALOG_NAME: "climatedt-o25.1"
  MARENOSTRUM5-TRANSFER:
    TOTALJOBS: 0

EXPERIMENT:
  MEMBERS: "fc0"
  SPLITSIZEUNIT: day
  DATELIST: "20170101" #Startdate
  CHUNKSIZEUNIT: month
  CHUNKSIZE: 1
  NUMCHUNKS: 1
  CALENDAR: standard

JOBS:
        SIM:
                RETRIALS: 1
                WALLCLOCK: "3:30"
        DN:
                #SPLITSIZE: 10
                TOTALJOBS: 5
        OPA_HYDROMET:
                #SPLITSIZE: 10
                RETRIALS: 1
        OPA_HYDROLAND:
                #SPLITSIZE: 10
                RETRIALS: 1
        OPA_ENERGY_INDICATORS:
                #SPLITSIZE: 10
                RETRIALS: 1
        OPA_ENERGY_OFFSHORE:
                #SPLITSIZE: 10
                RETRIALS: 1
        OPA_WILDFIRES_FWI:
                #SPLITSIZE: 10
                RETRIALS: 1
        OPA_WILDFIRES_WISE:
                #SPLITSIZE: 10
                RETRIALS: 1
        APP_HYDROMET:
                RETRIALS: 1
        APP_HYDROLAND:
                RETRIALS: 1
        APP_ENERGY_INDICATORS:
                RETRIALS: 1
        APP_ENERGY_OFFSHORE:
                RETRIALS: 1
        APP_WILDFIRES_FWI:
                RETRIALS: 1
        APP_WILDFIRES_WISE:
                RETRIALS: 1
        AQUA_ANALYSIS:
                WALLCLOCK: 02:00
        LRA_GENERATOR:
                RETRIALS: 1
                WALLCLOCK: 02:00
REQUEST:
    EXPVER: "0001"
                
#WRAPPERS:
#  WRAPPER_WILDFIRES_WISE_OPA:
#    TYPE: vertical
#    JOBS_IN_WRAPPER: OPA_WILDFIRES_WISE
#  WRAPPER_WILDFIRES_WISE_APP:
#    TYPE: vertical
#    JOBS_IN_WRAPPER: APP_WILDFIRES_WISE 
```

minimal.yml

```
CONFIG:
  # Current version of Autosubmit.
  # I UPDATED THIS VERSION!
  AUTOSUBMIT_VERSION: "4.1.14"
  # Total number of jobs in the workflow.
  TOTALJOBS: 300
  # Maximum number of jobs permitted in the waiting status.
  MAXWAITINGJOBS: 300
DEFAULT:
  # Job experiment ID.
  EXPID: "a31o"
  # Default HPC platform name.
  HPCARCH: "MareNostrum5"
  #hint: use %PROJDIR% to point to the project folder (where the project is cloned)
  # Custom configuration location.
  CUSTOM_CONFIG: "%PROJDIR%/conf/bootstrap/"
PROJECT:
  # Type of the project.
  PROJECT_TYPE: git
  # Folder to hold the project sources.
  PROJECT_DESTINATION: git_project
GIT:
  PROJECT_ORIGIN: "https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow"
  PROJECT_BRANCH: "184-t7-2-add-wrappers-int-the-applications-side-of-the-end-to-end-workflow"
  PROJECT_COMMIT: ''
  PROJECT_SUBMODULES: 'data-portfolio catalog'
  FETCH_SINGLE_BRANCH: true
````

Image

### Expected Behaviour

An exception instead of an inf loop.

### PR DESCRIPTION
**Pull request by ntorqulu**
Closes #2749 

**Things done**
- Add a general `_validate_platform_config` method in `paramiko_submitter.py` in order to check platform parameters.
- Currently it validates that parameters `TOTALJOBS` and `MAX_WAITING_JOBS` are different from 0. Raises AutosubmitCritical if any of them is 0.
- Warn user and disable create wrappers flag if any configuration file carries errors to avoid inf loops.

**Check List**

- [ ] Documentation updated.

### CODE REVIEW

File: autosubmit/platforms/paramiko_submitter.py [hunk 1]
```diff
@@ -91,6 +91,26 @@ def _get_host(section_host: str, add_project_to_host: bool, project: str) -> str
...
+    :raises AutosubmitCritical: If any configuration value is invalid.
+    """
+    if platform.total_jobs is not None and int(platform.total_jobs) == 0:
+        raise AutosubmitCritical(
+            f"PLATFORMS.{platform_name.upper()}.TOTALJOBS must be greater than 0. "
+            f"Current value: {platform.total_jobs}.", 7012)
+    if platform.max_waiting_jobs is not None and int(platform.max_waiting_jobs) == 0:
+        raise AutosubmitCritical(
+            f"PLATFORMS.{platform_name.upper()}.MAX_WAITING_JOBS must be greater than 0. "
+            f"Current value: {platform.max_waiting_jobs}.", 7012)
```
LuiggiTenorioK, line 111: There is a small incongruence between the error message and the check. It says `must be greater than 0`, but the check does `== 0` rather than `<=0`.

LuiggiTenorioK (changes requested):
Thanks @ntorqulu for the PR. The included validation looks good and sufficient to prevent the reported issue!

I will just request one extra change, because when the validated configuration is a negative number (e.g., `MAX_WAITING_JOBS: -1`) the infinite loop is triggered as well.

LuiggiTenorioK:
Just as a root cause analysis of this infinite loop, I saw that this issue might be triggered by this part of the code:

https://github.com/BSC-ES/autosubmit/blob/1fd1f07cbe392ce074e859217bae468b2f5f6da8/autosubmit/autosubmit.py#L1739-L1742

It seems that it will always have "ACTIVE" jobs that won't be "deactivated" by the `job_list.update_list` because the configuration and the fact that it is a dry run (`autosubmit inspect`) don't allow it.

Knowing this doesn't discard this PR, as it already solves the reported bug by not getting to that point. However, is a good to know in case another part of the code gets buggy for the same reason.

autosubmit/platforms/paramiko_submitter.py
ntorqulu: Yes, I had this same dilema when I was implementing the validation 😅
The problem is that `get_total_jobs()` and `get_max_waiting_jobs()` in `configcommon.py` return `-1`as the default value when the user does not set this values in the yaml files:

```
def get_total_jobs(self):
        """Returns max number of running jobs  from autosubmit's config file

        :return: max number of running jobs
        :rtype: int
        """
        return int(self.get_section(['CONFIG', 'TOTALJOBS'], -1))
```
```
    def get_max_waiting_jobs(self):
        """Returns max number of waiting jobs from autosubmit's config file

        :return: main platforms
        :rtype: int
        """
        return int(self.get_section(['CONFIG', 'MAXWAITINGJOBS'], -1))
```

So if the user leaves this parameters as empty in the yaml file, they are automatically filled with `-1`.
This implies that all the experiments that don't have set TOTALJOBS or MAXWAITINGJOBS would fail validation, which was previously not enforced.
This is why I checked only the `==0` case, when the user assigns an incorrect value to the parameter. When the user forgets to set it up I think it should not fail.

Another problem is what happens if the user sets the value to `-5` for instance... 
It should also raise the error? So the validation should be: `int(platform.total_jobs) == 0 or int(platform.total_jobs) < -1`.

But this overcomplicates things too much...

LuiggiTenorioK: > So if the user leaves this parameters as empty in the yaml file, they are automatically filled with `-1`.

Wait, so why is the infinite loop triggered when `TOTALJOBS: -1`, but not when it is empty? :face_with_spiral_eyes:

ntorqulu: Okay, I see the error! I was confusing `PLATFORM.TOTALJOBS` with `CONFIG.TOTALJOBS`. They are diff things.

When parameter `PLATFORM.TOTALJOBS` or `PLATFORM.MAXWAITINGJOBS` is not defined in `platforms.yml`, AS retrieves this value from `CONF.TOTALJOBS` or `CONF.MAXWAITINGJOBS`. This last ones are enforced to be always non-emtpy.

Current errors I see, I list them to have them written down to fix.

1. In `autosubmit.yml` is mandatory to define parameters `TOTALJOBS` and `MAXWAITINGJOBS`. The only restriction enforced for these parameters is to be non-empty:
```
if parser_data["CONFIG"].get('MAXWAITINGJOBS', -1) == -1:
                self.wrong_config["Autosubmit"] += [['config',
                                                     "MAXWAITINGJOBS parameter not found or non-integer"]]
            if parser_data["CONFIG"].get('TOTALJOBS', -1) == -1:
                self.wrong_config["Autosubmit"] += [['config',
                                                     "TOTALJOBS parameter not found or non-integer"]]
```
Should be changed to strictly positive integers.
I've finally set up locally the slurm container from giovtorres, I'm doing local tests with it.
If platform `SLURM-LOCAL` does not have `TOTALJOBS` defined and `CONFIG.TOTALJOBS` is defined as 0 or negative ( `CONF.TOTALJOBS = -5`) it also enters an inf loop. I'll fix that.
`CONF.TOTALJOBS` and `CONF.MAXWAITINGJOBS` should be always defined in `autosubmit.yml` and its values should be strictly positive.

2. In `platforms.yml` it should be enforced  that `PLATFORM.TOTALJOBS` and `PLATFORM.MAXWAITINGJOBS` must be **strictly positve or empty** (because then it uses the default values from autosubmit.yml)

3. At first I thought of moving the validation to `check_platforms_conf` to have all the validation into a single file... But then I saw that errors that do not concern the main platform are discarted. I will keep the validation of `MAXWAITINGJOBS` and `TOTALJOBS` into `paramiko_platform` though it would be better validating it in `configcommon`.

test/regression/workflows/auto-monarch-op1/conf/conf.yml
```diff
@@ -1,6 +1,9 @@
 DEFAULT:
   EXPID: t0jm
   HPCARCH: TEST_SLURM
+CONFIG:
+  TOTALJOBS: 20
+  MAXWAITINGJOBS: 20
```
ntorqulu: I'm not sure if this new implementation is too restrictive now...
CONFIG must define parameters TOTALJOBS and MAXWAITINGJOBS and they must be positive.

LuiggiTenorioK: Yes, it seems really restrictive now. I think this validation might not apply to every command, since commands like `autosubmit run` work without these values assigned.

At this point, do you think it is better to fix the loop or stay with the validation solution?

ntorqulu: I would implement both fixes, relaxing the validation in order to avoid changing the configuration files for the regression tests.
This way, with an invalid value of `TOTALJOBS` or `MAXWAITINGJOBS` AS fails fast during the validation. And only in edge cases it fails in the loop of active jobs.

I would do the following:
1. In `configcommon.py`, validate `CONFIG.MAXWAITINGJOBS` and `CONFIG.TOTALJOBS` to have its section defined, be non-empty and be strictly positive integers. The only extra validation added here that was not present before is the strictly positive condition, all the others were already enforced in `check_autosubmit_conf`.
```
            if int(parser_data["CONFIG"].get('MAXWAITINGJOBS', -1)) <= 0:
                self.wrong_config["Autosubmit"] += [['config',
                                                     "MAXWAITINGJOBS parameter not found or not strictly positive integer"]]
            if int(parser_data["CONFIG"].get('TOTALJOBS', -1)) <= 0:
                self.wrong_config["Autosubmit"] += [['config',
                                                     "TOTALJOBS parameter not found or not strictly positive integer"]]
```
3. In `paramiko_submitter.py` just validate that `PLATFORM.TOTALJOBS` and `PLATFORM.MAXWAITINGJOBS` are non-empty and diff from 0.
```
    if platform.total_jobs is not None and int(platform.total_jobs) == 0:
        raise AutosubmitCritical(
            f"PLATFORMS.{platform_name.upper()}.TOTALJOBS must be greater than 0. "
            f"Current value: {platform.total_jobs}.", 7012)
    if platform.max_waiting_jobs is not None and int(platform.max_waiting_jobs) == 0:
        raise AutosubmitCritical(
            f"PLATFORMS.{platform_name.upper()}.MAX_WAITING_JOBS must be greater than 0. "
            f"Current value: {platform.max_waiting_jobs}.", 7012)
```
This way we allow automatically-filled `-1`, meaning unlimited totaljobs and maxwaitingjobs.

The only problem I see with this implementation is when the user deliberately sets a negative integer in the `platfroms.yaml` configuration for TOTALJOBS or MAXWAITINGJOBS. In this case, it still enters an inf loop.
I think this is the case that should be handled in the active jobs loop, this way we avoid changing the regression tests config files.
What do you think?

LuiggiTenorioK: > 1. In `configcommon.py`, validate `CONFIG.MAXWAITINGJOBS` and `CONFIG.TOTALJOBS` to have its section defined, be non-empty and be strictly positive integers. The only extra validation added here that was not present before is the strictly positive condition, all the others were already enforced in `check_autosubmit_conf`.

Seems good! :+1:

> This way we allow automatically-filled `-1`, meaning unlimited totaljobs and maxwaitingjobs.

Can we avoid interpreting `-1` as unlimited? I think Dani made it like that to be similar to what is done when programming in C, but we have the more pythonic option to have it as `None`. No problem if it is too complicated.


> The only problem I see with this implementation is when the user deliberately sets a negative integer in the `platfroms.yaml` configuration for TOTALJOBS or MAXWAITINGJOBS. In this case, it still enters an inf loop.

Preferably, this should be solved, but it is not critical if it is not done, since it is something provoked by the user's misconfiguration. 

> I think this is the case that should be handled in the active jobs loop, this way we avoid changing the regression tests config files.

Yes, this is relevant to keep in mind and not move more things that are not directly related. More specifically, we have to avoid adding more tasks to the user when they move from one version of Autosubmit to a newer one.

ntorqulu: > Can we avoid interpreting -1 as unlimited? 

Yes, I think is a good idea but I'll enforce it only for `TOTALJOBS` and `MAXWAITINGJOBS` parameters. There're others that also follow this convention but I think it's out of the scope of this pr.
Maybe open a new issue for the other ones to keep consistency? I list the ones I have caught reading `configcommon.py`:

- `MAX_PROCESSORS`
- `DELAY_RETRY_TIME`
- `PROJECT_SUBMODULES_DEPTH`
- `CHUNKSIZE`
- `NUMCHUNKS`

I can add a TODO for each occurrence in the code.
We should enforce a unique default value for all of them, now there'll be a mismatch between `None` and `-1` meaning the same but with diff encodings.

I think I've caught a bug while setting `TOTALJOBS` and `MAXWAITINGJOBS` default value to `None` in method `get_section`:
In this method you can define which default value (`d_value`) you want the method to return in case the section is not found. The problem is that previously if you wanted to assign a `d_value = None` it always entered the raise-AutosubmitCritical condition:
```
if current_level is None or (
                not isinstance(current_level, numbers.Number) and len(current_level) == 0) and must_exists:
            raise AutosubmitCritical(
                f"{section_str} must exists. Check that subsection {str(current_level)} exists.", 7014)
```
due to logic operator precedence: 
- `must_exist = False`
- `not isisintance(...) and len(...) = True`
- `current_level is None = True`
TRUE OR TRUE AND FALSE = TRUE OR FALSE = TRUE
Always enter the condition if d_value is None and section not specified.
Solved, but not 100% sure about this one and if changing it will carry new problems... Currently changed and passing all the tests.

> Preferably, this should be solved, but it is not critical if it is not done, since it is something provoked by the user's misconfiguration.

Solved defining variables `total_jobs_limit` and `max_waiting_jobs_limit` to avoid the inf loop in `calculate_job_limits`.
What I've done is define a max threshold of inf if `total_jobs` < 0 or `total_jobs` is `None`.  Same applies for `max_waiting_jobs`.
Note that negative values are also handled as inf. Not sure about this one, but I think that if the user has already ran the workflow and it's in this stage the workflow should run, not validate. Can be changed.

LuiggiTenorioK: > There're others that also follow this convention but I think it's out of the scope of this pr.

Agree +1

> I think I've caught a bug while setting `TOTALJOBS` and `MAXWAITINGJOBS` default value to `None` in method `get_section`: In this method you can define which default value (`d_value`) you want the method to return in case the section is not found. The problem is that previously if you wanted to assign a `d_value = None` it always entered the raise-AutosubmitCritical condition:
> 
> ```
> if current_level is None or (
>                 not isinstance(current_level, numbers.Number) and len(current_level) == 0) and must_exists:
>             raise AutosubmitCritical(
>                 f"{section_str} must exists. Check that subsection {str(current_level)} exists.", 7014)
> ```
> 
> due to logic operator precedence:
> 
> * `must_exist = False`
> * `not isisintance(...) and len(...) = True`
> * `current_level is None = True`
>   TRUE OR TRUE AND FALSE = TRUE OR FALSE = TRUE
>   Always enter the condition if d_value is None and section not specified.
>   Solved, but not 100% sure about this one and if changing it will carry new problems... Currently changed and passing all the tests.

That's an interesting insight! It really seems that you successfully captured the bug by quickly testing it and setting the `TOTALJOBS` and `MAXWAITINGJOBS` to negative.

I'll give a closer look later to see if there are no unintended side effects, but for now, it looks good.

autosubmit/config/configcommon.py
```diff
@@ -203,8 +203,8 @@ def get_section(self, section: list[str], d_value: Union[str, Any] = "", must_ex
                         raise AutosubmitCritical(
                             f"[INDEX ERROR], {section_str} must exists. Check that {str(current_level)} is an section that exists.",
                             7014)
-        if current_level is None or (
-                not isinstance(current_level, numbers.Number) and len(current_level) == 0) and must_exists:
+        if (current_level is None or (
+                not isinstance(current_level, numbers.Number) and len(current_level) == 0)) and must_exists:
```
LuiggiTenorioK: Really good catch :+1:

To make it better, you can put the `must_exists` first before the parentheses start, so if `must_exists = False`, then the rest won't be evaluated.

```suggestion
        if must_exists and (current_level is None or (
                not isinstance(current_level, numbers.Number) and len(current_level) == 0)):
```

autosubmit/job/job_packager.py
LuiggiTenorioK, line 573: This is the tricky part, because before the update, it sets `self._max_jobs_to_submit` to 0 since it will be negative in the second-to-last line of this method. Then, the last `self.max_jobs` will be 0.

Now, it will be set to inf, which will allow submitting more jobs, which makes more sense, understanding that total jobs and max waiting jobs default to infinity capacity when not defined or negative.

I think it is correct. However, I also think @dbeltrankyl might have to give an extra look to be sure that this will not break something else.

autosubmit/config/configcommon.py
ntorqulu: Sure, changed it!

autosubmit/job/job_packager.py
ntorqulu: Yes, for this pr I think it makes sense to set them to inf, but I'm not sure if with this change we can break things. If so, we should include more tests then.
I'll tag him as a reviewer; thanks @LuiggiTenorioK!

autosubmit/platforms/paramiko_submitter.py
dbeltrankyl: > CONF.TOTALJOBS and CONF.MAXWAITINGJOBS should be always defined in autosubmit.yml and its values should be strictly positive.

I agree, but only when the user tries to use the create -cw flag, inspect, and run, no?

I can see a scenario where this configuration is not defined in minimal.yml because it is meant to be set once the `proj` folder is pulled from the remote. So there could be a race condition in which the refresh code is not triggered yet

Edit: Or we can just check if the proj/ exists or not before validating the conf, and once it exists, if the proj/ is generated by that command, we call the validation.

Also, the  archive/unarchive/migrate/delete should never validate the conf because it may not exist or is not relevant

autosubmit/job/job_packager.py
dbeltrankyl: Looks alright to me, thanks @ntorqulu !

dbeltrankyl (approved):
Thanks for the work! 

Just check that you can generate and run a minimal experiment without issues, and we can merge

autosubmit/platforms/paramiko_submitter.py
ntorqulu: Okay, thanks for the feedback @dbeltrankyl!
The validation of `CONF.TOTALJOBS` and `CONF.MAXWAITINGJOBS` defined in `autosubmit.yml` is done in `check_conf_files` -> `check_mandatory_parameters` -> `check_autosubmit_conf`.

If it's called with `running_time` set to `False` it will treat configuration errors as warnings, non-blocking. They will be logged but it won't stop the execution. 
E.g. 
```
[WARNING] On Configuration files:
Critical Issues on [Autosubmit] config file:
[config] TOTALJOBS parameter not found or not strictly positive integer
```
If it's called with `running_time` set to `True` it will raise an AutosubmitError` and stop the execution.
E.g.
```
Error code: 7014

 [CRITICAL] On Configuration files:
Critical Issues on [Autosubmit] config file:
[config] TOTALJOBS parameter not found or not strictly positive integer
 [eCode=7014]
```
Depending on the command that calls this method, `running_time` is set to `True` (validates strictly) or `False` (only warns).
So, from all the commands listed:

1. **check_conf_files(running_time=True)** -> raise AutosubmitError
- run -> `MAXWAITINGJOBS` and `TOTALJOBS` must be defined and non-negative in autosubmit.yml.
- inspect -> same
- recovery -> same

2. **check_conf_files(running_time=False)** -> logs warning
- create with/without  minimal.yml -> logs a warning, non blocking. Execution continues.
- `create -cw` -> same
- monitor -> same

3. Not call `check_conf_files`
- archive, unarchive, migrate, delete -> do not call this method

---

So, regarding the issues:
> I agree, but only when the user tries to use the create -cw flag, inspect, and run, no?

It will stop the execution for run, inspect and recovery. For create with the -cw flag it will only log a warning... Is this as intended?

> I can see a scenario where this configuration is not defined in minimal.yml because it is meant to be set once the proj folder is pulled from the remote. So there could be a race condition in which the refresh code is not triggered yet

In this case the user can do:
- `autosubmit create`  ->  with a warning-log while executing create (non-stopping)
- `autosubmit refresh` -> does not validate config files
- `autosubmit run` -> will merge the configuration files of `as_conf/` files (loaded as CUSTOM_CONF PRE) and `conf/`. If for instance `TOTALJOBS` is defined in both, the `conf/` version wins. This means that: `as_conf.TOTALJOBS: 10` + `conf.TOTALJOBS: 0` will still raise an error. Is this intended?

> Edit: Or we can just check if the proj/ exists or not before validating the conf, and once it exists, if the proj/ is generated by that command, we call the validation.

Currently in `refresh` the configuration is not validated. It is validated once the user executes `run`, raising an error. Or in `create`, with a log. Is this intended?

> Also, the archive/unarchive/migrate/delete should never validate the conf because it may not exist or is not relevant

Yes, they are commands that do not call the validation of the config files.

ntorqulu: Also, I was editing a custom config in my own github project to test the `refresh` command.
Reading the documentation, I though that everytime you execute `autosubmit refresh <EXPID>` the github files are synchronized, but it was not the case.
Is there a command to synchronize the local config files with the ones from the remote repo?

dbeltrankyl: > For create with the -cw flag it will only log a warning... Is this as intended?

Hm, maybe we have to disable the part of the -cw, when there is an invalid configuration and show an error?  

At the end, the -cw runs the same loop as the autosubmit run and inspect to form the wrappers, and it may hang forever

> Currently in refresh the configuration is not validated

That's intended, the project data must be gatherable always

> It is validated once the user executes run, raising an error. Or in create, with a log. Is this intended?

I think we only show a log in the create because we had race conditions issues with the minimal <-> versioned configuration.

What would be perfect is to show a critical error, not a warning, when the experiment is badly configured. We can do it in another issue/PR as it is out-scope of this one. 

For this one, the important thing is that the validation still works as it was working before 

> I though that everytime you execute autosubmit refresh <EXPID> the github files are synchronized, but it was not the case.

I also thought that :thinking: 

I'm not sure if we did some change lately, but what it was doing before was to mv the current $expid/proj/project to $expid/proj/project_randomstring unless there are pending commits; if there are pending commits, it is disabled with an error, and then we had a new git clone $expid/proj/git_project,(for example) with the remote changes

ntorqulu: > Hm, maybe we have to disable the part of the -cw, when there is an invalid configuration and show an error?

What I've done at the end is log a warning saying that the wrappers will not be generated until the configuration issues are fixed (any configuration issues, not just the `TOTALJOBS/MAXWAITINGJOBS` errors). 

```
Saving the jobs list...
[WARNING] There are errors in the configuration files. Wrappers will not be generated. Please fix the errors and run the command again.
```

This way I think is easier for them to fix it if they can see the plots. Not sure about this one, can be easily changed to an AutosubmitCritical error.

> What would be perfect is to show a critical error, not a warning, when the experiment is badly configured. We can do it in another issue/PR as it is out-scope of this one.

Yep, I can open another issue. Badly configured means that `len(as_conf.wrong_config) > 0` or we should only focus on certain errors to stop the execution? Is there a priority?

> I though that everytime you execute autosubmit refresh the github files are synchronized, but it was not the case.

I'll investigate further `autosubmit refresh` an open an issue if the error persists, I think it's handy to have a command to automatically synchronize the changes with the remote branch.

dbeltrankyl: > What I've done at the end is log a warning saying that the wrappers will not be generated until the configuration issues are fixed (any configuration issues, not just the TOTALJOBS/MAXWAITINGJOBS errors).

:+1: 

> Yep, I can open another issue. 

We have some validators already on the code in python but in this issue https://github.com/BSC-ES/autosubmit/issues/871  @kinow mentioned about the possibility of validating the yaml itself not sure if that can validate only parameters that are present or we can define which parameters must exists etc. 

We can use that Issue ( or open a fresh one since it was open in 2022)  to discuss if we validate the yaml in that way, we extend our current validator, a mix of both, another lib etc...


> Badly configured means that len(as_conf.wrong_config) > 0 or we should only focus on certain errors to stop the execution? Is there a priority?


Maybe we can have different levels of strictness configurable through the yaml files config_validator: "strict/dont_validate/other..." we can further discuss that in that issue


> I'll investigate further autosubmit refresh an open an issue if the error persists, I think it's handy to have a command to automatically synchronize the changes with the remote branch.

:+1: 

Thanks!

ntorqulu: Great, thanks! 
I've been reading through the issue and I think is better to avoid creating a new one, as it is really well-structured and collects key info for improving the yaml-validator. 
I'll investigate a little bit about the libraries @kinow proposed in the issue and continue the discussion of the validation there.
We can have defined multiple schemas depending on how strict we want the validation to be.

ntorqulu:
> Just check that you can generate and run a minimal experiment without issues, and we can merge

Tested the minimal experiment + git project with the following commands:
- `autosubmit create` -> only warns user, continues execution
- `autosubmit create -cw` -> warns user + disables flag, continues execution
- `autosubmit run` -> raise AutosubmitCritical
- `autosubmit inspect` -> raise AutosubmitCritical
- `autosubmit setstatus` -> raise AutosubmitCritical
- `autosubmit recovery` -> raise AutosubmitCritical
- `autosubmit monitor` -> warns user
- `autosubmit refresh` -> execution without warnings nor errors, but does not synchronize the remote if there are changes, TODO, out of scope
- `autosubmit describe, archive, unarchive, report, delete` ->execution without warnings nor errors

autosubmit/platforms/paramiko_submitter.py
kinow: The idea of that issue is to replace any validation in the code. To have something like

```
as_conf = config.load('some_path')
as_conf.validate()  # Raises an exception if invalid
...
# From this point onward, you can trust the data is loaded + match type-hint
```

With that, we reduce the amount of code & the amount to test, and the rest of the code runs with the assumption that if it passed validation, then we can trust the data is in the expected format [^1].

JsonSchema, Yamale, that Facebook lib (hydra?), and others offer ways to validate documents based on some declarative approach (i.e. no `if`'s written, less tests needed). But someone needs to invest some time to test, and verify which one offers the best features & is likely to be still maintained in 2/5+ years (there are heaps of smaller libs, but we need something future-proofed).

[^1]: This concept is old, and I've worked on projects that did this back in 00's, 2010's, but there's a more recent essay that explains this well, you probably have already read it: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

dbeltrankyl:
Thanks for the work @ntorqulu !

I believe this can be merged now, no? The outcome is the expected https://github.com/BSC-ES/autosubmit/pull/3046#issuecomment-4845707838, and the rest is out of scope of this issue, as you mentioned

PR #3050, opened 2026-06-08
Fix bool type for Python 3.9

### LINKED ISSUE #3009: Possible runtime bug  due to invalid type (False)
### Description

Found in `autosubmit_git.py` in 4.1.17 master:

https://github.com/BSC-ES/autosubmit/blob/4350503ddfba4896be6c0af67b816774bd599201/autosubmit/git/autosubmit_git.py#L178

### Reproducible Example

NA

kinow:
Closes #3009

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/autosubmit/git/autosubmit_git.py
+++ b/autosubmit/git/autosubmit_git.py
@@ -23,7 +23,7 @@ import subprocess
 from pathlib import Path
 from shutil import rmtree
 from time import time
-from typing import List, Optional, Union
+from typing import Optional, Union
 
 from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.config.configcommon import AutosubmitConfig
@@ -175,7 +175,7 @@ class AutosubmitGit:
         git_remote_project_path = as_conf.get_git_remote_project_root()
 
         git_project_commit = as_conf.get_git_project_commit()
-        git_project_submodules: Union[False, List[str]] = as_conf.get_submodules_list()
+        git_project_submodules: Union[list[str], bool] = as_conf.get_submodules_list()
         git_project_submodules_depth = as_conf.get_project_submodules_depth()
         max_depth = -1
         if len(git_project_submodules_depth) > 0:


kinow:
A 2 minutes job :grimacing: It's not broken-broken, as in git clones, even if this is invalid in py3.9 :shrug: Maybe @ntorqulu or @rajeshsuraj can review it? Thanks!

autosubmit/git/autosubmit_git.py
```diff
@@ -175,7 +175,7 @@ def clone_repository(as_conf, force, hpcarch):
         git_remote_project_path = as_conf.get_git_remote_project_root()
 
         git_project_commit = as_conf.get_git_project_commit()
-        git_project_submodules: Union[False, List[str]] = as_conf.get_submodules_list()
+        git_project_submodules: Union[bool, List[str]] = as_conf.get_submodules_list()
```
ntorqulu: Hi @kinow! Thanks for fixing the return type.
Checking method `get_submodules_list()`, I see it has the following signature:
```
    def get_submodules_list(self) -> Union[list[str], bool]:
```
I would keep the same order (first list, then bool) and typing style (list instead of List) to be consistent, although it does not change code behavior. This way I think we can avoid importing `List` from `typing` in `autosubmit_git.py`.

ntorqulu (commented):
Just a small reordering to be easier to read and avoid an import.

autosubmit/git/autosubmit_git.py
kinow: Hi @ntorqulu ! I hadn't seen the `List`. And after you mentioned the order of the types in the `Union`, I cannot unsee it. +1 to making them both equal. It doesn't break anything to leave the other way, but this way it appeases my OCD :rofl: 

Let me edit my commit.

kinow: Done!

ntorqulu: jajaj great, thanks!

ntorqulu:
Thanks for the changes! I've re-triggered the integration tests as they were failing due to timeout.
Feel free to merge once they pass!

kinow:
They passed (gotta fix that later). Thanks @ntorqulu ! Merging... :rocket:

PR #3007, opened 2026-05-14
Check jobs, wrappers rework

dbeltrankyl:
This should fix the wrapper's issues. I'll try to summarize the changes in a comment later.
fixes #2985
fixes #2469
fixes #2935
TODO:

* [x] There are two new tests failing (wip) due to a misconfiguration of the test itself.
* [x] Test ( and probably fix) ECplatform.
* [x] Maybe add some more tests.

**Check List**

- [x] I have read `CONTRIBUTING.md`.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to `pyproject.toml`.
- [x] Tests are included (or explain why tests are not needed).
- [x] Changelog entry included in `CHANGELOG.md` if this is a change that can affect users.
- [x] Documentation updated.
- [x] If this is a bug fix, PR should include a link to the issue (e.g. `Closes #1234`).

PR #3053, opened 2026-06-09
don't crash on JOBS_IN_WRAPPER references to undefined jobs

### LINKED ISSUE #3006: Refresh to be independent from wrapper definition in main
This issue comes from https://gitlab.earth.bsc.es/digital-twins/de_340-2/workflow/-/issues/1613 

Dear AS developers @dbeltrankyl @kinow @ntorqulu et al.

### Problem

When trying to do the first autosubmit create with a wrapper configuration set in conf/main.yml but not the joblist set in conf/main.yml, autosubmit complains because it does not find any joblist and it crashes.

I understand this behaviour in `AS create` but I noticed that it also happens at the first `AS refresh` command. I think it would be great to have `AS refresh` independent from anything that should be taken from `proj/` , as I see `AS refresh` as a way to get the remote code of the workflow into the vm (something like "autosubmit clone")

### Proposed Solution

Make `AS refresh` independent from the configuration under conf/main.yml. I understand that the only configuration that AS refresh would need is the information in conf/minimal.yml, in which the details on the workflow repository and workflow dependencies (e.g. submodules) are listed.

or

Create an AS command that does what I mention above, in the case AS refresh is not easily decouplable from AS create.

===========

NOTE: This was detected with wrappers and for now I do not see more cases in which this can be a problem but they may exist as well

I'll be happy reading your opinions!
Thanks in advance!

dbeltrankyl:
fixes #3006

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@
 - Wrapper databases migrated from pickle to SQLAlchemy, with upsert operations and chunked batch inserts for SQLite compatibility #3007
 - Jobs now validated individually via STAT files in check_jobs to confirm completion or failure, for both wrapped and non-wrapped jobs #3007
 - Expanded ECPlatform, paramiko platform support, and platform code cleanup #3007
+- Changed JOBS_IN_WRAPPER unknown job from critical error to warning #3006
 
 ### 4.1.16: Postgres (experimental) support, bug fixes, and enhancements
 
--- a/autosubmit/config/configcommon.py
+++ b/autosubmit/config/configcommon.py
@@ -496,7 +496,7 @@ class AutosubmitConfig(object):
                 elif len(element) == 0:
                     raise AutosubmitCritical(f"JOBS_IN_WRAPPER in WRAPPERS.{wrapper} contains empty job names ( check for double ,, or double && )", 7014)
                 elif element not in job_sections:
-                    raise AutosubmitCritical(f"JOBS_IN_WRAPPER in WRAPPERS.{wrapper} contains job: {element} that is not defined in JOBS section", 7014)
+                    Log.warning(f"JOBS_IN_WRAPPER in WRAPPERS.{wrapper} contains job: {element} that is not defined in JOBS section")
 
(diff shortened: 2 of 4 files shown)


dbeltrankyl:
When validating the experiment configuration, Autosubmit used to crash immediately if a wrapper's JOBS_IN_WRAPPER mentioned a job that wasn't defined in the JOBS section. Now it just warns you and moves on.

This way, if a wrapper references jobs from a different config layer or a definition that's not yet loaded, autosubmit won't crash. 

This is an issue when the user has all the configuration loaded and tries to refresh (for example, #3006).


It won't crash at runtime either. If the section doesn't exist, the AS will ignore it. With the warning in the log and the usage of -cw it should be enough to diagonist why a job wasn't being wrapped 

Ready to review @ntorqulu , @LuiggiTenorioK ( this was also set for 4.1.17)

dbeltrankyl:
Note: We can improve this when we have a proper config validator; there are issues for that...

dbeltrankyl:
Ready to review (it broke a related integration test, the change is the same as the unit one )

dbeltrankyl:
hm > FAILED test/integration/commands/run/test_run_slurm_scenarios.py::test_run_interrupted[Failure with wrapper] - AssertionError: Job_name: t004_20000101_fc0_1_JOB with counter: 2, Field start>=submit missing 

Ok, that's not right, and it is important, but it is not related to this PR I think. 

Seems a flaky failure however it shouldn't  happen.. I'll investigate if it is pollution issues or something else

LuiggiTenorioK (approved):
Seems good! Thanks @dbeltrankyl!

I tested it as in https://github.com/BSC-ES/autosubmit/pull/3007#issuecomment-4486394206 with different `JOBS_IN_WRAPPER` names. Tried with a name that wasn't in `JOBS_IN_WRAPPER` and with a list of sections with names that exist and don't exist in the sections.

The error messages do capture the missing sections correctly and don't crash it.

dbeltrankyl:
Thanks! I'm merging it then

PR #3057, opened 2026-06-10
Flaky bug wrappers

dbeltrankyl:
Fixes a flaky bug derived from #3007 

When an experiment is interrupted ... sometimes the database could end up with job entries that have a submit timestamp but no start or finish time. Because the async recovery process never completed writing them. 

On the next run, those entries were stuck because no code was looking for them.

This PR adds a function called recover_stale_job_data that runs once at the start of every experiment run,  in the create ( before generating a new run ), and during recovery commands.

It queries the database for those incomplete records, fetches the STAT files from the remote platforms (which contain the real timestamps), and writes the missing start and finish times directly into the database.

The submit time is never stale, because it is written by the main process. 

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -63,7 +63,7 @@ from autosubmit.experiment.experiment_common import (
 )
 from autosubmit.git.autosubmit_git import AutosubmitGit
 from autosubmit.git.autosubmit_git import check_unpushed_changes, clean_git
-from autosubmit.helpers.utils import check_jobs_file_exists, get_rc_path, user_yes_no_query
+from autosubmit.helpers.utils import check_jobs_file_exists, get_rc_path, recover_stale_job_data, user_yes_no_query
 from autosubmit.history.experiment_history import ExperimentHistory
 from autosubmit.history.experiment_status import ExperimentStatus
 from autosubmit.job.job import Job
@@ -2221,6 +2221,7 @@ class Autosubmit:
                 while pending_logs:
                     pending_logs = job_list.recover_logs()
                 job_list.save()
+                recover_stale_job_data(expid, as_conf, {p.name: p for p in platforms_to_test})
                 while job_list.get_active():
                     try:
                         if profile is not None:
@@ -2379,6 +2380,7 @@ class Autosubmit:
                         p.log_recovery_process.join(timeout=300)
                         if p.log_recovery_process.is_alive():
                             Log.warning(f"Log recovery process for {p.name} did not terminate within timeout.")
+                    p.clean_log_recovery_process()
 
                 Autosubmit.process_historical_data_iteration(job_list, job_changes_tracker, expid)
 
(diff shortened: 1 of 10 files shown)


dbeltrankyl:
Hm, this Pr has commits from another branch, let me fix that

dbeltrankyl:
Okay, fixed the undesired commits and rebased . Ready for review @LuiggiTenorioK  (the CI/CD should pass, but I'll check it in 15min ~)

The usual flaky  test ( not related to wrappers rework) appeared, re-running it

autosubmit/autosubmit.py
```diff
@@ -2221,6 +2221,8 @@ def run_experiment(expid: str, start_time: Optional[str] = None, start_after: Op
                 while pending_logs:
                     pending_logs = job_list.recover_logs()
                 job_list.save()
+                with suppress(Exception):
+                    recover_stale_job_data(expid, as_conf, {p.name: p for p in platforms_to_test})
```
LuiggiTenorioK: [nitpick] Could it be better to not suppress the exception but log it at least at debug level?

autosubmit/history/database_managers/experiment_history_db_manager.py
LuiggiTenorioK, line 476: [nitpick] I know `get_stale_rows` and `update_job_data_values` will be used by forcing SQLAlchemy. However, to keep consistent with the Protocol, it will be nice to have the methods also in the non-SQLAlchemy manager, at least raising a `NotImplementedError`.

LuiggiTenorioK (approved):
Changes seem good! The code seems pretty straightforward and not complicated to follow, and the tests do cover good cases.

I tested it by running an experiment with remote jobs and manually deleting the start/finish dates to see if the process recovers the dates. It does it successfully! 🎉 

I only added some nitpicks that are minor.

Thanks @dbeltrankyl!

autosubmit/autosubmit.py
dbeltrankyl: Sure

autosubmit/history/database_managers/experiment_history_db_manager.py
dbeltrankyl: Yup, I thought that we would delete that anyway and didn't think about it.

Ideally, we would want to move everything to SQLAlchemist, but I think it is not realistic to have the refactor for the 4.2.0, raising a NotImplementedError sounds good. I'll add that

dbeltrankyl:
Thanks @LuiggiTenorioK ! 

Added a commit addressing  the feedback, should be ready to merge after the PR checks

dbeltrankyl:
Ok @LuiggiTenorioK, as I've commented to you in person, the build_and_connect is changed so it actually loads the appropriate subclass 

I also added the stale_fix at the end of an autosubmit run just in case, there is a note regarding that

LuiggiTenorioK (approved):
New code added seems clean with no issue.

Tested the same manual case again, and with the LOCAL platform as well. Both worked correctly and recovered the datetimes as expected 💯

PR #3059, opened 2026-06-12
4.1.17-Release branch

dbeltrankyl:
**Check List**

- [x] I have read `CONTRIBUTING.md`.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to `pyproject.toml`.
- [x] Tests are included (or explain why tests are not needed).
- [x] Changelog entry included in `CHANGELOG.md` if this is a change that can affect users.
- [x] Documentation updated.
- [x] If this is a bug fix, PR should include a link to the issue (e.g. `Closes #1234`).

PR #3065, opened 2026-06-16
Detect and add missing columns to job_data/experiment_run tables

### LINKED ISSUE #3062: 4.1.17 issue: setstatus is not working when upgrading an existent experiment
### Description

Relevant info: 

hubs: 4.1.17-dev2 

Set status broken when upgrading an existing experiment from 4.1.16 -> 4.1.17

```
 [CRITICAL] Unexpected error: (sqlite3.OperationalError) no such column: job_data.split
[SQL: SELECT job_data.id, job_data.counter, job_data.job_name, job_data.created, job_data.modified, job_data.submit, job_data.start, job_data.finish, job_data.status, job_data.rowtype, job_data.ncpus, job_data.wallclock, job_data.qos, job_data.energy, job_data.date, job_data.section, job_data.member, job_data.chunk, job_data.last, job_data.platform, job_data.job_id, job_data.extra_data, job_data.nnodes, job_data.run_id, job_data."MaxRSS", job_data."AveRSS", job_data.out, job_data.err, job_data.rowstatus, job_data.children, job_data.platform_output, job_data.workflow_commit, job_data.split, job_data.splits, job_data.fail_count 
FROM job_data JOIN _tmp_job_names_1781600416635092621 ON job_data.job_name = _tmp_job_names_1781600416635092621.job_name 
WHERE job_data.last = ?]
[parameters: (1,)]
(Background on this error at: https://sqlalche.me/e/20/e3q8).
 Please report it to Autosubmit Developers through Git: https://github.com/BSC-ES/autosubmit/issues
```

My guess is that the split column is not in the job_data db, and the alter table is not triggered 

### Reproducible Example

> git checkout v4.1.16
> autosubmit expid ...
> autosubmit create ...
> autosubmit run ...
> git checkout master
> autosubmit expid
> autosubmit setstatus ...

dbeltrankyl:
Fixes #3062

There are some issues with the DB when upgrading an existing experiment. This PR is to add the missing columns 

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - Fixed inconsistent log paths in status txt output #3040
 - Follow-up fix for #3007: recover job times that were lost when an experiment was interrupted.
 - Fixed `num_max_spits` unit conversion to months when `split_unit` is "month" #2944
+- Fixed migration of pre-v4.1.17 historical databases missing columns (split, fail_count, out, err) #3062
 - Fixed infinite loop when STAT file stays on RUNNING, now falls back to scheduler status after IO_SAFE_WAIT #3059
 
 **Enhancements:**
--- a/autosubmit/history/data_classes/job_data.py
+++ b/autosubmit/history/data_classes/job_data.py
@@ -69,8 +69,8 @@ class JobData(object):
         # DB VERSION 15 attributes
         self.MaxRSS = MaxRSS
         self.AveRSS = AveRSS
-        self.out = out
-        self.err = err
+        self.out = out if out else ""
+        self.err = err if err else ""
         self.rowstatus = rowstatus
         self.children = children  # DB 17
         self.platform_output = platform_output  # DB 17
--- a/autosubmit/history/database_managers/experiment_history_db_manager.py
+++ b/autosubmit/history/database_managers/experiment_history_db_manager.py
@@ -47,7 +47,6 @@ DB_VERSION_SCHEMA_CHANGES = 12
 DEFAULT_DB_VERSION = 10
 DEFAULT_MAX_COUNTER = 0
 
-
 class ExperimentHistoryDbManager(DatabaseManager):
     """ Manages actions directly on the database.
     """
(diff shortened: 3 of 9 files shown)


dbeltrankyl:
Ready to review, this is necessary for #3059 @kinow , @LuiggiTenorioK

autosubmit/database/tables.py
```diff
@@ -125,8 +125,8 @@
     Column("run_id", Integer),
     Column("MaxRSS", Float, nullable=False, default=0.0),
     Column("AveRSS", Float, nullable=False, default=0.0),
-    Column("out", Text, nullable=False),
-    Column("err", Text, nullable=False),
+    Column("out", Text, nullable=False, default=""),
+    Column("err", Text, nullable=False, default=""),
```
LuiggiTenorioK: Probably not worth it to add a default to an empty string. The migration function just makes `ADD COLUMN`, but does not alter the definition of the column, and making it do so will be messy. Because of that, relying on the fact that these columns will always have a default value is not safe.

Maybe it could be better to evaluate if it is null after executing a select and changing to the default value if that is the case.

LuiggiTenorioK (commented):
Thanks for the fix @dbeltrankyl. It looks great!

I tried with an old experiment, and it added the new columns successfully.

I just added an observation in the change of the data schema that might cause problems in the future. The same issue might happen if any column definition is changed and the migration is not properly handled. I think it is better to keep it simple and only add new columns without changing existing ones, and handle new rules on the data at runtime.

autosubmit/database/tables.py
dbeltrankyl: I've changed it back and in the init, there is now a out/err = "" if it is None

dbeltrankyl:
Thanks  @LuiggiTenorioK 

changes done

dbeltrankyl:
Ah, didn't adapt the tests, let me do it ...

dbeltrankyl:
> Ah, didn't adapt the tests, let me do it ...

Should pass now

dbeltrankyl:
Ready to re-review, btw ci/cd is passing

LuiggiTenorioK (approved):
I re-ran the same tests from the last review, and everything works as expected!

The code seems good, and the comments have been resolved. So, we can approve this ✔️ 

Thanks @dbeltrankyl. Great work!

dbeltrankyl:
rebased due to the changelog conflict, merging, thanks!

PR #3066, opened 2026-06-17
Fixes hpc2020 issues on the hubs

dbeltrankyl:
Objective: Apply SSL handshake retry to all ECACCESS commands previously; only `ecaccess-job-submit` passed  `-retry` for SSL handshake retries. 

Now every ECACCESS command gets the retry flag, sourced from the new `ECACCESS_RETRIES` platform config key (default 100, matching the old hardcoded value on submit).

docs:
https://confluence.ecmwf.int/display/ECAC/ecaccess-file-get

Note:
* 0 is a valid number for the --retry
* All commands accept this flag ( transport layer flag)

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@
 - Jobs now validated individually via STAT files in check_jobs to confirm completion or failure, for both wrapped and non-wrapped jobs #3007
 - Expanded ECPlatform, paramiko platform support, and platform code cleanup #3007
 - Changed JOBS_IN_WRAPPER unknown job from critical error to warning #3006
+- Added ``ECACCESS_RETRIES`` platform config key to configure SSL retries for ecaccess platforms. Defaults to 100 #3059
 
 ### 4.1.16: Postgres (experimental) support, bug fixes, and enhancements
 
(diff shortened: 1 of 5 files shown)


dbeltrankyl:
Waiting for the CI/CD, but it should be **ready** to review. 

I've assigned @kinow because he reviewed the --retry flag when I added it in the submit. 

Discussing with @eferre1, the fix worked, but then another command failed, so I've applied the --retry flag to all the commands

I would like to do a dev3 release today, so if  @kinow has time to review it.. :pray: if not, I can assign it to another person

kinow (approved):
Looks good to me. The tests cannot cover all of ecplatform as it's quite hard to test. But the fall back to 100 doesn't seem to be tested, and I guess that one could be via a quick mock in a unit test, perhaps, @dbeltrankyl ? if not, all good too. I assume you tested this in the hpc2020. Leaving it as approved! Thanks!

dbeltrankyl:
Thanks @kinow 

Added!, I'll merge this if the cov reports 100% and CI/CD passes

kinow:
> Thanks @kinow
> 
> Added!, I'll merge this if the cov reports 100% and CI/CD passes

I think there was another part that was not tested, but it didn't seem worth the effort to cover that, and low risk if you ran it on hpc2020. This other if that you added I think could break if it failed (although I think it'd be hard/not possible... but still). If you get a green build and the if you added is now covered, I'm +1 even if there are a couple of other lines not covered in the ecplatform :+1:  Thanks Dani!

PR #3069, opened 2026-06-17
4.1.17 - infinite loop fix 

dbeltrankyl:
@agarci3 reported seeing jobs get permanently stuck in RUNNING status. 

The root cause is that when a job finishes but its STAT file still shows RUNNING (typically because the job's trap failed and did not write the final COMPLETED/FAILED line, which is another separated issue cc: @kinow, @agarci3 will open an issue for that), the old code trusted the STAT file unconditionally. It returned RUNNING no matter what the scheduler said. The IO_SAFE_WAIT fallback that should catch this never triggered because RUNNING was grouped together with COMPLETED and FAILED in the same early return branch.

I've extracted the logic outside the check_job function to add an unit test

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - Fixed inconsistent log paths in status txt output #3040
 - Follow-up fix for #3007: recover job times that were lost when an experiment was interrupted.
 - Fixed `num_max_spits` unit conversion to months when `split_unit` is "month" #2944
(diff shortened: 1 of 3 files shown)


dbeltrankyl:
Ready to review @kinow , @LuiggiTenorioK

kinow (approved):
Fix looks good. The new function appears to do most of what was done before, with an extra check if the status is running, and it's also setting the finished time to `None`. I think it it's all it takes to get the scheduler to update the jobs correct, and avoid the infinite loop. Tests appear to test that too, so +1. Thanks @dbeltrankyl !

PR #3074, opened 2026-06-23
updated documentation for platform connections under developers guide 

### LINKED ISSUE #2590: Rename PJM to Technical Compute Suite?
@manuel-g-castro found out that the actual name of the PJM scheduler appears to be Technical Compute Suite (source: https://earth.bsc.es/gitlab/mgimenez/phd-project-management/-/issues/18#note_389610).

If confirmed, then I guess we should rename our PJM to ... TCM? 😅

### LINKED ISSUE #3042: [documentation] Platform connections in Autosubmit
I'm opening this issue to document aspects related to platform connections in Autosubmit.

We have some experiments that will become operational at AEMET, and they need to know the following:

1) How many times is the `permission_checker_azxbyc` file created?

As far as I know, this file is created in the `Platform.check_remote_permissions` function, which is called from `Autosubmit.restore_platforms`. This function is executed in the following cases:

- **`autosubmit setstatus`**: The file is created for every platform that has jobs with status QUEUING, SUBMITTED or RUNNING (regardless of the specified job filter). We run `autosubmit setstatus` every day before `autosubmit create`. Since all jobs have status WAITING or READY, we think this case is not affecting our operational runs.
- **`autosubmit run`**:
  - At the start of the run.
  - For every retry if there is an error in the Autosubmit main loop. We suspect the file is created twice here, as the code contains a call to `prepare_run`, which calls `restore_platforms`.
  - After the main loop, if there are logs to transfer, Autosubmit calls this function before attempting to transfer the platform logs.
- **`autosubmit stop`**: Only when cancelling Slurm jobs.

2) How many connections are made during a run with Autosubmit? If retries are enabled, are these connections stopped between one retry and the next?

@kinow helped me a lot with this (thanks again :) ), the information above is a summary of what I learned with him. If you notice anything missing or something that is not well explained, feel free to add whatever you need.

Thanks!

rajeshsuraj:
Closes #3042
Closes #2590

**Check List**

- [x] Documentation updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@ man/
 venv/
 .ipynb_checkpoints
 .atl
+.benchmarks/artifacts/performance-metrics-4.2.0.csv
+:memory:
--- a/.gitignore
+++ b/.gitignore
@@ -26,10 +26,6 @@ test/coverage.xml
 coverage.xml
 test/htmlcov
 
-# Benchmark files
-.benchmarks/artifacts
-:memory:
-
 # Docker
 dockerfiles/id_rsa*
 dockerfiles/authorized_keys
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,10 @@ test/coverage.xml
 coverage.xml
 test/htmlcov
 
+# Benchmark files
+.benchmarks/artifacts
+:memory:
+
 # Docker
 dockerfiles/id_rsa*
 dockerfiles/authorized_keys
@@ -36,5 +40,3 @@ man/
 venv/
 .ipynb_checkpoints
 .atl
-.benchmarks/artifacts/performance-metrics-4.2.0.csv
-:memory:
--- a/autosubmit/platforms/headers/pjm_header.py
+++ b/autosubmit/platforms/headers/pjm_header.py
@@ -21,7 +21,7 @@ import textwrap
 
 
 class PJMHeader(object):
-    """Class to handle the PJM headers of a job"""
+    """Class to handle the PJM (also known as TCM Technical Computing Suite) headers of a job"""
 
     # noinspection PyMethodMayBeStatic,PyUnusedLocal
     def get_queue_directive(self, job, parameters):
--- a/autosubmit/platforms/headers/pjm_header.py
+++ b/autosubmit/platforms/headers/pjm_header.py
@@ -21,7 +21,7 @@ import textwrap
 
 
 class PJMHeader:
-    """Class to handle the PJM headers of a job"""
+    """Class to handle the PJM (also known as TCM Technical Computing Suite) headers of a job"""
 
     # noinspection PyMethodMayBeStatic,PyUnusedLocal
     def get_queue_directive(self, job, parameters):
--- a/docs/source/devguide/platforms/index.rst
+++ b/docs/source/devguide/platforms/index.rst
@@ -315,10 +315,15 @@ a directory, not a file; the name it uses depends on the platform type:
      - The local platform does not perform a write-permission check.
 
 .. note::
-    Operators searching log files or the remote scratch filesystem need to
-    look for all three directory names, because each platform type uses a
-    different one. A search restricted to ``permission_checker_azxbyc`` will
-    miss probes from ``ps`` and ``ecaccess`` platforms.
+    The probe directory is transient: Autosubmit creates it, confirms that
+    the create+delete succeeded, and removes it in the same call. Under
+    normal operation there is nothing left on the remote filesystem to
+    find afterwards. The directory names above are documented so that
+    operators searching Autosubmit's local logs (or investigating a rare
+    case where a probe crashed between the create and delete steps and
+    left the directory behind) know that each platform type uses a
+    different name — a grep restricted to ``permission_checker_azxbyc``
+    will miss references from ``ps`` and ``ecaccess`` platforms.
 
 When the write-permission check runs
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -333,15 +338,18 @@ frequency depends on the command:
    * - Command
      - When the probe runs
    * - ``autosubmit run``
-     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error and it is recovered, the probe runs again for each new attempt on each of the platforms. The number of retries is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``), so a prolonged outage can produce many probe entries in the logs.
+     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error, Autosubmit enters a reconnection loop that retries the SSH connection up to ``CONFIG.RECOVERY_RETRIALS`` **times** (a count of attempts, not a time budget; default ``3650``, with a per-attempt sleep that grows from 15s to a cap of 120s, so the default corresponds to roughly 72–122 hours of outage tolerance). The probe only runs on the retry where the connection is actually re-established, not on the failing attempts. In practice this means a prolonged outage produces many SSH retry log entries but at most one probe entry per platform per recovery episode.
    * - ``autosubmit setstatus``
      - Once per platform that currently has jobs in ``QUEUING``, ``SUBMITTED`` or ``RUNNING`` state. When ``setstatus`` is run before ``autosubmit create`` (the typical operational case, where all jobs are still ``WAITING`` or ``READY``) no probe is created.
    * - ``autosubmit stop --cancel``
      - Once per configured platform. This applies to any scheduler, not only Slurm; the trigger is the ``--cancel`` flag, not the platform type.
 
-The probe does **not** run during the main loop of ``autosubmit run`` after
-startup, during ``autosubmit recovery``, during ``autosubmit create``, or
-when Autosubmit reconnects after a transient network error mid-run.
+The probe does **not** run during the normal main loop of ``autosubmit
+run`` after startup, during ``autosubmit recovery``, during ``autosubmit
+create``, or when an individual SSH command inside the main loop fails
+with a network error and its channel is silently re-established (that
+reconnect path bypasses the ``restore_platforms`` step and therefore
+skips the probe).
 
 .. _stale_job_data_recovery:
 
@@ -355,11 +363,12 @@ recorded), Autosubmit will try to fetch the missing values from the remote
 ``STAT`` file for that job. This recovery runs during ``autosubmit run``,
 ``autosubmit create``, ``autosubmit setstatus`` and ``autosubmit recovery``.
 
-The path opens SSH sessions to the affected platforms when needed, but it
-does not run the write-permission probe. ``autosubmit create`` is the case
-worth flagging here: it normally appears to be a local operation, but if
-there are stale rows in the history database it will open one SSH session
-per affected platform.
+When Autosubmit performs this stale-job-data recovery it opens an SSH
+session to each affected platform, but it does not run the
+write-permission probe. ``autosubmit create`` is the case worth flagging
+here: it normally appears to be a local operation, but if there are stale
+rows in the history database it will open one SSH session per affected
+platform.
 
 Connections opened during a run
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -408,18 +417,17 @@ before exiting.
     The close operations are best-effort. If the underlying transport is
     already broken, the close will fail silently and the next connection
     will be opened anyway. This prevents a stuck close from blocking
-    recovery, but it means operators who want to confirm that no SSH
+    recovery, but it means users who want to confirm that no SSH
     sessions have leaked should spot-check the login node directly (for
     example with ``who`` or ``ss``) rather than rely on Autosubmit's
     logs alone.
 
 The reconnection behaviour described above applies after the workflow has
 started running. If the very first connection attempt at the start of
-``autosubmit run`` fails — for example because the HPC login node is
-briefly unreachable — Autosubmit makes two internal retries within a
-single connection attempt, but does not currently retry the whole
-``autosubmit run`` startup. If both internal retries fail, the run
-aborts. Adding a workflow-level startup retry is tracked separately.
+``autosubmit run`` fails, Autosubmit makes two internal retries within a
+single connection attempt and then aborts if both fail — see
+:ref:`userguide/run/index:retries` for the retry configuration Autosubmit
+exposes at the workflow level.
 
 Filesystem operations during a run
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -428,28 +436,41 @@ Beyond handling write-permission probes and SSH sessions, Autosubmit performs se
 
 .. list-table::
    :header-rows: 1
-   :widths: 32 68
+   :widths: 26 18 56
 
    * - Operation
+     - Command(s)
      - When it happens
    * - Upload of ``.cmd`` job scripts
+     - ``autosubmit run``
      - One upload per job script per submission cycle. A submission cycle happens for every main loop iteration that has new ``READY`` jobs to dispatch.
    * - Creation of the remote log directory
+     - ``autosubmit run``
      - Once per platform at startup, and then once per upload as part of the upload itself. ``ecaccess`` platforms create one directory per path level because their ``mkdir`` does not have a recursive option.
    * - Listing of ``_COMPLETED`` marker files
+     - ``autosubmit run``
      - Once per platform per status-check cycle, to detect newly completed jobs.
    * - Download of ``.out`` / ``.err`` log files
+     - ``autosubmit run``
      - Two downloads per completed job (one for standard output, one for standard error). These run in the per-platform log-recovery process, in parallel with the main loop.
-   * - Download of job ``STAT`` files
-     - One download per job at the end of execution. Additional opportunistic downloads happen when the history database contains incomplete rows.
-   * - Reads of remote ``STAT`` files
+   * - Download of ``STAT`` files at job end
+     - ``autosubmit run``
+     - One download per job at the end of execution, as part of the completed-job handling in the main loop.
+   * - Download of ``STAT`` files for stale history rows
+     - ``autosubmit run``, ``autosubmit create``, ``autosubmit setstatus``, ``autosubmit recovery``
+     - Opportunistic download of missing ``STAT`` files when the history database contains rows without start/end timestamps. Runs before and after the main loop of ``autosubmit run``, at the start of ``autosubmit create`` and ``autosubmit setstatus``, and during ``autosubmit recovery``.
+   * - Batched reads of remote ``STAT`` files
+     - ``autosubmit run``
      - One batched read per status-check cycle, covering all relevant ``STAT`` files in a single round-trip.
    * - Cleanup of previous-run marker files
-     - Once per platform at the start of ``autosubmit run``, scoped to the experiment's log directory.
+     - ``autosubmit run``
+     - Once per platform at startup, scoped to the experiment's log directory.
    * - Compression of remote logs
-     - One compression command per log file, only when ``COMPRESS_REMOTE_LOGS`` is enabled on the platform.
+     - ``autosubmit run``
+     - One compression command per log file, only when ``COMPRESS_REMOTE_LOGS`` is enabled on the platform. Runs in the log-recovery process, in parallel with the main loop.
    * - Write-permission probe
-     - See above.
+     - ``autosubmit run``, ``autosubmit setstatus``, ``autosubmit stop --cancel``
+     - See :ref:`When the write-permission check runs <platform_connections>` above.
 
 A few patterns worth knowing when reasoning about filesystem load:
 
@@ -463,10 +484,6 @@ A few patterns worth knowing when reasoning about filesystem load:
 - **``STAT`` files are written by the jobs, not by Autosubmit.** The remote
   job script appends a Unix epoch timestamp at start and finish; Autosubmit
   only reads these files.
-- **Wrappers reduce per-job I/O.** A wrapped package submits one script for
-  many jobs and produces consolidated log and ``STAT`` output, which cuts
-  the per-job upload and download counts roughly in proportion to the
-  number of jobs in the wrapper.
 
 .. note::
     The table above describes which operations Autosubmit performs and
--- a/docs/source/devguide/platforms/index.rst
+++ b/docs/source/devguide/platforms/index.rst
@@ -23,16 +23,16 @@ platform class, for which first we need to identify which existing platform is t
       directly on the machine where Autosubmit is running. No SSH, no
       scheduler. Used for local preparation steps in a workflow.
     - :mod:`PS Platform <autosubmit.platforms.psplatform>` — a remote host with
-      no batch scheduler. Connects via SSH and tracks jobs as remote OS
+      no scheduler, as it uses ``PowerShell``. Connects via SSH and tracks jobs as remote OS
       processes (``ps``).
     - :mod:`EC Platform <autosubmit.platforms.ecplatform>` — ECMWF systems
       reached through the ``ecaccess`` toolchain. The underlying scheduler can
       be ``pbs``, ``loadleveler`` or ``slurm``, selected via the platform's
       ``VERSION`` field.
     - :mod:`PJM Platform <autosubmit.platforms.pjmplatform>` — Fujitsu's PJM
-      scheduler, used on Fugaku and similar systems. Connects via SSH.
-    - :mod:`PBS Platform <autosubmit.platforms.pbsplatform>` — PBS Pro /
-      OpenPBS scheduler. Connects via SSH.
+      scheduler, used on Fugaku and others large-scale supercomputers. Connects via SSH.
+    - :mod:`PBS Platform <autosubmit.platforms.pbsplatform>` — Portable Batch System (PBS) / PBS Pro /
+      OpenPBS scheduler used on Miyabi. Connects via SSH.
     - :mod:`Slurm Platform <autosubmit.platforms.slurmplatform>` — the Slurm
       workload manager (MareNostrum, Leonardo, LUMI, etc.). Connects via SSH.
       Used as the worked example in this guide.
@@ -157,7 +157,7 @@ the timestamp to TOTAL_STATS file and jobs_data.db properly.
 
     if job_data_dc and type(self.platform) is not str and (self.platform.type in ["slurm", "example"]):
         thread_write_finish = Thread(target=ExperimentHistory(self.expid, jobdata_dir_path=BasicConfig.JOBDATA_DIR, historiclog_dir_path=BasicConfig.HISTORICAL_LOG_DIR).write_platform_data_after_finish, args=(job_data_dc, self.platform))
-        thread_write_finish.name = "JOB_data_{}".format(self.name)
+        thread_write_finish.name = f"JOB_data_{self.name}"
         thread_write_finish.start()
 
 ``autosubmit/job/job.py`` in `line 2817 <https://github.com/BSC-ES/autosubmit/blob/v4.1.13/autosubmit/job/job.py#L2817>`_ add a new validation for the validation of the queue
@@ -284,19 +284,19 @@ Platform Connections
 ----------------------
 
 This section documents how Autosubmit opens connections to the configured
-platforms, when it verifies that it has write permissions on the remote
-filesystem, and how those connections are torn down when the workflow ends or
-recovers from an error.  The intended audience is operators running Autosubmit
-in production who need to reason about the connection
+platforms, when it verifies that it has to write permissions on the remote
+filesystem, and how it cleans up and closes those connections upon completion or
+recovers from an error.  The intended audience are operators running Autosubmit
+in production who need to understand about the connection
 footprint that Autosubmit places on a login node.
 
 The permission probe directory
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To verify that Autosubmit can write to the configured ``SCRATCH_DIR``, each
-platform exposes a ``check_remote_permissions`` method.  The method creates a
+platform exposes a ``check_remote_permissions`` method. The method creates a
 small probe **directory** (not a file) under
-``<scratch_dir>/<project>/<user>/`` and immediately removes it.  The name of
+``<scratch_dir>/<project>/<user>/`` and immediately removes it. The name of
 the probe directory and the mechanism used to create it depend on the platform
 type:
 
@@ -323,7 +323,7 @@ type:
 .. note::
     Operators searching log files or running ``find`` on the remote scratch
     filesystem must check for **all three** directory names depending on which
-    platform types are in use.  A grep that looks only for
+    platform types are in use. A grep that looks only for
     ``permission_checker_azxbyc`` will miss every ``ps`` and ``ecaccess``
     platform probe.
 
@@ -331,7 +331,7 @@ When the permission probe runs
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The probe is only created from inside ``Autosubmit.restore_platforms``, which
-is reached from four user-facing commands.  For each call site below, the
+is reached from four user-facing commands. For each call site below, the
 ``check_remote_permissions`` method is invoked once per platform listed in
 ``platforms_to_test``.
 
@@ -344,10 +344,10 @@ is reached from four user-facing commands.  For each call site below, the
      - Conditions and per-run frequency
    * - ``autosubmit run``
      - ``Autosubmit.prepare_run`` (only when ``recover=False``)
-     - Once per platform at the start of the run.  ``prepare_run`` is **not** called recursively on retry, so the probe is not duplicated through this path during recovery.
+     - Once per platform at the start of the run. ``prepare_run`` is **not** called recursively on retry, so the probe is not duplicated through this path during recovery.
    * - ``autosubmit run`` (recovery)
      - ``Autosubmit.run_experiment`` recovery block
-     - Whenever the main loop catches an ``AutosubmitError``, a dedicated reconnection loop calls ``restore_platforms`` once per iteration until reconnection succeeds.  The loop is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``).  Under sustained platform outages this is the dominant contributor to the probe count.
+     - Whenever the main loop catches an ``AutosubmitError``, a dedicated reconnection loop calls ``restore_platforms`` once per iteration until reconnection succeeds. The loop is set by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``). Under sustained platform outages this is the dominant contributor to the probe count.
    * - ``autosubmit setstatus``
      - ``Autosubmit.set_status``
      - Once per platform whose jobs include at least one job in status ``QUEUING``, ``SUBMITTED`` or ``RUNNING``.  If ``setstatus`` is invoked when all jobs are ``WAITING`` or ``READY`` (the typical operational case immediately before ``autosubmit create``), ``platforms_to_test`` is empty and no probe runs.
@@ -357,8 +357,8 @@ is reached from four user-facing commands.  For each call site below, the
 
 The probe is **not** created by:
 
-- The main running loop of ``autosubmit run`` after startup.  Status checks,
-  job submission, file transfers and log retrieval all reuse the existing
+- The main running loop of ``autosubmit run`` after startup. "Status checks",
+  "job submission", "file transfers", and "log retrieval all" reuse the existing
   connection without re-running the probe.
 - Mid-loop SSH reconnections triggered by ``ParamikoPlatform.exec_command``
   when a command fails with a network error.  These call ``restore_connection``
@@ -367,7 +367,7 @@ The probe is **not** created by:
   ``platform.test_connection`` directly, opening the SSH connection but not
   invoking ``restore_platforms``.
 - The stale job-data recovery path (see :ref:`stale_job_data_recovery` below).
-- The log transfer that happens at the end of a run.  Log recovery uses a
+- The log transfer that happens at the end of a run. Log recovery uses a
   separate, already-connected subprocess and does not run the probe.
 
 .. _stale_job_data_recovery:
@@ -377,7 +377,7 @@ Stale job-data recovery (separate connection path)
 
 The helper :func:`autosubmit.helpers.utils.recover_stale_job_data` is invoked
 during ``autosubmit run`` (before and after the main loop), ``autosubmit
-create``, ``autosubmit setstatus`` and ``autosubmit recovery``.  When stale
+create``, ``autosubmit setstatus`` and ``autosubmit recovery``. When stale
 rows are found in the historical job-data database, it builds a fresh
 platform object via ``build_and_connect_platform`` and calls
 ``platform.restore_connection`` to download the missing STAT files.
@@ -401,27 +401,27 @@ For platforms backed by Paramiko (``slurm``, ``pjm``, ``pbs``, ``ps``), at
 startup of ``autosubmit run`` Autosubmit opens **two SSH sessions per
 platform**:
 
-1. The **main session**, owned by the main Autosubmit process.  This is the
+1. The **main session**, owned by the main Autosubmit process. This is the
    ``paramiko.SSHClient`` together with its ``Transport`` and the
    ``SFTPClient`` (``self._ftpChannel``).  It is created by
    ``restore_platforms`` → ``test_connection`` → ``restore_connection`` →
    ``connect``.
 2. A **log-recovery session**, owned by a per-platform background subprocess
-   spawned by ``spawn_log_retrieval_process``.  The subprocess calls its own
+   spawned by ``spawn_log_retrieval_process``. The subprocess calls its own
    ``restore_connection`` against the same platform and is responsible for
    downloading job log files asynchronously while the main loop continues.
 
 .. note::
     The log-recovery subprocess is **only** spawned during ``autosubmit run``.
     The spawn is guarded by ``AS_COMMAND == "run"`` in the experiment misc
-    configuration.  Commands such as ``setstatus``, ``recovery``, ``stop
+    configuration. Commands such as ``setstatus``, ``recovery``, ``stop
     --cancel`` and the stale-job-data path open only the main session per
     platform — no per-platform subprocess.
 
 For ``ecaccess`` platforms, neither ``connect`` nor ``restore_connection``
-opens an SSH session.  They run ``ecaccess-gateway-connected`` locally via
-``subprocess`` and set ``self.connected`` from the output.  All subsequent
-operations are shelled out as local ``ecaccess-*`` commands.  An ``ecaccess``
+opens an SSH session. They run ``ecaccess-gateway-connected`` locally via
+``subprocess`` and set ``self.connected`` from the output. All subsequent
+operations are shelled out as local ``ecaccess-*`` commands. An ``ecaccess``
 platform therefore contributes zero SSH sessions to the login node but does
 contribute gateway traffic.
 
@@ -440,8 +440,8 @@ opening a new one (see below).
 Are connections closed between retries?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Yes.  Whenever Autosubmit needs to re-establish a connection, the previous
-session is explicitly torn down before the new one is opened.  The teardown
+Yes. Whenever Autosubmit needs to re-establish a connection, the previous
+session is explicitly interrupted before the new one is opened. The tear-down
 chain in ``ParamikoPlatform`` is:
 
 ``test_connection`` (when ``self.connected`` is ``False``) →
@@ -451,7 +451,7 @@ both ``Transport`` objects, and the ``SSHClient``) →
 ``with suppress(Exception): self.reset()`` (defensive second close) →
 ``self.connect(...)`` (opens a new ``SSHClient``).
 
-The log-recovery subprocess follows the same discipline on its side.  When the
+The log-recovery subprocess follows the same discipline on its side. When the
 main process calls ``Autosubmit.refresh_log_recovery_process`` (once per main
 loop iteration), the helper invokes ``p.clean_log_recovery_process()`` for
 each platform if its subprocess has died.  ``clean_log_recovery_process``
@@ -467,7 +467,7 @@ At the end of a successful run, ``autosubmit.py`` issues an explicit
 
 .. warning::
     All ``close_connection`` calls are wrapped in ``with suppress(Exception):``
-    blocks.  If the close itself fails silently (for example because the
+    blocks. If the close itself fails silently (for example because the
     transport is already half-dead), the next connection is opened anyway.
     This is by design — it prevents a stuck close from blocking recovery —
     but it means operators should not rely on Autosubmit logs alone to detect
--- a/docs/source/devguide/platforms/index.rst
+++ b/docs/source/devguide/platforms/index.rst
@@ -284,9 +284,7 @@ Platform Connections
 This section describes how Autosubmit interacts with the platforms it has been
 configured to use: when it opens connections, when it checks that it has
 write access to the remote filesystem, and what it does when a connection has
-to be re-established mid-run. It is written for operators and workflow
-developers who need to understand the footprint Autosubmit places on a remote
-login node, not for Autosubmit developers reading the source.
+to be re-established mid-run.
 
 The write-permission check
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -427,12 +425,7 @@ aborts. Adding a workflow-level startup retry is tracked separately.
 Filesystem operations during a run
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Beyond the write-permission probe and the SSH sessions themselves,
-Autosubmit performs a number of file operations against the remote scratch
-filesystem during a run. The table below describes which operations happen
-and when. The absolute numbers depend on the workflow shape (number of
-jobs, retries, wrapper usage and recovery events), so the descriptions are
-qualitative.
+Beyond handling write-permission probes and SSH sessions, Autosubmit performs several file operations against the remote ``<SCRATCH_DIR>`` during each run. The exact number of these operations can vary based on the workflow's complexity, including factors such as the number of jobs, retries, wrapper usage, and recovery events. The table below outlines the types of operations performed at different stages, with descriptions provided in qualitative terms to reflect this variability.
 
 .. list-table::
    :header-rows: 1
--- a/docs/source/devguide/platforms/index.rst
+++ b/docs/source/devguide/platforms/index.rst
@@ -289,7 +289,7 @@ This section documents how Autosubmit opens connections to the configured
 platforms, when it verifies that it has write permissions on the remote
 filesystem, and how those connections are torn down when the workflow ends or
 recovers from an error.  The intended audience is operators running Autosubmit
-in production (for example AEMET) who need to reason about the connection
+in production who need to reason about the connection
 footprint that Autosubmit places on a login node.
 
 The permission probe directory
@@ -475,6 +475,92 @@ At the end of a successful run, ``autosubmit.py`` issues an explicit
     but it means operators should not rely on Autosubmit logs alone to detect
     leaked sessions on the gateway.  Spot-checking with ``who`` or ``ss`` on
     the login node is recommended for long-running operational experiments.
+
+The reconnection logic described in this subsection applies to connection
+losses that occur **after** the workflow has started running.  If the very
+first connection attempt to a platform fails at the start of ``autosubmit
+run`` — for example because the HPC login node is briefly unreachable —
+the ``restore_connection`` method inside ``ParamikoPlatform`` retries
+twice internally, but ``Autosubmit.run_experiment`` does not currently
+wrap the initial ``prepare_run`` call in a retry loop, so the run aborts
+if both internal retries fail.  Adding a startup-time retry loop is
+tracked separately.
+
+Filesystem I/O footprint
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Beyond the permission probe and the SSH connections themselves, Autosubmit
+performs a number of file operations against the remote scratch filesystem
+during a run.  Operators concerned with filesystem load can use the table below to map each operation
+to the code path that triggers it and to the cadence at which it runs.
+The frequencies are described qualitatively because the absolute count
+per run depends on the workflow shape — number of jobs, retries, wrapper
+usage, recovery events — rather than on Autosubmit alone.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 26 22 52
+
+   * - Operation
+     - Code path
+     - When it happens
+   * - Upload ``.cmd`` job scripts
+     - ``Platform.send_file`` (called from ``JobPackageBase.send_files`` inside ``Platform.prepare_submission``)
+     - One SFTP ``put`` per job script per submission cycle.  A submission cycle happens for every main loop iteration that has ``READY`` jobs to dispatch.
+   * - Create remote log directory
+     - ``ParamikoPlatform.check_remote_log_dir`` (called inside ``send_file``)
+     - One ``mkdir -p`` (or equivalent) per ``send_file`` call.  For ``ecaccess`` platforms, one ``ecaccess-file-mkdir`` per path level (no ``mkdir -p``).
+   * - Cancel-script upload
+     - ``Platform.remove_multiple_files``
+     - One SFTP ``put`` of ``multiple_delete_previous_run.sh`` per platform when stale files from a previous run need to be removed at startup.
+   * - Find ``_COMPLETED`` marker files
+     - ``ParamikoPlatform.check_completed_files`` / ``get_completed_job_names``
+     - One ``find`` (SSH) or ``ecaccess-file-dir`` (ecaccess) per platform per check cycle.  Used to detect newly completed jobs.
+   * - Download ``.out`` / ``.err`` log files
+     - ``ParamikoPlatform.get_file`` (called from the log-recovery subprocess via ``recover_job_log`` → ``Job.retrieve_logfiles``)
+     - Two SFTP ``get`` operations per completed job (one for stdout, one for stderr).  These run in the per-platform log-recovery subprocess, asynchronously to the main loop.
+   * - Download ``STAT`` files
+     - ``ParamikoPlatform.get_stat_file`` and ``recover_stale_job_data``
+     - One SFTP ``get`` per job at the end of execution (run-time STAT) plus opportunistic downloads via ``recover_stale_job_data`` when historical rows are incomplete.
+   * - Read ``STAT`` files in place
+     - ``ParamikoPlatform.confirm_done_jobs_via_stat`` and ``ParamikoPlatform.set_start_time_from_remote_stat_file``
+     - One batched ``awk``/``head`` SSH command per ``check_all_jobs`` cycle, reading all relevant STAT files in a single round-trip.
+   * - Delete previous-run marker files
+     - ``ParamikoPlatform.delete_previous_run_files_by_job_names`` and ``delete_previous_stat_files_by_job_names``
+     - One ``find … -delete`` per platform at startup of ``autosubmit run``, scoped to the experiment's log directory.
+   * - Compress remote logs (optional)
+     - ``ParamikoPlatform.compress_file``
+     - One ``xz`` or ``gzip`` SSH command per log file when ``COMPRESS_REMOTE_LOGS`` is enabled on the platform.
+   * - Cleanup probe (write-permission check)
+     - ``Platform.check_remote_permissions``
+     - See :ref:`platform_connections` above — once per platform per ``restore_platforms`` call.
+
+A few patterns worth knowing when reasoning about load:
+
+- **Job submission and log retrieval are decoupled.**  The main loop
+  submits and queries job status; the per-platform log-recovery subprocess
+  performs all log downloads in parallel.  Filesystem activity from the
+  two sources runs concurrently against the same remote directory.
+- **Status queries are batched.**  ``check_all_jobs`` runs a single
+  scheduler query (``squeue``, ``pjstat``, ``qstat``, or an
+  ``ecaccess-job-list`` invocation) per platform per cycle, regardless of
+  how many jobs are in flight.  ``confirm_done_jobs_via_stat`` similarly
+  issues a single batched ``awk`` command rather than one per file.
+- **STAT files are written by the job, not by Autosubmit.**  The remote
+  job script appends a Unix epoch timestamp to its STAT file at start and
+  another at finish; Autosubmit only reads these files.
+- **Wrappers reduce per-job I/O.**  A wrapped package submits one ``.cmd``
+  file for many jobs and produces consolidated log and STAT output, which
+  cuts the per-job upload and download counts roughly in proportion to
+  wrapper size.
+
+.. note::
+    The table above describes **which** operations Autosubmit performs and
+    **when**, but does not give absolute numbers (bytes transferred, IOPS,
+    syscall counts) for a given workflow.  Producing those numbers
+    requires either instrumenting Autosubmit with an I/O tracer or
+    running a representative workflow under ``strace`` / ``iostat``
+    against a controlled HPC.
 
 How to generate a new experiment
 ------------------------------------
--- a/docs/source/devguide/platforms/index.rst
+++ b/docs/source/devguide/platforms/index.rst
@@ -26,7 +26,7 @@ platform class, for which first we need to identify which existing platform is t
       current Atos systems; ``pbs`` is also supported for older
       configurations).
     - :mod:`PJM Platform <autosubmit.platforms.pjmplatform>` — Fujitsu's PJM
-      scheduler, used on Fugaku and other large-scale supercomputers.
+      scheduler, used on Fugaku and other large-scale supercomputers. (also known as TCM Technical Computing Suite)
       Connects via SSH.
     - :mod:`PBS Platform <autosubmit.platforms.pbsplatform>` — Portable Batch
       System (PBS) / PBS Pro / OpenPBS scheduler, used on systems such as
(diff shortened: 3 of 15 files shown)


docs/source/devguide/platforms/index.rst
```diff
@@ -6,23 +6,39 @@ Platforms
...
-    |br| :ref:`Local Platform <Local Platform>` :mod:`Platform <autosubmit.platforms.platform>`
-    |br| :mod:`EC Platform <autosubmit.platforms.ecplatform>` :mod:`PJM Platform <autosubmit.platforms.pjmplatform>`
-    |br| :mod:`Slurm Platform <autosubmit.platforms.slurmplatform>`
+
+    - :mod:`Local Platform <autosubmit.platforms.locplatform>` — runs jobs
+      directly on the machine where Autosubmit is running. No SSH, no
+      scheduler. Used for local preparation steps in a workflow.
+    - :mod:`PS Platform <autosubmit.platforms.psplatform>` — a remote host with
+      no batch scheduler. Connects via SSH and tracks jobs as remote OS
+      processes (``ps``).
```
VindeeR:
```suggestion
    - :mod:`PS Platform <autosubmit.platforms.psplatform>` — a remote host with
      no scheduler, as it uses ``PowerShell``. Connects via SSH and tracks jobs as remote OS
      processes (``ps``).
```

docs/source/devguide/platforms/index.rst
```diff
@@ -6,23 +6,39 @@ Platforms
...
+    - :mod:`EC Platform <autosubmit.platforms.ecplatform>` — ECMWF systems
+      reached through the ``ecaccess`` toolchain. The underlying scheduler can
+      be ``pbs``, ``loadleveler`` or ``slurm``, selected via the platform's
+      ``VERSION`` field.
+    - :mod:`PJM Platform <autosubmit.platforms.pjmplatform>` — Fujitsu's PJM
+      scheduler, used on Fugaku and similar systems. Connects via SSH.
```
VindeeR:
```suggestion
    - :mod:`PJM Platform <autosubmit.platforms.pjmplatform>` — Fujitsu's PJM
      scheduler, used on Fugaku and others large-scale supercomputers. Connects via SSH.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -6,23 +6,39 @@ Platforms
...
+    - :mod:`PBS Platform <autosubmit.platforms.pbsplatform>` — PBS Pro /
+      OpenPBS scheduler. Connects via SSH.
```
VindeeR:
```suggestion
    - :mod:`PBS Platform <autosubmit.platforms.pbsplatform>` — Portable Batch System (PBS) / PBS Pro /
      OpenPBS scheduler used on Miyabi. Connects via SSH.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -141,8 +157,8 @@ the timestamp to TOTAL_STATS file and jobs_data.db properly.
 
     if job_data_dc and type(self.platform) is not str and (self.platform.type in ["slurm", "example"]):
         thread_write_finish = Thread(target=ExperimentHistory(self.expid, jobdata_dir_path=BasicConfig.JOBDATA_DIR, historiclog_dir_path=BasicConfig.HISTORICAL_LOG_DIR).write_platform_data_after_finish, args=(job_data_dc, self.platform))
-            thread_write_finish.name = "JOB_data_{}".format(self.name)
-            thread_write_finish.start()
+        thread_write_finish.name = "JOB_data_{}".format(self.name)
+        thread_write_finish.start()
```
VindeeR:
```suggestion
        thread_write_finish.name = f"JOB_data_{self.name}"
        thread_write_finish.start()
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
 
+
+.. _platform_connections:
+
+Platform Connections
+----------------------
+
+This section documents how Autosubmit opens connections to the configured
+platforms, when it verifies that it has write permissions on the remote
+filesystem, and how those connections are torn down when the workflow ends or
```
VindeeR:
```suggestion
platforms, when it verifies that it has to write permissions on the remote
filesystem, and how it cleans up and closes those connections upon completion or
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+recovers from an error.  The intended audience is operators running Autosubmit
```
VindeeR:
```suggestion
recovers from an error.  The intended audience are operators running Autosubmit
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+in production who need to reason about the connection
```
VindeeR:
```suggestion
in production who need to understand about the connection
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+   * - Command
+     - Call site
+     - Conditions and per-run frequency
+   * - ``autosubmit run``
+     - ``Autosubmit.prepare_run`` (only when ``recover=False``)
+     - Once per platform at the start of the run.  ``prepare_run`` is **not** called recursively on retry, so the probe is not duplicated through this path during recovery.
+   * - ``autosubmit run`` (recovery)
+     - ``Autosubmit.run_experiment`` recovery block
+     - Whenever the main loop catches an ``AutosubmitError``, a dedicated reconnection loop calls ``restore_platforms`` once per iteration until reconnection succeeds.  The loop is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``).  Under sustained platform outages this is the dominant contributor to the probe count.
```
VindeeR:
```suggestion
     - Whenever the main loop catches an ``AutosubmitError``, a dedicated reconnection loop calls ``restore_platforms`` once per iteration until reconnection succeeds. The loop is set by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``). Under sustained platform outages this is the dominant contributor to the probe count.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+     - Once per platform whose jobs include at least one job in status ``QUEUING``, ``SUBMITTED`` or ``RUNNING``.  If ``setstatus`` is invoked when all jobs are ``WAITING`` or ``READY`` (the typical operational case immediately before ``autosubmit create``), ``platforms_to_test`` is empty and no probe runs.
+   * - ``autosubmit stop --cancel``
+     - ``Autosubmit.stop`` → ``Autosubmit.prepare_run``
+     - Once per platform.  This applies to any scheduler, not only Slurm; the trigger is the ``--cancel`` flag, not the platform type.
+
+The probe is **not** created by:
+
+- The main running loop of ``autosubmit run`` after startup.  Status checks,
+  job submission, file transfers and log retrieval all reuse the existing
+  connection without re-running the probe.
```
VindeeR:
```suggestion
- The main running loop of ``autosubmit run`` after startup. "Status checks",
  "job submission", "file transfers", and "log retrieval all" reuse the existing
  connection without re-running the probe.
```

VindeeR: or

```suggestion
- The main running loop of ``autosubmit run`` after startup. ``Status checks``,
  ``job submission``, ``file transfers``, and ``log retrieval all`` reuse the existing
  connection without re-running the probe.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+    --cancel`` and the stale-job-data path open only the main session per
+    platform — no per-platform subprocess.
+
+For ``ecaccess`` platforms, neither ``connect`` nor ``restore_connection``
+opens an SSH session.  They run ``ecaccess-gateway-connected`` locally via
+``subprocess`` and set ``self.connected`` from the output.  All subsequent
+operations are shelled out as local ``ecaccess-*`` commands.  An ``ecaccess``
+platform therefore contributes zero SSH sessions to the login node but does
+contribute gateway traffic.
+
```
VindeeR:
```suggestion
For ``ecaccess`` platforms, neither ``connect`` nor ``restore_connection``
opens an SSH session.  They run ``ecaccess-gateway-connected`` locally via
``subprocess`` and set ``self.connected`` from the output.  All subsequent
operations are treated as local ``ecaccess-*`` commands.  An ``ecaccess``
platform therefore initializes no SSH sessions but it does generates gateway traffic.

```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+SSH command fails with a network exception; that path calls
+``self.restore_connection(None)`` which closes the existing session before
+opening a new one (see below).
+
+Are connections closed between retries?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Yes.  Whenever Autosubmit needs to re-establish a connection, the previous
+session is explicitly torn down before the new one is opened.  The teardown
+chain in ``ParamikoPlatform`` is:
```
VindeeR:
```suggestion
Yes. Whenever Autosubmit needs to re-establish a connection, the previous
session is explicitly interrupted before the new one is opened. The tear-down
chain in ``ParamikoPlatform`` is:
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+.. note::
+    The log-recovery subprocess is **only** spawned during ``autosubmit run``.
+    The spawn is guarded by ``AS_COMMAND == "run"`` in the experiment misc
+    configuration.  Commands such as ``setstatus``, ``recovery``, ``stop
+    --cancel`` and the stale-job-data path open only the main session per
+    platform — no per-platform subprocess.
+
+For ``ecaccess`` platforms, neither ``connect`` nor ``restore_connection``
+opens an SSH session.  They run ``ecaccess-gateway-connected`` locally via
```
VindeeR:
```suggestion
For ``ecaccess`` platforms, neither ``connect`` nor ``restore_connection``
opens an SSH session. They run ``ecaccess-gateway-connected`` locally via
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+``subprocess`` and set ``self.connected`` from the output.  All subsequent
+operations are shelled out as local ``ecaccess-*`` commands.  An ``ecaccess``
+platform therefore contributes zero SSH sessions to the login node but does
+contribute gateway traffic.
```
VindeeR:
```suggestion
``subprocess`` and set ``self.connected`` from the output. All subsequent
operations are shelled out as local ``ecaccess-*`` commands. An ``ecaccess``
platform therefore contributes zero SSH sessions to the login node but does
contribute gateway traffic.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+   spawned by ``spawn_log_retrieval_process``.  The subprocess calls its own
+   ``restore_connection`` against the same platform and is responsible for
+   downloading job log files asynchronously while the main loop continues.
+
+.. note::
+    The log-recovery subprocess is **only** spawned during ``autosubmit run``.
+    The spawn is guarded by ``AS_COMMAND == "run"`` in the experiment misc
+    configuration.  Commands such as ``setstatus``, ``recovery``, ``stop
+    --cancel`` and the stale-job-data path open only the main session per
+    platform — no per-platform subprocess.
```
VindeeR:
```suggestion
    The log-recovery subprocess is **only** spawned during ``autosubmit run``.
    The spawn is guarded by ``AS_COMMAND == "run"`` in the experiment misc
    configuration. Commands such as ``setstatus``, ``recovery``, ``stop
    --cancel`` and the stale-job-data path open only the main session per
    platform — no per-platform subprocess.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+1. The **main session**, owned by the main Autosubmit process.  This is the
+   ``paramiko.SSHClient`` together with its ``Transport`` and the
+   ``SFTPClient`` (``self._ftpChannel``).  It is created by
+   ``restore_platforms`` → ``test_connection`` → ``restore_connection`` →
+   ``connect``.
+2. A **log-recovery session**, owned by a per-platform background subprocess
+   spawned by ``spawn_log_retrieval_process``.  The subprocess calls its own
+   ``restore_connection`` against the same platform and is responsible for
+   downloading job log files asynchronously while the main loop continues.
```
VindeeR:
```suggestion
1. The **main session**, owned by the main Autosubmit process. This is the
   ``paramiko.SSHClient`` together with its ``Transport`` and the
   ``SFTPClient`` (``self._ftpChannel``).  It is created by
   ``restore_platforms`` → ``test_connection`` → ``restore_connection`` →
   ``connect``.
2. A **log-recovery session**, owned by a per-platform background subprocess
   spawned by ``spawn_log_retrieval_process``. The subprocess calls its own
   ``restore_connection`` against the same platform and is responsible for
   downloading job log files asynchronously while the main loop continues.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+Stale job-data recovery (separate connection path)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The helper :func:`autosubmit.helpers.utils.recover_stale_job_data` is invoked
+during ``autosubmit run`` (before and after the main loop), ``autosubmit
+create``, ``autosubmit setstatus`` and ``autosubmit recovery``.  When stale
+rows are found in the historical job-data database, it builds a fresh
+platform object via ``build_and_connect_platform`` and calls
+``platform.restore_connection`` to download the missing STAT files.
```
VindeeR:
```suggestion
The helper :func:`autosubmit.helpers.utils.recover_stale_job_data` is invoked
during ``autosubmit run`` (before and after the main loop), ``autosubmit
create``, ``autosubmit setstatus`` and ``autosubmit recovery``. When stale
rows are found in the historical job-data database, it builds a fresh
platform object via ``build_and_connect_platform`` and calls
``platform.restore_connection`` to download the missing STAT files.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+  connection without re-running the probe.
+- Mid-loop SSH reconnections triggered by ``ParamikoPlatform.exec_command``
+  when a command fails with a network error.  These call ``restore_connection``
+  directly, bypassing ``restore_platforms``.
+- ``autosubmit recovery``.  This command uses ``online_recovery`` which calls
+  ``platform.test_connection`` directly, opening the SSH connection but not
+  invoking ``restore_platforms``.
+- The stale job-data recovery path (see :ref:`stale_job_data_recovery` below).
+- The log transfer that happens at the end of a run.  Log recovery uses a
+  separate, already-connected subprocess and does not run the probe.
```
VindeeR:
```suggestion
- The log transfer that happens at the end of a run. Log recovery uses a
  separate, already-connected subprocess and does not run the probe.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+     - Conditions and per-run frequency
+   * - ``autosubmit run``
+     - ``Autosubmit.prepare_run`` (only when ``recover=False``)
+     - Once per platform at the start of the run.  ``prepare_run`` is **not** called recursively on retry, so the probe is not duplicated through this path during recovery.
+   * - ``autosubmit run`` (recovery)
+     - ``Autosubmit.run_experiment`` recovery block
+     - Whenever the main loop catches an ``AutosubmitError``, a dedicated reconnection loop calls ``restore_platforms`` once per iteration until reconnection succeeds.  The loop is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``).  Under sustained platform outages this is the dominant contributor to the probe count.
+   * - ``autosubmit setstatus``
+     - ``Autosubmit.set_status``
+     - Once per platform whose jobs include at least one job in status ``QUEUING``, ``SUBMITTED`` or ``RUNNING``.  If ``setstatus`` is invoked when all jobs are ``WAITING`` or ``READY`` (the typical operational case immediately before ``autosubmit create``), ``platforms_to_test`` is empty and no probe runs.
```
VindeeR: have you confirmed the typical operational workflow with the operational?

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+.. list-table::
+   :header-rows: 1
+   :widths: 24 26 50
+
+   * - Command
+     - Call site
+     - Conditions and per-run frequency
+   * - ``autosubmit run``
+     - ``Autosubmit.prepare_run`` (only when ``recover=False``)
+     - Once per platform at the start of the run.  ``prepare_run`` is **not** called recursively on retry, so the probe is not duplicated through this path during recovery.
```
VindeeR:
```suggestion
     - Once per platform at the start of the run. ``prepare_run`` is **not** called recursively on retry, so the probe is not duplicated through this path during recovery.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+    ``permission_checker_azxbyc`` will miss every ``ps`` and ``ecaccess``
+    platform probe.
+
+When the permission probe runs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The probe is only created from inside ``Autosubmit.restore_platforms``, which
+is reached from four user-facing commands.  For each call site below, the
+``check_remote_permissions`` method is invoked once per platform listed in
+``platforms_to_test``.
```
VindeeR:
```suggestion
The probe is only created from inside ``Autosubmit.restore_platforms``, which
is reached from four user-facing commands. For each call site below, the
``check_remote_permissions`` method is invoked once per platform listed in
``platforms_to_test``.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+   * - ``local``
+     - (none)
+     - ``check_remote_permissions`` is a no-op that returns ``True`` without touching the filesystem.
+
+.. note::
+    Operators searching log files or running ``find`` on the remote scratch
+    filesystem must check for **all three** directory names depending on which
+    platform types are in use.  A grep that looks only for
+    ``permission_checker_azxbyc`` will miss every ``ps`` and ``ecaccess``
+    platform probe.
```
VindeeR:
```suggestion
    Operators searching log files or running ``find`` on the remote scratch
    filesystem must check for **all three** directory names depending on which
    platform types are in use. A grep that looks only for
    ``permission_checker_azxbyc`` will miss every ``ps`` and ``ecaccess``
    platform probe.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+The permission probe directory
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To verify that Autosubmit can write to the configured ``SCRATCH_DIR``, each
+platform exposes a ``check_remote_permissions`` method.  The method creates a
+small probe **directory** (not a file) under
+``<scratch_dir>/<project>/<user>/`` and immediately removes it.  The name of
+the probe directory and the mechanism used to create it depend on the platform
+type:
+
```
VindeeR:
```suggestion
``<scratch_dir>/<project>/<user>/`` and immediately removes it. The name of
the probe directory and the mechanism used to create it depend on the platform
type:

```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+recovers from an error.  The intended audience is operators running Autosubmit
+in production who need to reason about the connection
+footprint that Autosubmit places on a login node.
+
+The permission probe directory
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To verify that Autosubmit can write to the configured ``SCRATCH_DIR``, each
+platform exposes a ``check_remote_permissions`` method.  The method creates a
+small probe **directory** (not a file) under
```
VindeeR:
```suggestion
To verify that Autosubmit can write to the configured ``SCRATCH_DIR``, each
platform exposes a ``check_remote_permissions`` method. The method creates a
small probe **directory** (not a file) under
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+chain in ``ParamikoPlatform`` is:
+
+``test_connection`` (when ``self.connected`` is ``False``) →
+``self.reset()`` → ``self.close_connection()`` (closes SFTP, the SSH agent,
+both ``Transport`` objects, and the ``SSHClient``) →
+``self.restore_connection(as_conf)`` →
+``with suppress(Exception): self.reset()`` (defensive second close) →
+``self.connect(...)`` (opens a new ``SSHClient``).
+
+The log-recovery subprocess follows the same discipline on its side.  When the
```
VindeeR: The phrasing is confusing, what did you mean by this?
```suggestion
The log-recovery subprocess follows the same discipline on its side. When the
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+with ``atexit`` when it is set up, so its SSH session is closed when the
+subprocess exits.
+
+At the end of a successful run, ``autosubmit.py`` issues an explicit
+``p.close_connection()`` for every platform in ``platforms_to_test``, plus
+``p.clean_log_recovery_process()`` to wind down the per-platform subprocesses.
+
+.. warning::
+    All ``close_connection`` calls are wrapped in ``with suppress(Exception):``
+    blocks.  If the close itself fails silently (for example because the
```
VindeeR:
```suggestion
    All ``close_connection`` calls are wrapped in ``with suppress(Exception):``
    blocks. If the close itself fails silently (for example because the
```

docs/source/devguide/platforms/index.rst
```diff
@@ -6,23 +6,39 @@ Platforms
...
+
+    - :mod:`Local Platform <autosubmit.platforms.locplatform>` — runs jobs
+      directly on the machine where Autosubmit is running. No SSH, no
+      scheduler. Used for local preparation steps in a workflow.
+    - :mod:`PS Platform <autosubmit.platforms.psplatform>` — a remote host with
+      no batch scheduler. Connects via SSH and tracks jobs as remote OS
+      processes (``ps``).
+    - :mod:`EC Platform <autosubmit.platforms.ecplatform>` — ECMWF systems
+      reached through the ``ecaccess`` toolchain. The underlying scheduler can
+      be ``pbs``, ``loadleveler`` or ``slurm``, selected via the platform's
```
kinow: I think loadleveler is no more. Can't tell if that will work? Maybe it worked in the past, before we removed... can you confirm this?

docs/source/devguide/platforms/index.rst
```diff
@@ -6,23 +6,39 @@ Platforms
...
+      ``VERSION`` field.
+    - :mod:`PJM Platform <autosubmit.platforms.pjmplatform>` — Fujitsu's PJM
+      scheduler, used on Fugaku and similar systems. Connects via SSH.
+    - :mod:`PBS Platform <autosubmit.platforms.pbsplatform>` — PBS Pro /
+      OpenPBS scheduler. Connects via SSH.
+    - :mod:`Slurm Platform <autosubmit.platforms.slurmplatform>` — the Slurm
+      workload manager (MareNostrum, Leonardo, LUMI, etc.). Connects via SSH.
+      Used as the worked example in this guide.
```
kinow: worked example here sounds odd to me?

docs/source/devguide/platforms/index.rst
```diff
@@ -6,23 +6,39 @@ Platforms
 
 
 .. note::
-    This documentation is based on the v4.1.13 branch, and can only guarantee reproducibility in this context
+    This documentation is based on the v4.1.17 branch, and can only guarantee reproducibility in this context
```
kinow: "v4.1.17 branch" this doesn't look very future-proof (not your fault). Maybe we could either fetch it from the current version in the VERSION file or somewhere else, or just say that this is the latest... or--the option I'm more inclined to I think-- remove this note. ReadTheDocs should be showing the "Latest" somewhere in the bottom of the page, with the option for the user to switch versions. If we are to tell the user this docs about the platforms is the latest.... then why don 't we alert that in the config... in the networking part... in the templates part???

docs/source/devguide/platforms/index.rst
```diff
@@ -6,23 +6,39 @@ Platforms
...
+      scheduler, used on Fugaku and similar systems. Connects via SSH.
+    - :mod:`PBS Platform <autosubmit.platforms.pbsplatform>` — PBS Pro /
+      OpenPBS scheduler. Connects via SSH.
+    - :mod:`Slurm Platform <autosubmit.platforms.slurmplatform>` — the Slurm
+      workload manager (MareNostrum, Leonardo, LUMI, etc.). Connects via SSH.
+      Used as the worked example in this guide.
 
 Composing the Extended Platform Class
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
kinow: I normally keep the length of the header marker the same as the text in rst. Was there a reason for changing this? https://matthewsetter.com/blog/item/restructuredtext-headers

> Thirdly, you have to remember to underline all of the text in the header. You can’t underline part of it, though you can underline more if you wish.

Normally, though, you see people matching the length in most Python projects, including the Python project itself -- https://raw.githubusercontent.com/python/cpython/refs/heads/main/Doc/extending/index.rst

If we are to change it for a reason, then we might better use the same everywhere, i.e., if there's a reason for using more then the title length, then let's do it everywhere for consistency.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+Platform Connections
+----------------------
+
+This section documents how Autosubmit opens connections to the configured
+platforms, when it verifies that it has write permissions on the remote
+filesystem, and how those connections are torn down when the workflow ends or
+recovers from an error.  The intended audience is operators running Autosubmit
+in production who need to reason about the connection
+footprint that Autosubmit places on a login node.
```
kinow: I think we can remove the last sentence about the audience. It will be useful for others as well.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+The permission probe directory
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To verify that Autosubmit can write to the configured ``SCRATCH_DIR``, each
+platform exposes a ``check_remote_permissions`` method.  The method creates a
```
kinow: > each platform exposes a ``check_remote_permissions`` method

This might change, and then the docs would be outdated. I think we don 't need to tell users about the method name...

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+     - SFTP ``mkdir`` then ``rmdir`` on the existing Paramiko channel.
+   * - ``ps``
+     - ``ps_permission_checker_azxbyc``
+     - Shell ``rm -rf``, ``mkdir -p``, ``rm -rf`` sequence executed over the existing SSH connection.
+   * - ``ecaccess``
+     - ``_permission_checker_azxbyc``
+     - Local ``ecaccess-file-mkdir`` / ``ecaccess-file-rmdir`` commands against the ECMWF gateway. Five parent ``ecaccess-file-mkdir`` calls precede the probe to compensate for the lack of ``mkdir -p`` in ``ecaccess-file-mkdir``.
+   * - ``local``
+     - (none)
+     - ``check_remote_permissions`` is a no-op that returns ``True`` without touching the filesystem.
```
kinow: This might change too. Our docs were very outdated, and we automated the configurations to fix that. We should try to keep it future-proof & automated as much as we can to avoid having to deal with an issue like that again, IMHO. So maybe we can leave the Python details out, or fetch it from the code (sphinx has access to the Python code being documented if needed)

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+.. note::
+    Operators searching log files or running ``find`` on the remote scratch
+    filesystem must check for **all three** directory names depending on which
+    platform types are in use.  A grep that looks only for
+    ``permission_checker_azxbyc`` will miss every ``ps`` and ``ecaccess``
```
kinow: Digressing, but IMHO, we should probably use a UUID or some other random file name instead of a fixed file name to check for permissions. That would probably avoid potential (probably rare, not impossible) filesystem/concurrency issues...

docs/source/devguide/platforms/index.rst
kinow: I think the above is very likely to change in 2->5 years. It might be better to document really when the probe runs. Is it when Autosubmit first configures the experiment? When it runs? In which commands. I believe this is the kind of information users would be able to understand (e.g., a sysadmin from DestinE doesn't know about `restore_platform`, so these docs seem more useful/aimed at AS developers).

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+wrap the initial ``prepare_run`` call in a retry loop, so the run aborts
+if both internal retries fail.  Adding a startup-time retry loop is
+tracked separately.
+
+Filesystem I/O footprint
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Beyond the permission probe and the SSH connections themselves, Autosubmit
+performs a number of file operations against the remote scratch filesystem
+during a run.  Operators concerned with filesystem load can use the table below to map each operation
```
VindeeR: I believe that the `scratch` might not be the best description in here since, it is just the directory in which the platforms will store the experiment, and it can change from platform to platform, scratch is mostly for `Marenostrum`, but other projects and `HPCs` have their own filesystem structure.

I believe it is worthy clarifying it at the beginning and using, something more generic like the variable name <SCRATCH_DIR>

and the note could be something like
` For the purpose of facilitating the explanation the Path to the scratch directory for the machine, which holds the experiments in a project will be refereed as <SCRATCH_DIR>`

docs/source/devguide/platforms/index.rst
```diff
@@ -263,6 +277,289 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+   * - ``autosubmit run`` (recovery)
+     - ``Autosubmit.run_experiment`` recovery block
+     - Whenever the main loop catches an ``AutosubmitError``, a dedicated reconnection loop calls ``restore_platforms`` once per iteration until reconnection succeeds.  The loop is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``).  Under sustained platform outages this is the dominant contributor to the probe count.
+   * - ``autosubmit setstatus``
+     - ``Autosubmit.set_status``
+     - Once per platform whose jobs include at least one job in status ``QUEUING``, ``SUBMITTED`` or ``RUNNING``.  If ``setstatus`` is invoked when all jobs are ``WAITING`` or ``READY`` (the typical operational case immediately before ``autosubmit create``), ``platforms_to_test`` is empty and no probe runs.
+   * - ``autosubmit stop --cancel``
+     - ``Autosubmit.stop`` → ``Autosubmit.prepare_run``
+     - Once per platform.  This applies to any scheduler, not only Slurm; the trigger is the ``--cancel`` flag, not the platform type.
```
kinow: I think this needs to be simplified. Maybe drop the call site. No need to document the function used. It seems like the info was written automatically based on the code -- while what we need is some user-friendly information for sysadmin, managers (like Francesca having to answer about the number of connections/IO ops), BSC workflow devs working with AEMET (e.g. Laura), etc.

The info at this level is probably more useful for AS devs, and I think we can use tools to generate it automatically for us if needed, or use this information and the IDE to locate which parts it's talking about?

kinow (commented):
I stopped reviewing mid-way through the platforms index.rst.

I think the docs are great for AS devs, but we need to inform the platform connections & I/O usage to multiple users. My other concern is that this would likely get outdated very quickly, as the code-base is not that stable yet (will probably remain like that over AS4, until we have AS5, I think).

docs/source/devguide/platforms/index.rst
rajeshsuraj: image
under ecplatform we still have loadleveler as a valid scheduler but i think the code needs to be updated as this section was written 10 years ago and loadleveler was discontinued in 2020, was it kept for backwards compatibility ?

VindeeR (changes requested):
I've done a few observations or improvements I believe it can be done on the text, overall it looks really good for a explanation on the platforms, but I think there y be some places going too much in depth, which is not necessarily a bad thing, it just depends on the user.

I too gave up on signaling all the double spaces after the "." if you could remove them I would appreciate 😬

Thanks for the hard working going through the code to understand and explain the platforms 🫶

docs/source/devguide/platforms/index.rst
rajeshsuraj: I think its better to remove this section as well and i agree its not future proof as the previous contributor had it at version 4.1.13.

docs/source/devguide/platforms/index.rst
rajeshsuraj: there is no elaborate reason, will change it match everywhere and be concise.

rajeshsuraj:
Thank you @kinow and @VindeeR for your comments and suggestions , I have made changes based on your comments.

I agree that this would most likely outdated by 4.2 so solution here can be to raise a issue and make the documentation change sections based on the actual platforms code i don't know how well that can work ?

I have corrected grammatical mistakes , removed double spaces and content wise reduced overall complexity and made it simpler since both your suggestions mentioned that.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,209 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+1. A **main session** used by the running workflow for submission, status
+   queries and file transfers.
+2. A **log-recovery session** owned by a per-platform background process,
+   used to download job log files in parallel while the main loop carries
+   on with other work.
+
+The log-recovery session is only created during ``autosubmit run``. Other
+commands (``setstatus``, ``recovery``, ``stop --cancel``, ``create``)
+open at most the main session per platform.
+
```
VindeeR, line 389: maybe we can put just stop, instead the command and argument, it simplifies the explanation.


```suggestion
The log-recovery session is only created during ``autosubmit run``. Other
commands (``setstatus``, ``recovery``, ``stop``, ``create``)
open at most the main session per platform.

```

docs/source/devguide/platforms/index.rst
```diff
@@ -17,9 +13,27 @@ platform class, for which first we need to identify which existing platform is t
...
+      no batch scheduler. Connects via SSH and tracks jobs as remote OS
+      processes (using the Unix ``ps`` command).
+    - :mod:`EC Platform <autosubmit.platforms.ecplatform>` — ECMWF systems
+      reached through the ``ecaccess`` toolchain. The underlying scheduler is
+      selected via the platform's ``VERSION`` field (``slurm`` on ECMWF's
+      current Atos systems; ``pbs`` is also supported for older
+      configurations).
+    - :mod:`PJM Platform <autosubmit.platforms.pjmplatform>` — Fujitsu's PJM
+      scheduler, used on Fugaku and other large-scale supercomputers.
+      Connects via SSH.
```
VindeeR, line 30: This is something we came across the other day and it makes sense to add it on this PR.

```suggestion
    - :mod:`PJM Platform <autosubmit.platforms.pjmplatform>` — Fujitsu's PJM
      scheduler, used on Fugaku and other large-scale supercomputers. (also known as TCM Technical Computing Suite)
      Connects via SSH.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,209 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+.. _platform_connections:
+
+Platform Connections
+--------------------
+
+This section describes how Autosubmit interacts with the platforms it has been
+configured to use: when it opens connections, when it checks that it has
+write access to the remote filesystem, and what it does when a connection has
+to be re-established mid-run.
```
VindeeR: removing wording repetition making it easier to follow and keep track of information

```suggestion
This section describes how Autosubmit interacts with the platforms it is
configured to use: when a connections is opened, Autosubmit checks if it has the
write access to the remote filesystem, and what to be done when a connection has
to be re-established mid-run.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,209 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+The write-permission check
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before relying on a remote platform, Autosubmit verifies that it can create
+and delete entries under the configured ``SCRATCH_DIR``. It does this by
+creating a small probe directory under
+``<scratch_dir>/<project>/<user>/`` and immediately removing it. The probe is
+a directory, not a file; the name it uses depends on the platform type:
```
VindeeR:
```suggestion
Before relying on a remote platform, Autosubmit verifies that entries can be created
and deleted under the configured ``SCRATCH_DIR``. It is done by creating a small probe directory under
``<scratch_dir>/<project>/<user>/`` and immediately removing it. The probe is
a directory, not a file; the name it uses depends on the platform type:
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,209 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+frequency depends on the command:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 28 72
+
+   * - Command
+     - When the probe runs
+   * - ``autosubmit run``
+     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error and recovers, the probe runs again once per reconnection attempt against each platform, until reconnection succeeds. The number of reconnection attempts is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``), so a prolonged outage can produce many probe entries in the logs.
```
VindeeR:
```suggestion
     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error and it is recovered, the probe runs again for each new attempt on each of the platforms. The number of retries is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``), so a prolonged outage can produce many probe entries in the logs.
```

Other than simplifying the test I'm a bit skeptical on this line

`so a prolonged outage can produce many probe entries in the logs.`, are you sure it'll keep creating and deleting folders until the end of retries is reached? Or once it manages to create and delete the folder it'll continue without trying all the possible retires it has left.

If so we might need to change this behavior I believe, it could be better discussed later

VindeeR (commented):
Just added/changed a few things to make it more fluid and easily readable.

as we spoke on slack I would like you to add a small thing related to the `PJM`, as per added in the review, and another one as bellow

at the `autosubmit/platforms/headers/pjm_header.py:24`, I would like you to add the following to the description of the platform.
```
class PJMHeader(object):
    """Class to handle the PJM (also known as TCM Technical Computing Suite) headers of a job"""
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+     - Local ``ecaccess-file-*`` commands against the ECMWF gateway. Several parent directories are created first because ``ecaccess-file-mkdir`` has no recursive option.
+   * - ``local``
+     - (none)
+     - The local platform does not perform a write-permission check.
+
+.. note::
+    Operators searching log files or the remote scratch filesystem need to
+    look for all three directory names, because each platform type uses a
+    different one. A search restricted to ``permission_checker_azxbyc`` will
+    miss probes from ``ps`` and ``ecaccess`` platforms.
```
kinow: Are we keeping the `permission_checker_azxbyc` file, @rajeshsuraj ? I thought AS created it, confirmed whether it had any issues or not, and then deleted it?

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+frequency depends on the command:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 28 72
+
+   * - Command
+     - When the probe runs
+   * - ``autosubmit run``
+     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error and it is recovered, the probe runs again for each new attempt on each of the platforms. The number of retries is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``), so a prolonged outage can produce many probe entries in the logs.
```
kinow: This is reading as if we retry 3650 times, but I think in reality we retry up to 3650 seconds by default? Or I think it was supposed to be that way...

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Autosubmit keeps a history database with one row per job execution. If a job
+ran but the row for that execution is incomplete (for example because
+Autosubmit was interrupted before the start and end timestamps were
+recorded), Autosubmit will try to fetch the missing values from the remote
+``STAT`` file for that job. This recovery runs during ``autosubmit run``,
+``autosubmit create``, ``autosubmit setstatus`` and ``autosubmit recovery``.
+
+The path opens SSH sessions to the affected platforms when needed, but it
```
kinow: The path opens SSH sessions?

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+   * - ``autosubmit run``
+     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error and it is recovered, the probe runs again for each new attempt on each of the platforms. The number of retries is bounded by ``CONFIG.RECOVERY_RETRIALS`` (default ``3650``), so a prolonged outage can produce many probe entries in the logs.
+   * - ``autosubmit setstatus``
+     - Once per platform that currently has jobs in ``QUEUING``, ``SUBMITTED`` or ``RUNNING`` state. When ``setstatus`` is run before ``autosubmit create`` (the typical operational case, where all jobs are still ``WAITING`` or ``READY``) no probe is created.
+   * - ``autosubmit stop --cancel``
+     - Once per configured platform. This applies to any scheduler, not only Slurm; the trigger is the ``--cancel`` flag, not the platform type.
+
+The probe does **not** run during the main loop of ``autosubmit run`` after
+startup, during ``autosubmit recovery``, during ``autosubmit create``, or
+when Autosubmit reconnects after a transient network error mid-run.
```
kinow: Is this confirmed? Did you write this reading the AS code in autosubmit.py and in the platforms package?

I think during the main loop of `autosubmit run`, after startup, if there's an error it will call `prepare_run` again, which should call `restore_platforms`, I think. And that function would check the permissions again. Can you confirm?

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+before exiting.
+
+.. warning::
+    The close operations are best-effort. If the underlying transport is
+    already broken, the close will fail silently and the next connection
+    will be opened anyway. This prevents a stuck close from blocking
+    recovery, but it means operators who want to confirm that no SSH
+    sessions have leaked should spot-check the login node directly (for
+    example with ``who`` or ``ss``) rather than rely on Autosubmit's
+    logs alone.
```
kinow, line 423: >  This prevents a stuck close from blocking
>  recovery, but it means operators who want to confirm that no SSH
>  sessions have leaked should spot-check the login node directly (for
>  example with ``who`` or ``ss``) rather than rely on Autosubmit's
>  logs alone.

Not necessarily operators (workflow engineers, workflow developers, etc.) -- we have the operator role in DestinE, which may be confusing here. Maybe "users" would be fine here, I think.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+The reconnection behaviour described above applies after the workflow has
+started running. If the very first connection attempt at the start of
+``autosubmit run`` fails — for example because the HPC login node is
+briefly unreachable — Autosubmit makes two internal retries within a
```
kinow: I can't recall now if it's configurable or if it's hard-coded (we were trying to fix that). Users reading this may have this questions of whether it's always 2 or not. Maybe we should check that and... I don't know if we should document it here or not 😥 Maybe we should link the documentation about retries here.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+single connection attempt, but does not currently retry the whole
+``autosubmit run`` startup. If both internal retries fail, the run
+aborts. Adding a workflow-level startup retry is tracked separately.
+
+Filesystem operations during a run
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Beyond handling write-permission probes and SSH sessions, Autosubmit performs several file operations against the remote ``<SCRATCH_DIR>`` during each run. The exact number of these operations can vary based on the workflow's complexity, including factors such as the number of jobs, retries, wrapper usage, and recovery events. The table below outlines the types of operations performed at different stages, with descriptions provided in qualitative terms to reflect this variability.
```
kinow, line 435: This text looks great, @rajeshsuraj ! 👏

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+   * - Download of job ``STAT`` files
+     - One download per job at the end of execution. Additional opportunistic downloads happen when the history database contains incomplete rows.
+   * - Reads of remote ``STAT`` files
+     - One batched read per status-check cycle, covering all relevant ``STAT`` files in a single round-trip.
+   * - Cleanup of previous-run marker files
+     - Once per platform at the start of ``autosubmit run``, scoped to the experiment's log directory.
+   * - Compression of remote logs
+     - One compression command per log file, only when ``COMPRESS_REMOTE_LOGS`` is enabled on the platform.
+   * - Write-permission probe
+     - See above.
```
kinow: I haven't checked these items above, but I think these happen on different commands? I mean, the stat files I think recovery tries to recover them... there was something at the end of the main loop that tried to copy something too, I think? I can read these with more calm later and confirm (my brain is slowly entering weekend-mode).

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+- **Status queries are batched.** A single scheduler query per platform per
+  cycle covers all jobs in flight, regardless of how many there are. Reads
+  of remote ``STAT`` files are similarly batched into a single round-trip.
+- **``STAT`` files are written by the jobs, not by Autosubmit.** The remote
+  job script appends a Unix epoch timestamp at start and finish; Autosubmit
+  only reads these files.
+- **Wrappers reduce per-job I/O.** A wrapped package submits one script for
+  many jobs and produces consolidated log and ``STAT`` output, which cuts
+  the per-job upload and download counts roughly in proportion to the
+  number of jobs in the wrapper.
```
kinow: Does it? I haven't read it, so this is something new to me 😬 @pablogoitia or @manuel-g-castro or @dbeltrankyl might be able to confirm if this paragraph is correct.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+.. note::
+    The table above describes which operations Autosubmit performs and
+    when, but not absolute numbers (bytes transferred, IOPS, syscall
+    counts) for a given workflow. Producing those numbers requires either
+    instrumenting Autosubmit with an I/O tracer or running a representative
+    workflow under ``strace`` / ``iostat`` against a controlled HPC.
```
kinow, line 493: 🙆‍♂️ Thanks!

docs/source/devguide/platforms/index.rst
dbeltrankyl: **During submit:**  Autosubmit compresses all jobs inside a wrapper locally and then, through send_command, it sends + uncompress ( in a single call) them 
 

The wrapper cmd contains only logic to orchestrate the templates; the templates behave exactly as if they were submitted individually. 

In 4.1.17, each job/template writes its own STAT, completed file as before, so the I/O is more or less the same. I believe that we only gain at the moment of submission, but it is pretty negligible 

Also, the wrapper has its own cmd, stat, logs ( which are not downloaded, there is an issue for that), and completed files

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,208 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+A few patterns worth knowing when reasoning about filesystem load:
+
+- **Submission and log retrieval run in parallel.** The main loop submits
+  jobs and queries their status; a separate per-platform process downloads
+  the logs of completed jobs. Activity from both runs concurrently against
+  the same remote directory.
+- **Status queries are batched.** A single scheduler query per platform per
+  cycle covers all jobs in flight, regardless of how many there are. Reads
+  of remote ``STAT`` files are similarly batched into a single round-trip.
+- **``STAT`` files are written by the jobs, not by Autosubmit.** The remote
```
dbeltrankyl, line 484: This is true, in addition we also have the final status in the STAT ( failed/completed)


Also, the submission time is written by autosubmit, but in the TOTAL_STAT file and in the db

docs/source/devguide/platforms/index.rst
kinow: Thanks Dani! I think it's probably better to remove the wrappers here, @rajeshsuraj . That way, readers (like a sysadmin, workflow developer, or new workflow user) doesn't have to search what are wrappers and how they work, if they have wrappers in the workflow, etc.

docs/source/devguide/platforms/index.rst
rajeshsuraj: yes, that's how it works , i will reword this section to be more clear.

docs/source/devguide/platforms/index.rst
rajeshsuraj: yeah the count is 3650, it's not a time budget as the default ``3650`` is the count from a per-attempt sleep that grows from 15s to a cap of 120s, so the default corresponds to roughly 72–122 hours of outage tolerance)
image

docs/source/devguide/platforms/index.rst
rajeshsuraj: sorry for the misleading wording here, i meant the code path technically or autosubmit here.

docs/source/devguide/platforms/index.rst
rajeshsuraj: Yes, I did read the code — this was actually one of the specific things I dug into for this doc, because issue reporter raised the same suspicion in the original issue.
prepare_run only calls restore_platforms inside its if not recover: branch — I re-checked this today against autosubmit.py. On the recovery path, prepare_run(recover=True) returns without touching restore_platforms. The restore_platforms invocation that fires on retry is done by a separate reconnection loop inside run_experiment (the while not reconnected: loop that comes after the while not recovery: loop), not by prepare_run. So the probe fires at most once per platform per recovery episode.

docs/source/devguide/platforms/index.rst
rajeshsuraj: image
it is hardcoded to be 2 under the paramiko platforms class. yeah i can reduce the info here and give a hyper link to the section on retries on the userguide section.

docs/source/devguide/platforms/index.rst
rajeshsuraj: I have added more context , hopefully this helps.

docs/source/devguide/platforms/index.rst
kinow: Ah, I remember that constant now. Thanks @rajeshsuraj !

kinow:
I think I used all my mental capacity to review PRs today. Will do this one tomorrow morning!

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
     (e.g.: <Project_Dir>/<Project_Name_Folder>/<USER>)
 
+
+.. _platform_connections:
+
+Platform Connections
+--------------------
+
+This section describes how Autosubmit interacts with the platforms it is
+configured to use: when a connections is opened, Autosubmit checks if it has the
```
kinow: "when a connections" I think the article is not matching the noun (plural).

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+write access to the remote filesystem, and what to be done when a connection has
+to be re-established mid-run.
+
+The write-permission check
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before relying on a remote platform, Autosubmit verifies that entries can be created
```
kinow: s/entries/files ? It sounds like a DB, but in reality I think we are talking about a file system. (And a directory is a file on Linux... on common Linux distros, pretty much everything is a file https://en.wikipedia.org/wiki/Everything_is_a_file, so files should be OK...)

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+frequency depends on the command:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 28 72
+
+   * - Command
+     - When the probe runs
+   * - ``autosubmit run``
+     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error, Autosubmit enters a reconnection loop that retries the SSH connection up to ``CONFIG.RECOVERY_RETRIALS`` **times** (a count of attempts, not a time budget; default ``3650``, with a per-attempt sleep that grows from 15s to a cap of 120s, so the default corresponds to roughly 72–122 hours of outage tolerance). The probe only runs on the retry where the connection is actually re-established, not on the failing attempts. In practice this means a prolonged outage produces many SSH retry log entries but at most one probe entry per platform per recovery episode.
```
kinow: "recovery episode" sounds weird. Maybe recovery retry? or simply "per retry". I normally use episode for TV series, or something like a illness/etc.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+recorded), Autosubmit will try to fetch the missing values from the remote
+``STAT`` file for that job. This recovery runs during ``autosubmit run``,
+``autosubmit create``, ``autosubmit setstatus`` and ``autosubmit recovery``.
+
+When Autosubmit performs this stale-job-data recovery it opens an SSH
+session to each affected platform, but it does not run the
+write-permission probe. ``autosubmit create`` is the case worth flagging
+here: it normally appears to be a local operation, but if there are stale
+rows in the history database it will open one SSH session per affected
+platform.
```
kinow, line 371: """
is the case worth flagging
here: it normally appears to be a local operation, but if there are stale
rows in the history database it will open one SSH session per affected
platform.
"""

👏 👍

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+During the main loop, Autosubmit does not open new connections.
+Submission, status checks and file transfers all reuse the SSH session
+that was opened at startup, and keepalive packets are sent on the
+existing channel to keep it open.
+
+Reconnection on error
+~~~~~~~~~~~~~~~~~~~~~
+
+If a connection is lost mid-run, Autosubmit closes the existing session
+before opening a replacement: the SFTP channel, the SSH transport and the
```
kinow: s/replacement/new one?

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+     - ``autosubmit run``
+     - Once per platform at startup, scoped to the experiment's log directory.
+   * - Compression of remote logs
+     - ``autosubmit run``
+     - One compression command per log file, only when ``COMPRESS_REMOTE_LOGS`` is enabled on the platform. Runs in the log-recovery process, in parallel with the main loop.
+   * - Write-permission probe
+     - ``autosubmit run``, ``autosubmit setstatus``, ``autosubmit stop --cancel``
+     - See :ref:`When the write-permission check runs <platform_connections>` above.
+
+A few patterns worth knowing when reasoning about filesystem load:
```
kinow: maybe "about Autosubmit filesystem load". Not general filesystem load.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+- **``STAT`` files are written by the jobs, not by Autosubmit.** The remote
+  job script appends a Unix epoch timestamp at start and finish; Autosubmit
+  only reads these files.
+
+.. note::
+    The table above describes which operations Autosubmit performs and
+    when, but not absolute numbers (bytes transferred, IOPS, syscall
+    counts) for a given workflow. Producing those numbers requires either
+    instrumenting Autosubmit with an I/O tracer or running a representative
+    workflow under ``strace`` / ``iostat`` against a controlled HPC.
```
kinow, line 493: Hmm, maybe we could create these as a simple test, and link it here, @rajeshsuraj . If you have already used eBPF/strace, or if you are interested. We could use the dummy workflow, run it once and collect metrics, and use that as a simple example, warning users it's a simple workflow, and users may try a similar approach if they need exact numbers. This way we have the required documentation, with general instructions (what you already did here 💪 ), and an extra use case.

autosubmit/platforms/headers/pjm_header.py
```diff
@@ -20,8 +20,8 @@
 import textwrap
 
 
-class PJMHeader:
-    """Class to handle the PJM headers of a job"""
+class PJMHeader(object):
```
kinow: I think the recommendation is to remove the `object` now. It's not necessary anymore (since... py 3.1? or some early Py3 version, I think)

kinow (commented):
I think if you remove the object, ruff will pass.

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+This section describes how Autosubmit interacts with the platforms it is
+configured to use: when a connection is opened, Autosubmit checks if it has the
+write access to the remote filesystem, and what to be done when a connection has
+to be re-established mid-run.
+
+The write-permission check
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before relying on a remote platform, Autosubmit verifies that entries(a probe file in this context) can be created
+and deleted under the configured ``SCRATCH_DIR``. It is done by creating a small probe directory under
```
VindeeR:
```suggestion
Before relying on a remote platform, Autosubmit verifies that entries can be created and deleted, under the configured 
``SCRATCH_DIR``. This is done by creating a small probe directory under
```

rajeshsuraj:
message from @kinow on slack , 

- If you look at the root of our project, we have ./scripts/. There, you will find scripts used to perform some simple tasks, or test something (like notification, I think(.
- You can locally create a dummy workflow (with that --dummy flag), and run it with ebpf/strace/that other tool I used/etc. And then add the required scripts used in that folder, documenting in-place or elsewhere and linking the files in ./scripts/
- Or, if you do find that it'd be easier to have a pytest, that'd be OK too. As long as it's not brittle and doesn't take too long to run & complete on GH Actions :slightly_smiling_face:

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+    case where a probe crashed between the create and delete steps and
+    left the directory behind) know that each platform type uses a
+    different name — a grep restricted to ``permission_checker_azxbyc``
+    will miss references from ``ps`` and ``ecaccess`` platforms.
+
+When the write-permission check runs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The check is triggered by a small number of user-facing commands, and the
+frequency depends on the command:
```
VindeeR: what do you mean by `user-facing commands`? maybe you meant just `commands`

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 28 72
+
+   * - Command
+     - When the probe runs
+   * - ``autosubmit run``
+     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error, Autosubmit enters a reconnection loop that retries the SSH connection up to ``CONFIG.RECOVERY_RETRIALS`` **times** (a count of attempts, not a time budget; default ``3650``, with a per-attempt sleep that grows from 15s to a cap of 120s, so the default corresponds to roughly 72–122 hours of outage tolerance). The probe only runs on the retry where the connection is actually re-established, not on the failing attempts. In practice this means a prolonged outage produces many SSH retry log entries but at most one probe entry per platform per recovery session.
```
VindeeR:
```suggestion
     - Once per configured platform at the start of the run, before the main loop begins. If the run later encounters a connection error, Autosubmit enters a reconnection loop that retries the SSH connection up to ``CONFIG.RECOVERY_RETRIALS`` **times** (the default are ``3650`` attempts, with a per-attempt sleep that grows from 15s to a cap of 120s, so the default corresponds to roughly 72–122 hours of outage tolerance). The probe only runs on the retry where the connection is actually re-established, not on the failing attempts. In practice this means a prolonged outage produces many SSH retry log entries but at most one probe entry per platform per recovery session.
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+For ``ecaccess`` platforms, no SSH session is opened at all. Every
+interaction with ECMWF is a local ``ecaccess-*`` command against the
+ECMWF gateway, so the connection cost is gateway traffic rather than
+login-node SSH sessions.
+
+For the ``local`` platform, no connection of any kind is opened.
+
+During the main loop, Autosubmit does not open new connections.
+Submission, status checks and file transfers all reuse the SSH session
+that was opened at startup, and keepalive packets are sent on the
```
VindeeR:
```suggestion
that was opened at startup, and keep alive packets are sent on the
```

docs/source/devguide/platforms/index.rst
```diff
@@ -263,8 +275,225 @@ You must input the information suitable for your project (e.g.: user, host, plat
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Autosubmit keeps a history database with one row per job execution. If a job
+ran but the row for that execution is incomplete (for example because
+Autosubmit was interrupted before the start and end timestamps were
+recorded), Autosubmit will try to fetch the missing values from the remote
+``STAT`` file for that job. This recovery runs during ``autosubmit run``,
+``autosubmit create``, ``autosubmit setstatus`` and ``autosubmit recovery``.
+
+When Autosubmit performs this stale-job-data recovery it opens an SSH
```
VindeeR:
```suggestion
When Autosubmit performs this data recovery job it opens an SSH
```

scripts/measure_platform_io.sh
```diff
@@ -0,0 +1,228 @@
...
+        -e trace=openat,connect,read,write,execve \
+        -o "${TRACE}" \
+        autosubmit run "${EXPID}" > "${STDOUT_LOG}" 2>&1 || true
+
+    summarise "${TRACE}" "${EXPID}" "${TS}" | tee "${SUMMARY}"
+fi
+
+echo
+echo "Summary         : ${SUMMARY}"
+echo "Workflow output : ${STDOUT_LOG}"
```
VindeeR: EoF

scripts/measure_platform_io.sh
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env bash
+#
```
VindeeR, line 4:
```suggestion
#
# Copyright 2015-2025 Earth Sciences Department, BSC-CNS
#
# This file is part of Autosubmit.
#
# Autosubmit is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# Autosubmit is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
#
```

scripts/measure_platform_io.sh
```diff
@@ -0,0 +1,228 @@
...
+            }
+            row("Opens of local SQLite DBs", db_desc)
+
+            row("Bytes written to disk", bytes(bytes_written))
+            row("Bytes read from disk (includes subprocess-startup overhead)", \
+                bytes(bytes_read))
+            row("Total syscalls", comma(syscalls))
+            rule()
+
+            if (jobs > 0) printf "\n  Jobs detected in trace: %d\n", jobs
```
VindeeR: This line wasn't printing when I executed the script that was minimal, and rerunning the script without the create first wont print it either

VindeeR (commented):
I guess for the missing line you can either add a warning to it, but I believe it would be better to fix it and make it work at all scenarios, still the create before running the script might always be needed , so the warning would still be valid

.gitignore
```diff
@@ -33,3 +33,5 @@ man/
 venv/
 .ipynb_checkpoints
 .atl
+.benchmarks/artifacts/performance-metrics-4.2.0.csv
+:memory:
```
VindeeR: I think it is better to create it's own section for the benchmark files and add them above the coverage files for logical organization IMO,
 
```suggestion

# Benchmark files
.benchmarks/artifacts
:memory:

```

.gitignore
VindeeR:
```suggestion

# Benchmark files
.benchmarks/artifacts
:memory:

```

VindeeR (changes requested):

.gitignore
VindeeR, line 1: Rebase this branch and have it as master

rajeshsuraj: rebased and pushed

VindeeR (approved):
It looks good to go to me, just don't forget to update your branch and squash the commits before merging

PR #3080, opened 2026-06-25
[Bug] - PJM job submission not correctly submitted 

### LINKED ISSUE #2977: Fugaku (PJM) not working
PJM check_all_jobs() and submit? Functions are broken due to an unknown reason cc: @Lerriola 

Confirmed versions not working: 

4.1.16, 4.1.16.1, master, pjm-fix (pr)

master output 1 ( check all jobs)
```
[CRITICAL] 2026-04-23 12:45:38,660 Permanent PJM error: PJM reported an error; check directives or resource group [eCode=7014]
    platform.check_all_jobs(platform_jobs, as_conf)
  File "/home/Earth/larriola/Development/Autosubmit/hanami-ev/lib/python3.12/site-packages/autosubmit/platforms/paramiko_platform.py", line 1094, in check_all_jobs
    self._check_for_unrecoverable_errors()
  File "/home/Earth/larriola/Development/Autosubmit/hanami-ev/lib/python3.12/site-packages/autosubmit/platforms/pjmplatform.py", line 464, in _check_for_unrecoverable_errors
autosubmit.log.log.AutosubmitCritical: Permanent PJM error: PJM reported an error; check directives or resource group
autosubmit.log.log.AutosubmitCritical: Permanent PJM error: PJM reported an error; check directives or resource group
[CRITICAL] 2026-04-23 12:49:05,201 Permanent PJM error: PJM reported an error; check directives or resource group [eCode=7014]
```

master output( submit?):

```
2026-04-23 12:52:22,480 Executing command pjstat -v --choose jid,jnam --filter "jnam=aaph_20120101_fc11_9_SIMULATION", retry #1 out of 3
2026-04-23 12:52:23,518 Saving JobList: /esarchive/autosubmit/aaph/pkl/job_list_aaph.pkl.tmp
2026-04-23 12:52:23,874 JobList saved in /esarchive/autosubmit/aaph/pkl/job_list_aaph.pkl
[ERROR] 2026-04-23 12:52:23,876 Trace: None
[ERROR] 2026-04-23 12:52:23,878 Failed to retrieve job IDs for submitted jobs. Submission output:  [eCode=6005]
```

4.1.16 output: 

```
[ERROR] 2026-04-23 14:37:23,778 Trace: Bad Parameters.
[ERROR] 2026-04-23 14:37:23,781 [err.] pjm 0204 pjstat invalid option argument: --filter 48541572.
[err.] pjm 0204 pjstat invalid option argument: --filter 48541572.
 [eCode=7014]
```

Testing: 4.1.15

VindeeR:
Closes #2977

The submission of jobs for PJM stopped working for the versions 4.1.16, 4.1.16.1, master 

**Check List**

- [x] Documentation updated.


autosubmit/platforms/pjmplatform.py
```diff
@@ -354,7 +370,7 @@ def _get_job_names_cmd(self, job_names: list[str]) -> str:
         if not job_names:
             return ""
 
-        commands = " ; ".join(
+        commands = "+".join(
```
kinow, line 396: A plus sign? I thought it was joining linux commands... `ls /tmp ; cd /tmp ; ...`. Interesting.

autosubmit/platforms/pjmplatform.py
```diff
@@ -233,6 +233,22 @@ def parse_queue_reason(self, output, job_id):
...
+        :type script_name: str
+        :param pre: command part to be placed before the script name, e.g. timeout, export, executable
+        :type pre: str
+        :param post: command part to be placed after the script name, e.g. redirection of stdout and stderr
+        :type post: str
+        :param x11_options: x11 options to run the script, if any
+        :type x11_options: str
+        :return: command to submit a job
+        """
+        return f"{self._submit_command_name} --no-check-directory {script_name} {x11_options} {post} & echo $!"
```
kinow: Is it working as expected? 

```
$ cat test.sh
#!/bin/bash

set -eu -o pipefail

echo "OK" & lsdklds & echo $?
$ bash test.sh
0
OK
test.sh: line 5: lsdklds: command not found
$ echo $?
0
```

VindeeR: I believe so, I tried not to change too much of the command from how it was behaving before, I'm not sure, with the new changes, how the AS will be handling the return of the output, so changing it might have a bigger impact.

autosubmit/platforms/pjmplatform.py
VindeeR: There is a few other places that wore refactored to handle the "+" too, so I was a bit unsure if I should touch it, since it is a change done by @dbeltrankyl and I believe it might have a bigger impact, and since the code seems to be working fine on the tests I've decided against it.

dbeltrankyl: If I recall correctly, if this comes from the lumi-testbranch I had, I was looking at it with Leo in person, and + is used to concatenate jobs there, but double-check it

VindeeR:
@Lerriola I executed a few simpler tests and it worked, could you test this branch on your workflow to validate if this fix the issue you were facing plase.

I ran the one you suggested `aaph`, locally, and it executed fine until it run into a workflow issue, the other workflow I create by myself was much simpler but worked fine without running into errors.

kinow (commented):
This one appears to be fully tested, and ready for a final review? Is there anything else pending on this PR, @VindeeR ? Besides the changelog...

VindeeR:
I believe it can be reviewed @kinow I just wanted to get the approval of @Lerriola for more complex experiments just to ensure it is working, although I tested with the experiment mentioned on the issue and it seemed to work fine, although I didn't finish running it.

Lerriola:
Hi! Hope you're doing well, sorry I couldn't check this earlier. I just got back from holidays yesterday. I'll set up an experiment asap with the hanami workflow and let you know : )

Lerriola:
Okay! Not really a regular test, but wanted to see if it would work. I wanted to recover the old experiment I was running in Fugaku before the issues, `experiment aaph`.  

What I did is to set up the python venv and pjm-fix repositories and then run the recovery command:

```bash
autosubmit recovery aaph --all -s -v
YAML configuration loaded:
  /esarchive/autosubmit/aaph/proj/git_project/config/bootstrap
  /esarchive/autosubmit/aaph/proj/git_project/config/bootstrap/include.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/general.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/jobs/jobs-precompiled.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/platforms.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/models/nicoco/fugaku_test.yml
  /home/Earth/larriola/hanami/workflow/user_platforms.yml
  /esarchive/autosubmit/aaph/conf/as_misc.yml
  /esarchive/autosubmit/aaph/conf/main.yml
  /esarchive/autosubmit/aaph/conf/version.yml
The aaph experiment 4.1.16 version is being updated to 4.1.17 for match autosubmit version
Autosubmit is running with 4.1.17

Checking configuration files...
YAML configuration loaded:
  /esarchive/autosubmit/aaph/proj/git_project/config/bootstrap
  /esarchive/autosubmit/aaph/proj/git_project/config/bootstrap/include.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/general.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/jobs/jobs-precompiled.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/platforms.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/models/nicoco/fugaku_test.yml
  /home/Earth/larriola/hanami/workflow/user_platforms.yml
  /esarchive/autosubmit/aaph/conf/as_misc.yml
  /esarchive/autosubmit/aaph/conf/main.yml
  /esarchive/autosubmit/aaph/conf/version.yml
Expdef config file is correct
Platforms sections: OK
Jobs sections OK
Configuration files OK

Recovering experiment aaph
Loading JobList
Load finished
Creating jobs...
Adding dependencies to the graph..
Adding dependencies to the job..
Transitive reduction...
Looking for edgeless jobs...

Checking configuration files...
YAML configuration loaded:
  /esarchive/autosubmit/aaph/proj/git_project/config/bootstrap
  /esarchive/autosubmit/aaph/proj/git_project/config/bootstrap/include.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/general.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/jobs/jobs-precompiled.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/platforms.yml
  /esarchive/autosubmit/aaph/proj/git_project/config/models/nicoco/fugaku_test.yml
  /home/Earth/larriola/hanami/workflow/user_platforms.yml
  /esarchive/autosubmit/aaph/conf/as_misc.yml
  /esarchive/autosubmit/aaph/conf/main.yml
  /esarchive/autosubmit/aaph/conf/version.yml
Expdef config file is correct
Platforms sections: OK
Jobs sections OK
Configuration files OK

Restoring SSH connection...
Using /home/Earth/larriola/.ssh/config as SSH configuration file
Restoring SSH connection...
Using /home/Earth/larriola/.ssh/config as SSH configuration file
```
Then it just gets stuck there, and it has been there for some time now.

Lerriola:
Good news! I did a copy of the experiment aaph, changed the settings so it ran only for a chunk and it worked :tada: 

```
Start of the NICOCO and regridding mpiexec runs - Tue Jul 14 18:02:32 JST 2026
End of the NICOCO and regridding mpiexec runs - Tue Jul 14 18:11:09 JST 2026
All logfiles have been generated correctly.
---------------------------------
- Chunk simulation has finished -
---------------------------------
 - Chunk start date 20120101
 - Chunk end date 20120201
 - Chunk rundir /home/ra000012/data/u12688/abcx/nicoco/run/20120101/fc00
```

And hey! Actually just now, while I was writing this, the recovery just worked :heart_decoration:  It just took extremely long. To be fair, the experiment has ~3700 jobs so it makes sense.

Thank you very much @VindeeR and @kinow :heart:

Lerriola:
Mnm okay when running the actual experiment aaph I run into the following issue:

```bash
  File "/shared/earth/easybuild/rocky/9.5/x86_64/software/Miniconda3/24.7.1-0/lib/python3.12/_strptime.py", line 333, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data '0' does not match format '%Y%m%d%H%M%S'

 [CRITICAL] Unexpected error: time data '0' does not match format '%Y%m%d%H%M%S'.
 Please report it to Autosubmit Developers through Git: https://github.com/BSC-ES/autosubmit/issues
```
Any ideas?

Lerriola:
Hello! I guess the time data issue is something different, Should I open another issue? So at least you can progress with this?

VindeeR:
> Hello! I guess the time data issue is something different, Should I open another issue? So at least you can progress with this?

Sorry for the delay in replying, if the issue seems to be coming from a different place then yes, it would be better to create a different issue, so this way we can merge this PR.

Again, sorry for the delay in replying to this. @kinow I'll be rebasing this issue now, I believe it is ready for a final review

kinow:
> Again, sorry for the delay in replying to this. @kinow I'll be rebasing this issue now, I believe it is ready for a final review

GH Actions is a bit unhappy I think.

kinow:
Ruff :grimacing: @VindeeR

VindeeR:
> Ruff 😬 @VindeeR

not sure why the ruff decided to point at so many issues, and so many of them to places I didn't even change
but that a good opportunity to address more of those issues

kinow:
Oh, looks like something changed recently that made ruff more picky? It'd be a shame to have to ignore mypy and now ruff. I can try to locate where it changed later (fighting pytest/coverage/containers this morning).

VindeeR:
@kinow okay I found the issue, ruff was updated from version 0.15.22 to 0.16.0, which changes the test behavior, getting a lot more issue and sort of ignoring the files it was being requested to check.

I'll create a small PR with just this change in the setting so it can be fixed at master.

kinow:
> @kinow okay I found the issue, ruff was updated from version 0.15.22 to 0.16.0, which changes the test behavior, getting a lot more issue and sort of ignoring the files it was being requested to check.
> 
> I'll create a small PR with just this change in the setting so it can be fixed at master.

Thank you for looking into this, @VindeeR !

kinow (commented):
One ruff error 😬

VindeeR:
I believe this PR is ready for review @kinow @dbeltrankyl

test/unit/test_utils/__init__.py
```diff
@@ -17,6 +17,8 @@
 
 """Unit test utilities."""
 
+from __future__ import annotations
```
kinow: Is this needed?

CHANGELOG.md
```diff
@@ -79,10 +79,11 @@ This release also includes several bug fixes and enhancements to improve the ove
 - Fixed infinite loop when STAT file stays on RUNNING, now falls back to scheduler status after IO_SAFE_WAIT #3059
 - Fixed `HETSIZE` whenever it had values bigger than 0 without a `CURRENT_QUEUE` or skipping the order of info #2660
 - Fixed `CPU per task` for new version of autosubmit #2897
-- Fixed issue overwritting expid config variables with the ones from the github repo #2877
+- Fixed issue overwriting expid config variables with the ones from the github repo #2877
 - Fixed inspect infinite loop when PLATFORMS.TOTALJOBS or PLATFORMS.MAX_WAITING_JOBS is set to 0 #2749
 - Fixed wrappers on Lumi platform #3059 
 - Fixed HPC2020 platform #3059
+- Fixed submission of jobs for PJM platform
```
kinow: Include the PR or issue ID here, in case we need to go back to it from the changelog entry 👍

autosubmit/job/job_list.py
```diff
@@ -791,7 +790,7 @@ def _add_all_jobs_edge_info(self, dic_jobs, option="DEPENDENCIES"):
             for job in jobs_gen:
                 self._apply_jobs_edge_info(job, dependencies)
 
-    def _deep_map_dependencies(self, section, jobs_data, option, dependency_list=set(),
+    def _deep_map_dependencies(self, section, jobs_data, option, dependency_list=(),
```
kinow: I think both are bad here? A set is mutable. The replacement here is an empty tuple. I think this should actually default to `None` and then we check if it's None, and do the assignment to an empty set if it's.

kinow (commented):
Haven't finished the review, just had a super fast look in GH UI. Tomorrow I can check it out with the IDE and review it properly, but maybe you can already have a look at these comments. Thanks!

autosubmit/job/job_list.py
```diff
@@ -204,11 +204,10 @@ def _delete_edgeless_jobs(self):
...
         # indices to delete
         for job in self.job_list[:]:
-            if job.dependencies is not None and job.dependencies not in ["{}", "[]"]:
-                if ((len(job.dependencies) > 0 and not job.has_parents() and not
-                job.has_children()) and str(job.delete_when_edgeless).casefold() ==
-                        "true".casefold()):
-                    self.graph.remove_node(job.name)
+            if (job.dependencies is not None and job.dependencies not in ["{}", "[]"] and
+                    ((len(job.dependencies) > 0 and not job.has_parents() and not job.has_children())
+                     and str(job.delete_when_edgeless).casefold() == "true".casefold())):
```
kinow: I think we can change this if `ruff` is complaining. Otherwise, if ruff is not angry with the code, we can leave it like it was.

This if statement is a bit confusing, I think. Or at least could be simplified... `len(job.depedencies) and `job.dependencies is not None` in the same if normally can be shortened to `if job.dependencies`.

These two, `job.has_parents()` and `job.has_children()` could be replaced by a new function (if there's none already) in `Job` that does something like,

```py
def edgeless():
    return not has_parents and not has_children
```

And then the last part about delete_when_edgeless, if used only here, it could be updated to return a `bool` instead of string, and do this casefold comparison in the class.

The `self.joblist[:]` is also not needed as the property returns a new list comprehension every call. And not the networkx self.graph elements. So we don't need yet another list, I believe.

That would allow us to write something similar to:

```
for job in self.job_list:
  if job.dependencies and job.dependencies not in ["{}", "[]"] and job.edgeless and job.delete_when_edgeless: 
    self.graph.remove_node(job.name)
```

Which I think wouldn't need the parentheses. But for a future change, when we need to review the `job_list.py`, which is also nearing the 4.2K lines now.

autosubmit/job/job_list.py
```diff
@@ -1251,38 +1250,37 @@ def _unify_to_filter(self, unified_filter, filter_to, filter_type, splits=None)
             value_list = self._member_list
         elif filter_type == "CHUNKS_TO":
             value_list = self._chunk_list
-        if "all".casefold() not in unified_filter[filter_type].casefold():
-            aux = str(filter_to.pop(filter_type, None))
-            if aux:
-                if "," in aux:
-                    split_aux_list = aux.split(",")
+        aux = str(filter_to.pop(filter_type, None))
+        if "all".casefold() not in unified_filter[filter_type].casefold() and aux:
```
kinow: Move the `aux` earlier in the if, `if aux and ...`, so if that fails we immediately skip it?

No need to `.casefold()` "all" here. It's already lowercase. Only the other unknown text in `unified_filter` needs to be lower-cased.

autosubmit/job/job_list.py
```diff
@@ -1251,38 +1250,37 @@ def _unify_to_filter(self, unified_filter, filter_to, filter_type, splits=None)
...
-                            else:
-                                check_whole_string = str(ele) + ","
-                            if str(check_whole_string) not in unified_filter[filter_type]:
-                                unified_filter[filter_type] += check_whole_string
+                        if extra_data:
+                            check_whole_string = str(ele) + extra_data + ","
+                        else:
+                            check_whole_string = str(ele) + ","
+                        if str(check_whole_string) not in unified_filter[filter_type]:
+                            unified_filter[filter_type] += check_whole_string
```
kinow: You are changing this block for ruff? If it's to improve the code, I'd leave as it was.

The for loop is creating a new regex within each loop iteration (we can compile the regex outside, improves performance, more readable -- this is a pattern we have everywhere in the code, @rajeshsuraj , @VindeeR related to what we spoke Tuesday about these code smells/patterns we have).

The `.split(",")` should be able to handle both splitting by comma, or `[aux]`. i.e., the result of `.split` should be enough.

The code is assuming `.findall` will have at least one result, and will raise `IndexError` if no matches found. Safer to use `search` as it doesn't allocate objects, better performance too.

That `skip = True`, is another pattern that we have, where a variable is used to hold a bool, passed around to compute the final value and then used in a `return` or `if`. The issue with that is that a developer following what happens in the code when they see that variable, they need to keep that in their mind or debug session, and keep adding more variables...

Normally `var = True; etc etc` can be simplified. I **think** we could have something like

```py
FILTER_RE = re.compile(r"([\[:\]a-zA-Z0-9._-]+)")
"""Regex pattern for alphanumeric parts."""

aux = str(filter_to.pop(filter_type, None))

if aux and "all" not in unified_filter[filter_type].casefold():
    for element in filter(None, aux.split(",")):
        # Get only the first alphanumeric part and [:] chars
        match = FILTER_RE.search(element)
        if match is None:
            continue

        parsed = match.group(1).lower()
        extra_data = element[len(parsed):]

        parsed_elements = JobList._parse_filter_to_check(
            parsed,
            value_list=value_list,
            level_to_check=filter_type,
            splits=splits,
        )

        # replacing the skip = False/True... just skip it here instead...
        if (
            unified_filter[filter_type]
            and any(
                isinstance(ele, str)
                and ele.lower() in {"natural", "none"}
                for ele in parsed_elements
            )
        ):
            continue

        # Now you have 1 place to set your breakpoint, and then you can check the other values in your debug session if you need.
        for ele in parsed_elements:
            check_whole_string = f"{ele}{extra_data},"
            if check_whole_string not in unified_filter[filter_type]:
                unified_filter[filter_type] += check_whole_string
```

But again, I think this is probably better to be done in another PR, and leave the PJM changes here, and simple cosmetics/ruff. `job_list.py` is a dangerous place to change without good tests. So we will need to confirm these are well covered too.

kinow (commented):
Reviewed some more, but I have to disconnect for a bit now.

autosubmit/job/job_list.py
VindeeR: As suggested I changed it back to how it was and I'll create a new issue to handle this

autosubmit/job/job.py
```diff
@@ -1375,6 +1375,15 @@ def has_parents(self):
         """
         return self.parents.__len__()
 
+    def edgeless(self):
+        """
+        Returns true if job has is edgless, else return false
```
kinow: Move pydoc one line up 😬

autosubmit/job/job.py
```diff
@@ -1375,6 +1375,15 @@ def has_parents(self):
...
+
+        :return: true if job has is edgless, otherwise return false
+        :rtype: bool
```
kinow: The rtype here is probably ignored by mypy. Probably better to have it in a single place, in the function signature, e.g., `def edgeless(self) -> bool`.

autosubmit/platforms/pjmplatform.py
```diff
@@ -294,12 +343,11 @@ def check_file_exists(self, filename: str, wrapper_failed: bool = False, sleepti
                     retries = retries + 1
                 else:
                     retries = 9999
-            except Exception as e:
-                if "garbage" in str(e).lower():
-                    if not wrapper_failed:
-                        sleep(sleeptime)
-                        sleeptime = sleeptime + 5
-                        retries = retries + 1
+            except SSHException as e:  # Unrecoverable error
```
kinow, line 342: Note to self: double-check the exception returned (but good improvement, thanks)

autosubmit/platforms/pjmplatform.py
kinow: Thanks Dani.

Note-to-self: check it later before final review.

autosubmit/platforms/pjmplatform.py
```diff
@@ -212,23 +236,42 @@ def get_check_all_jobs_cmd(self, jobs_id):
...
+         This needs to be adapted to each scheduler, the default assumes that is being launched directly.
+
+        :param script_name: name of the script to submit
+        :type script_name: str
+        :param pre: command part to be placed before the script name, e.g. timeout, export, executable
+        :type pre: str
+        :param post: command part to be placed after the script name, e.g. redirection of stdout and stderr
+        :type post: str
+        :param x11_options: x11 options to run the script, if any
+        :type x11_options: str
```
kinow: Probably better to skip adding the types here. We have a few places where type hints are out of sync with the pydoc types.

IDEs, mypy, and sphinx are able to parse type hints from function signature. But the types from pydocs are not supported everywhere. So I'd drop the pydoc, and keep the types in a single place.

autosubmit/platforms/pjmplatform.py
```diff
@@ -108,24 +113,29 @@ def check_remote_log_dir(self):
...
-                                          f"Could not create the DIR {self.remote_log_dir} on HPC {self.host}'.format(self.remote_log_dir, self.host)")
-            except BaseException as e:
-                raise AutosubmitError(
-                    "SFTP session not active ", 6007, str(e))
+                    raise AutosubmitError(
+                        "SFTP session not active ",
+                        6007,
+                        f"Could not create the DIR {self.remote_log_dir} on HPC {self.host}'.format(self.remote_log_dir, self.host)",
+                    )
+            except SSHException as e:
```
kinow, line 125: ❤️ I think the `BaseException` was dangerous here. I will do a final review later confirming the change is safe (to the best that I can 😬 ). Ta

autosubmit/job/job_list.py
```diff
@@ -3681,8 +3661,8 @@ def _wrapper_job_dict(self, wrapper_job: 'WrapperJob') -> tuple[dict[str, Any],
 
         :param wrapper_job: The wrapper job instance to serialize.
         :type wrapper_job: WrapperJob
-        :return: Tuple containing a dictionary of wrapper job attributes (including run_id) and a list of inner job dicts (each with run_id).
-        :rtype: Tuple[Dict[str, Any], List[Dict[str, Any]]]
+        :return: tuple containing a dictionary of wrapper job attributes (including run_id) and a list of inner job dicts (each with run_id).
+        :rtype: tuple[Dict[str, Any], List[Dict[str, Any]]]
```
kinow: The pydoc has Dict and List, but we are moving away from those types. In reality, it's probably better to drop the types from pydocs, and leave them just in the function signature + mypy, this way IDE/Mypy/other tools all work correctly, and we have a single place to maintain them. See my other comment.

autosubmit/job/job_list.py
```diff
@@ -1529,58 +1526,44 @@ def _calculate_natural_dependencies(self, dic_jobs, job, dependency, date, membe
...
+                        distance = 0
+                        for aux_ in self.dependency_map_with_distances.get(aux, []):
+                            if "-" in aux_ and job.section == aux_.split("-")[0]:
+                                distance = int(aux_.split("-")[1])
+                            elif "+" in aux_ and job.section == aux_.split("+")[0]:
+                                distance = int(aux_.split("+")[1])
+                            if distance >= max_distance:
+                                skip = True
+                    # get max value in distances_of_current_section.values
+                    if not skip and job.running == "chunk" and parent.chunk <= (chunk_list[-1] - max_distance):
```
kinow: Note to self: review it with care if this is not covered (will confirm later).

kinow (commented):
@VindeeR I'm splitting the autosubmit.py class this week, but I'm trying to keep up with the reviews by doing quick reviews (skimming the changes quickly top to bottom in GH UI), and later doing a better review on another day. I left a few comments, but I think this is already much better. Thanks!

VindeeR:
> autosubmit/job/job_list.py

nah, I appreciate the extra effort and thanks for the quick reviews

autosubmit/job/job.py
VindeeR: took the liberty of doing the same for the other function,

autosubmit/job/job.py
VindeeR: same as the last comment

autosubmit/job/job_list.py
VindeeR: I just had a quick look at this section and it is covered.

image

autosubmit/job/job_list.py
VindeeR: same as the previous ones

VindeeR:
@kinow the CI runned just fine, but I had to remove the files generated by the benchmark and `:memory:` I added the fix to create the `:memory:` file on temp but I was a bit unsure on how to handle the benchmark, maybe we could add the artifacts folder to the `.gitignore`

autosubmit/job/job_list.py
kinow: > As suggested I changed it back to how it was and I'll create a new issue to handle this

I think GH UI is still showing the modified block of code. IMO it's probably safer to modify the job_list in the separate PR, unless you need something here for PJM :)

autosubmit/job/job_list.py
kinow, line 1: @VindeeR can you point what part of `job_list.py` you had to change for PJM? Depending on what you needed here we might be able to keep some cosmetic changes, leaving only the ones that are a bit more risky to a separate PR (for git bisect, etc.).

autosubmit/platforms/pjmplatform.py
kinow: @VindeeR I had a look at this new exception handling statement.

`SSHException` extends `Exception`. The `get_mkdir_cmd` call is safe. The `send_command` internally handles `SSHException`, trying to prevent it from being raised.

Image

Image

But after sending the command, the function reads the stream/channel.

Image

That might raise a `TimeoutError`,

Image

which is an `OSError`.

Image

Maybe replace the `SSHException` here by an `OSError`. That would indicate something wrong happened after the command was sent, and the platform cannot check remote logs dir, and must retry/abort/fail/etc.

Feel free to have a look at it too. Unless we are using a really good and stable library, it's always challenging to trust the exceptions documented (or sometimes not even documented). And capturing the wrong exceptions normally only gets noticed in operationals, when we have some weird behaviour of a workflow during a network instability, disk crashes, and other weird OS errors.

autosubmit/platforms/pjmplatform.py
kinow: @VindeeR this looks like a bug to me. Even if this is the old code, or if we have this in Slurm.

The Bash shell template in AS will use `-xuve` and `-o pipefail`, which is why I used it in my example above. But I am not sure if this command even gets added to a Bash template. It looks like this may be submitted directly via Paramiko.

So, for simplicity, we can ignore the `-eu -o pipefail` I used before.

If we use this example file, `test.sh`:

```bash
#!/bin/bash

false & echo $?
```

We are saying, run a command `false` that exits with `1`, "background" it, and then print the result code of sending the code to background.

Whereas, it looks like what would make sense here is in reality to run the command and print the exit code. As a one-liner, I **think** this is what would have to be used (if this command is ever used within a shell with xuve/pipefail).

```bash
if false | true; then rc=0; else rc=$?; fi; printf '%d\n' "$rc"
```

You can probably test it simulating that the PJM platform sends any `pjsub` that fails. With the current code, I believe you will always get successful submission, simply because of the `&` is sending the `pjsub` to the background. And this action results in an exit `0`. Whether the background'ed command was successful or not, that's never captured.

kinow (commented):
Haven't addressed all notes to self I left yesterday 😅 but here's a few more comments, @VindeeR . Cheers

autosubmit/platforms/pjmplatform.py
VindeeR: In the `send_command` function is handled the exception with the `OSError` should I remove the `try...exception` then?

Image

autosubmit/job/job_list.py
VindeeR: It was mostly organizational and cosmetic I realized some of the code could be rewritten, when I rebased there was a few conflicts in there which let me to try and organize it better, I will move this risky changes to a different commit.

autosubmit/platforms/pjmplatform.py
kinow: Ah, it already catches the `OSError` in `send_command`. So it looks to me like the `BaseException` can be replaced by an `Exception`, as that would be in case we get any exception that was not `AttributeError` nor `OSError`, for whatever reason (an unknown error).

VindeeR:
@kinow I ran the command with a broken job and it gave me 0 with the error twice.

one with and one without the `& echo $?`

image

kinow:
> @kinow I ran the command with a broken job and it gave me 0 with the error twice.
> 
> one with and one without the `& echo $?`

Yup. `pjsub` will give you an exit code, which can be anything. I think in your examples `pjsub` is submitting something successfully and returning `0`? Can you make it return `1` or greater, and test with and without `& echo $?`?

VindeeR:
> > @kinow I ran the command with a broken job and it gave me 0 with the error twice.
> > one with and one without the `& echo $?`
> 
> Yup. `pjsub` will give you an exit code, which can be anything. I think in your examples `pjsub` is submitting something successfully and returning `0`? Can you make it return `1` or greater, and test with and without `& echo $?`?

The first command is being tested without the `& echo $?`

I'll try the to make it execute 1 or greater

VindeeR:
> > @kinow I ran the command with a broken job and it gave me 0 with the error twice.
> > one with and one without the `& echo $?`
> 
> Yup. `pjsub` will give you an exit code, which can be anything. I think in your examples `pjsub` is submitting something successfully and returning `0`? Can you make it return `1` or greater, and test with and without `& echo $?`?

After chatting with @dbeltrankyl he explained that the `& echo $?` is only for the platforms without scheduler. which is not the case here, and I can base the command on the slurm, so I changed the command to reflect it.

Thanks for catching it :raised_hands:

autosubmit/platforms/pjmplatform.py
```diff
@@ -71,11 +73,10 @@ def __init__(self, expid, name, config):
...
-        self.job_status['RUNNING'] = ['RNO', 'RNE', 'RUN']
-        self.job_status['QUEUING'] = ['ACC', 'QUE', 'RNA', 'RNP', 'HLD']  # TODO NOT SURE ABOUT HOLD HLD
-        self.job_status['FAILED'] = ['ERR', 'CCL', 'RJT']
+        self.job_status = {"COMPLETED": ["EXT"],
+                           "RUNNING": ["RNO", "RNE", "RUN"],
+                           "QUEUING": ["ACC", "QUE", "RNA", "RNP", "HLD"],
+                           "FAILED": ["ERR", "CCL", "RJT"]}
         self._pathdir = "\\$HOME/LOG_" + self.expid
         self._allow_arrays = False
         self._allow_wrappers = True  # NOT SURE IF WE NEED WRAPPERS
```
dbeltrankyl: Are the wrappers working? If not, we can set this to False until they're ready ( if they will ever be)

autosubmit/platforms/pjmplatform.py
```diff
@@ -108,24 +113,29 @@ def check_remote_log_dir(self):
...
-            self.scratch, self.project_dir, self.user, self.expid)
+            self.scratch, self.project_dir, self.user, self.expid
+        )
         self.remote_log_dir = os.path.join(self.root_dir, "LOG_" + self.expid)
         self.cancel_cmd = "pjdel"
         self._checkhost_cmd = "echo 1"
-        self._submit_cmd = f'cd {self.remote_log_dir} ; pjsub'
+        self._submit_cmd = f"cd {self.remote_log_dir} ; pjsub"
         self._submit_command_name = "pjsub"
-        self._submit_hold_cmd = f'cd {self.remote_log_dir} ; pjsub'
```
dbeltrankyl, line 128: Hold stuff can be dropped as it is no longer supported, but maybe it is wiser to do it in a different PR and remove it from every platform

autosubmit/platforms/pjmplatform.py
```diff
@@ -108,24 +113,29 @@ def check_remote_log_dir(self):
...
+            except SSHException as e:
+                raise AutosubmitError("SFTP session not active ", 6007, str(e))
 
     def update_cmds(self):
         """Update commands for platforms."""
         self.root_dir = os.path.join(
-            self.scratch, self.project_dir, self.user, self.expid)
+            self.scratch, self.project_dir, self.user, self.expid
+        )
         self.remote_log_dir = os.path.join(self.root_dir, "LOG_" + self.expid)
```
dbeltrankyl: I know these os.path.join comes from the early PJM stages, when we copied all the functions from the Slurm platform. Perhaps we can change it to use Pathlib throughout this file?

autosubmit/platforms/pjmplatform.py
```diff
@@ -137,33 +147,47 @@ def get_remote_log_dir(self):
         return self.remote_log_dir
 
     def queuing_reason_cancel(self, reason):
```
dbeltrankyl: Hm, about this function, I don't see it in other platforms, only in the job.py. Do you know why? Also, I think it is not being called anywhere, so maybe we can drop it?

autosubmit/platforms/pjmplatform.py
```diff
@@ -173,14 +197,14 @@ def parse_job_list(self, job_list: list[list['Job']]) -> str:
         """
         job_list_cmd = ""
         for job in job_list:
-            job_list_cmd += str(job.id) + "+"
+            job_list_cmd += str(job.id) + "+" if hasattr(job, "id") else ""
```
dbeltrankyl: Why is it necessary to have hasattr(job,"id") here? 

I don't see how it is possible in normal execution. If that is happening, perhaps we're hiding some bug?

autosubmit/platforms/pjmplatform.py
dbeltrankyl: Did we check this? I guess it works because Leo experiment is working but just in case

dbeltrankyl (commented):
Thanks @VindeeR 

Some minor comments, mainly misc, but looks very good! thanks

I miss some integration test, tho 

We can't do a run because we don't have a container for that platform, but  maybe we could  add an integration test using the inspect command?

autosubmit/platforms/pjmplatform.py
VindeeR: Not really, and we might try to re implement them later on if needed, I'll add a warning to reactivate it in case we add the wrappers

autosubmit/platforms/pjmplatform.py
VindeeR: I'll create an issue for this

autosubmit/platforms/pjmplatform.py
VindeeR: oops didn't notice it, thanks for the catch

autosubmit/platforms/pjmplatform.py
VindeeR: I can't really recall why this was added here, or if it was a test that I forgot to remove, since I'm not sure on what happened I'll remove it and make some tests.

VindeeR:
@dbeltrankyl I tested it locally and it worked for the most part, I believe, there was a change with the groups or something like that, which breaks the submission so I couldn't fully test.

So it submits a job but the job `Fails` due to the group and only the remote fails and I can't progress the experiment, but running manually a generated `.cmd` it runs normally.

So I believe it works.

@Lerriola could you help me with that part?

Regarding the `+` I just tested manually and it doesn't work, but as it is now, it doesn't fail either because of the filter, apparently it is not filtering properly so I need to have a look at it 
image

Other than that I believe the comments were addressed, I'll investigate it further to confirm why when I use the filter it doesn't return anything.

VindeeR:
Okay the `--filter` is working, I was trying to use the name of the job but because it doesn't show the full name I couldn't find it, but since when creating the command it uses the full name it works.

image

And the `+` might work if it is only one job to look for so yes it's better to use the `;`

Lerriola:
Hello all, thanks for all the fixes.

> So it submits a job but the job `Fails` due to the group and only the remote fails and I can't progress the experiment, but running manually a generated `.cmd` it runs normally.
> 
> @Lerriola could you help me with that part?

@VindeeR  Should I test by running an experiment in Fugaku with this latest version of the branch ? Do you want me to test anything else?

VindeeR:
> Hello all, thanks for all the fixes.
> 
> > So it submits a job but the job `Fails` due to the group and only the remote fails and I can't progress the experiment, but running manually a generated `.cmd` it runs normally.
> > @Lerriola could you help me with that part?
> 
> @VindeeR Should I test by running an experiment in Fugaku with this latest version of the branch ? Do you want me to test anything else?

Just testing the latest version of the branch should be enough thanks for the help

Lerriola:
Hi @VindeeR, I ran a simple experiment with only one member and one chunk using this branch and it seems to have run well :heavy_check_mark: 

I used the experiment abmx:

- [Fugaku] `/vol0004/ra000012/data/u12688/abmx`
- [AS VM]  `/esarchive/autosubmit/abmx`

Do you want me to try multiple members and chunks?

autosubmit/platforms/pjmplatform.py
```diff
@@ -71,22 +72,23 @@ def __init__(self, expid, name, config):
...
-        self.job_status['QUEUING'] = ['ACC', 'QUE', 'RNA', 'RNP', 'HLD']  # TODO NOT SURE ABOUT HOLD HLD
-        self.job_status['FAILED'] = ['ERR', 'CCL', 'RJT']
+        self.job_status = {"COMPLETED": ["EXT"],
+                           "RUNNING": ["RNO", "RNE", "RUN"],
+                           "QUEUING": ["ACC", "QUE", "RNA", "RNP", "HLD"],  # TODO NOT SURE ABOUT HOLD HLD
+                           "FAILED": ["ERR", "CCL", "RJT"]}
         self._pathdir = "\\$HOME/LOG_" + self.expid
         self._allow_arrays = False
-        self._allow_wrappers = True  # NOT SURE IF WE NEED WRAPPERS
+        self._allow_wrappers = False  # WRAPPERS aARE CURRENTLY NOT ACTIVE FOR THIS PLATFORM
```
kinow: s/aARE/ARE.. or maybe remove the comment? The variable itself is self-explanatory IMO

autosubmit/platforms/pjmplatform.py
```diff
@@ -71,22 +72,23 @@ def __init__(self, expid, name, config):
...
         self.update_cmds()
         self.config = config
-        exp_id_path = os.path.join(self.config.get("LOCAL_ROOT_DIR"), self.expid)
-        tmp_path = os.path.join(exp_id_path, "tmp")
-        self._submit_script_path = os.path.join(
-            tmp_path, self.config.get("LOCAL_ASLOG_DIR"), "submit_" + self.name + ".sh")
-        self._submit_script_base_name = os.path.join(
-            tmp_path, self.config.get("LOCAL_ASLOG_DIR"), "submit_")
+        exp_id_path = Path(self.config.get("LOCAL_ROOT_DIR", ""), self.expid)
```
kinow: > self.config.get("LOCAL_ROOT_DIR", "")

This looks a bit dangerous. If it can be `None` or empty, then better to raise an error if so. Or just leave like the other, without the default value.

autosubmit/platforms/pjmplatform.py
```diff
@@ -71,22 +72,23 @@ def __init__(self, expid, name, config):
...
+        tmp_path = Path(exp_id_path, "tmp")
+        self._submit_script_path = Path(
+            tmp_path, self.config.get("LOCAL_ASLOG_DIR", ""), "submit_" + self.name + ".sh"
+        )
+        self._submit_script_base_name = Path(
+            tmp_path, self.config.get("LOCAL_ASLOG_DIR", ""), "submit_"
```
kinow: Ditto on these other ones... the path shouldn't be empty, or concatenating that could lead to some issues...

autosubmit/platforms/pjmplatform.py
```diff
@@ -97,7 +99,9 @@ def submit_error(self, output):
         :param output: output of the submit cmd
         :return: boolean
         """
-        return not all(part.lower() in output.lower() for part in ["pjsub", "[INFO] PJM 0000"])
+        return not all(
+            part.lower() in output.lower() for part in ["pjsub", "[INFO] PJM 0000"]
```
kinow, line 104: Out of curiosity, is [INFO] the log level? Any chance changing the log level would change this too to [WARN] or something? And the 0000 is some string PJM always outputs?

autosubmit/platforms/pjmplatform.py
```diff
@@ -108,24 +112,29 @@ def check_remote_log_dir(self):
...
                 else:
-                    raise AutosubmitError("SFTP session not active ", 6007,
-                                          f"Could not create the DIR {self.remote_log_dir} on HPC {self.host}'.format(self.remote_log_dir, self.host)")
-            except BaseException as e:
-                raise AutosubmitError(
-                    "SFTP session not active ", 6007, str(e))
+                    raise AutosubmitError(
+                        "SFTP session not active ",
+                        6007,
+                        f"Could not create the DIR {self.remote_log_dir} on HPC {self.host}'.format(self.remote_log_dir, self.host)",
```
kinow: Is the `.format(` call here correct? It appears to be mixing f-strings and .format (which are still useful to lazily resolve something..., but not sure that was intended here)

autosubmit/platforms/pjmplatform.py
```diff
@@ -136,34 +145,17 @@ def get_mkdir_cmd(self):
...
-        status = ""
-        try:
-            status = [x.split()[1] for x in output.splitlines()
-                      if x.split()[0] == str(job_id)]
-        except BaseException:
-            pass
+        status = []
+        if output is not None:
+            status = [
+                x.split()[1] for x in output.splitlines() if x.split()[0] == str(job_id)
```
kinow, line 153: What if the output is empty but not `None`, or if it has a single word? Is it copying the exact behaviour of Slurm here? Or can we enhance and make it more defensive for PJM?

autosubmit/platforms/pjmplatform.py
kinow: @VindeeR I think here the exception expected to increase retries is an `SFTPError`, which extends `Exception`.

https://github.com/paramiko/paramiko/blob/142f593e40ad767c5e3556cbace66dc84589620c/paramiko/sftp.py#L216

Image

Image

That'd be safer than catching Exception since a random Exception with some text like "garbage collection" could trigger the retries. But SSHException is a different class, also extending `Exception`. So this block would never get called from Paramiko 👍

autosubmit/platforms/pjmplatform.py
kinow: Sorry, I think my last comment was not answering your question, @VindeeR .

`send_command` is already handling `OSError` and `AttributeError`. That's fine.

It may still raise `SSHException`, which is correctly captured in the current code. So I think we don't have to do any changes here.

autosubmit/platforms/pjmplatform.py
VindeeR: `[INFO] PJM 0000` is a log that returns from the platform, iirc always, saying that things went right with the submission|deletion

autosubmit/platforms/pjmplatform.py
VindeeR: IIRC the output will never be empty or just one word but I can try and make it more defensive

.gitignore
```diff
@@ -15,6 +15,10 @@ dist/
 build/
 .cache
 
+# Benchmark files
+.benchmarks/artifacts
+:memory:
```
kinow: The :memory: can be removed, as this could hide actual bugs. :memory: may be used to indicate in-memory sqlite, and accidentally be used instead to write a file. The other benchmarks/artifacts looks like a useful addition (combined with the change to prevent the test from running altogether by default, @VindeeR ).

kinow (commented):
@VindeeR maybe move the gitignores to another PR? Everything else is looking good

.gitignore
kinow: If you remove them, then we can kick CICD again, have a final quick look and then I'll hit the approve button :boom: _

kinow (approved):
Code looks great. No git conflicts. Leo tested it, just pending CICD now. Leaving approved, feel free to merge if the tests succeed. Thanks @VindeeR !

.gitignore
```diff
@@ -14,7 +14,6 @@ docs/build/
 dist/
 build/
 .cache
-
```
kinow, line 17: Maybe leave this blank space? 😬

kinow (approved):

Issue #3047, opened 2026-06-04
Refactor `_get_platform_by_type` to use an enum


ntorqulu:
Looking at `paramiko_submitter.py`, I think `_get_platform_by_type` could use a `PlatformType` enum instead of strings in order to conditionally separate the different platforms.
And use a dict dispatch to replace the if/elif chain. 
This way is easy to maintain if new platforms are added.

```
def _get_platform_by_type(platform_type: str, expid: str, platform_name: str, experiment_data: dict,
                          platform_version: str, auth_password: Optional[str]) -> Optional['ParamikoPlatform']:
    if platform_type == 'ps':
        return PsPlatform(expid, platform_name, experiment_data)
    elif platform_type == 'ecaccess':
        return EcPlatform(expid, platform_name, experiment_data, platform_version)
    elif platform_type == 'slurm':
        return SlurmPlatform(expid, platform_name, experiment_data, auth_password=auth_password)
    elif platform_type == 'pjm':
        return PJMPlatform(expid, platform_name, experiment_data)
    elif platform_type == 'pbs':
        return PBSPlatform(expid, platform_name, experiment_data)

    return None
```

ntorqulu:
Closed as completed via #3086

PR #3086, opened 2026-06-26
Use constants to check for platform type in job_packages

### LINKED ISSUE #3081: timeout guard is silently disabled for login/local jobs: v4.1.17-dev3-c2a87d0
Hi @kinow @dbeltrankyl 
jobs on no-scheduler platforms (`p`s / `local`) are supposed to be launched wrapped in a shell timeout, so they can't run past their wallclock. Right now that wrapper is never added, because the code compares the platform type against the wrong casing.

The platform types are set in lowercase:

```
platforms/psplatform.py:36 → self.type = 'ps'
platforms/locplatform.py:46 → self.type = 'local'
```
But this check looks for uppercase:

```
# job_packages.py:79-80  — current (buggy)
self.timeout = max(job.wallclock_in_seconds for job in jobs) \
    if jobs[0].platform.type in ["PS", "LOCAL"] else None
```
'ps' in ["PS", "LOCAL"] is always False, so package.timeout ends up None for every login/local job. That None then flows through and drops the wrapper:

```
# paramiko_platform.py
:1487  timeout = package.timeout if package.timeout else 0          # -> 0
:1490  self.get_call(abs_path, timeout, …)                          # timeout = 0
:1551  pre = f"timeout {timeout} " if float(timeout) > 0 else ""    # -> ""  (no wrapper)
```
Result: the job is launched as nohup bash …_DVC_SETUP.cmd & echo $! — with no time bound.

These platforms have no scheduler to enforce the wallclock, so the timeout wrapper is the only thing bounding a job's lifetime. Without it, an interrupted or hung login-node job becomes an immortal nohup process that leaks on the shared login node and can hold shared locks (e.g. the DVC repo lock) indefinitely — which is what causes the DVC_SETUP "Unable to acquire lock" failures.

Fix: normalise the comparison:

```
# minimal
... if jobs[0].platform.type in ["ps", "local"] else None
```

```
# safer (casing is inconsistent across the codebase: PBS/SLURM are upper, ps/local lower)
... if jobs[0].platform.type.lower() in ["ps", "local"] else None
```

test with this fix: https://github.com/MuhammadShafeeque/autosubmit_sfq/blob/4.1.17-fix-ps-local-timeout/autosubmit/job/job_packages.py#L79-L80 
```bash
mushaf001@climatedt-wf:~
❯ grep -h "DVC_SETUP.cmd >" /appl/AS/AUTOSUBMIT_DATA/t0sf/tmp/ASLOGS/*run.log   | grep -oE "nohup( timeout [0-9]+)? bash" | sort -u
nohup timeout 4680 bash
                                                                                                                                          [11:27:49]
mushaf001@climatedt-wf:~
❯ grep -h "DVC_SETUP.cmd >" /appl/AS/AUTOSUBMIT_DATA/t0s7/tmp/ASLOGS/*run.log   | grep -oE "nohup( timeout [0-9]+)? bash" | sort -u
nohup bash
```

### LINKED ISSUE #3108: Autosubmit Git module may use a string instead of a Platform object
Found by @ntorqulu on https://github.com/BSC-ES/autosubmit/pull/3086#discussion_r3518742675

Image

In the screnshot above, `hpcarch` is set to a string. I only fixed it to use a string enum. However, Núria correctly identified that in reality the `hpcarch` instance type is `Platform`. I'm fixing the bug in my pull request where Núria identified it, and created this issue so we know when/where it was fixed.

@LuiggiTenorioK , @f-macchia some time ago we had an operational issue found by Cheng, that could have been caught by mypy type checker, for free and very quickly on GitHub Actions. @VindeeR is slowly tackling the mypy issues but I suspect we still have a few thousand places to fix it #2820 #2999 . This bug that Núria found is identified by mypy, but we are still ignoring these warnings in our GH Actions, and the only way to get these issues is in operations/testing suite/or in code reviews 😥 For me it'd be good to prioritise fixing mypy.

kinow:
Closes #3081 

Closes #3108

- Fixes the issue identified by Muhammad using an enum
- Moves code out of Autosubmit class
- Adds more tests to code changed, and to some nearby code
- Fixes brittle tests in CICD
- Fixes bug in Autosubmit Git found by Núria
- Adds more types for mypy

**Check List**

- [ ] Documentation updated.


autosubmit/job/job_packages.py
```diff
@@ -27,72 +27,86 @@
...
-def threaded(fn):
-    def wrapper(*args, **kwargs):
-        thread = Thread(target=fn, args=args, kwargs=kwargs)
-        thread.name = "data_processing"
-        thread.start()
-        return thread
-
-    return wrapper
-
-
```
kinow, line 54: Not used.

autosubmit/job/job_packages.py
```diff
@@ -27,72 +27,86 @@
...
+        - All jobs have the same platform
+        """
+        if not jobs:
+            raise ValueError('No jobs given')
+
+        first_job = jobs[0]
+        # ATTENTION: Do not use ``first_job.platform`` -- that is a property.getter with
+        # a different logic, returning the inner ``platform.serial_platform`` sometimes.
+        # noinspection PyProtectedMember
+        self.platform = first_job._platform
```
kinow, line 70: The code validated if jobs were not given, but in another part, which is odd. Now it immediately fails if no jobs are provided. It also expects all jobs to have the same platform, so I added that below to fail quickly.

autosubmit/job/job_packages.py
```diff
@@ -27,72 +27,86 @@
...
+        self.executable = first_job.executable
+        self.x11_options = first_job.x11_options
         # Scheduler manages the timeout, this is for platforms without scheduler
         # Wrappers are only allowed within a scheduler and timeout is calculated differently there
-        self.timeout = max(job.wallclock_in_seconds for job in jobs) if jobs[0].platform.type in ["PS",
-                                                                                                  "LOCAL"] else None
-        self.x11 = jobs[0].x11
+        ps_local_platforms = [PsPlatform.TYPE, LocalPlatform.TYPE]
+        is_ps_local = first_job.platform.type in ps_local_platforms
+        self.timeout: Optional[int] = None if not is_ps_local else max(job.wallclock_in_seconds for job in jobs)
```
kinow: These three lines, ☝️ , are the fix.

autosubmit/job/job_packages.py
```diff
@@ -153,16 +166,16 @@ def check_job_files_exists(self, configuration: 'AutosubmitConfig', only_generat
...
 
             for additional_file in job.additional_files:
                 if not additional_file or not (project_dir / additional_file).exists():
                     if not only_generate:
                         raise AutosubmitCritical(
-                            f"[section:{job.section}]: Additional file:{additional_file} does not exists", 7014)
+                            f"[section: {job.section}]: Additional file:{additional_file} does not exist", 7014)
                     Log.warning(
-                        f"[section:{job.section}]: Additional file:{additional_file} does not exists, skipping check")
+                        f"[section: {job.section}]: Additional file: {additional_file} does not exist, skipping check")
```
kinow, line 177: Typos, tidying up the messages given to users.

autosubmit/job/job_packages.py
```diff
@@ -200,10 +212,11 @@ def _delete_previous_run_files(self):
         self.platform.delete_previous_stat_files_by_job_names([job.name for job in self.jobs if job.fail_count == 0])
 
     def _create_scripts(self, configuration: 'AutosubmitConfig'):
-        raise Exception('Not implemented')
+        raise NotImplementedError  # pragma: no cover
 
     def send_files(self):
         """ Send local files to the platform. """
+        raise NotImplementedError  # pragma: no cover
```
kinow, line 218: This was awkwardly empty (returns `None`, I think?), and the one above was raising `Exception` for not implemented. Fixed, without coverage for these.

autosubmit/platforms/slurmplatform.py
```diff
@@ -64,6 +64,8 @@
 class SlurmPlatform(ParamikoPlatform):
     """Class to manage jobs to host using SLURM scheduler."""
 
+    TYPE = 'slurm'
```
kinow: Note to self: check on Monday if it's easy to make `Paramiko` or base `Platform` an abc, add `TYPE: str`. Also, why no type on EcPlatform? Shouldn't `ParamikoPlatform` be abc too? 💫

test/unit/job/test_job_packages.py
```diff
@@ -284,3 +320,72 @@ def test_job_package_send_files_uses_current_public_api(mocker, local, tmp_path)
...
+        platform.serial_platform = serial_platform
+        job.platform = platform
+        # ATTENTION: Internally, Autosubmit increases 30% the wall-time, so we will have 6 seconds (floor).
+        job.wallclock = '00:00:05'
+
+    job_packages = JobPackageSimple(jobs)
+    if expected_timeout is None:
+        assert job_packages.timeout is None
+    else:
+        assert job_packages.timeout == expected_timeout
```
kinow, line 396: Test for the fix.

kinow:
Will continue fixing the tests on Monday. The `job_packages.py` has a lot of code that's not very clear why it was designed the way it is. I am trying to add some comments, and fix typos in messages presented to users.

The object hierarchies of the platforms & packages are also a bit odd. No abstract class, lots of duplicated and tangled/tightly-coupled logic. Will try to just make the tests pass with the current changes and then mark it ready for review, and leave the rest for other issues.

autosubmit/log/utils.py
```diff
@@ -86,44 +82,3 @@ def compress_gzip(
...
-    for filename in all_files:
-        # Match the regex pattern if provided
-        if pattern and not re.match(pattern, str(filename.name)):
-            continue
-
-        # Check if the file is not compressed
-        if not is_xz_file(str(filename)) and not is_gzip_file(str(filename)):
-            result.append(str(filename))
-
-    return result
```
kinow, line 129: I did a quick pass through a few files (while GH Actions was running) to review if there was any unused code, and found these that are test helper functions that were added to the source tree. Moved to the test folder.

autosubmit/monitor/utils.py
kinow, line 1: Not used.

kinow:
Hi @ntorqulu ! You commented in the issue linked here, maybe you could review this one if you have some time this week? Thanks!

test/unit/job/test_job_packages.py
```diff
@@ -284,3 +320,72 @@ def test_job_package_send_files_uses_current_public_api(mocker, local, tmp_path)
...
+)
+def test_job_package_timeout(platform_class , expected_timeout, create_platform):
+    """Test that the ``timeout`` is calculated correctly.
+
+    The ``timeout`` is calculated ONLY in ``ps`` and ``local`` platforms
+    as the ``max`` of ``job.wallclock_in_seconds``. It receives a job
+    list, which is assumed to be non-empty.
+
+    We use a fixed list of four jobs, with wallclocks of 1, 2, 3, and 4
+    seconds. Jobs get assigned a platform from the ``list_of_platforms``.
```
kinow: Argh. @ntorqulu , I will update this docstring. I had the test working like this initially, but changed that. I'll fix the docstring to reflect what's in the code later (I'm rebasing a PR right now). Sorry

ntorqulu: No problem @kinow! I had not reached the tests yet. 
It will take me a while as I'm not familiar with the platfroms.

autosubmit/job/job_packages.py
```diff
@@ -27,72 +27,86 @@
...
-        self.x11_options = jobs[0].x11_options
+        self.export = first_job.export
+        self.executable = first_job.executable
+        self.x11_options = first_job.x11_options
         # Scheduler manages the timeout, this is for platforms without scheduler
         # Wrappers are only allowed within a scheduler and timeout is calculated differently there
-        self.timeout = max(job.wallclock_in_seconds for job in jobs) if jobs[0].platform.type in ["PS",
-                                                                                                  "LOCAL"] else None
-        self.x11 = jobs[0].x11
+        ps_local_platforms = [PsPlatform.TYPE, LocalPlatform.TYPE]
```
ntorqulu: Is it better to define an boolean attribute on the platforms like `no_scheduler_platform` set to `TRUE` for LocalPlatform and PsPlatform and false for the others?
This way we don't have to manually maintain this list if new platforms that do not have a scheduler are added.

autosubmit/job/job_packages.py
```diff
@@ -27,72 +27,86 @@
...
+        is_ps_local = first_job.platform.type in ps_local_platforms
```
ntorqulu: In the ATTTENTION comment above (line 65) it is said to do not use `first_job.platfom` as it's a property getter. But the code does this, which can return serial_platform for serial jobs. Should be changed to `first_job._platform.type`?

autosubmit/job/job_packages.py
```diff
@@ -124,15 +138,14 @@ def jobs(self) -> list[Job]:
...
 
     @property
-    def platform(self):
-        """
-        Returns the platform
-
-        :return: platform
-        :rtype: Platform
-        """
+    def platform(self) -> 'Platform':
```
ntorqulu, line 141: Which should be the conventions for the docstrings in getters?
Should we not include the return in the comment?

autosubmit/job/job.py
```diff
@@ -2334,7 +2334,7 @@ def init_platform(self, as_conf: AutosubmitConfig) -> None:
         if not self.platform:
             submitter = ParamikoSubmitter(as_conf=as_conf)
             if not self.platform_name:
-                self.platform_name = as_conf.experiment_data.get("DEFAULT", {}).get("HPCARCH", "LOCAL")
+                self.platform_name = as_conf.experiment_data.get("DEFAULT", {}).get("HPCARCH", LocalPlatform.TYPE.upper())
             self.platform = submitter.platforms.get(self.platform_name)
 
     def update_content_extra(self, as_conf: AutosubmitConfig, files: list[str]) -> list[str]:
```
ntorqulu, line 2359: Line 2864 of `job.py` also uses a string as platform comparison:
```
        # Launch second as threaded function only for slurm
        if job_data_dc and type(self.platform) is not str and self.platform.type == "slurm":
```

ntorqulu (commented):
Hi @kinow! 
Thanks for the improvements on how we compare the platforms types. This way we can get rid of the string comparisons.

For keeping track which platforms have a scheduler and which one don't I'm not sure if keeping them in a list will introduce errors in the future, as adding new platforms will imply updating also the list `ps_local_platforms`. 
What do you think about keeping it as an attribute?

Also, just a small thing about formatting, there are strings that use single quotes and others using double quotes. Could you apply the linter to the changes? 
Thanks!

kinow:
Good questions, @ntorqulu ! Let me think about the scheduler one. About the linter, initially we had nothing, then @VindeeR added the initial checks, but I think we didn't enable some (line-length, single/double quote) because we had some major PRs that would break.

Are we going to adopt only single or double quotes (or at least primarily unless we need to combine/mix them)? I can adjust the ones I used here, but it might be easier to maybe switch everything to one option + enable ruff/linter in CICD... maybe we can do it now during the summer break.

autosubmit/job/job.py
kinow: Great catch!! I will check for the other platforms. I only checked ps + local :-O Thanks Núria!

autosubmit/job/job_packages.py
kinow: Good question. I think we should include the return. I can' t see in GitHub UI if I removed, but if I did it was my mistake. The only thing I'm avoiding it duplicating the types in docstrings + type hints, as I found (still find...) places where the docs were outdated, and they cannot be automatically checked as with mypy/type-checkers with the type hints. So I'm using just the type-hints for types.

autosubmit/job/job_packages.py
kinow: Let me check. I wrote that while troubleshooting the text, I may have to adjust the message. It's quite confusing, TBH. You must use that when you want the serial platform in some cases, I think :dizzy:

ntorqulu:
> Are we going to adopt only single or double quotes (or at least primarily unless we need to combine/mix them)? I can adjust the ones I used here, but it might be easier to maybe switch everything to one option + enable ruff/linter in CICD... maybe we can do it now during the summer break.

Okay! I though we used Ruff as linter and we had to format the code before merging it. Ruff formats the single quotes to double ones. 
Nevertheless, it's not that important and enabling ruff in CICD for the code changed seems like a good idea.

autosubmit/job/job_packages.py
kinow: Ha! I didn' t consider removing the list at all. I was focusing only in the constants. Excellent question. Let me think about this one...

ntorqulu:
Also, I think this pr can close the following issue too: #3047 as `_get_platform_by_type` no longer uses strings for checking the different platform types.

autosubmit/platforms/slurmplatform.py
kinow: Ah, forgot to reply here. I tried this on Monday, but it was a nightmare to make both Platform and ParamikoPlatform abstract.

1) there are tests instantiating these classes, which is no major block but 2) the functions that raise NotImplementedError are not all implemented in the subclasses.

The 2) is really annoying. There are functions in Slurm, for example, that have the NotImplementedError or just don't do anything. I think if the classes were initially abstract, the devs wouldn't have done that as abc would fail and they would have to rethink how they were designing the functions.

So making both of those classes abstract will stay for a future issue. For now I added it only to Platform.

ntorqulu: Yes, I imagin! I think is a good idea to leave just the Platform as abstract for now in order to start introducing small improvements into the platforms design and handle the abstract methods in another pr.
Also, while checking the structure I noticed some things that made noise for me, not sure if is worth adding them in the future issue:
- `LocalPlatform` and `EcPlatform` both inherit from `ParamikoPlatform`. They are overriding ssh-methods when they don't need them. Maybe better inherit from Platform instead?
- For when we handle the abstract methods, `check_file_exists` and `_check_for_unrecoverable_errors` have diff signatures across the diff platforms and `get_submitted_job_id` diff return types.
- Also there are some methods defined in Platforms that could carry problems in the future: 
```
def check_all_jobs(self, job_list, as_conf, retries=5):
    """Checks jobs running status"""
```
Someone could create a new subclass and forget to override this.

```
def check_file_exists(self, src, wrapper_failed=False, sleeptime=5, max_retries=3):
    return True
```
And the same here, but then all files will be reported as existing.

autosubmit/autosubmit.py
```diff
@@ -1440,7 +1465,7 @@ def expid(description, hpc="", copy_id='', dummy=False, minimal_configuration=Fa
             create_required_folders(exp_id, exp_folder)
         except OSError as e:
             with suppress(Exception):
-                Autosubmit._delete_expid(exp_id, True)
+                delete_experiment(exp_id, True)
```
kinow, line 1467: Not part of this change, but this was broken. This function had been replaced by `delete_experiment`. Somehow it was added back. I will write a few tests to cover it so it doesn't happen again. Right now I believe this would trigger a runtime error, ignored by the try/catch probably.

test/unit/platforms/test_paramiko_submitter.py
```diff
@@ -373,3 +373,59 @@ def test_add_invalid_platform(autosubmit_config):
...
+                "SCRIPT": "echo 'hello world'",
+            },
+        }
+    }
+], ids=["Undefined", "Unsupported"])
+def test_load_platforms(autosubmit_config, config):
+    experiment_id = 'random-id'
+    as_conf = autosubmit_config(experiment_id, config)
+    with pytest.raises(AutosubmitCritical):
+        ParamikoSubmitter(as_conf=as_conf)
```
kinow, line 474: These functions were in test/unit/test_paramiko_submitter.py. But we also had test/platforms/test_paramiko_submitter.py. I kept the one that matches the package structure.

autosubmit/platforms/slurmplatform.py
```diff
@@ -244,14 +238,6 @@ def get_submitted_jobs_by_name(self, script_names: list[str]) -> list[int]:
...
         return submitted_job_ids
 
-    def get_check_job_cmd(self, job_id: str) -> str:
-        """Generates sacct command to the job selected.
-
-        :param job_id: ID of a job.
-        :return: Generates the sacct command to be executes.
-        """
-        return f'sacct -n -X --jobs {job_id} -o "State"'
-
```
kinow, line 254: Not used.

autosubmit/platforms/slurmplatform.py
```diff
@@ -178,16 +182,6 @@ def get_remote_log_dir(self) -> str:
...
 
-    def parse_job_output(self, output: str) -> str:
-        """Parses check job command output, so it can be interpreted by autosubmit.
-
-        :param output: output to parse.
-        :type output: str
-        :return: job status.
-        :rtype: str
-        """
-        return output.strip().split(' ')[0].strip()
```
kinow, line 189: Not used.

autosubmit/platforms/platform_type.py
kinow, line 1: @ntorqulu you linked that issue about an enum. I think we have another issue about replacing the job status and other enumeration fields that are using ints/strings by enums, but we never do that... so, here's a tentative at introducing enums to AS code base. Let me know if you think this change grew too much. I can simplify it and move the rest to a separate branch and send another PR later on.

autosubmit/platforms/execution_mode.py
```diff
@@ -0,0 +1,34 @@
...
+Batch execution means that the jobs are scheduler on a batch scheduler
+(meta-scheduling).
+
+Direct execution means that the jobs are executed directly in the platform.
+"""
+
+
+class ExecutionMode(str, Enum):
+    BATCH = "batch"
+    DIRECT = "direct"
```
kinow, line 33: @ntorqulu today I thought about your other comment about keeping a list of platforms, versus having an attribute... I first went with a `is_batch_scheduler`, setting this to False on local+ps, and True elsewhere. But then I realized that info "belongs" to the class.

If you have a SlurmPlatform, you immediately know it's of type slurm, and it executes jobs in a batch scheduler. So I introduced a `EXECUTION_MODE` to the `PLATFORM`, and used this new enum for that.

This way we have in the signature of platforms that they must have a type & execution mode, and that they must declare the type to be a valid value from the enumeration, and the same for the execution mode.

Based on the execution mode (batch vs. direct) we know whether we want to calculate the timeout, or if we will delegate to the batch server.

If a platform is removed, we never have to update any list. Adding things too, you add everything in a new class, without having to update too many places (you still need to update the function to get the platform from a string/type, but we can solve that one in a future PR too.

kinow (commented):
I still have a few more comments from @ntorqulu to go through, and a few (4-10) lines to write tests to. Addressed just a couple of comments today.

autosubmit/platforms/slurmplatform.py
kinow: And I think it should be safe to remove as the API probably doesn't use platforms code (I will check anyway later).

autosubmit/platforms/paramiko_platform.py
```diff
@@ -129,11 +128,24 @@ def _get_user_config_file(
...
+    """Retrieve the system event notification mechanism.
+
+    Uses `poll()` on Linux and `kqueue()` on BSD-based systems.
+
+    References:
+        Python select: https://docs.python.org/3/library/select.html
+        poll(2): https://man7.org/linux/man-pages/man2/poll.2.html
+        kqueue(2): https://man.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2
+    """
+    return select.kqueue() if sys.platform != "linux" else select.poll()
```
kinow, line 141: Actually.... I'll read more about it tomorrow, but I think we should ditch this code and just use `select.DefaultSelector`. Maybe in a future PR. I think it will use `epoll` on Linux, which should perform better than `select.poll`. And less code for us to write tests/maintain.

autosubmit/config/configcommon.py
```diff
@@ -1914,7 +1921,7 @@ def load_current_hpcarch_parameters(self, parameters: Optional[dict] = None) ->
             target["HPCROOTDIR"] = Path(scratch) / project / user / self.expid
             target["HPCLOGDIR"] = target["HPCROOTDIR"] / f"LOG_{self.expid}"
         # Default local paths.
-        elif hpcarch.upper() == "LOCAL":
+        elif hpcarch.lower() == LocalPlatform.TYPE:
             target["HPCROOTDIR"] = Path(BasicConfig.LOCAL_ROOT_DIR) / self.expid / BasicConfig.LOCAL_TMP_DIR
             target["HPCLOGDIR"] = target["HPCROOTDIR"] / f"LOG_{self.expid}"
```
ntorqulu, line 1968: A little nitpicking, just for completeness. 
In `configcommon.py` 

1. line 1580:
```
if platform_name == 'LOCAL':
```

2. line 2123:
```
except KeyError:
                job.platform = submitter.platforms["LOCAL"]
```
Replace for TYPE

autosubmit/job/job_list.py
```diff
@@ -2209,7 +2209,7 @@ def get_waiting_remote_dependencies(self, platform_type='slurm'.lower()):
         :rtype: list
         """
         waiting_jobs = [job for job in self._job_list if (
```
ntorqulu: `platform_type='slurm'.lower()` in the method signature could be replaced by the PlatformType enum

autosubmit/helpers/utils.py
```diff
@@ -344,7 +344,7 @@ def build_and_connect_platform(platform_name: str, as_conf: 'AutosubmitConfig',
     :param expid: Experiment identifier.
     :return: Connected platform instance.
     """
-    if platform_name.upper() == "LOCAL":
+    if platform_name.lower() == LocalPlatform.TYPE.value:
```
ntorqulu: `LocalPlatfrom.TYPE.value` is equal to `LocalPlatform.TYPE` right? Should be better to keep it just as TYPE to make it cleaner?

test/integration/commands/run/test_run_local_scenarios.py
```diff
@@ -796,7 +795,7 @@ def test_build_and_connect_platform_local(autosubmit_exp, general_data):
         "EXPERIMENT": {"MEMBERS": "fc0", "NUMCHUNKS": "1"},
     }
     as_exp = autosubmit_exp(experiment_data=experiment_data, include_jobs=False, create=True)
-    plat = build_and_connect_platform("LOCAL", as_exp.as_conf, as_exp.expid)
+    plat = build_and_connect_platform(PlatformType.LOCAL.value, as_exp.as_conf, as_exp.expid)
     assert isinstance(plat, LocalPlatform)
-    assert plat.type == "local"
+    assert plat.TYPE == PlatformType.LOCAL.value
```
ntorqulu: LocalPlatfrom.TYPE.value is equal to LocalPlatform.TYPE right? Should be better to keep it just as TYPE to make it cleaner?

autosubmit/job/job_packages.py
```diff
@@ -27,72 +27,82 @@
...
-        self.x11_options = jobs[0].x11_options
-        # Scheduler manages the timeout, this is for platforms without scheduler
-        # Wrappers are only allowed within a scheduler and timeout is calculated differently there
-        self.timeout = max(job.wallclock_in_seconds for job in jobs) if jobs[0].platform.type in ["PS",
-                                                                                                  "LOCAL"] else None
-        self.x11 = jobs[0].x11
+        self.export = first_job.export
+        self.executable = first_job.executable
+        self.x11_options = first_job.x11_options
+        is_batch_platform = first_job.platform.EXECUTION_MODE == ExecutionMode.BATCH
```
ntorqulu: `Job.platform` can return `_platform.serial_platform`and serial_platform returns itself when is None or a different platform when one is configured:
```
    @property
    def serial_platform(self):
        """Platform to use for serial jobs.

        :return: platform's object
        :rtype: platform
        """
        if self._serial_platform is None:
            return self
        return self._serial_platform
```
I think that if in `platforms.yml` you have defined a serial platform (ps) and a non-serial platform (slurm) then:
1. first_job.platform -> calls Job.platform property
2. slurm.serial_platform -> calls Platform.serial_platform
3. It returns ps
4. So EXECUTION_MODE -> ExecutionMode.DIRECT, which is compared against BATCH -> False

I would substitute it by:
`self.platform.EXECUTION_MODE == ...`

autosubmit/platforms/paramiko_platform.py
ntorqulu: Yes, from what I've understood `selectors.DefaultSelector()` auto-picks epoll/kqueue/poll/select and then you just need one one API call, which simplifies the code and makes it more optimal (O(1) instead of O(n).
Is worth implementing it in another pr.

autosubmit/autosubmit.py
```diff
@@ -4513,10 +4535,10 @@ def _copy_code(as_conf: AutosubmitConfig, expid: str, project_type: str, force:
                 hpcarch = submitter.platforms[as_conf.get_platform()]
             except AutosubmitCritical as e:
                 Log.warning(f"{e.message}\nRemote git cloning is disabled")
-                hpcarch = "local"
+                hpcarch = LocalPlatform.TYPE
```
ntorqulu: I think it needs to return the LocalPlatform instance instead of the enum string as `hpcarch` later calls `send_command()` etc.).
I think it should be `submitter.platforms[LocalPlatform.TYPE]`

autosubmit/autosubmit.py
ntorqulu: Perfect, thanks!

autosubmit/job/job.py
```diff
@@ -2861,7 +2869,7 @@ def write_end_time(self, completed, count=-1):
                                                     fail_count=count if count >= 0 else self.fail_count)
 
         # Launch second as threaded function only for slurm
-        if job_data_dc and type(self.platform) is not str and self.platform.type == "slurm":
+        if job_data_dc and type(self.platform) is not str and self.platform.TYPE is PlatformType.SLURM:
```
ntorqulu, line 2883: Better use == for enum comparion to compare enum-to-enum

autosubmit/platforms/execution_mode.py
ntorqulu: Great, thanks! Yes, there's still the problem with the `_get_platform_by_type`. 
Maybe can be solve it with decorators, but for another pr 😅

autosubmit/platforms/execution_mode.py
```diff
@@ -0,0 +1,34 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
+from enum import Enum
+
+"""The platform execution mode.
```
ntorqulu: docstring is not attached to the class. As with platform_type, I would move the docstring inside the class.

ntorqulu (commented):
Thanks for incorporating enums @kinow!
The only real issue I see is in `autosubmit.py`. in `_copy_code`, where the fallback was passing an enum instead of a Platform object.
All the other changes are more cosmetic, just to normalize things.

About when to use TYPE vs TYPE.value for the PlatformType enum, at the end both return the same right? When should one be used instead of the other?
I started to get confused about it mid review...
Should be .value for serialization things and .TYPE for all the other?

autosubmit/platforms/paramiko_platform.py
kinow: Great! I will create an issue for that. Thanks Núria!

EDIT: https://github.com/BSC-ES/autosubmit/issues/3098

kinow:
Rebased and fixed conflicts.

test/integration/commands/run/test_run_local_scenarios.py
kinow: Yes! Thanks.

autosubmit/helpers/utils.py
kinow: True that. I keep using `.value` because I got used to how enums worked before stringenum & the (enum, str) were widely used. I'll switch to use the enum type directly. Excellent catch, thanks Núria!

autosubmit/config/configcommon.py
```diff
@@ -2156,7 +2164,7 @@ def load_section_parameters(self, job_list, as_conf, submitter):
             try:
                 job.platform = submitter.platforms[job.platform_name]
             except KeyError:
-                job.platform = submitter.platforms["LOCAL"]
+                job.platform = submitter.platforms[PlatformType.LOCAL.upper()]
```
kinow, line 2165: The `ParamikoSubmitter` creates a dictionary with key-value where the key is in upper.

Image

I will not change that here 👍

autosubmit/platforms/platform.py
```diff
@@ -573,7 +580,7 @@ def add_parameters(self, as_conf: 'AutosubmitConfig'):
         as_conf.experiment_data['HPCBUDG'] = self.budget
         as_conf.experiment_data['HPCRESERVATION'] = self.reservation
         as_conf.experiment_data['HPCEXCLUSIVITY'] = self.exclusivity
-        as_conf.experiment_data['HPCTYPE'] = self.type
+        as_conf.experiment_data['HPCTYPE'] = self.TYPE.value
```
kinow, line 578: Just because the dict is/will be probably a generic `[str, str]`, so it doesn't hurt to use `.value` which is `str`, instead of `self.TYPE` which is `PlatformType` (inherently a `str` too, but probably better for type-checkers & IDEs a str here).

autosubmit/platforms/slurmplatform.py
kinow: > LocalPlatform and EcPlatform both inherit from ParamikoPlatform. They are overriding ssh-methods when they don't need them. Maybe better inherit from Platform instead?

Definitely. Local at least I am sure we document as a local-only, no SSH, platform. So it doesn't make sense that it inherits from ParamikoPlatform.  https://github.com/BSC-ES/autosubmit/issues/3102

> For when we handle the abstract methods, check_file_exists and _check_for_unrecoverable_errors have diff signatures across the diff platforms and get_submitted_job_id diff return types.

Yeah, I noticed that before. My IDE screams every time I run analysis check over that part of the code. I believe it will be fixed through https://github.com/BSC-ES/autosubmit/issues/1117

> Also there are some methods defined in Platforms that could carry problems in the future:

Agreed, although I am quite sure this same problem is present in most of the other places where OOP classes were introduced. Not sure if we need an issue for this, or if this will be fixed when we work on AS5.

autosubmit/job/job_packages.py
kinow: image

I think only the type is not specified here. For now I am omitting the `:return:` (as mentioned above) as there are multiple places where it's out of sync, and we have no way to detect that. Whereas with type hints we have a possible way to do so (if we implement it correctly).

image

autosubmit/job/job.py
kinow: Did a search for "local 'local "slurm 'slurm, etc., repeating with UPPER CASE. I think/hope I got all the string values.

autosubmit/config/configcommon.py
kinow: Both fixed! Thanks!

autosubmit/job/job.py
kinow: Isn't that the other way around?

image

I think both work, but if we ever have another enum with a similar name, e.g., PlatformVariety, and that type has a `"slurm"` (i.e. now we have hypothetically PlatformType.SLURM and PlatformVariety.SLURM), if we do a `PlatformType.SLURM == PlatformVariety.SLURM`, then that cause a bug where that's evaluated to `True`.

image

Both should work, but in the long term I guess `is` is better?

So probably safer to keep the `is` here?

autosubmit/platforms/execution_mode.py
kinow: Opsie :laughing:

kinow: Fixed! Thanks.

kinow:
@ntorqulu I think I replied all your comments, resolving the ones where I fixed what you pointed was wrong (thanks!). I am waiting for the GH Actions run to complete so I can check what I have to finish writing tests for :+1: I hope to have all tests done this week (today, maybe tomorrow).

autosubmit/job/job_packages.py
kinow: Hm, I still have to check this. For later.

kinow:
About the enum question, @ntorqulu , 

> Should be .value for serialization things and .TYPE for all the other?

I was just used to using `.value`. I've already modified to use the enum both as enum & str. But to use as `str` when it improves type-checking (.e.g., `dict[str, str]` where you set the enum as key or value).

autosubmit/job/job_packages.py
kinow: Comment updated!

kinow:
@ntorqulu I bit the bullet and updated that function _get_platform_by_type, it's now using a dictionary/map to return the class and then creates it. It fails if the type is invalid.I think ti's fully covered, but it might need a review when you have some time to spare with calm. Tried to add tests to other parts of the code touched, or that were nearby but didn't have tests.

kinow:
Argh. One more place missing a test. But I won't touch the code (I hope), just add one or two new tests.

autosubmit/job/job_packages.py
kinow: Missed that one. For tomorrow/later too!

autosubmit/autosubmit.py
kinow: Another one missed, sorry. Will update it Wed->Fri I think.

kinow: Oh, boy. Yup. This is definitely a bug (that's not introduced by this change, but we must fix it!). Excellent catch Núria!!!! Let me fix this, then write test, etc... :hammer: :hammer_and_wrench: :construction_worker: This might take a while and make this PR a bit bigger too :sweat_smile: but it's what it it's. ta

kinow:
@ntorqulu found a bug, not caused by this change, while reviewing the code. I've confirmed the bug, and pushed a commit (the last one) with the fix. I will see how tests + coverage go, and improve the tests a bit if needed. I will also create & link another issue in our repo so we have a record of what changed/why/where.

autosubmit/autosubmit.py
kinow: https://github.com/BSC-ES/autosubmit/issues/3108 fixed here now.

kinow:
Oh boy, had 1 place missing... fixed the bug with hpcarch from #3108,

image

autosubmit/job/job_packages.py
kinow: Ok. I read this comment, and took me a couple of minutes to put all variables in my brain and simulate and confirm if this could happen. I believe you are correct, and this is possibly another bug, although not sure if introduced by this change or not.

What I will do now is,

1. Write a simple unit test (mocking things if needed) that creates the scenario you explained
2. Fix it with the self.platform
3. Add a small comment above explaining why we don't use (and shouldn't use) the first_job platform
4. Add more tests if needed for coverage/code safety

Thanks!!

kinow: Excellent review, @ntorqulu !

Reproduced the scenario described by Núria. Doing that, I found some bugs in the platform, that I will open separate issue for (basically we require UPPERCASE everywhere for platforms, using lower case and mixing it caused several issues in my tests, so AS is not case-insentive).

Núria, here we can see the `assert platform.EXECUTION_MODE is BATCH` passing, but then failing the test because the `timeout` is set. That's wrong, because the first_job's platform is a Slurm (which happens to have a serial platform too).

image

After it failed, I just did as you suggested, and used `self.platform` instead, and now the test passes fine :-) 

image

kinow:
All done. Coverage report is missing because of a codecov outage that Luiggi informed others on Slack. Tomorrow I will kick the build again.

autosubmit/job/job.py
ntorqulu: True! I mixed them up in my head; is safer to use `is` instead of `==`. Thanks!

kinow: I do that too. Thanks for being thorough with the review :bow:

autosubmit/job/job.py
ntorqulu: Yes, I did the same. I think they're all covered now!

kinow: Hooray!

ntorqulu:
Great, thanks for the work @kinow. I'll check all the new changes today, probably finish tomorrow!

kinow:
> Great, thanks for the work @kinow. I'll check all the new changes today, probably finish tomorrow!

Thank you for the review, @ntorqulu . I think (and hope) we solved the problem with lower/upper case found by Muhammad and also did a few improvements in autosubmit_git/enums/etc :muscle: :rocket:

autosubmit/platforms/platform.py
```diff
@@ -575,9 +575,11 @@ def add_parameters(self, as_conf: 'AutosubmitConfig'):
         as_conf.experiment_data['HPCBUDG'] = self.budget
         as_conf.experiment_data['HPCRESERVATION'] = self.reservation
         as_conf.experiment_data['HPCEXCLUSIVITY'] = self.exclusivity
-        as_conf.experiment_data['HPCTYPE'] = self.type
+        as_conf.experiment_data['HPCTYPE'] = self.TYPE
```
ntorqulu: When running `autosubmit run <EXPID>` the following warning appears:
```
autosubmit run a060

Loading JobList
Load finished
Creating jobs...
Adding dependencies to the graph..
Adding dependencies to the job..
Transitive reduction...
Looking for edgeless jobs...
[WARNING] Failed to save experiment_data.yml: cannot represent an object: <PlatformType.LOCAL: 'local'>
Autosubmit is running with v4.1.17
Checking the connection to all platforms in use
No more jobs to run.
Waiting for all logs to be updated
Run successful
ROCRATE not present in experiment YAML configuration. No RO-Crate archive created.
```
I think it's because we're storing in experiment_data.HPCTYPE the enum object instead of the string. This causes an warning and `experiment_data.yml` file is not stored correctly.
I think it should be changed for `self.TYPE.value` to store the platform`s string name.

autosubmit/config/configcommon.py
```diff
@@ -99,10 +102,10 @@ def platforms_data(self) -> dict[str, Any]:
         try:
             hpcarch = str(
                 self.experiment_data.get("DEFAULT", {}).get("HPCARCH", "")
-            ).upper()
+            ).lower()
             platforms = self.experiment_data.get("PLATFORMS")
             if platforms is None:
-                if hpcarch == "LOCAL":
+                if hpcarch == LocalPlatform.TYPE:
```
ntorqulu: Maybe is better to use `PlatformType.LOCAL`. This way we avoid importing both `LocalPlatform` and `PlatformType`.

autosubmit/config/configcommon.py
```diff
@@ -1359,7 +1367,7 @@ def check_platforms_conf(self, no_log=False):
         """Checks experiment's platforms configuration file."""
         parser_data = self.experiment_data.get("PLATFORMS", {})
         main_platform_found = False
-        if self.hpcarch == "LOCAL":
+        if self.hpcarch.lower() == LocalPlatform.TYPE:
```
ntorqulu: Same, I would import just PlatformType and use PlatformType.LOCAL.

autosubmit/config/configcommon.py
```diff
@@ -1372,7 +1380,7 @@ def check_platforms_conf(self, no_log=False):
                     self.wrong_config["Platform"] += [[section, "Mandatory TYPE parameter not found"]]
                 else:
                     platform_type = platform_type.lower()
-                if platform_type != 'ps':
+                if platform_type.lower() != PsPlatform.TYPE:
```
ntorqulu: Same, to avoid importing PsPlatform into configcommon I would use PlatformType.PS.

autosubmit/config/configcommon.py
```diff
@@ -1956,7 +1964,7 @@ def load_current_hpcarch_parameters(self, parameters: Optional[dict] = None) ->
             target["HPCROOTDIR"] = Path(scratch) / project / user / self.expid
             target["HPCLOGDIR"] = target["HPCROOTDIR"] / f"LOG_{self.expid}"
         # Default local paths.
-        elif hpcarch.upper() == "LOCAL":
+        elif hpcarch.lower() == LocalPlatform.TYPE:
```
ntorqulu: Same, to avoid multiple imports.

autosubmit/config/configcommon.py
```diff
@@ -1938,7 +1946,7 @@ def load_current_hpcarch_parameters(self, parameters: Optional[dict] = None) ->
         :param parameters: Dictionary to populate with HPC values. If None, use self.experiment_data.
         """
         platforms = self.experiment_data.get("PLATFORMS", {})
-        hpcarch: str = self.experiment_data.get("DEFAULT", {}).get("HPCARCH", "LOCAL")
+        hpcarch: str = self.experiment_data.get("DEFAULT", {}).get("HPCARCH", LocalPlatform.TYPE)
```
ntorqulu: Same here to avoid multiple platform's imports.

autosubmit/git/autosubmit_git.py
```diff
@@ -124,210 +133,223 @@ def clean_git(as_conf: AutosubmitConfig) -> bool:
...
-                command_1 += "git checkout; "
+            git_version = subprocess.check_output("git --version", shell=True)
+            git_version = git_version.decode(locale.getlocale()[1]).split(" ")[-1].strip("\n")
+
+            version_int = ""
+            for number in git_version.split("."):
+                version_int += number
+            git_version = int(version_int)
+        except Exception:
+            git_version = 2251
```
ntorqulu, line 270: Why this hardcoded git version by default? Should we better use constant `GIT_CONFIG_GLOBAL`?

autosubmit/git/autosubmit_git.py
```diff
@@ -124,210 +133,223 @@ def clean_git(as_conf: AutosubmitConfig) -> bool:
...
+        if git_remote_project_path == '':
+            command_0 = "cd {0} ; {1}".format(project_path, command_0)
+            subprocess.check_output(command_0, shell=True)
+        else:
+            command_0 = "cd {0} ; {1}".format(project_path, command_0)
+            platform.send_command(command_0)
+        # command 1
+
+        if os.path.exists(os.path.join(git_path, ".githooks")) and git_version > 2136:
```
ntorqulu, line 279: Why this git version?

autosubmit/helpers/utils.py
```diff
@@ -344,7 +344,7 @@ def build_and_connect_platform(platform_name: str, as_conf: 'AutosubmitConfig',
     :param expid: Experiment identifier.
     :return: Connected platform instance.
     """
-    if platform_name.upper() == "LOCAL":
+    if platform_name.lower() == LocalPlatform.TYPE:
```
ntorqulu: Same, I would use the generic PlatformType instead of LocalPlatform to avoid in the future unnecessary imports

autosubmit/job/job_list.py
```diff
@@ -2200,7 +2200,7 @@ def get_waiting(self, platform=None, wrapper=False):
             return [job for job in waiting_jobs if job.packed is False]
         return waiting_jobs
 
-    def get_waiting_remote_dependencies(self, platform_type='slurm'.lower()):
+    def get_waiting_remote_dependencies(self, platform_type=PlatformType.SLURM.lower()):
```
ntorqulu: I think this method is never called, nor is used by the api... Should be deleted?

autosubmit/platforms/ecplatform.py
```diff
@@ -104,11 +112,11 @@ def _check_jobid_in_queue(self, ssh_output: str, job_list_cmd: str) -> bool:
     def __init__(self, expid, name, config, scheduler):
         ParamikoPlatform.__init__(self, expid, name, config)
         # version=scheduler
-        if scheduler == 'pbs':
+        if scheduler.lower() == PlatformType.PBS.lower():
             self._header = EcCcaHeader()
-        elif scheduler == 'loadleveler':
+        elif scheduler == PlatformType.LOAD_LEVELER.lower():
```
ntorqulu: `scheduler.lower()` as a safeguard too for LOADLEVELER & SLURM comparison too?

autosubmit/config/configcommon.py
```diff
@@ -2837,11 +2845,17 @@ def is_valid_jobs_in_wrapper(self, wrapper=None) -> bool:
...
+        origin = str(self.experiment_data["GIT"].get('PROJECT_ORIGIN', ""))
         branch = self.get_git_project_branch()
         commit = self.get_git_project_commit()
-        return origin_exists and (
-                (branch is not None and len(str(branch)) > 0) or (commit is not None and len(str(commit)) > 0))
+        return bool(
+            origin
+            and (
+                (branch is not None and len(str(branch)) > 0)
+                or (commit is not None and len(str(commit)) > 0)
```
ntorqulu, line 2856: Just to take it into account (no need fixing). I think that `(branch is not None and len(str(branch)) > 0)` is always `True` because branch defaults to `master` if no branch defined. So in the end, is dead code.
It's just to have in mind that everything after the `and` currently will always be `True`.

autosubmit/platforms/paramiko_submitter.py
```diff
@@ -201,7 +245,6 @@ def load_platforms(self, as_conf: 'AutosubmitConfig', auth_password: Optional[st
                 return
 
             # Set the type and version of the platform found
-            remote_platform.type = platform_type
             remote_platform._version = platform_version
 
             # Concatenating the host with a project and adding to the object
```
ntorqulu, line 251: `_get_serial_platforms` receives `platform_used` (uppercased) but looks up into `platform_data`, which comes from YAML config files (can be whatever the user enters).
If the users know that platforms can only be defined with upper-cases no problem. If they usually enter lower-cased platforms I would upper-case `platforms_data` like `platform_used` does.

```
platforms_data_upper = {k.upper(): v for k, v in platforms_data.items()}
platforms_serial_in_parallel = _get_serial_platforms(
    platforms_used=platforms_used,
    platforms_data=platforms_data_upper
)
```

ntorqulu (commented):
Hi @kinow! Review of the latest changes.

The main issue is while running `autosubmit run <EXPID>`. 
This causes a warning and `experiment_data.yml` file is not stored correctly. Probably due to the platform being stored as an enum instead of a string. Proposed a solution in one of the comments.

All the other comments are more related to future maintainability to avoid unnecessary imports.

I've seen a few more inconsistencies while reviewing the code not related to this pr in:
- `platform.py`: searches for TOTALJOBS as a child of PLATFORMS (e.g. PLATFORMS['TOTAL_JOBS']). `platform_total_jobs = self.config.get("PLATFORMS", {}).get('TOTAL_JOBS', config_total_jobs)`
- `platform.py`: `as_conf.platforms_data.get(self.name, {})` uses `self.name` without upper-case it.
- `paramiko_platform`: `get_job_id_by_job_name_cmd` returns NotImplementedError instead of raising it.

I'll open issues for these ones.

Thanks for the work!

kinow:
> I'll open issues for these ones.

Thanks @ntorqulu ! We do have quite a few inconsistencies in the code. Always cheaper for us and better for users if we can spot those ourselves and fix it before something breaks in the testing suite,.... or worst, in an operational run. I will go over the comments slowly today/tomorrow. Thanks again!

autosubmit/platforms/ecplatform.py
kinow: That won't hurt! I'll adjust it and mark as resolved once I've pushed and GH UI has shown it as outdated.

autosubmit/job/job_list.py
kinow: I'll check if it's not used elsewhere in the code, and if not I'll delete it. Thanks Núria!

autosubmit/helpers/utils.py
kinow: Argh. You are correct. Will fix it. Thanks!

autosubmit/git/autosubmit_git.py
kinow: Beats me. I think this was the version used when the code was written, so this was added as a safeguard in case previous versions of Git didn't have git hooks or some feature from git hook we needed. I think unfortunately there was no comment in the code explaining the magic-numbers. Sorry.

autosubmit/git/autosubmit_git.py
kinow: I think so, although it'd be good to understand why these specific versions are used in the code, and move that to a constant. I'll create an issue for this (I think you are creating issues for platforms, not for this right? Just so we don't create duplicated issues).

autosubmit/config/configcommon.py
kinow: :+1: great catch!

autosubmit/config/configcommon.py
kinow: Definitely, my bad. Thanks!!!

autosubmit/config/configcommon.py
ntorqulu: Will try to investigate more about this here: #3101.
In theory we can let git clone auto-detect the default branch instead of using `master` by default.

autosubmit/platforms/platform.py
kinow: Oh, yeah. I think somewhere else I used the `.value` to avoid this, exactly in a dictionary, but I missed this part. Will fix it.

autosubmit/helpers/utils.py
kinow: Yup, even though we use `LocalPlatform` in other parts of the code, it doesn't hurt to use `PlatformType` and make it more future-proofed. Thanks!

autosubmit/config/configcommon.py
kinow: I think this if/else logic was just moved (or at least I don't remember writing it). There are a few other if statements that can be simplified (in configcommon but also in autosubmit.py/and job classes). I'll add a TODO here.

```
        # TODO: Review the if/else logic; the branch name defaults to "master", so
        #       maybe a validator+typechecker will prevent empty/none from happening.
```

autosubmit/platforms/paramiko_submitter.py
kinow: Yeah, this is the bug I mentioned last Friday in the Slack room. This is not the only place where using lower case or upper case breaks in AS config. In the end, AS is case-sensitive, and prefers/favors UPPERCASE VALUES. I still have to create the issue for that, but I will fix this one here in the meantime, then verify if the issue occurs on this branch (or master after we merge this) with the other tests that I did on master last Friday.

autosubmit/job/job_list.py
kinow: Deleted! Along with a few other methods never used...

autosubmit/platforms/paramiko_submitter.py
kinow: Applied the change as suggested!

kinow:
@ntorqulu I believe I addressed all changes. I marked things as Resolved after confirming again everything was addressed. You can review the changes in this GH UI page: https://github.com/BSC-ES/autosubmit/pull/3086/changes/3fd93b0a4794feecf75076d482e8757734ae21a7

There's another diff shown above this comment (if you click on "Compare"), but that's because I also rebased the branch. The link I sent above is one above, for the 3fd93b0 commit.

image


Ready for another review :muscle:

autosubmit/git/autosubmit_git.py
kinow: Done! https://github.com/BSC-ES/autosubmit/issues/3116

autosubmit/git/autosubmit_git.py
kinow: https://github.com/BSC-ES/autosubmit/issues/3116

autosubmit/git/autosubmit_git.py
ntorqulu: Thanks!

autosubmit/git/autosubmit_git.py
ntorqulu: No problem! We can enforce a minimum git version needed for AS to work globally (https://github.com/BSC-ES/autosubmit/issues/3110) or define a min threshold depending on the case. Thanks for opening the issue!

ntorqulu:
Great, thanks for the work @kinow! All reviewed, feel free to merge :sparkles:

kinow:
Thank you for the great review @ntorqulu

PR #3088, opened 2026-06-30
Fix multi-wrapper issues (v4.1.17! destine) #3084

### LINKED ISSUE #3084: Possible bug in wrappers
### Description

* Autosubmit version: autosubmit/4.1.17-dev3-c2a87d0
* Machine/VM/environment name (if applicable): lumi_vm
* Experiment ID (if applicable): a3op
  * e.g. `/appl/AS/AUTOSUBMIT_DATA/a3op/tmp/ASLOGS/20260625_164056_inspect.log --> log after inspecting., similarly for the last AS create

### Reproducible Example

copy and run a3op, lumi

### Expected Behaviour

I expect that autosubmit to wrapp the jobs in packs of 10 or 15, but only a single wrapper at the end of the simulation is created, of 3-4 jobs.

The metadata file to check the wrapper and job conf can be found here: /appl/AS/AUTOSUBMIT_DATA/a3op/conf/metadata/experiment_data.yml

cc @mromeu-bsc please add anything that I can be missing

dbeltrankyl:
Fixes #3084

- Fixed wrapper deadlock false positive when non-wrapped jobs are still
  active or could become ready later #3084
- Fixed leftover wrapper batch stuck on MIN_WRAPPED when remaining
  jobs are blocked by the same package #3084

is_deadlock  only checked SIMPLE jobs that happened to be READY in that iteration, so it was missing active SIMPLE chains  that still had jobs RUNNING, QUEUING, or WAITING.
 

Additionally, when this was fixed another issue raised when the last batch of jobs couldn't meet MIN_WRAPPED (e.g., 8 remaining with MIN_WRAPPED=10), process_not_wrappeable_packages counted remaining WAITING jobs as "can still progress" even when they were blocked by the unwrappable package itself, creating a deadlock 

**Check List**

- [x] Documentation updated.


dbeltrankyl:
imagen

experiment info: 

```yaml

PLATFORMS:
  LUMI:
    TYPE: slurm
    HOST: lumi-cluster
    PROJECT: project_465002727
    USER: francerou
    SCRATCH_DIR: /scratch
    MAX_WALLCLOCK: 48:00
    MAX_PROCESSORS: 99999
    EXCLUSIVE: 'True'
    APP_PARTITION: standard
    CLEAN_PARTITION: small
    PARTITION: standard
    PROCESSORS_PER_NODE: 128

  LUMI-LOGIN:
    TYPE: ps
    HOST: lumi-cluster
    PROJECT: project_465002727
    USER: francerou
    SCRATCH_DIR: /scratch
    MAX_WALLCLOCK: 44:00

JOBS:
  DN:
    FILE: templates/dn.sh
    PLATFORM: LUMI-login
    RUNNING: chunk
    WALLCLOCK: 02:00
    SPLITS: auto
    TOTALJOBS: 10
    DEPENDENCIES:
      SIM:
        STATUS: RUNNING
        ANY_FINAL_STATUS_IS_VALID: true
      DN:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: previous
      DN-1: {}

  APP_ENERGY_INDICATORS:
    FILE: templates/application.sh
    PLATFORM: LUMI
    RUNNING: chunk
    WALLCLOCK: 00:20
    SPLITS: auto
    DEPENDENCIES:
      OPA_ENERGY_INDICATORS:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: '[1:auto]*\1'
      OPA_ENERGYTDIG2:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: '[1:auto]*\1'
      OPA_ENERGYTDIG1:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: '[1:auto]*\1'
      APP_ENERGY_INDICATORS:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: previous
      APP_ENERGY_INDICATORS-1: {}

  OPA_ENERGY_INDICATORS:
    FILE: templates/opa.sh
    PLATFORM: LUMI
    RUNNING: chunk
    WALLCLOCK: 00:20
    SPLITS: auto
    DEPENDENCIES:
      DN:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: '[1:auto]*\1'
      OPA_ENERGY_INDICATORS:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: previous
      OPA_ENERGY_INDICATORS-1: {}

  OPA_ENERGYTDIG1:
    FILE: templates/opa.sh
    PLATFORM: LUMI
    RUNNING: chunk
    WALLCLOCK: 00:20
    SPLITS: auto
    DEPENDENCIES:
      DN:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: '[1:auto]*\1'
      OPA_ENERGYTDIG1:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: previous
      OPA_ENERGYTDIG1-1: {}

  OPA_ENERGYTDIG2:
    FILE: templates/opa.sh
    PLATFORM: LUMI
    RUNNING: chunk
    WALLCLOCK: 00:20
    SPLITS: auto
    DEPENDENCIES:
      DN:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: '[1:auto]*\1'
      OPA_ENERGYTDIG2:
        SPLITS_FROM:
          ALL:
            SPLITS_TO: previous
      OPA_ENERGYTDIG2-1: {}

WRAPPERS:
  POLICY: strict
  WRAPPER_OPA_ENERGYTDIG1:
    TYPE: vertical
    JOBS_IN_WRAPPER:
    - OPA_ENERGYTDIG1
    MIN_WRAPPED: 10
    MAX_WRAPPED: 15
  WRAPPER_OPA_ENERGYTDIG2:
    TYPE: vertical
    JOBS_IN_WRAPPER:
    - OPA_ENERGYTDIG2
    MIN_WRAPPED: 10
    MAX_WRAPPED: 15
  WRAPPER_OPA_ENERGY_INDICATORS:
    TYPE: vertical
    JOBS_IN_WRAPPER:
    - OPA_ENERGY_INDICATORS
    MIN_WRAPPED: 10
    MAX_WRAPPED: 15
  WRAPPER_APP_ENERGY_INDICATORS:
    TYPE: vertical
    JOBS_IN_WRAPPER:
    - APP_ENERGY_INDICATORS
    MIN_WRAPPED: 10
    MAX_WRAPPED: 15
```

dbeltrankyl:
Ready to review

Cc: @LuiggiTenorioK , @franra9 ( this would be the full solution to your issue)

I would assign @kinow because of the urgency, as this is blocking the release

If you're not available, this mainly affects build_packages and the policies of the wrappers, so I think @ntorqulu knows a bit of this part of the code since she recently worked on the TOTALJOBS (not entirely related to this issue tho)

autosubmit/job/job_packager.py
```diff
@@ -305,21 +305,88 @@ def check_packages_respect_wrapper_policy(self, built_packages_tmp, packages_to_
...
+                if isinstance(sections, str):
+                    sections = sections.split()
+                for section in sections:
+                    wrapped.add(section.upper())
+        elif isinstance(jobs_in_wrapper, str):
+            for section in jobs_in_wrapper.split():
+                wrapped.add(section.upper())
+        return wrapped
+
+    def _has_blocking_non_wrapped_jobs(self) -> bool:
```
pablogoitia, line 335: For sure an interesting solution to catch out false positive deadlocks! I find the logic coherent and complete (I cannot guess any other case where a wrapper could be externally blocked forever).

autosubmit/job/job_packager.py
```diff
@@ -305,21 +305,88 @@ def check_packages_respect_wrapper_policy(self, built_packages_tmp, packages_to_
...
+        :return: True if there are non-wrapped jobs that block deadlock detection.
+        :rtype: bool
+        """
+        all_wrapped = self._get_wrapped_sections(self.jobs_in_wrapper)
+
+        for job in self._jobs_list._job_list:
+            if job.section.upper() in all_wrapped:
+                continue
+            if job.status in (Status.READY, Status.RUNNING, Status.QUEUING,
+                              Status.SUBMITTED, Status.SUSPENDED):
```
pablogoitia: Just a quick question. Do suspended mean that an already allocated job has been "paused" because of a preemptive scheduling configuration?

autosubmit/job/job_packager.py
```diff
@@ -305,21 +305,88 @@ def check_packages_respect_wrapper_policy(self, built_packages_tmp, packages_to_
...
-                and len(self._jobs_list.get_in_queue()) == 0
-                and len(not_wrappeable_package_info) == len(built_packages_tmp)
-        )
+        if any_simple_packages:
+            return False
+        if len(self._jobs_list.get_in_queue()) > 0:
+            return False
+        if self._has_blocking_non_wrapped_jobs():
+            return False
+        return len(not_wrappeable_package_info) == len(built_packages_tmp)
```
pablogoitia: This is a matter of taste, but maybe inverting these last two conditions would be a bit more efficient in most cases?

autosubmit/job/job_packager.py
pablogoitia, line 526: Here `remaining_can_progress` seems to be used as if it were a bool, so it could be changed accordingly, WDYT Dani?

Another option (which I prefer if my previous observation remains true) is just to merge the conditions so that 512 replaces 508, and 514-521 replaces 510. Same logic, less lines. We could also leave comments saying "remaining can/cannot progress" so the conditions can be better understood

pablogoitia (changes requested):
Hi @dbeltrankyl. I didn't want to miss the chance to take a look into a wrappers' PR. I found the solution very complete and interesting, with no visible uncovered cases. I just left few questions or comments that do not impact the implementation. I have not already had the opportunity to deeply check the tests, but I skimmed them and look complete.

autosubmit/job/job_packager.py
dbeltrankyl: The person launching the experiment could decide not to run some jobs through a set status; in these cases, we use the "SUSPENDED" status. Usually, the job was waiting -> suspended, not allocated anywhere ( for that, we had the HOLD status, but it is now deprecated). It is not a natural status of the workflow... 

When I included this... I was considering that, while the experiment runs, the person could use setstatus or update_list to unsuspend the job... But it is safer not to include it, so the experiment doesn't keep iterating if the user forgets about it

I'll change that. Thanks for the question, it let me reconsider the inclusion of it

autosubmit/job/job_packager.py
dbeltrankyl: Na, you're right, I didn't pay attention to the order

I'll change it  for
```python
        if any_simple_packages:
            return False
        if len(self._jobs_list.get_in_queue()) > 0:
            return False
        if len(not_wrappeable_package_info) != len(built_packages_tmp):
            return False
        if self._has_blocking_non_wrapped_jobs():
            return False
        return True
```
 I could avoid the last return True, but I find this more legible, symetrical and easier to debug with breakpoints etc

autosubmit/job/job_packager.py
dbeltrankyl: True, thanks for noticing it. I just removed that remaining_can_progress variable, it is not necessary

dbeltrankyl:
Changes applied! thanks @pablogoitia

dbeltrankyl:
I've added a commit to use 3.12 instead of 3.9 for integrations

Also, the misc was using 3.10 and the docs 3.9, changed to 3.12 as well

autosubmit/job/job_packager.py
pablogoitia: Ohh, I see. Thanks for the explanation!

> But it is safer not to include it, so the experiment doesn't keep iterating if the user forgets about it

Haven't thought about that, but agreed. I would also consider this situation as a deadlock. Anyway, the user will always have the option to "unsuspend" the conflicting jobs and resume the workflow whenever wanted

autosubmit/job/job_packager.py
pablogoitia: Sure. There would be many ways but this is the easier to understand from my POV. Thanks for checking it!

autosubmit/job/job_packager.py
pablogoitia: Thanks!

dbeltrankyl:
Hello! 

So I guess this can be merged? @pablogoitia, @kinow

kinow:
Having a quick look now before lunch...

.github/workflows/ci.yaml
```diff
@@ -82,17 +82,14 @@ jobs:
       matrix:
         # NOTE: We use 3.11 and not the latest 3.12 as the debugger of 3.12 is buggy
         #       and 3.11 is the current version used at the BSC ES department.
-        python-version: ["3.9", "3.11"]
+        python-version: ["3.11", "3.12"]
```
kinow: We had 3.9 as that was the oldest version, to ensure we didn't use anything that was not supported (e.g., `Union` from typing, or those string-enums from 3.11+). Maybe we should keep the lowest and the highest values?

.github/workflows/ci.yaml
```diff
@@ -192,9 +189,7 @@ jobs:
 
     strategy:
       matrix:
-        # NOTE: We use 3.11 and not the latest 3.12 as the debugger of 3.12 is buggy
-        #       and 3.11 is the current version used at the BSC ES department.
-        python-version: ["3.9", "3.11"]
+        python-version: ["3.11", "3.12"]
```
kinow: Ditto here.

.github/workflows/ci.yaml
```diff
@@ -82,17 +82,14 @@ jobs:
...
         pytest-markers: ["not postgres and not slurm"]
         # Here we include an extra build with the lowest version of Python and Postgres.
         # We do it this way to avoid wasting CICD resources since we already tested the
         # code with SQLite. So we run just one build with Postgres (normally newer versions
         # of Python may succeed while older fail -- e.g. using newer features in a code that
         # must support older versions).
         # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-adding-configurations
-        include:
-          - python-version: '3.9'
-            pytest-markers: 'postgres and not slurm'
```
kinow, line 95: Not needed anymore? Maybe it was moved to pytest or somewhere else? I think the idea was to have the lowest version testing it... the comment above is for that too.

autosubmit/job/job_packager.py
```diff
@@ -305,21 +305,90 @@ def check_packages_respect_wrapper_policy(self, built_packages_tmp, packages_to_
...
+
+    def _has_blocking_non_wrapped_jobs(self) -> bool:
+        """Check if there are non-wrapped jobs that can still progress, preventing a false deadlock.
+
+        With ``strict`` or ``mixed`` wrapper policies, ``is_deadlock`` must distinguish between:
+        - A genuine deadlock where nothing can ever progress.
+        - A temporary state where non-wrapped jobs are still active and will
+          eventually complete, unblocking the wrapper chain.
+
+        A non-wrapped job is considered to block the deadlock declaration if:
```
kinow: What does it mean to block the deadlock declaration? It makes it invalid? The part above is clear, this one I couldn't follow it.

kinow: Ah! Reading the `:return:` part, I think s/declaration/detection ?

kinow: And is it "the deadlock detection if the non-wrapped job status is:" ?

autosubmit/job/job_packager.py
```diff
@@ -305,21 +305,90 @@ def check_packages_respect_wrapper_policy(self, built_packages_tmp, packages_to_
...
-        )
+        if any_simple_packages:
+            return False
+        if len(self._jobs_list.get_in_queue()) > 0:
+            return False
+        if len(not_wrappeable_package_info) != len(built_packages_tmp):
+            return False
+        if self._has_blocking_non_wrapped_jobs():
+            return False
+        return True
```
kinow, line 398: Much easier to read! 👏

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 **Bug fixes:**
 
+- Fixed wrapper deadlock false positive when non-wrapped jobs are still
+  active or could become ready later #3084
+- Fixed leftover wrapper batch stuck on MIN_WRAPPED when remaining
+  jobs are blocked by the same package #3084
+
+
```
kinow: The extra lines are not necessary, I think?

kinow:
Going to have lunch, will finish quickly after I return. Just need to check one thing and then will probably press the approve button! I submitted a few comments I made reading the code in PyCharm. But the changes seem to be improving, and matching with the docs, and fixing the bug. Thanks Dani!

dbeltrankyl:
Let me know when you finish, so I can fix the feedback all at once. Thanks!

kinow:
Sorry, had a meeting with Suraj & Marti. Let me finish it now.

kinow (approved):
All good, @dbeltrankyl . I only mentioned the extra blank lines in the changelog, and I think we should buikd+test with the oldest Python version too. There's a dangling comment in the ci yaml from one part that was removed but I am not sure if it should be removed. But these comments are not directly related to this change, so leaving it approved as the code fix looks good! Thanks!

.github/workflows/ci.yaml
dbeltrankyl: Ah, I wrongly read it, thought it was a duplicate of `pytest-markers: not postgres and not slurm`
        
But it is actually postgres without the (not) I'll use 3.10 for it then

autosubmit/job/job_packager.py
dbeltrankyl: Hm, reading it again, for the WAITING part, this is only checking one level; if the grandparent fails, this could make autosubmit run endlessly, I believe.... I'll check tomorrow and add a function that checks if the non-wrapped parent chain can still progress if that's the case

dbeltrankyl: Because in autosubmit, if a job fails, the run still progresses until there are no more jobs to submit

pablogoitia (approved):

.github/workflows/ci.yaml
dbeltrankyl: Done

.github/workflows/ci.yaml
dbeltrankyl: done

.github/workflows/ci.yaml
dbeltrankyl: done

autosubmit/job/job_packager.py
dbeltrankyl: I think I was overcomplicating it, I changed the docstring based on Bruno inputs but the logic remained

dbeltrankyl:
Sorry for the delay, 

After more testing, I think I was overcomplicating the scenarios, but I found a bug that I thought I had resolved.

When a wrapper section has fewer READY jobs than MIN_WRAPPED (e.g., 8 jobs vs MIN_WRAPPED: 10) and no other jobs exist in the experiment for that section, the strict policy still raises the deadlock in some multi-wrapper cases

I added an `or`  that is called before checking the deadlock at all, which checks if there are more non_in_current_package of the sections wrapped. If there are no more possible jobs, it submits the wrapper as it is


Do you mind checking that @kinow :pray: it should be the last thing

>

kinow:
Not a problem, @dbeltrankyl . Just finishing late lunch. Will review it when I'm back in <hour. Thanks

autosubmit/job/job_packager.py
```diff
@@ -283,10 +283,17 @@ def check_packages_respect_wrapper_policy(self, built_packages_tmp, packages_to_
...
+            if (len(p.jobs) >= wrapper_limits["real_min"] and min_v >= wrapper_limits["min_v"] and min_h >=
+                wrapper_limits["min_h"] and not failed_innerjobs) or \
+                (not failed_innerjobs and self._remaining_blocked_by_package(
+                    self._jobs_list.get_jobs_by_section(
+                        self.jobs_in_wrapper[self.current_wrapper_section],
+                        [job.name for job in p.jobs],
+                        True
+                    ),
+                    p.jobs
+                )):
```
kinow, line 296: Not needed for this change, but maybe later we could break this if into multiple bool vars?

Something like

```py
if not failed_innerjobs:
  # If the quantity is enough
  has_min_quantity = (
    len(p.jobs) >= wrapper_limits["real_min"]
    and min_v >= wrapper_limits["min_v"]
    and min_h >= wrapper_limits["min_h"]
  )
  
  # If the rest of the jobs are blocked, i.e., nothing else can progress
  remaining_are_blocked = self._remaining_blocked_by_package(
      self._jobs_list.get_jobs_by_section(
          self.jobs_in_wrapper[self.current_wrapper_section],
          [job.name for job in p.jobs],
          True
      ),
      p.jobs
  )

  if has_min_quantity or remaining_are_blocked:
    for job in p.jobs:
      job.wrapper_type = p.wrapper_type
```

Maybe that's a bit easier to read... although I don't really trust code I write after 2PM on a Friday... 😅

kinow (approved):
Reviewed the diff in GitHub UI, looks good to me! Left one comment, but more for future improvement. Thanks Dani!!!

autosubmit/job/job_packager.py
dbeltrankyl: Thanks @kinow ! I'll open an issue for this so as not to delay dev4 even more

dbeltrankyl:
I've disabled some tests for hybrid wrappers. The fixes on this branch discovered* more issues with hybrid wrappers, which are really out of scope for this PR. We don't recommend using hybrid wrappers in 4.1 in general, so leaving it for 4.2. 

When the pipeline passes, I'll merge this

kinow:
Thanks Dani!

PR #3089, opened 2026-06-30
Simplify coverage setup

kinow:
A test cc @VindeeR 

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,8 +4,10 @@
 [run]
 branch = True
 cover_pylib = False
-concurrency = thread,multiprocessing
-data_file = .coverage
+# multiprocessing is not needed here because we are using spawn, and not fork.
+# So even with multiprocessing, the patched code (from coveragepy) would not
+# be used because of the new interpreter being launched.
+concurrency = thread
 disable_warnings =
     trace-changed
     module-not-python
@@ -15,9 +17,15 @@ disable_warnings =
 omit =
     tests/*
     autosubmit/job/template/common.py
-source = autosubmit
+source =
+    autosubmit
+    */autosubmit
+    /home/runner/work/autosubmit/autosubmit/autosubmit
 timid = False
 sigterm = True
+parallel = True
+patch =
+  _exit
 
 [report]
 exclude_lines =
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ permissions:
 jobs:
   test-unit:
     env:
-      COVERAGE_PROCESS_START: .coveragerc
+      COVERAGE_PROCESS_START: ${{ github.workspace }}/.coveragerc
     # needs: lint
     runs-on: ubuntu-latest
     timeout-minutes: 10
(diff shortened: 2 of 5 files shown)


kinow:
Apparently this is still not enough. I confirmed there are tests for  pmip notifier that are not showing the reports yet.

kinow:
Passing `remove=True`, or using int's as dictionary keys are not working in newer testcontainers.

test/integration/log_recovery/test_log_recovery.py
```diff
@@ -67,8 +69,8 @@ def test_log_recovery_keep_alive(prepare_test, local, mocker, as_conf):
     assert local.log_recovery_process.is_alive()
     local.work_event.set()
     assert local.log_recovery_process.is_alive()
-    # there is new auto-fixture that doesn't allow sleeps higher than 1 while testing
-    deadline = time.monotonic() + 1.2
+    # Give the child enough time to drain the queue and exit after the keep-alive timeout.
```
kinow, line 72: This might still be too short, but 5 seconds is probably safer than 1.2, without risking increasing the tests time too much.

test/integration/test_cpmip_metrics.py
```diff
@@ -29,9 +29,6 @@
 )
 from test.integration.test_utils.docker import get_mailhog_messages
 
-
-pytestmark = [pytest.mark.docker, pytest.mark.slurm, pytest.mark.ssh]
-
```
kinow, line 34: Took me a while to realise the test had these marks! The rest of the code is using @pytest.mark, so I adjusted this code to match the rest of the code base.

.coveragerc
```diff
@@ -15,9 +17,15 @@ disable_warnings =
 omit =
     tests/*
     autosubmit/job/template/common.py
-source = autosubmit
+source =
+    autosubmit
+    */autosubmit
+    /home/runner/work/autosubmit/autosubmit/autosubmit
```
kinow, line 23: This is to avoid possible issues if in GH action a coverage report file has /home/runner/.../autosubmit for some files, and autosbumit/.../ for other files. This could cause the same file to appear on both. This is common in python projects with coveragepy.

.github/workflows/ci.yaml
```diff
@@ -16,7 +16,7 @@ permissions:
 jobs:
   test-unit:
     env:
-      COVERAGE_PROCESS_START: .coveragerc
+      COVERAGE_PROCESS_START: ${{ github.workspace }}/.coveragerc
```
kinow, line 19: To avoid risks of using .coveragerc files found elsewhere. That probably wasn't happening, but doesn't hurt...

.github/workflows/ci.yaml
```diff
@@ -55,13 +55,24 @@ jobs:
...
+              if ! sqlite3 "$f" "SELECT version FROM coverage_schema;" >/dev/null 2>&1; then
+                echo "Invalid coverage file: $f"
+                mv "$f" ".ignored_cov_$(basename "$f")"
+              fi
+            fi
+          done
+          # A process terminated via os._exit() can leave an empty parallel data file.
+          if find . -maxdepth 1 -type f -name '.coverage.*' -size +0c -print -quit | grep -q .; then
+            coverage combine --append
+          fi
```
kinow, line 75: This isn't fixing the coverage, but I believe it's harmless to test before combining the coverage files.

.github/workflows/ci.yaml
```diff
@@ -133,13 +144,27 @@ jobs:
           
           pytest \
             -n $(nproc) \
-            --cov=autosubmit --cov-config=.coveragerc \
-            --cov-report=xml:test/coverage.xml --cov-append \
```
kinow, line 137: This was the fix. I tested locally, and without these I got the right coverage when exporting the env var to use .coveragerc. Everything else we had here is specified in that file. So there's no need to specify that here -- and by not doing it, we get the code from cpmip6 showing the correct coverage locally (in coverage.xml) and Codecov. Which was the goal here.

kinow:
@VindeeR the coverage was fixed in another PR, apparently. The code of @paparici is shown as covered now (the same code we looked before, from cpmip_metrics.py). The code of the log retrieval process is also shown as covered :tada: This means we are seeing all the covered tests in Codecov.

Can you have a look and review this one whenever you have time, please?

Thank you!

VindeeR (approved):
This changes looks good to me, thanks for fixing the coverage, no more losing coverage 🙌  thanks

kinow:
Thanks @VindeeR ! Merging!

Issue #2974, opened 2026-04-23
E-mail notifications causing experiment to fail


TracyMcBean:
### Description

* Autosubmit version: 4.1.16
* Machine/VM/environment name (if applicable): climatedt
* Experiment ID (if applicable): o02g
* Experiment tasks & log path if applicable):
`/appl/AS/AUTOSUBMIT_DATA/o02g/tmp/ASLOGS/20260422_151505_run.log`

### Reproducible Example

I am not sure I can make a reproducible example as this has actually been working before.

### Expected Behaviour

When using the e-mail notifications, I have so far gotten e-mails when a job fails. This is the correct behaviour. What now has happened is that my experiment fails with the following error message:

```
^[[38;5;214m[ERROR] Arguments provided: Namespace(version=False, logfile='DEBUG', logconsole='WARNING', command='run', expid='o02g', notransitive=False, update_version=False, start_time=None, start_after=None, run_only_members=None, profile=False)^[[0m^[[39m
^[[38;5;214m[ERROR] This is the experiment: <o02g> which had an issue with the command: <run> and it is currently using the Autosubmit Version: 4.1.16.^[[0m^[[39m
^[[1m ^[[31m[CRITICAL] Traceback (most recent call last):
  File "/appl/AS/4.1.16.1/lib64/python3.9/site-packages/autosubmit/scripts/autosubmit.py", line 105, in main
    return_value = Autosubmit.run_command(args)
  File "/appl/AS/4.1.16.1/lib64/python3.9/site-packages/autosubmit/autosubmit.py", line 761, in run_command
    return Autosubmit.run_experiment(args.expid, args.start_time, args.start_after, args.run_only_members,
  File "/appl/AS/4.1.16.1/lib64/python3.9/site-packages/autosubmit/autosubmit.py", line 2173, in run_experiment
    Autosubmit.job_notify(as_conf, expid, job, job_prev_status, job_changes_tracker)
  File "/appl/AS/4.1.16.1/lib64/python3.9/site-packages/autosubmit/autosubmit.py", line 1703, in job_notify
    Notifier.notify_status_change(MailNotifier(BasicConfig), expid, job.name,
  File "/appl/AS/4.1.16.1/lib64/python3.9/site-packages/autosubmit/notifications/notifier.py", line 25, in notify_status_change
    implementation.notify_status_change(exp_id, job_name, prev_status, status, notify_to)
[...]

PR #3097, opened 2026-07-03
JSON schema validation for AS configuration

### LINKED ISSUE #871: [enhancement] Validate Autosubmit configuration

ntorqulu:
Closes #871, #2974

**Check List**

- [ ] Documentation updated.


autosubmit/config/schema/autosubmit_schema.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
```
ntorqulu: Destine json schemas use Draft 2020-12. Red Hat YAML extension does not have full support for 2020-12: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml
*YAML Language support uses [JSON Schemas](https://json-schema.org/) to understand the shape of a YAML file, including its value sets, defaults and descriptions. The schema support is shipped with JSON Schema Draft 7.*

autosubmit/config/schema/autosubmit_schema.json
```diff
@@ -0,0 +1,38 @@
...
+  "$id": "https://earth.bsc.es/autosubmit/config-schema",
```
ntorqulu: Could be versioned, `/v1`, if more schema versions are expected

autosubmit/config/schema/autosubmit_schema.json
```diff
@@ -0,0 +1,109 @@
...
+                        "true",
+                        "false",
+                        "True",
+                        "False"
+                    ]
+                }
+            ],
+            "$description": "A boolean or a string representation of a boolean"
+        },
+        "customConfigSpec": {
```
ntorqulu: this one was tricky. From the documentation:
*It has two different syntaxes:

Simple a list of paths to the model or project yaml files. This can be a file or a folder. If it is a folder, all the files inside will be loaded in a non-recursive way.

Advanced a dictionary with two keys: PRE and POST. The PRE key is used to define the files that will be loaded before the $EXPID/CONF ones. The POST key is used to define user configuration.*

ntorqulu:
Used `experiment.schema.json` (https://gitlab.earth.bsc.es/digital-twins/de_340-3/workflow/-/blob/main/tests/schemas/experiment.schema.json) as reference for the EXPERIMENT section. A few inconsistencies in the DestinE schema, ask if they needed to be corrected or doesn't matter.
- `DATELIST` defined with `type: integer`. Should be `type: string; it's a space-separated string like "20260703 20260803", not a number
- `MEMBERS` not in `required`. Should be in `required`.
- `CHUNKSIZE` is not required in configcommon, whereas DestinE schema treats it as a required parameter.

**Questions**
- I'll omit validation of parameter `TWO_STEP_START` as it is deprecated from AS3
- `CHUNKSIZE`, `NUMCHUNKS` and `CHUNKINI` should be string or int? Accept both and normalize after? Currently users can enter both and internally it is normalized to an int.

**Errors in the documentation (FIXME)**
- `GIT.PROJECT_TAG`: Defined in the parameters, not defined anywhere in the code.
- In the expdef configuration
```
experiment:
    # Supply the list of start dates. Available formats: YYYYMMDD YYYYMMDDhh YYYYMMDDhhmm
    # Also you can use an abbreviated syntax for multiple dates with common parts:
    # 200001[01 15] <=> 20000101 20000115
    # DATELIST: 19600101 19650101 19700101
    # DATELIST: 1960[0101 0201 0301]
```

ntorqulu:
**How to use the auto-complete feature for AS config files:**
There are two possible ways to set up the auto-complete feature in VS Code using the schemas defined in this pr. The first one is easier to set up but will require publishing the schemas to the JSON Schema Store (https://www.schemastore.org/), which will be done only as the last step. So currently, only the second option is available. Listing both for future documentation.

### **Option 1: Publishing schemas into JSON Schema Store**
Once all the schemas have been correctly defined (expdef_schema.json, autosubmit_schema.json, jobs_schema.json, platforms_schema.json, minimal_schema.json) we'll publish them into JSON Schema Store. This way, they'll be accessible just with an url.
After this, for the user's point of view they'll just have to follow this steps:
1. Install YAML Red Hat extension (https://www.ibm.com/links?url=https%3A%2F%2Fmarketplace.visualstudio.com%2Fitems%3FitemName%3Dredhat.vscode-yaml) in their VS Code app.
2. For each yaml configuration file where they need the auto-complete feature, add the following tag at the top:
`yaml-language-server: $schema="url of the schema from the JSON Schema Store"`
3. We'll put in the documentation the url's of those schemas for easy access, or we can insert them by default commented (or not) when the files are automatically generated.
4. Take into account that each config file will need its own url, as we'll define diff JSON Schemas
image

### **Option 2: Configure schema validation via VS Code Settings**
This option requires more steps from the user but does not need the schemas published into JSON Schema Store. Useful for testing, we can give both options to the user and let them choose.
1. Install YAML Red Hat extension (https://www.ibm.com/links?url=https%3A%2F%2Fmarketplace.visualstudio.com%2Fitems%3FitemName%3Dredhat.vscode-yaml) in their VS Code app.
2. Open the `yaml.schemas` property in `settings.json`: Click the gear wheel in the bottom left of the VS Code editor and click Settings
image
3. Search for `Yaml: Schemas` and press the `Edit in settings.json` label to open `settings.json`
image
4. Add Schema File Paths as Properties: Add all the schemas needed for validation under the `yaml.schemas`property as shown
image
Where `/home/ntorquet/Documents/autosubmit/autosubmit/autosubmit/config/schema/` needs to be replaced for the path to the schema.
You can add multiple file patters related to a schema. Currently only using `name-of-the-config-file_*.yml`.

ntorqulu:
Closing this pr as @pablogoitia will continue it.
Currently implemented the json schema for `autosubmit.yml` and `expdef.yml` if you want to look it for reference.
Missing wrapper section specifications in `autosubmit.yml`, as it was a bit complicated to define.
Also included a brief explanation on how to use the json schema for autocompletion in VSCode, tested.

ntorqulu:
pd. @pablogoitia in issue https://github.com/BSC-ES/autosubmit/issues/871 you'll find the original discussion about the necessity of an AS validator and diff libraries that could be used very well explained by @kinow.

PR #3103, opened 2026-07-07
Postgres connection DB pool

### LINKED ISSUE #2973: Connection pool not used in Postgres
I remember that when we were working on https://github.com/BSC-ES/autosubmit/issues/2627 in the last coding sprint, @dbeltrankyl noticed some latency during some tests.

While checking the code, I noticed that we are not using any Connection Pool strategy while creating connections to the Postgres DB. Instead, every time we need data, we are creating a new connection instead of reusing existing ones from a pool.

This issue is not noticed when we use SQLite, where it is preferred not to use it to avoid open persistent file descriptors. An issue that we already had in the API.

### Task

We have to update every `create_engine` call from SQLAlchemy that uses the `DATABASE_CONN_URL` (meaning that is using Postgres) to use a Pool Strategy and be persisted for every call (Singleton pattern).

LuiggiTenorioK:
Closes #2973 

This PR changed the `create_engine` function from the session module to handle a persistent instance of the engine used for Postgres to enforce the usage of the Pool strategy and reutilize the connections. In the case of SQLite, this should behave the same.

Additionally, the function to create the experiment status row has been updated to do a native upsert.

**Check List**

- [x] Documentation updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,18 @@ This release also includes several bug fixes and enhancements to improve the ove
 
 - **Database Method Enhancements**: Support for both PostgreSQL and SQLite with optimized save/load mechanisms. Improved wrapper data storage and load.
 - **Testing Efforts**: Tests have been added for every rework to enhance reliability, with a particular effort on integration tests.
+
+- Enforce connection pool usage when using PostgreSQL as database backend #2973
+
+
+### 4.1.18: Unreleased
+
+**Bug fixes:**
+
+- Fix timeout guard is silently disabled for login/local jobs #3081
+
+**Enhancements:**
+
 - Auto-detect git default branch when `-b` flag not specified #3101
 - Removed `files` arguments from autosubmit sub-commands, and moved code to upgrade scripts out of `autosubmit.py` #2711
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,18 +33,7 @@ This release also includes several bug fixes and enhancements to improve the ove
 
 - **Database Method Enhancements**: Support for both PostgreSQL and SQLite with optimized save/load mechanisms. Improved wrapper data storage and load.
 - **Testing Efforts**: Tests have been added for every rework to enhance reliability, with a particular effort on integration tests.
-
 - Enforce connection pool usage when using PostgreSQL as database backend #2973
-
-
-### 4.1.18: Unreleased
-
-**Bug fixes:**
-
-- Fix timeout guard is silently disabled for login/local jobs #3081
-
-**Enhancements:**
-
 - Auto-detect git default branch when `-b` flag not specified #3101
 - Removed `files` arguments from autosubmit sub-commands, and moved code to upgrade scripts out of `autosubmit.py` #2711
 
--- a/autosubmit/database/db_manager.py
+++ b/autosubmit/database/db_manager.py
@@ -28,7 +28,6 @@ from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.database import session
 from autosubmit.database.tables import TableRegistry, GENERALTABLES, Table
 
-
 class DbManager:
     """A database manager using SQLAlchemy.
 
@@ -36,24 +35,17 @@ class DbManager:
     as Postgres, Mongo, MySQL, etc.
     """
 
-    def __init__(self, connection_url: str, schema: str | None = None, historical: bool | None = False) -> None:
+    def __init__(self, db_path: str, schema: str | None = None, historical: bool | None = False) -> None:
         self.engine = None
         self.engine_historical = None
         if BasicConfig.DATABASE_BACKEND == "sqlite":
             if historical:
-                self.engine_historical = session.create_engine(connection_url)
-                if self.engine_historical.url.database and not Path(self.engine_historical.url.database).exists():
-                    Path(self.engine_historical.url.database).touch()
-                    Path(self.engine_historical.url.database).chmod(0o775)
+                self.engine_historical = session.get_engine(db_path)
             else:
-                self.engine = session.create_engine(connection_url)
-                # make file
-                if self.engine.url.database and not Path(self.engine.url.database).exists():
-                    Path(self.engine.url.database).touch()
-                    Path(self.engine.url.database).chmod(0o775)
+                self.engine = session.get_engine(db_path)
         else:
             # Postgres is unified
-            self.engine: Engine = session.create_engine(connection_url)
+            self.engine: Engine = session.get_engine(db_path)
             self.engine_historical = self.engine
 
         self.schema = schema if BasicConfig.DATABASE_BACKEND != "sqlite" else None
--- a/autosubmit/database/session.py
+++ b/autosubmit/database/session.py
@@ -15,20 +15,74 @@
 # You should have received a copy of the GNU General Public License
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
+import threading
+from pathlib import Path
+from typing import Union
+
+from autosubmit.config.basicconfig import BasicConfig
 from sqlalchemy import Engine, NullPool, create_engine as sqlalchemy_create_engine
 
+__all__ = ["_resolve_engine", "get_engine"]
 
-def create_engine(connection_url: str) -> Engine:
-    """Create SQLAlchemy Core engine.
+
+def _resolve_engine(connection_url: str) -> Engine:
+    """Create SQLAlchemy Core engine and resolves the connection pool class based on the backend.
 
     :param connection_url: A SQLAlchemy connection URL.
     """
     if not connection_url:
-        raise ValueError(f'Invalid SQLAlchemy connection URL: {connection_url}')
+        raise ValueError(f"Invalid SQLAlchemy connection URL: {connection_url}")
 
     is_sqlite = connection_url.startswith("sqlite")
     pool_class = NullPool if is_sqlite else None
     return sqlalchemy_create_engine(connection_url, poolclass=pool_class)
 
 
-__all__ = ["create_engine"]
+class PostgreSQLEngineSingleton:
+    """Singleton class to manage a single instance of the PostgreSQL engine."""
+
+    _instance: Engine = None
+    _lock: threading.Lock = threading.Lock()
+
+    @classmethod
+    def get_instance(cls) -> Engine:
+        """Get the singleton instance of the PostgreSQL engine."""
+        with cls._lock:
+            if cls._instance is None:
+                connection_url = BasicConfig.DATABASE_CONN_URL
+                cls._instance = _resolve_engine(connection_url)
+        return cls._instance
+
+
+def get_engine(db_path: Union[str, Path]) -> Engine:
+    """Get SQLAlchemy Core engine.
+    This will resolve which backend to use based on the AS configuration.
+
+    In case the backend is PostgreSQL, the connection URL will be read from the environment variable,
+    and the engine will be reutilized from the global variable to use the same connection pool.
+
+    In case the backend is SQLite, a new engine will be created for each call based on the provided database path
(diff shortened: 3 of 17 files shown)


autosubmit/database/session.py
```diff
@@ -15,20 +15,68 @@
...
+        connection_url = f"sqlite:///{db_path}"
+        return _resolve_engine(connection_url)
+    elif db_backend == "postgres":
+        # Get from singleton engine
+        return PostgreSQLEngineSingleton.get_instance()
+    else:
+        raise ValueError(f"Unsupported database backend: {db_backend}")
+
+
+__all__ = ["_resolve_engine", "get_engine"]
```
kinow: Later we can choose the best place for __all__. I used __all__ at the bottom in some files because I was used to that from a previous project. But I think the convention is to have it near the top of the file (all dunders are supposed to be found there -- https://peps.python.org/pep-0008/#module-level-dunder-names)

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -181,14 +186,29 @@ def get_experiment_status_row_by_exp_id(self, exp_id: int) -> Optional[Models.Ex
         return Models.ExperimentStatusRow(*row)
 
     def create_exp_status(self, exp_id: int, expid: str, status: str) -> int:
+        """
```
kinow: This line is the first line of the pydoc, and is blank. Most tools work this way, but some may fail, which is why normally it's

```
"""Short doc.

If this, that.

That causes this.

:return etc...
"""

But no biggie.

File: autosubmit/database/session.py [hunk 3]
```diff
@@ -15,20 +15,68 @@
...
-__all__ = ["create_engine"]
+class PostgreSQLEngineSingleton:
+    """Singleton class to manage a single instance of the PostgreSQL engine."""
+
+    _instance: Engine = None
+
+    @classmethod
+    def get_instance(cls) -> Engine:
+        """Get the singleton instance of the PostgreSQL engine."""
+        if cls._instance is None:
```
kinow: Not sure if we will stay away from threads or not (threads may become more useful later after the gil is optional, and if the log process is moved back from its own process -- due to that 2fa). We have thread in one or two places of the code, but I think those don't use the DB session. If we have threads again in the log recovery, calling this get_instance could return different instances. I guess we don't have to worry about that now, but just to remember about it later..... or we could use `functools.cache` from stdlib and won't require any threading/locking/etc.

kinow (commented):
A few comments but nothing too important. Sorry for the delay reviewing this. The code is much simpler after the change, and now we gonna have a DB pool.

@LuiggiTenorioK shouldn't we allow users to configure the pool size, and settings (timeout/recycle/etcetc)? If I understand correctly, here we are using the sqlalchemy default values, but if we need to modify this we'd need a new AS release, I think?

Thanks!

kinow:
There are also two parts of the new/modified code not tested. Normally that's OK, but given this is DB-related, and something new that may fail, I think it'd be worth adding those if possible :+1:

LuiggiTenorioK:
@kinow now the singleton class is thread-safe, and the `__all__` was moved to the top. The docstring might be good to standardize in another PR; however, I don't feel it is an issue for Autosubmit really.

About the last line that needs to be tested, it is an unreachable line because we have 2 different DB managers (sqlalchemy and non-sqlalchemy). I think it should be solved in another PR where we stick to just the sqlalchemy one and remove the previous one. I will write an issue about that later.

dbeltrankyl:
>  it is an unreachable line because we have 2 different DB managers (sqlalchemy and non-sqlalchemy)

I have one point regarding that [here](https://github.com/BSC-ES/autosubmit/issues/3109)

kinow (commented):
> @kinow now the singleton class is thread-safe, and the __all__ was moved to the top. The docstring might be good to standardize in another PR; however, I don't feel it is an issue for Autosubmit really.

👍 

>About the last line that needs to be tested, it is an unreachable line because we have 2 different DB managers (sqlalchemy and non-sqlalchemy). I think it should be solved in another PR where we stick to just the sqlalchemy one and remove the previous one. I will write an issue about that later.

I think the issue is that it's not testing sqlite + sqlalchemy. We will drop sqlite, but we want to keep sqlite + sqlalchemy, so I think adding the test now would help in the future.

Maybe this will cover that part.

```diff
diff --git a/autosubmit/history/database_managers/experiment_status_db_manager.py b/autosubmit/history/database_managers/experiment_status_db_manager.py
index 6401f72d4..98deced78 100644
--- a/autosubmit/history/database_managers/experiment_status_db_manager.py
+++ b/autosubmit/history/database_managers/experiment_status_db_manager.py
@@ -159,7 +159,7 @@ class SqlAlchemyExperimentStatusDbManager:
     def create_experiment_status_as_running(self, experiment):
         self.create_exp_status(experiment.id, experiment.name, Models.RunningStatus.RUNNING)
 
-    def get_experiment_status_row_by_expid(self, expid: str) -> Optional[Models.ExperimentRow]:
+    def get_experiment_status_row_by_expid(self, expid: str) -> Optional[Models.ExperimentStatusRow]:
         experiment_row = self.get_experiment_row_by_expid(expid)
         return self.get_experiment_status_row_by_exp_id(experiment_row.id)
 
diff --git a/test/integration/conftest.py b/test/integration/conftest.py
index 6739dd2fe..63a508e78 100644
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -468,6 +468,12 @@ def postgres_server(request: 'FixtureRequest') -> Generator[Optional[PostgresCon
 
             yield container
 
+
+@pytest.fixture(params=[False, True])
+def use_sqlalchemy(request):
+    return request.param
+
+
 @pytest.fixture(params=['postgres', 'sqlite'])
 def as_db(request: 'FixtureRequest', autosubmit: Autosubmit, tmp_path: 'LocalPath', postgres_server: 'DockerContainer',
           autosubmit_exp):
diff --git a/test/integration/history/database_managers/test_experiment_status_db_manager.py b/test/integration/history/database_managers/test_experiment_status_db_manager.py
index 2b0661c12..09bdee336 100644
--- a/test/integration/history/database_managers/test_experiment_status_db_manager.py
+++ b/test/integration/history/database_managers/test_experiment_status_db_manager.py
@@ -44,29 +44,41 @@ def test_create_experiment_status_db_manager_invalid_value():
 
 @pytest.mark.docker
 @pytest.mark.postgres
-def test_experiment_status_db_manager(tmp_path: 'LocalPath', as_db: str, get_next_expid):
+def test_experiment_status_db_manager(tmp_path: 'LocalPath', as_db: str, use_sqlalchemy: bool, get_next_expid):
+    if as_db == "postgres" and not use_sqlalchemy:
+        pytest.skip("Postgres only supports SQLAlchemy")
+
     expid = get_next_expid()
     options = {"expid": expid}
     tmp_test_dir = tmp_path / "test_status"
     tmp_test_dir.mkdir()
 
-    clazz = SqlAlchemyExperimentStatusDbManager
-
-    is_sqlalchemy = as_db == "sqlite"
-    if is_sqlalchemy:
-        clazz = ExperimentStatusDbManager
-
+    if as_db == "sqlite":
         options["db_dir_path"] = tmp_test_dir
         options["local_root_dir_path"] = tmp_test_dir
         options["main_db_name"] = "tests.db"
 
+    if as_db == "sqlite" and use_sqlalchemy:
+        database_manager = SqlAlchemyExperimentStatusDbManager()
+        clazz = SqlAlchemyExperimentStatusDbManager
+    else:
+        database_manager = create_experiment_status_db_manager(as_db, **options)
+        clazz = (
+            ExperimentStatusDbManager
+            if as_db == "sqlite"
+            else SqlAlchemyExperimentStatusDbManager
+        )
+
     # Assert type of database manager
-    database_manager = create_experiment_status_db_manager(as_db, **options)  # type: clazz
     assert isinstance(database_manager, clazz)
 
-    # Test initialization of the table (is possible is created by some previous test)
-    if is_sqlalchemy:
-        assert Path(cast(ExperimentStatusDbManager, database_manager)._as_times_file_path).exists()
+    if as_db == "sqlite" and not use_sqlalchemy:
+        assert Path(
+            cast(ExperimentStatusDbManager, database_manager)._as_times_file_path
+        ).exists()
+    elif as_db == "sqlite" and use_sqlalchemy:
+        inspector = inspect(database_manager.engine)
+        assert inspector.has_table(ExperimentStatusTable.name)
     else:
         inspector = inspect(database_manager.engine)
         assert inspector.has_table(ExperimentStatusTable.name, schema="public")
```

kinow:
Maybe isntead of the fixture, just a pytest.parametrize with True/False. That might actually be simpler, I guess. But this way we should be testing sqlite + sqlachemy, I think.

LuiggiTenorioK:
> I think the issue is that it's not testing sqlite + sqlalchemy. We will drop sqlite, but we want to keep sqlite + sqlalchemy, so I think adding the test now would help in the future.

Agree with that. I was trying to mean that above, but I wanted to avoid making more changes unrelated to this PR that will be solved in the future.

> Maybe this will cover that part.

It does! For now, I think this should be enough until #3114 is done.

kinow:
Thanks @LuiggiTenorioK ! Just waiting for CICD to complete to hit the approval button!

kinow (approved):
Thank you!

autosubmit/database/db_manager.py
```diff
@@ -28,32 +28,24 @@
...
-                if self.engine_historical.url.database and not Path(self.engine_historical.url.database).exists():
-                    Path(self.engine_historical.url.database).touch()
-                    Path(self.engine_historical.url.database).chmod(0o775)
+                self.engine_historical = session.get_engine(db_path)
             else:
-                self.engine = session.create_engine(connection_url)
-                # make file
-                if self.engine.url.database and not Path(self.engine.url.database).exists():
-                    Path(self.engine.url.database).touch()
-                    Path(self.engine.url.database).chmod(0o775)
```
LuiggiTenorioK, line 53: @dbeltrankyl is there any reason why sometimes the chmod is 775 or 770 (like in the `get_connection_url` function)?

I am generalizing everything to 770, but maybe 775 is better.

dbeltrankyl: Nope, no reason

I'm good with one or the other... Everyone using autosubmit should have group permissions I believe

LuiggiTenorioK:
@kinow this PR is now rebased, and tests should pass. I think the only relevant change that came from the rebase is the additional chmod in the db file check.

Other than that, this should be ready to merge once you re-approve it.

CHANGELOG.md
```diff
@@ -33,6 +33,18 @@ This release also includes several bug fixes and enhancements to improve the ove
...
+- Enforce connection pool usage when using PostgreSQL as database backend #2973
+
+
+### 4.1.18: Unreleased
+
+**Bug fixes:**
+
+- Fix timeout guard is silently disabled for login/local jobs #3081
+
+**Enhancements:**
```
kinow: Can you move these to 4.2.0? I fixed the changelog yesterday removing the 4.1.18 and replacing by 4.2.0.

autosubmit/history/database_managers/experiment_status_db_manager.py
kinow: Small one, but if you update the changelog maybe you can move this line up too @LuiggiTenorioK :grimacing:

kinow (commented):
Looking good @LuiggiTenorioK ! Left a comment about the changelog entry and one pydoc line 😬 Thanks Luiggi!

CHANGELOG.md
LuiggiTenorioK: All right, this line is not from this PR. I will remove it since it is already in 4.2.0 section.

kinow (approved):
💣 💥 thank you! Just pending the build to pass, but code looks grand.

kinow:
I think CICD needs a kick or two...

PR #3107, opened 2026-07-08
Fix %CHUNK% always resolving to 1 for SYNCHRONIZE: date jobs

### LINKED ISSUE #3087: Possible bug when using SYNCHRONIZE: date
### Description

Hi, I'm trying to run a workflow which have a determined list of dates which are used for the first part of the workflow. Later on, to run other part of the same workflow I would like to use chunks but I do not want them be related with the datelist so for example, I would like to launch a workflow for 10 years and set the amount of chunks as 5 so the idea should be to create for the first part 10 jobs and for the chunk part just 5. I'll manage to do that using ` SYNCHRONIZE: date`. However what I found is that even the name of the job contains the different chunks (e.g., aaqj_1_ANEMOI_LOAD, aaqj_2_ANEMOI_LOAD...) when I print the %CHUNK% variable for each job is always 1.

I don't know if it's a bug or a mistake with my config files. 

* Autosubmit version: 4.1.16
* Machine/VM/environment name (if applicable): NORD4
* Experiment ID (if applicable): aaqj
* Experiment tasks & log path if applicable): aaqj_X_ANEMOI_LOAD

### Reproducible Example

Running the experiment as it is right now, should fail at the finalise part but the issue is in the load part. There is an echo command in the script to identify that the chunks are not being well retrieved

### Expected Behaviour

The expected behaviour is that the number that appears in the name of the job aaqj_X_ANEMOI_LOAD corresponds to the chunk number instead of being always 1

rajeshsuraj:
Summary: calendar_chunk guarded CHUNK binding behind self.date is not None, so synchronized jobs (which have date=None) never overwrote the hardcoded CHUNK=1 default. Now binds CHUNK from self.chunk independently of date presence.

Closes #3087

**Check List**

- [ ] Documentation updated.


kinow (commented):
Change looks good. It's only missing a test? For issues like this I normally start by writing a test that fails. Then I write the fix, and confirm the test fixes. This way i) we know the fix is fixing, and ii) the fix + test will prevent regressions.

rajeshsuraj:
> Change looks good. It's only missing a test? For issues like this I normally start by writing a test that fails. Then I write the fix, and confirm the test fixes. This way i) we know the fix is fixing, and ii) the fix + test will prevent regressions.

I think this is better than what i used do , will follow this approach in the future. will add tests as i am still testing if the bug was fixed and also if something breaks as the hard coded value is now dynamic.

kinow:
Thank you Suraj!!

kinow (approved):
Thank you, @rajeshsuraj !

PR #3125, opened 2026-07-17
Improve docs `Platform variables` section

### LINKED ISSUE #3118: [documentation] HPC and CURRENT prefix variables description in documentation
Hi, 

I was trying to understand the behaviour of the variables with prefix `HPC` in the ClimateDT workflow. While checking the Autosubmit documentation, in particular the [Platform Variables section](https://autosubmit.readthedocs.io/en/latest/userguide/variables.html#platform-variables), I could not find an explanation on how the variables defined under the `PLATFORMS` section are translated to `HPC<variable>` in the templates. 

I found another issues mentioning that the keyword `CURRENT` follows a similar mechanism (https://github.com/BSC-ES/autosubmit/issues/2239). 

It would be useful to document this behaviour, as the current documentation only lists a predefined set of `HPC*` and `CURRENT*` variables that are defined, but not the generic mechanism (unless I'm overlooking it).

Thank you!

ntorqulu:
Closes #3118 

**Things done**

- Modified `Platforms variables`
- Deleted deprecated table of platform variables
- Explained the difference between raw keys, `HPC`-prefixed keys, and `CURRENT_`-prefixed keys.

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -1885,7 +1885,7 @@ class Job(object):
             self.shape = parameters.get("CURRENT_SHAPE", "")
             self.processors_per_node = parameters.get("CURRENT_PROCESSORS_PER_NODE", "1")
             self.nodes = parameters.get("CURRENT_NODES", "")
-            self.exclusive = parameters.get("CURRENT_EXCLUSIVE", False)
+            self.exclusive = parameters.get("CURRENT_EXCLUSIVITY", False)
             self.threads = parameters.get("CURRENT_THREADS", "1")
             self.tasks = parameters.get("CURRENT_TASKS", "0")
             self.reservation = parameters.get("CURRENT_RESERVATION", "")
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -1885,7 +1885,8 @@ class Job(object):
             self.shape = parameters.get("CURRENT_SHAPE", "")
             self.processors_per_node = parameters.get("CURRENT_PROCESSORS_PER_NODE", "1")
             self.nodes = parameters.get("CURRENT_NODES", "")
-            self.exclusive = parameters.get("CURRENT_EXCLUSIVITY", False)
+            # FIXME: Should be ``CURRENT_EXCLUSIVITY`` instead of ``CURRENT_EXCLUSIVE`` to match the platform parameter?
+            self.exclusive = parameters.get("CURRENT_EXCLUSIVE", False)
             self.threads = parameters.get("CURRENT_THREADS", "1")
             self.tasks = parameters.get("CURRENT_TASKS", "0")
             self.reservation = parameters.get("CURRENT_RESERVATION", "")
--- a/docs/source/userguide/variables.rst
+++ b/docs/source/userguide/variables.rst
@@ -160,6 +160,7 @@ In the output of :ref:`report <report>` you will find them with the pattern
     PLATFORMS.MARENOSTRUM5.SCRATCH_DIR=/gpfs/scratch
     PLATFORMS.SLURM.TYPE=slurm
     ...
+
 ---
 
 **2. HPC prefixed keys (default platform, global)**
--- a/docs/source/userguide/variables.rst
+++ b/docs/source/userguide/variables.rst
@@ -77,33 +77,125 @@ job, you can use ``CUSTOM_DIRECTIVES=--gpus-per-node=10``.
 Platform variables
 ==================
 
-These variables are relative to the platforms defined in each
-job configuration. The table below shows the complete set of variables
-available in the current platform. These variables appear in the
-output of the :ref:`report <report>` command with the pattern
-``JOBS.${JOB_ID}.${PLATFORM_VARIABLE}=${VALUE}``. They can be used in
-templates with ``%PLATFORM_VARIABLE%``.
+Platform variables come from the ``PLATFORMS``section of the configuration
+files. Each ``JOBS`` entry can reference any of these platforms via its
+``PLATFORM`` key. If no platform is specified, the job uses the experiment's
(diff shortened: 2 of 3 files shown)


ntorqulu:
I've not included a changelog entry for this issue as it's just a rephrasing with an example, nothing important.
@VindeeR, if you think it's better to include it let me know!

kinow:
I can review this one pretty quickly (I'm doing a quick check on PRs before reviewing Suraj's PR).

kinow (approved):
Looks great to me, @ntorqulu !

I think maybe we would need to clarify this part,

"""
    Certain variables (e.g. ``CURRENT_RESERVATION``,
    ``CURRENT_EXCLUSIVITY``) are only available for certain
    platforms (e.g. MareNostrum).
"""

I read it as if we had something like this in AS

```
if platform_name == 'MN5':
   create_those_variables...
```

But in reality I think users can set that to any platform. Looking at the code, I think exclusivity/current_exclusivity is never passed to Slurm, or used directly. It's up to users to read that value and add it to their templates? (In reality, it seems to me EXCLUSIVITY is not actually used at all? And instead --exclusive is used for Slurm via platforms/jobs?).

Leaving approved as this can be refined later and is a minor detail (and I may be wrong too 😆 ).

ntorqulu:
Yep, I was not sure about this part when I read it before changing the section, so I decided to leave it as it is (just in case) at the end:
```
    The variables ``_USER``, ``_PROJ`` and ``_BUDG``
    have no value on the LOCAL platform.

    Certain variables (e.g. ``_RESERVATION``,
    ``_EXCLUSIVITY``) are only available for certain
    platforms (e.g. MareNostrum).
```
I can investigate more before merging it and check if I find any reference in the code. I'll let you know!
If not, I think is better delete it as it can confuse users (and us!).

kinow:
> Yep, I was not sure about this part when I read it before changing the section, so I decided to leave it as it is (just in case) at the end:
> 
> ```
>     The variables ``_USER``, ``_PROJ`` and ``_BUDG``
>     have no value on the LOCAL platform.
> 
>     Certain variables (e.g. ``_RESERVATION``,
>     ``_EXCLUSIVITY``) are only available for certain
>     platforms (e.g. MareNostrum).
> ```
> 
> I can investigate more before merging it and check if I find any reference in the code. I'll let you know! If not, I think is better delete it as it can confuse users (and us!).

Sounds good to me! Thanks Núria!

docs/source/userguide/variables.rst
```diff
@@ -112,23 +205,109 @@ also available.
...
+      **automatically** filled by Autosubmit.
+    
+    * ``CURRENT_USER``, ``CURRENT_PROJ``, ``CURRENT_QUEUE``,
+    ``CURRENT_PARTITION`` and ``CURRENT_TEMP_DIR`` are only
+    filled if the corresponding YAML key (``USER``,
+    ``PROJECT``, ``QUEUE``, ``PARTITION``, ``TEMP_DIR``) is
+    defined in the platform configuration. On the LOCAL
+    platform they remain empty.
+
+    * ``CURRENT_BUDG``, ``CURRENT_RESERVATION`` and
```
ntorqulu, line 303: `CURRENT_BUDG`: Only reference found in `ec_cca_header.py`
`#PBS -l EC_billing_account=%CURRENT_BUDG%`

ntorqulu, line 303: `CURRENT_RESERVATION`: found in slurm and pbs headers
```
def get_reservation_directive(self, job, parameters, het=-1):
    if check_directive('RESERVATION', parameters):
        return f"SBATCH --reservation={parameters['RESERVATION']}"
```
```
def get_reservation_directive(self, job, parameters, het=-1):
    if parameters['RESERVATION'] != '':
        return f"PBS -W x={parameters['RESERVATION']}"
```

docs/source/userguide/variables.rst
```diff
@@ -112,23 +205,109 @@ also available.
...
+    ``CURRENT_EXCLUSIVITY`` are only populated if the
```
ntorqulu: `CURRENT_EXCLUSIVITY`: Only found in slurm header
```
def get_exclusive_directive(self, job, parameters, het=-1):
    if parameters is not None and 'EXCLUSIVE' in parameters and bool(parameters['EXCLUSIVE']):
        return "SBATCH --exclusive"
```

I think there's an error here (not 100% sure).
`slurm_header.py` reads `parameters['EXCLUSIVE']` with a different spelling (EXCLUSIVE vs EXCLUSIVITY).
The value comes from `update_platform_associated_parameters`, that reads `CURRENT_EXCLUSIVE`.
But `update_current_parameters` sets `CURRENT_EXCLUSIVITY` (with ivity).
So at the end in parameters there are two values that mean the same.

docs/source/userguide/variables.rst
kinow: I think there is a ticket about bringing the budget variable back... I think it was removed some time ago. But we just need to remember and update here if/when that issue is fixed

autosubmit/job/job.py
```diff
@@ -1885,7 +1885,8 @@ def update_platform_associated_parameters(self, as_conf: AutosubmitConfig, param
             self.shape = parameters.get("CURRENT_SHAPE", "")
             self.processors_per_node = parameters.get("CURRENT_PROCESSORS_PER_NODE", "1")
             self.nodes = parameters.get("CURRENT_NODES", "")
-            self.exclusive = parameters.get("CURRENT_EXCLUSIVITY", False)
+            # FIXME: Should be ``CURRENT_EXCLUSIVITY`` instead of ``CURRENT_EXCLUSIVE`` to match the platform parameter?
```
kinow, line 1888: 👍 thanks!

kinow (approved):

PR #3135, opened 2026-07-24
[Bug] - fixing ruff versioning

VindeeR:
Temporarily setting the ruff to be at a specific target version in order to avoid targeting unchanged files

**Check List**

- [ ] Documentation updated.


kinow (commented):
So the fix is to downgrade it? Isn't there a fix with the latest version instead?

VindeeR:
> So the fix is to downgrade it? Isn't there a fix with the latest version instead?

As I explained, in order to avoid getting the error all the time, and having people ignoring the errors, my idea will be use the previous version, until we manage to get the fix working, I'm still making some tests on the local to get it working, so we wouldnt risk letting it pass without check

.github/workflows/lint.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install Ruff
-        run: pip install ruff
+        run: pip install ruff==0.15.22
       - name: Run Ruff
         run: ruff check $(git diff --name-only --cached -- '*.py')
```
kinow: 0.15.22 was the last release of ruff in the 0.15.x line, https://github.com/astral-sh/ruff/blob/main/changelogs/0.15.x.md

@VindeeR I believe what's happening is described in their changelog - https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md

> Ruff now enables a much larger set of rules by default (413, up from 59). See the blog post for more details and the new [Default Rules](https://docs.astral.sh/ruff/default-rules/) page for a full listing of the enabled rules.

And

Image

So they simply added more rules, and promoted rules that were not being enforced.

I did this modification in one of my PRs:

```diff
diff --git a/.github/workflows/lint.yml b/.github/workflows/lint.yml
index 8cf8ace6c..78d1588b8 100644
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,26 +15,18 @@ jobs:
     runs-on: ubuntu-latest
     name: ruff run
     steps:
-      - uses: actions/checkout@v7
+      - name: Checkout code
+        uses: actions/checkout@v7
 
-      - uses: actions/setup-python@v7
+      - name: Set up Python
+        uses: actions/setup-python@v7
         with:
           python-version: ${{ matrix.python-version }}
 
-      - run: pip install ruff
-
-      - name: Get Python changed files
-        id: changed-py-files
-        uses: tj-actions/changed-files@v47
-        with:
-          files: |
-            *.py
-            **/*.py
-
+      - name: Install Ruff
+        run: pip install ruff
       - name: Run Ruff
-        if: steps.changed-py-files.outputs.any_changed == 'true'
-        run: |
-          ruff check ${{ steps.changed-py-files.outputs.all_changed_files }}
+        run: ruff check $(git diff --name-only --cached -- '*.py')
 
 
   mypy-changes-in-file:

```

We are already using the tj-actions elsewhere (mypy) so it makes sense to use it here too, I think?

Then, I compared what ruff is complaining about, and these are legit failures. I touched `db_common.py` and it was working, but now ruff warned me that I had to fix the imports that were out of order (which isort / pycodestyle would also have warned). I checked on PyCharm, and it indeed fixed imports.

So I think we shouldn't downgrade it. It's actually finding errors, and in the files modified. As it was intended. We probably are going to need to either ignore for a few days, until one of us has time to create a pull request. Maybe it'll be as simple as to create it automatically with ruff --fix and then review that the changes are good, that there's nothing wrong.

kinow: Or, we can also maybe update pyproject.toml to include a ruff section (if there's not one yet), where we ignore the newly added items. That way, the build would use the newest ruff, and pass. And we could slowly add back the recommended checks.

VindeeR: yeap this is partially what is happening, but not the issue I mentioned, at least not in my branches, for some reason the way we were sorting the files before, by finding the changed files, and thus run the validation just to the changed files, is broken on the new version. 

I'm trying to understand why locally I was able to fix but not on github, I believe it is related how the files are used for the ruff check command.

So the idea was to reverse only momentarily so we could still fix it, and people still could fix the appointed errors.

VindeeR:
Okay I finally found a way to go around the files issue we were having and I was able to implement this fix which allow us to you the latest version of ruff with the same behavior we had previously :raised_hands:

VindeeR:
I saw that the tj-actions/changed-files@v45 and prior had a security issue where it was trying to dump the secrets of the CI.
[changed-files](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised)

I saw that v47 and later versions (patches starting from v46/v47) contain fixes, but the credentials still need constant change and audit logs

VindeeR:
@kinow I think we might be having an issue with the port of the docker for the test containers

VindeeR:
@kinow the tests passed in here but I had to kick them 3 times for it to work, we probably will have to keep an eye on it to ensure this won't happen all the time

test/integration/conftest.py
```diff
@@ -127,9 +136,9 @@ def _get_next_expid() -> str:
 @pytest.fixture
 def autosubmit_exp(
         autosubmit: Autosubmit,
-        request: "FixtureRequest",
-        tmp_path: "LocalPath",
-        mocker: "MockerFixture",
+        request: FixtureRequest,
+        tmp_path: LocalPath,
```
kinow: Interesting! We don't need to use quotes for imports used with TYPE_CHECKING, apparently.

test/integration/conftest.py
```diff
@@ -16,17 +16,20 @@
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
 """Fixtures for integration tests."""
+from __future__ import annotations
```
kinow: Can we postpone adding this? i.e. disable it in ruff, and ignore if there are errors?

Image

Besides deferring the imports, it may also hide issues when we import the wrong file (example in the image above; you can try that without the __future__ annotations and then you get an error). And I personally think it's useful to know when you have something imported but that you are only using the type (IIRC, in TypeScript you can import the type alone, which would be useful in Python).

kinow (commented):
@VindeeR almost there! I only have a nitpick about future annotations... see my comment. The rest looks perfect. Thanks!!

test/integration/conftest.py
VindeeR: I can revert but Should I keep the `Optional`? Or can I keep using the new `Type | None`?
Should it be a problem in case use the new way?

I'll add the ignore rule `FA102 Missing from _ _future_ _ import annotations, but uses PEP 604 union` to the `.toml` file

test/integration/conftest.py
VindeeR: apparently this is due the __future__ annotation, so removing it I think we will have to put it back

test/integration/conftest.py
kinow: I think Type | None is OK. I use it in other projects, but in AS we were forced with have Optionals as we were locked in 3.9 for a while (climate dt vm).

VindeeR: I found this tread [bpo-46120](https://github.com/python/cpython/pull/30222#pullrequestreview-838738862) stating that using `Type | None` over the Optional
So I guess I know which to use :raised_hands:

test/integration/conftest.py
```diff
@@ -570,11 +578,11 @@ def copy_content_from_containers(request, log_name, path_to_docker=""):
     has_failures = request.session.testsfailed
     func_args = request.node.funcargs
     if has_failures and log_name in func_args and func_args[log_name]:
-        container_in_use: 'Container'
+        container_in_use: Container
```
kinow, line 581: Without future annotations, I think we need quotes here? I think the build is passing, which is strange. But not sure if this is being called in the current tests?

kinow: Well, the build is passing like this... let's merge it and if it breaks later we can fix it. Hitting the approve button! 💣 💥 

Thanks @VindeeR !

kinow (approved):

PR #3140, opened 2026-07-29
Fixing `DELAY_RETRY_TIME` to match with the documentation

### LINKED ISSUE #3138: DELAY_RETRY_TIME +N and *N produce wrong delays
### Description

Two bugs in the retry-delay computation at `autosubmit/job/job_list.py:2655-2661`.
Both confirmed as bugs by @kinow in #2992.

* Autosubmit version: master (HEAD `bb179c56`, `2992-documentation-improve-documentation-for-retries` branch of the audit)
* Machine/VM/environment name: N/A — bugs are in pure Python delay computation
* Experiment ID: N/A

**Bug 1 — `DELAY_RETRY_TIME: +N` strips the wrong character.**
`job_list.py:2656-2657` uses `int(delay_retry_time[:-1])`, which drops the *last* character instead of the `+` prefix. `"+11"` is parsed as `"+1"` → `1`. The intended slice is `[1:]` (as the `*N` branch correctly uses on line 2659).

**Bug 2 — `DELAY_RETRY_TIME: *N` multiplies by 11 each retry.**
The loop at `job_list.py:2660-2661` does `retry_delay += retry_delay * 10`, which is equivalent to `retry_delay *= 11`. So `*11` produces `11, 121, 1331, 14641, ...` — grows very fast. The intended behavior is unclear from the docs (three doc sources disagree), so this bug covers "the current behavior is almost certainly not what any reader would expect"; the exact intended formula should be decided as part of the fix.

### Expected Behaviour

* `+11`: reasonable interpretation is `11, 22, 33, 44` (linear growth by N).
* `*11`: intended formula needs decision — see docs discrepancy in #2992.
* Docs on both pages (`develop_a_project.rst`, `configure/index.rst`) should be updated to match whatever behavior is chosen. This will be covered in the docs consolidation under #2992.

rajeshsuraj:
Closes #3138 

**Check List**

- [ ] Documentation updated.


test/unit/test_job_list.py
```diff
@@ -686,3 +687,44 @@ def test_recover_last_data_on_old_schema(tmp_path, as_conf):
...
+        delays.append(round((job.delay_end - before).total_seconds()))
+    return delays
+
+
+@pytest.mark.parametrize(
+    "delay_retry_time, expected",
+    [
+        ("11", [11, 11, 11, 11]),
+        ("+11", [11, 22, 33, 44]),      # was [2, 3, 4, 5]
+        ("*11", [11, 110, 1100, 11000]),  # was [121, 1331, 14641, 161051]
```
VindeeR:
```suggestion
        ("*11", [11, 110, 1110, 11110]),  # was [121, 1331, 14641, 161051]
```

test/unit/test_job_list.py
```diff
@@ -686,3 +687,44 @@ def test_recover_last_data_on_old_schema(tmp_path, as_conf):
...
+        ("+5", [5, 10, 15, 20]),        # was ValueError from int("+")
+        ("*2", [2, 20, 200, 2000]),     # was [22, 242, 2662, 29282]
+    ],
+    ids=["constant", "linear", "factor_of_ten", "linear_single_digit", "factor_of_ten_single_digit"],
+)
+def test_delay_retry_time_matches_documented_sequence(mocker, delay_retry_time, expected):
+    """DELAY_RETRY_TIME must produce the sequences promised in the configuration reference."""
+    assert _retry_delays(mocker, delay_retry_time) == expected
+
```
VindeeR: EoF

autosubmit/job/job_list.py
```diff
@@ -3372,13 +3372,11 @@ def _update_failed_jobs(self, as_conf: AutosubmitConfig) -> bool:
...
-                        retry_delay = (job.fail_count * int(delay_retry_time[:-1]) + int(delay_retry_time[:-1]))
-                    elif "*" in delay_retry_time:
-                        retry_delay = int(delay_retry_time[1:])
-                        for retrial_amount in range(job.fail_count):
-                            retry_delay += retry_delay * 10
+                        delay_retry_time = str(job.retry_delay)
+                    if delay_retry_time.startswith("+"):
+                        retry_delay = int(delay_retry_time[1:]) * job.fail_count
+                    elif delay_retry_time.startswith("*"):
+                        retry_delay = int(delay_retry_time[1:]) * 10 ** (job.fail_count - 1)
```
VindeeR, line 3383: this math is wrong

VindeeR (changes requested):

autosubmit/job/job_list.py
rajeshsuraj: math aint mathing

docs/source/userguide/configure/develop_a_project.rst
```diff
@@ -131,7 +131,7 @@ Autosubmit configuration
         ##  Allows to put a delay between retries, of retries if a job fails. If not specified, it will be static
         # DELAY_RETRY_TIME:11
         # DELAY_RETRY_TIME:+11 # will wait 11,22,33,44...
-        # DELAY_RETRY_TIME:*11 # will wait 11,110,1110,11110...
+        # DELAY_RETRY_TIME:*11 # will wait 11,110,1100,11000...
```
VindeeR: Could you add here too the math behind this? it would be nice have documented not only the end result but the math behind it

test/unit/test_job_list.py
VindeeR, line 1: I'm not sure why you split the test into 2 function to me it makes more sense to have it all in only one function.

And I believe you could make a more robust test on `_update_failed_jobs` that would hit some of the missing lines like 3392 to 3394.

Although your changes don't impact there I believe this test could easily validate the whole function.

🙌 thanks for the hard work.

VindeeR (changes requested):
Almost there just a few quick suggestion

VindeeR:
Don't forget to update your branch too

test/unit/test_job_list.py
rajeshsuraj: I split the test because i needed a separate setup to build the failed jobs. after your comment i looked into it and found that building a fixture was the better approach and will be included in my next commit.

.gitignore
```diff
@@ -33,3 +33,5 @@ man/
 venv/
 .ipynb_checkpoints
 .atl
+:memory:
+.benchmarks/artifacts/performance-metrics-4.2.0.csv
```
kinow: We also have to fix the tests for these. We could add these to gitignore, but they shouldn' t be created in the first place. So maybe better to fix the root cause (or we will merge gitignore, and forge about it)

rajeshsuraj: should i remove it from gitignore ? i didnt want to accidentally push those files ?

VindeeR: it was added in other 2 PR to `.gitignore` hopefully we will be merging one of them soon until fixed

kinow: @VindeeR you fixed the :migrate: in a PR, no?

VindeeR: @rajeshsuraj you can remove this one as I just finished another PR fixing this and will shortly be merged, you'll just need to rabase later.
Thanks

.gitignore
VindeeR, line 1: once you rollback this file this PR is ready to be merged, Thanks for the hard work 🙌

VindeeR (changes requested):
If the CI finishes fine after the rollback of the changes in `.gitignore` you can merge it

.gitignore
rajeshsuraj: rolled back and rebased

VindeeR (changes requested):
you have a simple ruff issue you can merge after rebase and fixing that import order

VindeeR (approved):
it looks good to me thanks for the improvement 🙌

Issue #3139, opened 2026-07-29
PJM jobs not detected as COMPLETED


Lerriola:
Hello @VindeeR and @kinow ( cc @kellekai )

### Description

Hello the issue is that the jobs are being completed in the autosubmit VM but they're not being detected as such by Autosubmit amd neither by the API/GUI. Actually although the jobs generate the COMPLETED file, Autosubmit doesn't detect it and the logs are not closed either.

<img width="1416" height="435" alt="Image" src="https://github.com/user-attachments/assets/712d6fce-3981-4865-b50f-351b60e78293" />

- `FUGAKU PATH: /vol0004/ra000012/data/u12688/abfg/LOG_abfg`

* Autosubmit version: https://github.com/BSC-ES/autosubmit/tree/pjm-fix (using a python env)
* Machine/VM/environment name (if applicable): bscautosubmit04
* Experiment ID (if applicable): abfg
* Experiment tasks & log path if applicable): abfg_20120101_fc17_4_SIMULATION_COMPLETED and many more

### Reproducible Example

Creating an experiment with that pjm-fix branch and running. I think by creating a copy of abfg it should happen.

### Expected Behaviour

Jobs to be detected as completed so the simulation can go on an continue succesfully.

kellekai:
- **EXPID: abhq**
- **Folder on Fugaku: /vol0004/ra000012/data/a04454/abhq**


I think I might just have hit the issue in my experiment. I followed the steps described here: https://github.com/kellekai/bsc-ndse/issues/70#issuecomment-5116279780

I can see those files:

```
-rw-r--r-- 1 a04454 ra000012 0 Jul 31 12:52 /vol0004/ra000012/data/a04454/abhq/LOG_abhq/abhq_20120101_fc08_ENSEMBLE_INI_COMPLETED
-rw-r--r-- 1 a04454 ra000012 0 Jul 31 12:52 /vol0004/ra000012/data/a04454/abhq/LOG_abhq/abhq_20120101_fc04_ENSEMBLE_INI_COMPLETED
-rw-r--r-- 1 a04454 ra000012 0 Jul 31 12:52 /vol0004/ra000012/data/a04454/abhq/LOG_abhq/abhq_20120101_fc09_ENSEMBLE_INI_COMPLETED
[...]

PR #3144, opened 2026-07-31
[Bug] - Fix IO_SAFE_WAIT standarda value

VindeeR:
Related #3139

Fixing waiting time for files to be dealt with, this might be breaking the files handling of many jobs

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/test/unit/platforms/test_psplatform.py
+++ b/test/unit/platforms/test_psplatform.py
@@ -282,11 +282,21 @@ def _make_ps_job(name: str, status: Status, fail_count: int = 0):
     return job
 
 
-@pytest.mark.parametrize('stat_line,expected_final_status', [
-    ('COMPLETED', Status.COMPLETED),
-    ('FAILED', Status.FAILED),
-    ('', Status.COMPLETED),  # STAT not flushed, IO_SAFE_WAIT=0 → scheduler status wins
-])
+@pytest.mark.parametrize(
+    "stat_line,expected_final_status",
+    [
+        ("COMPLETED", Status.COMPLETED),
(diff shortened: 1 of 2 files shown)


VindeeR:
Before merging this PR I want to address as many of the `ruff + mypy` I can in this commit

dbeltrankyl (changes requested):
The test is failing because it was assuming an io_safe_wait of 0 seconds

Either patch io_safe_wait to 0 seconds in the test, or change the test so that the expected outcome is Status.RUNNING, and it should fix it, I believe ( or do both cases)

test/unit/platforms/test_psplatform.py
```diff
@@ -282,11 +282,21 @@ def _make_ps_job(name: str, status: Status, fail_count: int = 0):
...
-])
+@pytest.mark.parametrize(
+    "stat_line,expected_final_status",
+    [
+        ("COMPLETED", Status.COMPLETED),
+        ("FAILED", Status.FAILED),
+        (
+            "",
+            Status.COMPLETED,
+        ),  # STAT not flushed, IO_SAFE_WAIT=0 → scheduler status wins
```
dbeltrankyl: The comment should be IO_SAFE_WAIT=60, no?

dbeltrankyl (approved):
Just a question about the comment, thanks!

test/unit/platforms/test_psplatform.py
VindeeR: commit post lunch for got to change the comment thanks for the catch

kinow (approved):

PR #3162, opened 2026-08-05
Performance benchmark suite + metrics workflow

dbeltrankyl:
**Check List**

- [x] Documentation updated.

Added:

- Benchmark tests for create, run, recovery and setstatus (quick + full suites), with readable scenario IDs like 4m/2c/10s.
- .benchmarks/compare_results.py + thresholds.yml: compares runs against a baseline, flags regressions, and produces a markdown report + two heatmap plots ordered by speed.
- metrics workflow: runs as a required check on perf-benchmark-labeled PRs, with manual /metrics, /metrics_full and /metrics_promote commands. Results are posted as a comment with collapsible tables and artifact download links.
- Fix in the Profiler: MEMORY GROW was reported with the wrong unit when growth was small relative to final memory.
- Unit tests for the comparison logic.

Notes:

- Baselines are stored per CPU model (GitHub runners don't guarantee a fixed CPU), so several runs are needed to seed the baselines before comparisons kick in.
- The perf-benchmark label acts as an auth gate on its own: only the triage team can add it, so unlabeled PRs are a cheap no-op and can't run the benchmark.
- The manual /metrics* commands are also gated by an allowlist: only members of the BSC-ES/autosubmit team (or users listed as @username in MAINTAINERS.md on the default branch) can trigger them. That's why MAINTAINERS.md is added in this PR.
- The /metrics* commands won't work on this branch: issue_comment workflows run from the default branch, so the workflow changes only take effect once merged to master.
- You can see it in action on a demo PR here: https://github.com/dbeltrankyl/autosubmit-benchmarks/pull/11


dbeltrankyl:
Ruff fails, but because I deleted stuff

```
E902 No such file or directory (os error 2)
--> .benchmarks/benchmark_utils.py:1:1

E902 No such file or directory (os error 2)
--> .benchmarks/generate_markdown.py:1:1

E902 No such file or directory (os error 2)
--> .benchmarks/generate_plot.py:1:1

```

dbeltrankyl:
Note about the linked demo, the baselines on my fork are not updated, that's why you see some colors

Note: Apparently I can start the benchmark on this branch without merging into master :thinking: maybe it is because the metrics.yaml already existed  ( unfinished version from my 4.2.0 merge). Since there are no baselines it will be likely grey

dbeltrankyl:
I assigned as reviewer @kinow,  but everyone is invited to have a look

dbeltrankyl:
related to #2696 does not closes it because this only adds the mechanism to compare versions. Once this is merged, my plan is to generate a baseline for 4.1.17 and compare

dbeltrankyl:
rebased + resolved conflicts  ( it will trigger another benchmark because the label is set). As far as I can see, there are 4 CPUs in the pool: 2 AMD, 1 Intel, ~and 1 "linux"~ nvm. Looking at the JSON, it is the intel machine. 

I can manually dispatch the metrics workflow (after reviews + approval) through the action menu to seed the baseline

In the future, maybe we can have a dedicated runner...

.benchmarks/compare_results.py
```diff
@@ -0,0 +1,664 @@
...
+        --previous .benchmarks/reference \\
+        --thresholds .benchmarks/thresholds.yml \\
+        --version 4.2.0 \\
+        --output-dir .benchmarks/artifacts
+
+When no ``--previous`` is provided the report is rendered current-only (used
+for the first run after the baseline branch is created).
+"""
+
+from __future__ import annotations
```
kinow: Is this needed? We dropped annotations from ruff as it may hide unused imports in older py 3.10 (.11 too, I think? I think py 3.14 is safe), e.g.,

```
(autosubmit4) kinow@ranma:~$ python
Python 3.10.20 | packaged by conda-forge | (main, Jun 11 2026, 03:31:56) [GCC 14.3.0] on linux
>>> def f(x: Viking): ...
... 
  File "<stdin>", line 1, in <module>
NameError: name 'Viking' is not defined
```

But if we have `from __future__ import annotations`, then

```bash
(autosubmit4) kinow@ranma:~$ python
Python 3.10.20 | packaged by conda-forge | (main, Jun 11 2026, 03:31:56) [GCC 14.3.0] on linux
>>> from __future__ import annotations
>>> def f(x: Viking): ...
... 
>>> f(10)
>>> 
```

Was there a reason for using future annotations here that we cannot avoid it?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -17,3 +17,29 @@ Thanks for your contribution! Please:
 - [ ] Changelog entry included in `CHANGELOG.md` if this is a change that can affect users.
 - [ ] Documentation updated.
 - [ ] If this is a bug fix, PR should include a link to the issue (e.g. `Closes #1234`).
+
+<!--
+### Performance benchmarks
+
+Autosubmit runs a performance gate on PRs labeled `perf-benchmark`: the PR is
```
kinow: s/gate/check? Benchmark check? I'm not sure what gate means here, maybe gate is a term from performance? If so, feel free to ignore my comment 👍 Suggested check as that's what other GH Actions are called... 

Image

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -17,3 +17,29 @@ Thanks for your contribution! Please:
...
+benchmarked against the stored master baseline and the result is posted as a
+comment.
+
+* Add the `perf-benchmark` label if this PR can affect runtime performance
+  (e.g. job list, monitor loop, database layer, wrappers, platform handling).
+  Adding labels requires write/triage access -- if you can't, say so in the
+  description so a maintainer can label it.
+* Once labeled, the merge button stays disabled until the benchmark completes
+  (full suite, ~20 minutes). Add the label when the PR is ready; while it
+  is on, every push triggers another full run.
```
kinow, line 34: I guess the commit can ignore that with `noci`... any of those GitHub special commit keywords that disable GH Actions. But not important to have it here IMO.

autosubmit/profiler/profiler.py
```diff
@@ -327,27 +327,21 @@ def _report(self) -> None:
...
-        report += f"\nINITIAL MEMORY: {mem_init:.2f} {_UNITS[unit]}."
-        report += f"\nFINAL MEMORY: {mem_final:.2f} {_UNITS[unit]}."
+        def _to_units(value: float) -> tuple[float, str]:
+            abs_v = abs(value)
+            u = 0
+            while abs_v >= 1024 and u < len(_UNITS):
+                abs_v /= 1024
+                value /= 1024
+                u += 1
+            return value, _UNITS[u]
```
kinow, line 338: Note to self: I feel like I've seen another code in our repo doing the unit conversion, but I could be wrong.

test/integration/commands/test_performance.py
```diff
@@ -163,142 +178,90 @@ def prepare_yml(members, chunks, splits) -> dict:
...
+        return amount / 1024
+    if unit == "MiB":
+        return amount
+    if unit == "GiB":
+        return amount * 1024
+    if unit == "TiB":
+        return amount * 1024 * 1024
+    if unit == "PiB":
+        return amount * 1024 * 1024 * 1024
+    return amount / (1024 * 1024)  # B
```
kinow, line 198: Ah, maybe it was this code that I saw that was doing something similar, converting file sizes, units. I think we also have something similar in `autosubmit.profiler`... although not converting to one unit specifically. Maybe we cannot combine that.

.gitignore
```diff
@@ -30,3 +30,10 @@ dockerfiles/authorized_keys
 venv/
 .ipynb_checkpoints
 .atl
+
+# Benchmark ( reference lives in a separated branch )
+.actrc
```
kinow, line 35: Is this file needed? I cannot find it here. Maybe it only exists in the other branch (as per the comment)?

.gitignore
```diff
@@ -30,3 +30,10 @@ dockerfiles/authorized_keys
...
+
+.benchmarks/data/
+.benchmarks/artifacts/
+.benchmarks/reference/
```
kinow, line 39: Are these directories created when running the benchmarks locally? Asking as I haven't tried the tests locally yet 😬

CHANGELOG.md
```diff
@@ -7,13 +7,15 @@ This release also includes several bug fixes and enhancements to improve the ove
 **Bug fixes:**
 
 - Fix timeout guard is silently disabled for login/local jobs #3081
+- Fixed the Profiler reporting `MEMORY GROW` with the wrong unit when the growth was small relative to the final memory (e.g. ~0.5 MiB shown as 527 "MiB"). #3162
```
kinow: Grow is normally a verb, and growth as noun? The sentence has "the growth". Is it intended to say grow as in a verb here (and elsewhere)? Not important as this is internal AS code, not shared with users, so we don't need to be too rigorous with how we call things here, IMO. But maybe good in case we want to share it with other groups/teams/etc.

MAINTAINERS.md
kinow, line 1: Couldn't we use a file in `.benchmarks`? There is a `MAINTAINERS.md` file used in some projects, e.g., 

- https://toc.hyperledger.org/guidelines/MAINTAINERS-guidelines.html
- https://jungle.bsc.es/git/rarias/jungle/src/commit/4ff540c4c8121f00ba9a78a3620973d58cb3ec2d/doc/maintainers.md
- https://gitlab.com/gitlab-org/omnibus/-/blob/6.1.18/MAINTAINERS.md
- https://github.com/nexu-io/open-design/blob/main/MAINTAINERS.md
- https://sources.debian.org/src/llvm-toolchain-23/1%3A23.1.0-rc2-1~exp1/mlir/Maintainers.md

etc.

That file is used to list the maintainers of the project (or project with sub-projects/modules, where it's important to know who maintains what). But it looks, from the description, like this file is being used to list who can trigger the benchmarks tests, and I think the idea here is to allow someone external (e.g., from DestinE, another organisation, etc.) to trigger that in their PR. So we could have someone included that's not a maintainer, and the file currently doesn't follow the pattern of what's normally found in a `MAINTAINERS.md` file...

CONTRIBUTING.md
kinow, line 1: Thanks for the detailed docs!

dbeltrankyl:
Thanks for the feedback @kinow

I'll address them later. I will also  do a new commit to fine-tune the bechmark scenearios, 

The benchmark* already take 20 minutes ( as it is sequential)

So I plan to reduce the number of run scenarios and perhaps also the time_taken of some of them (fewer iterations)  so I can run them 3 times and take the median instead of a single run of each scenario... To see if I manage to reduce the noise in the comparison towards the baseline

.benchmarks/compare_results.py
dbeltrankyl: I was using 3.9 ( not anymore), and PyCharm or some automatic lint tool I have set up added it. Removed it, thanks!

.github/PULL_REQUEST_TEMPLATE.md
dbeltrankyl: Right, changed to check to be consistent

.gitignore
dbeltrankyl: It is when you trigger the action locally through `act` that we don't want to commit these files, as it would be noise. (On a side note, in the end I just used the fork I had, as it was way easier to test everything)

.gitignore
dbeltrankyl: Yes, we don't want that noise

CHANGELOG.md
dbeltrankyl: Renamed to growth everywhere

MAINTAINERS.md
dbeltrankyl: Yeah, it is confusing... I've moved it to .benchmarks/allowed-users.txt.. thanks!

dbeltrankyl:
Modified the scenarios; now we have the same sizes for all commands, and the benchmark runs 3 times 

The metrics job now takes 36m ( up from 20m)... waiting for the job to end to see if there are issues 

The quick run should be faster

I've removed the MEM_TEST one; I found it confusing to have a scenario with a bigger number of jobs ending almost at the same time as a shorter scenario in a benchmark test. Also, the number of jobs computed was different between runs which made the heatmap weirdier

Since the create tends to be fast, maybe I can just add a normal regression test that does create + run for this case with 10 iterations or less just to check the MEM usage. I will think about it in another issue

dbeltrankyl:
* Separated the mem-test into a new table that compares the mem and mem growth (only run) between all commands for the same scenario 
* Added crosscheck  ( run mem < other commands mem for the same scenario)
* Some qol, ruff,  etc

dbeltrankyl:
While checking the metrics, I find that some of them are still odd

* Added a function to force release the resources to Linux. I ran a profile for the same size for create + run, and now the run is less than the create; that is what I would expect

* Deciding to drop the obj growth from the render; the profiler itself is contaminating it, and the gc.collect is not freeing the objects.. I think it is not giving much data, to be honest

* Mem growth noticed in the autosubmit run; after the new function; I'll investigate why and if it is also the profiler contaminating it ( but in another pr ) 

* Crosscheck metric: not sure if I should compare the max rss peak of create,setstatus, recovery vs one point of the autosubmit run instead of the final memory of each one

Modified the heatmaps a bit more, waiting for the results

dbeltrankyl:
/metrics_full 

Testing if this is still working and also to check the last fixes because the last benchmark run was the first one with that machine and I was fixing the comparison towards the baseline

dbeltrankyl:
Action was triggered, but I think the comment indicating progress is missing because the metrics.yaml in master doesn't include it yet. Regardless, the benchmark results should be updated in 30 min

dbeltrankyl:
Ok I was wrong; that workflow is also using the path set in master. The only way of triggering it during this PR is through the label, did a dummy commit

dbeltrankyl:
Ok, I think this is ready to have a last review @kinow did the last one, but anyone is invited to review it

dbeltrankyl:
Rebased

kinow:
@dbeltrankyl , I think this is the next in the list to review. Sorry for the delay.

Can you rebase and fix the conflicts?

dbeltrankyl:
rebased, tomorrow morning I'll revise if the CI/CD had issues. Thanks!

kinow:
Thanks you @dbeltrankyl!

CHANGELOG.md
```diff
@@ -24,8 +25,9 @@ several bug fixes and enhancements to improve the overall user experience.
 - [enhancement] Allow recovery to update current running/ready jobs #1251
 - `autosubmit` Bash autocomplete #1227 #3171
 - Added "Did you mean 'run'" when an unknown sub-command is similar (e.g., "rum") to a valid one. #3194 #3171
+- Introduced a performance benchmark suite and a `metrics` workflow that profiles `create`, `run`, `recovery` and `setstatus`, compares each run against a per-CPU baseline stored on the `benchmark-reference` branch, and posts a regression report (tables, heatmap plots and artifact download links) on PRs labeled `perf-benchmark`. Maintainers can also trigger it manually with `/metrics`, `/metrics_full` and `/metrics_promote`. #2696 #3162
 
-**Migration from `job_list.pkl` to Database**
+**Migration from `job_list.pkl` to Database**
```
kinow: Unnecessary space at the end of the phrase?

autosubmit/helpers/utils.py
```diff
@@ -384,3 +386,24 @@ def build_and_connect_platform(platform_name: str, as_conf: 'AutosubmitConfig',
...
+    remaining free heap pages back so RSS reflects real usage instead of the
+    historical high-water mark. No-op on platforms without ``malloc_trim``.
+
+    :return: None
+    :rtype: None
+    """
+    gc.collect()
+    if _malloc_trim is not None:
+        with suppress(OSError):
+            _malloc_trim(0)
```
kinow: I think this sounds a bit dangerous.

Python keeps the memory around exactly because it's slow to request more memory. So in some cases, forcing the process to release memory, will seem like the process is using less memory, but it may have returned memory that's not from the jobs only, but also from other objects that are frequently re-allocated.

In that case, I believe Python would have to re-create arenas/pools, which is slower than doing so with the process doesn't need to ask Kernel for more memory...

Also, this won't work on Alpine (common in containers), only glibc users will have it.

Image

kinow: Found a recent comment in the Python discuss forum about it: https://discuss.python.org/t/using-malloc-trim-to-help-with-memory-management/107682/2

kinow:
I have a few comments in the review, but all pending as there are two that I want to review more carefully before hitting submit.

I was going to finish it today, but got sidetracked by issues at the ClimateDT VM, MN5, and Slack messages. Will try to finish it by Friday or Monday, not much :brain: left for this sorry, @dbeltrankyl . I will try to review something easier to send small PRs today.

PR #3163, opened 2026-08-06
Add timezone-awareness to jobs db table

LuiggiTenorioK:
There are some parts of the code that are not timezone-aware. This PR will try to fix the ones in the new `jobs` table that replaced the pkl file.

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -78,6 +78,7 @@ from autosubmit.git.autosubmit_git import (
     clone_repository,
     is_git_repo,
 )
+from autosubmit.helpers.datetime_utils import to_utc_iso
 from autosubmit.helpers.enums import ChunkUnit
 from autosubmit.helpers.utils import check_jobs_file_exists, get_rc_path
 from autosubmit.helpers.version import get_version
@@ -3273,8 +3274,12 @@ class Autosubmit:
                         Log.warning("The user does not exist anymore in the system, using id instead")
                         user = str(uid)
 
-                    created = datetime.datetime.fromtimestamp(
-                        Path(as_conf.conf_folder_yaml).stat().st_mtime)
+                    created = to_utc_iso(
+                        datetime.datetime.fromtimestamp(
+                            Path(as_conf.conf_folder_yaml).stat().st_mtime,
+                            tz=datetime.timezone.utc,
+                        )
+                    )
 
                     if as_conf.get_svn_project_url():
                         model = branch = as_conf.get_svn_project_url()
--- a/autosubmit/database/tables.py
+++ b/autosubmit/database/tables.py
@@ -33,6 +33,8 @@ from sqlalchemy import (
     UniqueConstraint,
 )
 
+from autosubmit.helpers import LOCAL_TZ
+
 metadata_obj = MetaData()
 
 ExperimentTable = Table(
@@ -191,8 +193,8 @@ JobsTable = Table(
     Column("packed", Boolean),
     Column("current_checkpoint_step", Integer, nullable=False, default=0),
     Column("platform_name", String),
-    Column("created", Text, nullable=False, default=lambda: datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d %H:%M:%S")),
-    Column("modified", Text, nullable=False, default=lambda: datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d %H:%M:%S"))
+    Column("created", Text, nullable=False, default=lambda: datetime.datetime.now(tz=LOCAL_TZ).isoformat(timespec='seconds')),
+    Column("modified", Text, nullable=False, default=lambda: datetime.datetime.now(tz=LOCAL_TZ).isoformat(timespec='seconds'))
 )
 
 """Table that holds the structure of the experiment jobs."""
--- a/autosubmit/database/tables.py
+++ b/autosubmit/database/tables.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2025 Earth Sciences Department, BSC-CNS
+# Copyright 2015-2026 Earth Sciences Department, BSC-CNS
 #
 # This file is part of Autosubmit.
 #
@@ -16,7 +16,6 @@
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
 
-import datetime
 from functools import cache, cached_property
 from typing import cast
 
@@ -33,7 +32,7 @@ from sqlalchemy import (
     UniqueConstraint,
 )
 
-from autosubmit.helpers import LOCAL_TZ
+from autosubmit.helpers.datetime_utils import utc_now_iso
 
 metadata_obj = MetaData()
 
@@ -62,7 +61,7 @@ ExperimentStatusTable = Table(
     Column("name", Text, nullable=False),
     Column("status", Text, nullable=False),
     Column("seconds_diff", Integer, nullable=False),
-    Column("modified", Text, nullable=False),
+    Column("modified", Text, nullable=False, default=utc_now_iso),
 )
 """Stores the status of the experiments."""
 
@@ -75,8 +74,8 @@ ExperimentRunTable = Table(
     "experiment_run",
     metadata_obj,
     Column("run_id", Integer, primary_key=True),
-    Column("created", Text, nullable=False),
-    Column("modified", Text, nullable=True),
+    Column("created", Text, nullable=False, default=utc_now_iso),
+    Column("modified", Text, nullable=True, default=utc_now_iso),
     Column("start", Integer, nullable=False),
     Column("finish", Integer),
     Column("chunk_unit", Text, nullable=False),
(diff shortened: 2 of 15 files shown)


LuiggiTenorioK:
@kinow this PR is ready for review.

Not sure why the coverage report didn't appear as a comment, but in codecov everything is covered.

The changelog entry is not included because it is part of something that is not released yet.

autosubmit/helpers/__init__.py
```diff
@@ -16,3 +16,7 @@
 # along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
 
 """Autosubmit Helpers."""
+
+from datetime import datetime, timezone
+
+LOCAL_TZ = datetime.now(timezone.utc).astimezone().tzinfo
```
kinow, line 22: If the VM is booted in Finland time, maybe daylight saving time, and then someone imports `import LOCAL_TZ from autosubmit.helpers`, and then the time is adjusted to another time zone, or to compensate for the daylight saving time change, the Python imported value is not changed while the code is running. So I think there's a small chance here that if somewhere in the AS scheduler main loop it uses `LOCAL_TZ`, it could be using an outdated value. I haven't tested it, but it should be possible to verify with Docker running ubuntu:latest or a Python image, creating LOCAL_TZ in a terminal, and in another one changing the system timezone. I believe LOCAL_TZ would be outdated.

kinow (commented):
Didn't have time to test things out, sorry @LuiggiTenorioK . Did a first pass reading the code. One concern I had was about the time zone being a constant -- I think the time zone may not be a constant over the course of a large simulation (e.g., change of time zone, or a user in a container doing some crazy setup, changing settings on the fly).

I also think we are still going to be use a text field in Postgres? Couldn't we use one of those timestamp + timezone in Postgres, and text in sqlite? Perhaps that's a bit too hard? 

But for sqlite, wouldn't UTC suffice?

What we will get in the text field will be something like +02:00. Which could be either Spain in summer or Finland in winter (Spain being +0, +1 & +2, and Finland +2 and +3). Then the UI can retrieve the UTC time from the DB and present to the user in any timezone.

LuiggiTenorioK:
> I also think we are still going to be use a text field in Postgres? Couldn't we use one of those timestamp + timezone in Postgres, and text in sqlite? Perhaps that's a bit too hard?

I am not sure about how safe it could be to change the data type. I will give it a try, because it is a new table and shouldn't have retrocompatibility issues.

> But for sqlite, wouldn't UTC suffice?

Is sufficient if the datetime has a timezone, no matter which one. Since it is in ISO format, it could be parsed and manipulated to any timezone by any programming language. That is why having it as a constant doesn't matter that much.

autosubmit/helpers/__init__.py
LuiggiTenorioK: I actually see it as a feature more than a bug :sweat_smile: 

I will be super confused reading the dates if I have 2 different time zones at least in the same run.

kinow:
> I am not sure about how safe it could be to change the data type. I will give it a try, because it is a new table and shouldn't have retrocompatibility issues.

:+1: otherwise, I think we may have to change that later, at some point (when you are not around :sob: :grimacing:  ).

Thank you, @LuiggiTenorioK !

LuiggiTenorioK:
I tried with [DateTime](https://docs.sqlalchemy.org/en/21/core/type_basics.html#sqlalchemy.types.DateTime) from SQLAlchemy, and it declares the right data type for Postgres and SQLite. However, when writing in SQLite, the timezone gets lost in the way.

I'm trying to find another solution, but it makes the code more complicated. So, I think it is better to keep it as it is.

LuiggiTenorioK:
@kinow, can you review it again pls

@ntorqulu will merge it if I am not here when it is approved

autosubmit/job/job.py
```diff
@@ -362,7 +363,7 @@ def __getstate__(self):
         job_data["remote_logs_err"] = self.remote_logs[1] if self.remote_logs[1] else ""
         if job_data["date"]:
             job_data["date"] = job_data["date"].isoformat()
-        job_data["modified"] = datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d %H:%M:%S")
+        job_data["modified"] = datetime.datetime.now(tz=LOCAL_TZ).isoformat(timespec='seconds')
```
kinow: @pablogoitia , @ntorqulu it'd be good if we could discuss this one at some point. I didn't have much time to discuss with Luiggi about this part. But my preference is to use timezone.utc everywhere in the workflow backend. And in the API, also store and send everything as UTC. Leaving the conversion to a timezone to the final part of the process, normally the web GUI, or someone running a script, notebook.

pablogoitia: I understand that the purpose was to facilitate the work on the GUI side, but a format conversion is still needed in some cases. For example, in the job history table the timestamps are displayed as they come from the API, but in the job detail cards, the durations are given in a more human-readable format.

I'm on your side. It would be better to have a standard format and make the conversions in the front to avoid problems such as the one raised a few weeks ago https://github.com/BSC-ES/autosubmit-gui/issues/449. However this would imply giving support to 3 timestamp formats (currently there are 2). Anyway I would prefer start moving to the definitive, standard one.

ntorqulu: I agree that the backend should store all dates as UTC without any local conversion. 
I think that the API should do the normalization of the different date formats on read: legacy rows stored as local (+0200) or as plain dates (yyyy-mm-dd hh:mm:ss) should be parsed and re-serialized as UTC in the API before being sent to the GUI. 
And that the GUI should convert to local for display; since the API agrees on the contract of returning everything as UTC, the GUI just have to parse it once and render it with the browser timezone.

autosubmit/database/tables.py
```diff
@@ -33,6 +33,8 @@
     UniqueConstraint,
 )
 
+from autosubmit.helpers import LOCAL_TZ
```
ntorqulu: I would remove the `LOCAL_TZ` constant and replace it with a time module in helpers; like `autosubmit/helpers/datetime_utils.py`. This way `history/`, `database/`, `job/` etc... can import without circular dependency risk.

And the timestamp helpers should have a function that returns the current UTC time as an ISO string, `utc_now_iso()`. This way all places that need utc encoding get it from the same source.

autosubmit/database/tables.py
```diff
@@ -191,8 +193,8 @@
     Column("packed", Boolean),
     Column("current_checkpoint_step", Integer, nullable=False, default=0),
     Column("platform_name", String),
-    Column("created", Text, nullable=False, default=lambda: datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d %H:%M:%S")),
-    Column("modified", Text, nullable=False, default=lambda: datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d %H:%M:%S"))
+    Column("created", Text, nullable=False, default=lambda: datetime.datetime.now(tz=LOCAL_TZ).isoformat(timespec='seconds')),
```
ntorqulu:
```suggestion
    Column("created", Text, nullable=False, default=utc_now_iso),
```

autosubmit/database/tables.py
```diff
@@ -191,8 +193,8 @@
...
+    Column("modified", Text, nullable=False, default=lambda: datetime.datetime.now(tz=LOCAL_TZ).isoformat(timespec='seconds'))
```
ntorqulu:
```suggestion
    Column("modified", Text, nullable=False, default=utc_now_iso),
```

autosubmit/job/job.py
ntorqulu:
```suggestion
        job_data["modified"] = utc_now_iso()
```

autosubmit/database/tables.py
```diff
@@ -60,7 +61,7 @@
     Column("name", Text, nullable=False),
     Column("status", Text, nullable=False),
     Column("seconds_diff", Integer, nullable=False),
-    Column("modified", Text, nullable=False),
+    Column("modified", Text, nullable=False, default=utc_now_iso),
```
ntorqulu: `ExperimentStatusTable.modified`
Backward compatibility to handle in the API: from local time with offset and no colon (%Y-%m-%dT%H:%M:%S%z'), to utc timezone aware ("2026-09-04T09:00:00+00:00")

autosubmit/database/tables.py
```diff
@@ -73,8 +74,8 @@
     "experiment_run",
     metadata_obj,
     Column("run_id", Integer, primary_key=True),
-    Column("created", Text, nullable=False),
-    Column("modified", Text, nullable=True),
+    Column("created", Text, nullable=False, default=utc_now_iso),
+    Column("modified", Text, nullable=True, default=utc_now_iso),
```
ntorqulu: `ExperimentRunTable.created & ExperimentRunTable.modified`
Backward compatibility to handle in the API: from local time with offset and no colon (%Y-%m-%dT%H:%M:%S%z'), to utc timezone aware ("2026-09-04T09:00:00+00:00")

autosubmit/autosubmit.py
```diff
@@ -3273,8 +3274,12 @@ def describe(input_experiment_list="*", get_from_user=""):
                         Log.warning("The user does not exist anymore in the system, using id instead")
                         user = str(uid)
 
-                    created = datetime.datetime.fromtimestamp(
-                        Path(as_conf.conf_folder_yaml).stat().st_mtime)
+                    created = to_utc_iso(
```
ntorqulu, line 3277: Not sure if the returned value from `describe` command for the experiment `created` date should be utc or local.
Maybe convert it to local time in `Log.result`? Or should be all kept as utc for consistency?

PR #3167, opened 2026-08-06
Fix GH actions ruff job to avoid linting deleted files

### LINKED ISSUE #3166: Ruff GH actions `HEAD^1` fails on deleted files or single-commit branches
### Description

Enforce the same ruff command to lint only changed files in AS backend. Follow a similar implementation as AS API.
Should be robust against:
- deleted-file linting
- single-commit branches
- pr that don't point to master
- manually triggered

E.g. of the one implemented in AS API:
```
  lint:
    runs-on: ubuntu-latest
    timeout-minutes: 5

    steps:
      - name: Checkout code
        uses: actions/checkout@v7
        with:
          fetch-depth: 0

      - name: Set up Python
        uses: actions/setup-python@v7
        with:
          python-version: "3.10"

      - name: Install dependencies
        run: |
          python -m pip install --upgrade pip
          pip install -e .[all]

      - name: Determine diff range
        id: range
        run: |
          if [ "$GITHUB_EVENT_NAME" = "pull_request" ]; then
            echo "range=origin/${GITHUB_BASE_REF}...HEAD" >> "$GITHUB_OUTPUT"
          elif [ "$GITHUB_EVENT_NAME" = "push" ] && [ "${{github.event.before}}" != "0000000000000000000000000000000000000000" ]; then
            echo "range=${{github.event.before}}...${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
          else
            echo "range=" >> "$GITHUB_OUTPUT"
          fi
      
      - name: Lint changed files
        run: |
          RANGE="${{ steps.range.outputs.range }}"
          if [ -n "$RANGE" ]; then
            git diff -z --name-only --diff-filter=ACMRT "$RANGE" -- '*.py' > changed.txt
          else
            git ls-files -z -- '*.py' > changed.txt
          fi

          if [ -s changed.txt ]; then
            xargs -0 -a changed.txt ruff check
          else
            echo "No Python files to lint."
          fi
```

ntorqulu:
Closes #3166
- Change `lint.yml` to lint only changed files.
- Fix `package-data` file resolution in `pyproject.toml`
- Follow pep 621/639 conventions

**Check List**

- [x] Documentation updated.


pyproject.toml
```diff
@@ -75,6 +75,7 @@ readme = {file = "README_PIP.md"}
 
 [project.optional-dependencies]
 postgres = ["psycopg", "psycopg2"]
+ruff = ["ruff==0.16.1"]
```
kinow: Right now we have ruff in tests. And other dependencies used for linting (are/were there... can't recall if we dropped some)

kinow: Not saying it's the right place! Just that we already have it there. Mypy was there and then was moved, I think. So we could do something like a lint group, moving ruff there, and any others if we have to add later (pycodestyle, flake, etc.). Maybe. :sweat_smile:

ntorqulu: Yep, I didn't look in tests. At the end I've moved ruff into its own lint group so we can add more dependencies later if they arise.

Looking with more detail at the pyproject toml file I've found things that maybe could be updated:

- The license was specified in the classifiers, but as `setuptools >= 77.0.3` follows PEP 639 standard, it deprecates the license classifier and recommends setting it in a separate field. Moved it outside `classifiers`.

- Changed `[project.entry-points.console_scripts]` to `[project.scripts]` following PEP 621. `[project.scripts]` is used for cli commands, while `[project.entry-points.<group>]` is more for generic entry-points groups (like plugins) I think.

kinow: > The license was specified in the classifiers, but as setuptools >= 77.0.3 follows PEP 639 standard, it deprecates the license classifier and recommends setting it in a separate field. Moved it outside classifiers.

+1

> Changed [project.entry-points.console_scripts] to [project.scripts] following PEP 621. [project.scripts] is used for cli commands, while [project.entry-points.<group>] is more for generic entry-points groups (like plugins) I think.

I'll have a look at that one later. The idea was to make the AS subcommands entrypoints (so another package could implement their own `autosubmit do-some-with-wrappers-for-instance`). But I haven't looked at the PEP & recent updates to build/setuptools/etc in a while.

Thanks @ntorqulu !

ntorqulu: And there are a few things that could be rewritten, but I'm not sure:

1. `vulture`
@kinow, are we using `vulture` anywhere in the codebase? I think the tool is not used anywhere.
Could be removed?
```
[tool.vulture]
exclude = ["*/build/*", "*/.venv/*", "*/docs/*"]
min_confidence = 70
```

2. `package-data`
I think it's not pointing to the correct files. Could be? I don't see any `.conf` files nor any `config/conf` directory.
```
[tool.setuptools]
package-data = { "autosubmit" = [
    "autosubmit/config/files/autosubmit.conf",
    "autosubmit/config/files/expdef.conf",
    "autosubmit/database/data/autosubmit.sql",
    "autosubmit/config/conf/files/*",
    "README",
    "CHANGELOG",
    "VERSION",
    "LICENSE",
    "docs/autosubmit.pdf"
] }
include-package-data = true
```

kinow: >  @kinow, are we using vulture anywhere in the codebase? I think the tool is not used anywhere.
>Could be removed?

We use it from time to time to inspect dead code. @VindeeR used it and delete a big chunk of code. That's done on demand, by devs, whenever they have time (i.e., not in CICD).

If it's causing any issue, we can remove it and add it back later if needed. If it's not breaking anything, I think leaving it might save us some time the next time we need to run it.

> package-data
> I think it's not pointing to the correct files. Could be? I don't see any .conf files nor any config/conf directory.

You might be right here. That was kept in AS3 to 4 migration, then setup.py to pyproject.toml migration... The conf/sql files are either moved, deleted, or replaced by YAML files. I think we should have the YAML files now from the AS config templates. I haven't checked if we are packaging those (but they should be, or autosubmit expid would fail I guess).

ntorqulu: Great, thanks!
I'll leave the vulture specifications then, just in case. I've added a comment just to remember that it's used from time to time.
About the files resolution in `package-data`, they were handled correctly because they are defined in `MANIFEST.in`. So, although the path was not correct in `pyproject.toml`, in `MANIFEST.in` it was correctly resolved because includes `autosubmit/config/files/*` and `*sql`, `*yml` and `*.conf` wildcards.

ntorqulu: p.d. in `MANIFEST.in`, the `*.conf` wildcard may be too broad. It also matches the test fixtures in `regression_to_be_cleaned`.
And there are two `include` lines that are dead:
- `include CHANGES` -> repo has `CHANGELOG.md`
- `include README` -> repo has `README_PIP.md` and `README.md`

kinow: > I'll leave the vulture specifications then, just in case. I've added a comment just to remember that it's used from time to time.

Thanks! We should have done it before! :sweat_smile: 

> About the files resolution in package-data, they were handled correctly because they are defined in MANIFEST.in. So, although the path was not correct in pyproject.toml, in MANIFEST.in it was correctly resolved because includes autosubmit/config/files/* and *sql, *yml and *.conf wildcards.

Great detective work! Looks like we have some work to do with packaging. 

Those include lines probably need to go or be updated to include the md files (unless they are included some other way).

No one has looked at the packaging in a while. Thanks a lot for going the extra mile on this PR, @ntorqulu !

ntorqulu: No problem! Modified the MANIFEST files to match the ones in the project, but I'm not confident on removing the `global-include` line. I think it does not affect the wheel but it doesn't hurt leaving it.

pyproject.toml
```diff
@@ -159,14 +157,6 @@ module = [
 ignore_missing_imports = true
 follow_untyped_imports = true
 
-[tool.vulture]
-exclude = ["*/build/*", "*/.venv/*", "*/docs/*"]
-min_confidence = 70
-
-[project.entry-points.console_scripts]
-autosubmit = "autosubmit.scripts.autosubmit:main"
```
kinow, line 167: I'm modifying exactly console_scripts, and adding another entry point. I **think** the move to script may stay, but I'll have to check if it works with the other entry points I'm adding for #973

.github/workflows/lint.yml
```diff
@@ -13,19 +13,47 @@ permissions:
...
+      - name: Setup Python
+        uses: actions/setup-python@v7
         with:
-          fetch-depth: 2
-      - name: Run Ruff
-        run: ruff check $(git diff --name-only HEAD^1 HEAD -- '*.py')
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: pip install -e .[lint]
```
kinow, line 30: 👏

.github/workflows/lint.yml
```diff
@@ -13,19 +13,47 @@ permissions:
...
+            git diff -z --name-only --diff-filter=ACMRT "$RANGE" -- '*.py' > changed.txt
+          else
+            git ls-files -z -- '*.py' > changed.txt
+          fi
+
+          if [ -s changed.txt ]; then
+            xargs -0 -a changed.txt ruff check
+          else
+            echo "No Python files to lint."
+          fi
```
kinow, line 56: I reviewed the rest, left just this part for later. I'll try to review it tomorrow morning (with calm & ☕ 😬 ). Thanks!

kinow:
Argh. Reviewed this one yesterday, but I forgot to hit the send button. Sorry, @ntorqulu .

kinow (approved):
@ntorqulu feel free to merge! I just had a look and I think this script might work, and if not, I will just revert explaining the reason why (will try to leave it the way you modified first). Cheers!

ntorqulu:
Great, thanks @kinow!
If the lint job fails, let me know as I would have to change it in the API repo too!

PR #3175, opened 2026-08-11
Fixes run_modes issues.

### LINKED ISSUE #2273: job_data->experiment_data table issues
Hello, 

With the #2212 additions, this table now exhibits an unexpected behavior.

There is a new entry each time the autosubmit create (and autosubmit monitor, #2272) is triggered.

### LINKED ISSUE #3151: start after isn't working with master/4.2.0
@f-macchia mentioned this feature in de-mwt Slack channel, so I decided to test it as I haven't used it in a while.

With the tag 4.1.17 both experiments work. You start one with --start-after, and then it monitors the other experiment. Once the other experiment is done, the experiment that was monitoring is initialised correctly.

With master branch, 4.2.0 dev, after 5 minutes nothing happened. The experiment initially launched correctly says it's monitoring the other experiment, and that there are 0 jobs. Then, after I start the other experiment, I see the monitoring experiment increases the number of jobs running. But that's it. Once the other experiment is finished, the number goes back to `0`, and nothing happens.

We only have a unit test with mocks for this. Probably safer to have an integration test (if we had one, I think the 4.2.0 branch would have failed and this bug would have been prevented).

dbeltrankyl:
closes #3151
closes #2273

Fix --start-after and --run-only-members

- --start-after: the waiting experiment never started. Run totals were recomputed from the in-memory job list, which gets emptied as jobs finish, so the monitored run ended up with total=0 and the trigger never fired. Totals now come from the jobs DB. Also pointing to a non-existent experiment now warns correctly

- --run-only-members: all members ran. The filter was applied after the graph was loaded and nothing pruned the already-loaded jobs. The filter is now applied at load time.

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/autosubmit.py
+++ b/autosubmit/autosubmit.py
@@ -1844,7 +1844,8 @@ class Autosubmit:
             exp_history.initialize_database()
             run_dc = exp_history.process_status_changes(job_list.get_job_list(), as_conf.get_chunk_size_unit(),
                                                         as_conf.get_chunk_size(),
-                                                        current_config=as_conf.get_full_config_as_json())
+                                                        current_config=as_conf.get_full_config_as_json(),
+                                                        status_counts=job_list.get_status_counts())
             job_list.run_id = run_dc.run_id if run_dc else None
             Autosubmit.database_backup(expid)
         except Exception:
@@ -1870,7 +1871,8 @@ class Autosubmit:
         """
         exp_history = ExperimentHistory(expid)
         if len(job_changes_tracker) > 0:
-            exp_history.process_job_list_changes_to_experiment_totals(job_list.get_job_list())
+            exp_history.process_job_list_changes_to_experiment_totals(job_list.get_job_list(),
+                                                                      status_counts=job_list.get_status_counts())
             Autosubmit.database_backup(expid)
 
     @staticmethod
@@ -1943,10 +1945,12 @@ class Autosubmit:
 
         if recover:
             Log.info("Recovering job_list")
+        # Check if the user has launch autosubmit run with -rom option.
+        allowed_members = AutosubmitHelper.get_allowed_members(run_only_members, as_conf)
         try:
             job_list = Autosubmit.load_job_list(
                 expid, as_conf, new=False, full_load=False, submitter=submitter,
-                check_failed_jobs=True, run_mode=False)
+                check_failed_jobs=True, run_mode=False, run_members=allowed_members)
 
         except OSError as e:
             raise AutosubmitError(
@@ -1994,12 +1998,10 @@ class Autosubmit:
         if not recover:
             Log.info(f"Autosubmit is running with v{Autosubmit.autosubmit_version}")
             # Before starting main loop, setup historical database tables and main information
-        # Check if the user has launch autosubmit run with -rom option ( previously named -rm )
-        allowed_members = AutosubmitHelper.get_allowed_members(run_only_members, as_conf)
+        # Check if the user has launch autosubmit run with -rom option ( previously named -rm ).
+        # run_members was already applied in load_job_list before generating the
+        # graph, so only report the restriction here.
         if allowed_members:
-            # Set allowed members after checks have been performed.
-            # This triggers the setter and main logic of the -rm feature.
-            job_list.run_members = allowed_members
             Log.result(f"Only jobs with member value in {str(allowed_members)} or no member will be allowed in this "
                        f"run. Also, those jobs already SUBMITTED, QUEUING, or RUNNING will be allowed to complete and"
                        f" will be tracked.")
@@ -5184,7 +5186,7 @@ class Autosubmit:
     # TODO: To be moved to utils
     @staticmethod
     def load_job_list(expid, as_conf, monitor=False, new=True, full_load=True, submitter=None,
-                      check_failed_jobs=False, run_mode=False) -> JobList:
+                      check_failed_jobs=False, run_mode=False, run_members: list[str] | None = None) -> JobList:
         """Load the JobList for a given experiment.
 
         :param expid: experiment id
@@ -5195,10 +5197,15 @@ class Autosubmit:
         :param submitter: submitter to be used
         :param check_failed_jobs: whether to check failed jobs or not
         :param run_mode: whether to load the job list in run mode or not
+        :param run_members: optional list of members to restrict the run to. Set
+            before loading the job graph so that only the allowed members are
+            loaded into memory and therefore submitted.
         :return: JobList object
         """
         rerun = as_conf.get_rerun()
         job_list = JobList(expid, as_conf, YAMLParserFactory(), run_mode=run_mode, submitter=submitter)
+        if run_members:
+            job_list.run_members = run_members
         date_list = as_conf.get_date_list()
         date_format = ''
         if as_conf.get_chunk_size_unit() == ChunkUnit.HOUR:
--- a/autosubmit/helpers/autosubmit_helper.py
+++ b/autosubmit/helpers/autosubmit_helper.py
@@ -72,8 +72,11 @@ def handle_start_after(start_after: str, expid: str) -> None:
             Log.info(
                 "Hey! What do you think is going to happen? In theory, "
                 "your experiment will run again after it has been completed. Good luck!")
-        # Check if experiment exists. If False or None, it does not exist
-        if not check_experiment_exists(start_after):
+        # Check if experiment exists. If False or None, it does not exist.
+        # error_on_inexistence is disabled because a missing experiment must not
+        # block the run: it is reported and ignored.
+        if not check_experiment_exists(start_after, error_on_inexistence=False):
+            Log.warning(f"Experiment {start_after} does not exist. Ignoring the start_after trigger.")
             return
         # Historical Database: We use the historical database to retrieve the current progress
         # data of the supplied expid (start_after)
--- a/autosubmit/helpers/autosubmit_helper.py
+++ b/autosubmit/helpers/autosubmit_helper.py
@@ -77,6 +77,8 @@ def handle_start_after(start_after: str, expid: str) -> None:
         # block the run: it is reported and ignored.
         if not check_experiment_exists(start_after, error_on_inexistence=False):
             Log.warning(f"Experiment {start_after} does not exist. Ignoring the start_after trigger.")
+            # Keep the warning readable before the console is cleared.
+            sleep(3)
             return
         # Historical Database: We use the historical database to retrieve the current progress
         # data of the supplied expid (start_after)
--- a/autosubmit/helpers/autosubmit_helper.py
+++ b/autosubmit/helpers/autosubmit_helper.py
@@ -90,16 +90,34 @@ def handle_start_after(start_after: str, expid: str) -> None:
                  f"suspended jobs becomes equal to the total number of jobs of experiment {start_after}, experiment "
                  f"{expid} will start. Querying every 60 seconds. Status format "
                  f"Completed/Queuing/Running/Suspended/Failed.")
-        while current_run := exp_history.manager.get_experiment_run_dc_with_max_id():
-            if (current_run.finish > 0 and current_run.total > 0
-                    and current_run.total == current_run.completed + current_run.suspended):
-                break
-            sys.stdout.write(
-                f"\rExperiment {start_after} ({current_run.total} total jobs) status {current_run.completed}/"
-                f"{current_run.queuing}/{current_run.running}/{current_run.suspended}/{current_run.failed}")
+        waiting_for_first_run_reported = False
+        current_run = exp_history.manager.get_experiment_run_dc_with_max_id_or_none()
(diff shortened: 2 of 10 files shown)


dbeltrankyl:
I changed a lot of calls, so Ruff is complaining across many files not related to my changes per se. I think this is not the appropriate PR to fix them

dbeltrankyl:
Ready to review if you don't mind @VindeeR (  Bruno  had a lot of PR revisions lately )

VindeeR:
the tests passed but the coverage failed in the integration :sweat_smile: it should be fine just to kick it again

autosubmit/history/experiment_history.py
```diff
@@ -315,8 +319,28 @@ def _verify_slurm_monitor(self, slurm_monitor, job_data_dc):
...
+        :type current_config: str
+        :param create: whether a new run must be created regardless of changes.
+        :type create: bool
+        :param status_counts: optional pre-computed status counts (e.g. from the
+            complete jobs database, which includes jobs unloaded from memory).
+            When omitted they are computed from ``job_list``.
+        :type status_counts: dict[str, int] | None
+        :return: the current ``ExperimentRun`` or None if the database failed.
+        :rtype: ExperimentRun | None
+        """
```
VindeeR, line 343: I believe we re moving away from the :type: and the :rtype: since they'll be just a repetition from the signature of the function.

aka: `pydocs` to `signature + mypy`

autosubmit/history/experiment_history.py
```diff
@@ -332,28 +356,40 @@ def process_status_changes(self, job_list=None, chunk_unit="NA", chunk_size=0, c
...
+                                                      status_counts: "dict[str, int] | None" = None) -> ExperimentRun | None:
+        """ Updates current experiment_run row with totals calculated from job_list.
+
+        :param job_list: the jobs of the experiment as a plain list.
+        :type job_list: list[Job] | None
+        :param status_counts: optional pre-computed status counts (e.g. from the
+            complete jobs database, which includes jobs unloaded from memory).
+            When omitted they are computed from ``job_list``.
+        :type status_counts: dict[str, int] | None
+        """
```
VindeeR, line 382: same as the previous one

autosubmit/history/experiment_history.py
```diff
@@ -370,9 +406,21 @@ def _chunk_config_has_changed(self, current_exp_run_dc, new_chunk_unit, new_chun
...
+        """ Return updated row as Models.ExperimentRun.
+
+        :param experiment_run_dc: the run to update.
+        :type experiment_run_dc: ExperimentRun
+        :param job_list: the jobs of the experiment as a plain list. Used to
+            compute the counts when ``status_counts`` is not provided.
+        :type job_list: list[Job] | None
+        :param status_counts: optional pre-computed status counts (e.g. from the
+            complete jobs database, which includes jobs unloaded from memory).
+        :type status_counts: dict[str, int] | None
```
VindeeR, line 420: ditto

autosubmit/history/experiment_history.py
```diff
@@ -389,15 +437,28 @@ def finish_current_experiment_run(self):
...
+    def create_new_experiment_run(self, chunk_unit: str = "NA", chunk_size: int = 0, current_config: str = "",
+                                  job_list: "list[Job] | None" = None,
+                                  status_counts: "dict[str, int] | None" = None) -> ExperimentRun:
+        """ Also writes the finish timestamp of the previous run.
+
+        :param job_list: the jobs of the experiment as a plain list.
+        :type job_list: list[Job] | None
+        :param status_counts: optional pre-computed status counts (e.g. from the
+            complete jobs database). When omitted they are computed from ``job_list``.
+        :type status_counts: dict[str, int] | None
```
VindeeR, line 449: ditto

autosubmit/job/job_list.py
```diff
@@ -2262,6 +2262,23 @@ def get_job_list(self):
...
+        """Return status counts of ALL jobs from the database.
+
+        The counts are computed from the persisted ``jobs`` table, which keeps
+        every job (including finished jobs unloaded from memory), so they are
+        complete. Keys follow the same convention as
+        ``ExperimentHistory.get_status_counts_from_job_list``
+
+        :return: dict with keys COMPLETED, FAILED, QUEUING, SUBMITTED, RUNNING,
+            SUSPENDED and TOTAL.
+        :rtype: dict[str, int]
```
VindeeR, line 2275: ditto

VindeeR (commented):
I'll finish reviewing it tomorrow, thanks for improving the quality of the code.

autosubmit/history/experiment_history.py
dbeltrankyl: Ah I missed that update

autosubmit/history/experiment_history.py
```diff
@@ -332,28 +356,40 @@ def process_status_changes(self, job_list=None, chunk_unit="NA", chunk_size=0, c
...
-            return self.update_counts_on_experiment_run_dc(current_experiment_run_dc, job_list)
+            return self.update_counts_on_experiment_run_dc(current_experiment_run_dc, job_list,
+                                                           status_counts=status_counts)
         except Exception as exp:
             self._log.log(str(exp), traceback.format_exc())
             Log.debug(f'Historical Database error: {str(exp)} {traceback.format_exc()}')
 
-    def should_we_create_a_new_run(self, job_list, changes_count, current_experiment_run_dc, new_chunk_unit,
-                                   new_chunk_size, create=False):
+    def should_we_create_a_new_run(self, job_list: "list[Job]", changes_count: int, current_experiment_run_dc,
```
VindeeR, line 391:
```suggestion
    def should_we_create_a_new_run(self, job_list: "list[Job]", changes_count: int, current_experiment_run_dc: ExperimentRun,
```

autosubmit/history/experiment_history.py
```diff
@@ -370,9 +406,21 @@ def _chunk_config_has_changed(self, current_exp_run_dc, new_chunk_unit, new_chun
             return True
         return False
 
-    def update_counts_on_experiment_run_dc(self, experiment_run_dc, job_list=None):
-        """ Return updated row as Models.ExperimentRun. """
-        status_counts = self.get_status_counts_from_job_list(job_list)
+    def update_counts_on_experiment_run_dc(self, experiment_run_dc, job_list: "list[Job] | None" = None,
```
VindeeR, line 409:
```suggestion
    def update_counts_on_experiment_run_dc(self, current_experiment_run_dc: ExperimentRun, job_list: "list[Job] | None" = None,
```

autosubmit/history/experiment_history.py
```diff
@@ -448,29 +509,24 @@ def _get_next_counter_by_job_name(self, job_name):
...
         """ Each item in the job_list must have attributes: date, member, status_str. """
         job_list = job_list if job_list else []
         return sum(1 for job in job_list if
                    job.date is not None and job.member is not None and job.status_str == HUtils.SupportedStatus.COMPLETED)
 
-    def get_status_counts_from_job_list(self, job_list):
-        """
-        Return dict with keys COMPLETED, FAILED, QUEUING, SUBMITTED, RUNNING, SUSPENDED, TOTAL.
-        """
+    def get_status_counts_from_job_list(self, job_list: "list[Job] | None" = None) -> "dict[str, int]":
```
VindeeR, line 518:
```suggestion
    def get_status_counts_from_job_list(self, job_list: "list[Job] | None" = None) -> dict[str, int]:
```

autosubmit/autosubmit.py
```diff
@@ -1943,10 +1945,12 @@ def prepare_run(
 
         if recover:
             Log.info("Recovering job_list")
+        # Check if the user has launch autosubmit run with -rom option.
```
VindeeR, line 1948:
```suggestion
        # Check if the user launched autosubmit run with -rom option.
```

autosubmit/autosubmit.py
```diff
@@ -1994,12 +1998,10 @@ def prepare_run(
         if not recover:
             Log.info(f"Autosubmit is running with v{Autosubmit.autosubmit_version}")
             # Before starting main loop, setup historical database tables and main information
-        # Check if the user has launch autosubmit run with -rom option ( previously named -rm )
-        allowed_members = AutosubmitHelper.get_allowed_members(run_only_members, as_conf)
+        # Check if the user has launch autosubmit run with -rom option ( previously named -rm ).
```
VindeeR, line 2001:
```suggestion
        # Check if the user launched autosubmit run with -rom option ( previously named -rm ).
```

VindeeR (commented):
The code itself looks clean and pretty much good to go. There are smaller correction related to spelling and the newer pattern of the code.

I've never ran any workflow this way tho, so as I was trying to test it I couldn't replicate the expected behavior of starting an experiment after another finishes.

To run I executed the following as instructed without modifications on the experiment, so if you realize something is amiss let me know.

```
autosubmit expid -dm -H "local" -d "test"
autosubmit expid -dm -H "local" -d "test"
autosubmit create a01i; autosubmit create a01j;
nohup autosubmit run a01i & autosubmit run a01j --start-after a01i
```

I tried without the nohup, tried running both of them separately just to confirm both experiments were fine and tried to follow the process created, but the second job never started although the first experiment would run as expected.

dbeltrankyl:
Thanks @VindeeR !

> To run I executed the following as instructed without modifications on the experiment, so if you realize something is amiss let me know.

I'm currently working on another issue; I probably won't have time to check this today (as I still have to lunch ). 

I'll let you know when I come back to this one ( probably tomorrow morning )

dbeltrankyl:
I forgot I still had this PR open. I'll address the feedback. Thanks!

dbeltrankyl:
Thanks @VindeeR !

Added the feedback and rebased on the new master

About the --start-after case you couldn't reproduce: 

```bash
autosubmit expid -dm -H "local" -d "test"
autosubmit expid -dm -H "local" -d "test"
autosubmit create a01i; autosubmit create a01j;
nohup autosubmit run a01i & autosubmit run a01j --start-after a01i
```

I think I found it. When you launch both experiments at the same time, B starts polling A before A has registered its first run yet. The monitor would hit "No Experiment Runs registered."

Now waits for a run row to show up and tells you it's waiting while it does. 

I also added integration tests for the "launched together" and "A hasn't started yet" scenarios.

autosubmit/helpers/autosubmit_helper.py
```diff
@@ -72,8 +72,11 @@ def handle_start_after(start_after: str, expid: str) -> None:
             Log.info(
                 "Hey! What do you think is going to happen? In theory, "
                 "your experiment will run again after it has been completed. Good luck!")
-        # Check if experiment exists. If False or None, it does not exist
-        if not check_experiment_exists(start_after):
+        # Check if experiment exists. If False or None, it does not exist.
+        # error_on_inexistence is disabled because a missing experiment must not
+        # block the run: it is reported and ignored.
+        if not check_experiment_exists(start_after, error_on_inexistence=False):
```
VindeeR, line 78: As I was testing this I kept trying to find where this warning was displayed, but it is erased as soon as the next one starts, maybe we could add a small sleep here just so the user could se the warning, in case of a typo or something of the sorts it would be easier to find instead of having to open the logs.

Or maybe something interaction with the user like confirming the execution can be finished with no problems?!

VindeeR (commented):
It all looks good to me 🙌 
Thanks for the hard work.

I just had one thing I would like to clear about the logging, but other than that is all good to go.

autosubmit/helpers/autosubmit_helper.py
dbeltrankyl: Hello @VindeeR , thanks for the review!


> As I was testing this I kept trying to find where this warning was displayed, but it is erased as soon as the next one starts, maybe we could add a small sleep here just so the user could se the warning, in case of a typo or something of the sorts it would be easier to find instead of having to open the logs.

I was looking at the log files, not the terminal, but adding a sleep doesn't hurt too much. Thanks for the suggestion

> Or maybe something interaction with the user like confirming the execution can be finished with no problems?!

Nah, that would conflict with nohup. The autosubmit run should remain uninteractive ( except for the 2FA because we don't have a workaround)

dbeltrankyl: Added! And patched the sleep in the test. 

Should be ready to merge now

dbeltrankyl:
Hii @VindeeR 

Can you confirm if this is ready to merge? thanks!

Also, this fixes #2273

autosubmit/helpers/autosubmit_helper.py
```diff
@@ -72,8 +72,13 @@ def handle_start_after(start_after: str, expid: str) -> None:
...
+            Log.warning(f"Experiment {start_after} does not exist. Ignoring the start_after trigger.")
+            # Keep the warning readable before the console is cleared.
+            sleep(3)
```
kinow, line 81: Does it disappear because of the system `clear` that we issue?

PR #3200, opened 2026-08-27
fix: pass tzinfo to datetime.datetime() calls (ruff DTZ001)

TrueFurina:
## Summary

Fixes all remaining ruff `DTZ001` violations (`datetime.datetime()` called without a `tzinfo` argument) and removes the rule from the ignored list in `pyproject.toml`.

## Changes

- `autosubmit/helpers/autosubmit_helper.py`: `target_date` now built via `datetime_now.replace(hour=..., minute=..., second=...)` instead of a bare `datetime.datetime(...)` constructor (preserves the naive/local semantics of the surrounding code)
- `test/scripts/git/stats_average_close_time_gitlab_github.py`: `since` and `last_day_2025` now pass `tzinfo=datetime.timezone.utc`; removed the now-redundant `utc.localize(...)` line
- `test/scripts/git/stats_new_features_gitlab_github.py`: same fix
- `pyproject.toml`: removed `"DTZ001"` from the ruff `ignore` list

## Verification

- `python -c "import ast"` syntax check passed for all three modified Python files
- `pyproject.toml` parses cleanly
- No remaining bare `datetime.datetime(...)` calls in the three files (grep verified)


autosubmit/helpers/autosubmit_helper.py
kinow, line 1: Hi @TrueFurina . I think your changes here are re-formatting the whole file? At least GH shows several changes. It'd be easier to review if you limited to the changes related to tzinfo.

kinow (changes requested):
Thanks for the PR, @TrueFurina . I think first we'd need to reduce the size of the changes in Git. Then, we can check if it's building fine in GH Actions.

PR #3232, opened 2026-09-07
Improve provenance documentation

martiortis:
This PR expands the Autosubmit provenance documentation with practical information about configuring the content of generated RO-Crates.

It documents:
- additional workflow inputs through `ROCRATE.INPUTS`
- automatically exported and available input sections
- additional workflow outputs through `ROCRATE.OUTPUTS`
- unmatched, duplicate, and automatically included output files
- JSON-LD metadata customization through `ROCRATE.PATCH`

This information was previously documented in the ClimateDT workflow documentation, but it is not specific to ClimateDT. Moving it to the main Autosubmit documentation to avoid duplication.

**Check List**

- [x] Documentation updated.


martiortis:
As discussed on Slack with @kinow, I've moved the generic provenance configuration documentation I wrote from the ClimateDT workflow docs to the main Autosubmit documentation.

I also tested building the documentation locally with `make html`, and it completed successfully without errors.

Let me know if there's anything that should be changed.

docs/source/userguide/provenance.rst
```diff
@@ -51,6 +51,276 @@ and WorkflowHub are issuers of `DOI`_'s (digital object identifiers),
...
+For example, given:
+
+.. code-block:: yaml
+
+    PROJECT:
+      PROJECT_TYPE: git
+      PROJECT_DESTINATION: git_project
+
+exporting ``PROJECT`` results in ``PROJECT_TYPE`` and
+``PROJECT_DESTINATION`` being represented as workflow inputs.
```
kinow: This is coming right after the `INPUTS`, @martiortis . Would it be 

```
ROCRATE:
    PROJECT:
      PROJECT_TYPE: git
      PROJECT_DESTINATION: git_project
```

or

```
ROCRATE:
  INPUTS:
    PROJECT:
      PROJECT_TYPE: git
      PROJECT_DESTINATION: git_project
```

If the latter ,maybe we can add that here to avoid users confused (as PROJECT: is also common in AS configuration, without any parent attribute/object).

docs/source/userguide/provenance.rst
```diff
@@ -51,6 +51,276 @@ and WorkflowHub are issuers of `DOI`_'s (digital object identifiers),
...
+Automatically exported inputs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The following sections are exported automatically and do not need to be added
+under ``ROCRATE.INPUTS``:
+
+* ``DEFAULT``
+* ``EXPERIMENT``
+* ``CONFIG``
+* ``PROJECT``
```
kinow, line 106: `PROJECT` is listed here, but also in the example above. Maybe it'd be better to use another in the example above to avoid confusion?

docs/source/userguide/provenance.rst
```diff
@@ -51,6 +51,276 @@ and WorkflowHub are issuers of `DOI`_'s (digital object identifiers),
...
+
+* represented as a workflow output;
+* referenced from the RO-Crate ``CreateAction``;
+* described as a ``File`` entity in the JSON-LD metadata;
+* physically included in the RO-Crate ZIP file.
+
+Patterns without matches
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+If an output pattern does not match any file, it is silently ignored.
```
kinow, line 176: Oh, not even logged? /me hides

docs/source/userguide/provenance.rst
```diff
@@ -51,6 +51,276 @@ and WorkflowHub are issuers of `DOI`_'s (digital object identifiers),
...
+
+Each file selected through ``OUTPUTS`` must only be selected once.
+
+If the same file is matched more than once, for example because:
+
+* the same pattern is specified twice;
+* two different patterns overlap;
+* the file was already added to the RO-Crate through another mechanism;
+
+RO-Crate generation currently fails.
```
kinow, line 192: Ohhh, good to know. I didn't recall that.

docs/source/userguide/provenance.rst
```diff
@@ -51,6 +51,276 @@ and WorkflowHub are issuers of `DOI`_'s (digital object identifiers),
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``ROCRATE.OUTPUTS`` does not define the complete contents of the generated ZIP.
+
+The RO-Crate generator also includes files required to describe the experiment,
+including experiment configuration and Autosubmit runtime information.
+
+Therefore, ``OUTPUTS`` should be understood as a mechanism for declaring
+additional workflow output files rather than as a complete list of files that will
+be present in the archive.
```
kinow, line 207: Beautifully written 👏

kinow (commented):
Looking great, @martiortis ! Thanks a lot for the pull request. Leaving a few comments, but nothing major.

docs/source/userguide/provenance.rst
martiortis: Neither of those. If we wanted to explicitly export `PROJECT` through `ROCRATE.INPUTS`, the syntax would be:

```yaml
ROCRATE:
  INPUTS:
    - PROJECT
  OUTPUTS:
```

In this example, I only wanted to explain what happens hen we add the name of a top-level configuration section with child keys to `ROCRATE.INPUTS`: the direct child keys are exported as individual workflow inputs.
For example, we can find in the `/conf/minimal.yml` that :

```yaml 
PROJECT:
  PROJECT_TYPE: git
  PROJECT_DESTINATION: git_project
```
If we export: 
``` yaml
ROCRATE:
  INPUTS:
    - PROJECT
  OUTPUTS:
```
the workflow inputs in `ro-crate-metadata.json` will be represented as:
```json
"input": [
  {
    "@id": "#PROJECT.PROJECT_TYPE-param"
  },
  {
    "@id": "#PROJECT.PROJECT_DESTINATION-param"
  }
]
``` 

So `PROJECT` itself is not represented as a single workflow input, its direct child keys are.

However, I realize `PROJECT` was not a good example here and makes this confusing. As mentioned in the section below, `PROJECT` is already exported automatically. If we explicitly added it again under `ROCRATE.INPUTS`, its child inputs would be duplicated.

I’ll replace this with a different KEY that is not exported by default.

docs/source/userguide/provenance.rst
martiortis: Yes, I agree. Using `PROJECT` in the example is confusing because, as shown here, it is already exported automatically.

I’ll replace it with a different section that is not exported by default, such as `PLATFORMS`, so the example better represents the intended use of `ROCRATE.INPUTS`.

docs/source/userguide/provenance.rst
martiortis: I checked the logs from the RO-Crate generation test i did, and I couldn’t find any message related to the non-existing output pattern or the missing file. So, as far as I can tell, it is currently skipped without being logged.

docs/source/userguide/provenance.rst
martiortis: Yeah, I only noticed this while testing the different `OUTPUTS` cases. I thought it was worth documenting since overlapping patterns or re-adding an already included file can make the RO-Crate generation fail.

docs/source/userguide/provenance.rst
kinow: Ah, got it!

> I’ll replace this with a different KEY that is not exported by default.

Thanks!

martiortis: Just updated the example accordingly and rebuilt the documentation locally. The build completed successfully.

kinow: Great, @martiortis ! I've enable dthe GH Actions pipeline. In GitHub, external collaborators are only allowed to have their pull requests triggering GH Actions automatically once they have committed something. And you are not in the team of AS developers (although perhaps you should :grimacing: ). Once this is merged, all your PRs will automatically trigger GH Actions and run the tests. Thanks!

kinow (approved):
Previewed the docs at https://autosubmit--3232.org.readthedocs.build/en/3232/userguide/provenance.html and they look great. Just waiting for CICD to hit the merge button. Thanks @martiortis !

kinow:
Actually, as this is docs, there's no need to wait for the rest of CICD.

PR #3248, opened 2026-09-08
Use a simple Python function to create the database engine, instead of class + function

### LINKED ISSUE #3246: Remove unnecessary singleton object
We can have a single function, without having to create a class, with function, etc. This should create a lighter object/memory use, and keep the code shorter.

Noticed this while reviewing #2947 .

### LINKED ISSUE #3252: [Bug] - CodeCov stopped reporting
As I have been going through PRs lately I realized that the CodeCov has stopped reporting the changes and the coverage.

It used to generate a comment that would show a diff in coverage from the current master branch to the working branch, but at some point it stopped reporting and I believe it can be investigated.

kinow:
Closes #3246 

Closes #3252 

Related #2957 (we are adding an enum here for the database type; previously the function was using "sqlite" and "postgres" strings).

Draft for tomorrow. Wrote quickly now so I don't forget. I have to test it a bit manually, and then add unit and/or integration tests.

**Check List**

- [ ] Documentation updated.


kinow:
I think changes here are done. While GH Actions was running, I finished reading Codecov quick-start, https://docs.codecov.com/docs/quick-start, and noticed that a) before we had nada in .codecov.yml, b) later we/I added comments: true, c) but quick-start docs actually show that as a YAML object with several properties.

So I matched the quick-start docs and pushed an extra commit mentioning the issue #3252 . Let's see if that fixes the comments...

image

test/unit/database/test_session.py
kinow, line 1: Updated to fix small lint issues, add docs, and add a few more tests to achieve 100% coverage of the session function (as that was pretty easy).

codecov.yml
```diff
@@ -27,5 +28,10 @@ flag_management:
...
 
-# turn off comments to pull requests
-comment: true
+comment:
+  layout: " diff, flags, files"
+  behavior: default
+  require_changes: false
+  require_base: false
+  require_head: true
+  hide_project_coverage: false
```
kinow, line 37: These changes (rename + properties) fix the missing coverage comment.

autosubmit/database/session.py
```diff
@@ -15,21 +15,34 @@
...
+
+
+class DatabaseType(str, Enum):
+    """Enum for the database engine."""
+
+    SQLITE = "sqlite"
+    """SQLite database."""
+
+    POSTGRES = "postgres"
+    """Postgres database."""
```
kinow, line 39: To replace the strings... #2957

autosubmit/database/session.py
```diff
@@ -39,40 +52,70 @@ def _resolve_engine(connection_url: str) -> Engine:
     return sqlalchemy_create_engine(connection_url, poolclass=pool_class)
 
 
-class PostgreSQLEngineSingleton:
```
kinow, line 42: In Python a module is an object. And you can't clone it or re-instantiate (well, I think it's technically doable... you could delete it, try to re-load it, making sure objects/dependencies are updated or destroyed... it'd be a hell lot of work, and very dirty... but I digress).

So a module is already a singleton in Python. That's why design patterns like Singleton, Factory, Builder, are normally not translated from Java or web languages like TypeScript directly into Python.

Here I'm simply using the module as the singleton, and adding a function that does something similar to the existing code.

Main differences are we create the new engine first, and if that works, then we reassign the objects (to avoid race condition issues), and for Postgres, if you modify the connection URL, we automatically discard the old and create a new one.

@ntorqulu assigned you as reviewer as I did this change after reading a change you had where we had to have a new function in the tests to dispose the engine. I think this might simplify those tests?

kinow:
@ntorqulu if you are busy with other changes, I can ask Suraj or Erick to review it :+1: Cheers

test/unit/database/test_session.py
```diff
@@ -66,8 +92,115 @@ def test_get_engine_postgres(mocker):
...
+    try:
+        assert database_path.parent.exists()
+        assert database_path.exists()
+        assert database_path.is_file()
+        assert engine.name == "sqlite"
+    finally:
+        engine.dispose()
+
+
+@pytest.mark.postgres
```
ntorqulu, line 138: Mmm one question @kinow, the postgres-marked tests under `unit` are run in CI?
The only place where the postgres markers are accepted is in the `test-integration` job, but these are the tests under the `integration` folder only.
So I think that are unit tests marked with postgres are not running in the CI.

test/unit/database/test_session.py
```diff
@@ -37,15 +44,34 @@ def test_resolve_engine(url: str, expected: str | Exception):
...
-def test_get_engine_sqlite(mocker):
+def test_get_engine_sqlite(tmp_path, mocker):
+    """Test that ``get_engine`` creates a SQLite engine.
+
+    :param tmp_path: Temporary directory provided by pytest.
+    :param mocker: Pytest-mock fixture.
+    """
     mocker.patch("autosubmit.config.basicconfig.BasicConfig.DATABASE_BACKEND", "sqlite")
     mocker.patch("autosubmit.config.basicconfig.BasicConfig.DATABASE_CONN_URL", None)
-    engine = get_engine(db_path=":memory:")
```
ntorqulu, line 43: Thanks! I didn't know the file was created here!

autosubmit/database/session.py
ntorqulu: Should be taken into account when doing the postgres migration and the standardization of the db layer, as there're a lot of files (`db_common.py`, `db_manager.py` etc...) that still use the string comparison.
Thanks for creating the enum!

test/unit/database/test_session.py
```diff
@@ -37,15 +44,34 @@ def test_resolve_engine(url: str, expected: str | Exception):
...
+    engine = get_engine(db_path=database_path)
+
+    try:
+        assert engine.name == "sqlite"
+    finally:
+        engine.dispose()
 
 
 @pytest.mark.postgres
 def test_get_engine_postgres(mocker):
```
ntorqulu, line 70: This tests leaves `session._postgres_engine` and `_postgres_connection_url` filled for the rest of the worker process.
This can be a problem, because it makes tests order-dependent.
As nothing resents this module globals, the results of the tests will depend on the execution order. And as we ran the tests with `-n auto` the order is not deterministic.


After calling `get_engine` the module holds the info stored.
So imagine the following test execution:
- `test_get_engine_postgres`: writes `_postgres_engine = Engine(postgresql://user:***@localhost:5432/autosubmit)` and `_postgres_connection_url = "postgresql://user:pass@localhost:5432/autosubmit"`
- after this, `test_first_postgres_call_creates_engine` is executed.
```
with _postgres_lock:
        new_engine = _resolve_engine(connection_url) 
        ...
    return _postgres_engine    
```
The if guard is false because the previous test stored a value for _postgres_engine and _postgres_connection_url.
So `_resolve_engine` is not called and `old_engine.dispose.assert_called_once_with()` reports that it was called 0 times.

autosubmit/database/session.py
```diff
@@ -39,40 +52,70 @@ def _resolve_engine(connection_url: str) -> Engine:
...
+            _postgres_engine = new_engine
+            _postgres_connection_url = connection_url
+
+            if old_engine is not None:
+                old_engine.dispose()
+
+        return _postgres_engine  # type: ignore
 
 
 def get_engine(db_path: str | Path) -> Engine:
```
ntorqulu, line 97: For PostgreSQL the db_path is ignored. And the tests need to provide a dummy value:
`get_engine(db_path="dummy_path")`
But this value has no meaning for PostgreSQL...
Maybe we can make it optional?
`def get_engine(db_path: str | Path | None = None) -> Engine:`
Ignore me if there're too many callers to update.

test/unit/database/test_session.py
```diff
@@ -37,15 +44,34 @@ def test_resolve_engine(url: str, expected: str | Exception):
...
+    :param mocker: Pytest-mock fixture.
+    """
     mocker.patch("autosubmit.config.basicconfig.BasicConfig.DATABASE_BACKEND", "sqlite")
     mocker.patch("autosubmit.config.basicconfig.BasicConfig.DATABASE_CONN_URL", None)
-    engine = get_engine(db_path=":memory:")
-    assert engine.name == "sqlite"
+
+    database_path = tmp_path / "autosubmit.db"
+
+    assert database_path.parent.exists()
```
ntorqulu, line 58: already defined one line above, it'll always be `tmp_path` (true) no?

ntorqulu (commented):
Hi @kinow, done a quick review.
Thanks for improving the coverage and the codecov problem!
I've not found any bugs in the implementation, my only concern is in the tests and the persistence values of the module attributes across tests.
As pytest does not re-import modules between tests in the same process, whatever on test writes in the module attributes will stay visible to every later test in the same worker process, which can cause tests to fail intermittently.

I would create some sort of fixture the resets `_postgres_engine` and `_postgres_connection_url` globals for every test in the module.

test/unit/database/test_session.py
kinow: I noticed it being created and displayed in my git status. Not sure if this is the only place 😅 ...

test/unit/database/test_session.py
kinow: I thought the unit tests would be running everything, with any markers, but I will check tomorrow to confirm :+1: (so that we can write tests for postgres that do not require a running DB, like this engine unit test)

test/unit/database/test_session.py
kinow: Ah, an interesting problem. Will think about this one. My preference would be for a solution that doesn't involve changing the "production" code just so our tests pass (i.e., we design for environments/use cases/features... not for unit/integration tests to simply pass). Thanks Núria!

autosubmit/database/session.py
kinow: That was to satisfy the [protocol](https://typing.python.org/en/latest/spec/protocol.html) where the path was always passed as in the previous design (this came from the time we started adding postgres + sqlalchmey, while keeping sqlite running... i.e. we could'nt change all the existing API calls).

test/unit/database/test_session.py
kinow: Will confirm tomorrow, but what this was doing was asserting just as a way to show that that parent HAD to exist.

There is an if in the code that checks if the database parent path exists, and if not, creates it -- if I recall correctly.

So there is one unit test that doesn't care about the parent, and one that does (this, I believe). Just so we cover all branches in that conditional.

test/unit/database/test_session.py
ntorqulu: Yes, looking at the `ci.yaml` file, the units test are run as:
```
pytest \
  -n $(nproc) \
  --verbose --verbosity="2" \
  test/unit \
  -m "not postgres and not profile and not profilelong"
```
So it excludes tests marked with `postgres`, `profile` or `profilelong`.

test/unit/database/test_session.py
ntorqulu: We could define a fixture that resets the postgres states, this way we make `_postgres_engine` and `_postgres_connection_url` as `None` for every test.

kinow:
@ntorqulu I moved it back to draft until I have time to finish updating the tests and reading it once again with calm on Monday.

I searched where the engines are created. Mainly on times db, metrics, jobs data, and autosubmit db. And apparently they are created once for each DB (which makes sense).

I'm deleting the whole singleton here, and documenting that callers/users must update/discard engines.

Then, while working on Postgres over next months until the deliverable in March, I will verify that it's indeed safe and we don't have any places where we are creating multiple engine instances in a loop or something like that.

If so, we can clean that up directly where it's done (assuming it makes sense to keep that behaviour), or modify it so only one instance exists. It shouldn't be allowed to users to modify the DB engine, as otherwise we could have one part of jobs updated in one postgres DB, and the other in another one... or half of the metrics in one place, and the rest in another.

So no need to review it yet :-) thanks!

PR #3253, opened 2026-09-09
Add a test to verify Job serialisation and recursion

### LINKED ISSUE #3197: autosubmit 4.1.16 recursion error
### Description

An experiment in Climate DT is giving the following error:
```
  File "/usr/lib64/python3.9/multiprocessing/queues.py", line 244, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/usr/lib64/python3.9/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
  File "/appl/AS/4.1.16.1/lib64/python3.9/site-packages/autosubmit/job/job.py", line 164, in __getstate__
    return dict([(k, getattr(self, k, None)) for k in self.__slots__ if k not in EXCLUDED])
RecursionError: maximum recursion depth exceeded
```
This does not cause the job to fail, but it still seems like a bug. 

* Autosubmit version: 4.1.16
* Machine/VM/environment name (if applicable): Climate DT
* Experiment ID (if applicable): a3hx

This is what @kinow already checked:
>Managed to craft a test that raises the RecursionError as above in 4.1.16.1, then tried on 4.1.17 tag, and the same happened.
>On master, 4.2.0.dev now, the exception did not happen.
>It occurs due to cyclic dependency in Python objects, during de/serialisation. Dani changed serialisation in 4.2.0, which seems to have fixed this error.

kinow:
Closes #3197 

@TracyMcBean this was fixed in the `master` branch, which contains code for 4.2.0. Versions 4.1.16.1 and 4.1.17.1 both still have this issue. This pull request is just to prevent it from happening in 4.2.0 or future versions of Autosubmit.

While it may seem harmless, it's hard to confirm it's not hiding any strange issue as the recursion error could cause a job to mysteriously have a wrong state, for instance.

**Check List**

- [ ] Documentation updated.

### DIFF
--- a/test/unit/job/test_job.py
+++ b/test/unit/job/test_job.py
@@ -1,0 +1,59 @@
+# Copyright 2015-2026 Earth Sciences Department, BSC-CNS
+#
+# This file is part of Autosubmit.
+#
+# Autosubmit is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Autosubmit is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Autosubmit.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Unit tests for ``autosubmit.job.job`` module."""
+
+import pickle
+from typing import TYPE_CHECKING
+
+from autosubmit.job.job import Job
+
+if TYPE_CHECKING:
+    from _pytest.monkeypatch import MonkeyPatch
+
+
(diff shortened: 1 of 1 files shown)


kinow:
Build passed (finally) @VindeeR . No codecov comment though, but ready for review :sweat_smile:

test/unit/job/test_job.py
```diff
@@ -0,0 +1,59 @@
...
+    and serialisation succeeds.
+
+    This test was manually run on 4.1.16.1, where it failed to run, and also
+    on 4.2.0 (commit 54058a7d2fe5e43731f124b096deaed677e5ada8) successfully.
+    """
+    job = Job(name="sim_parent_20200101_001_1", job_id=1, status=3)
+
+    class RecursiveAttr:
+        def __get__(self, instance, owner):
+            return instance.rerun_only
```
kinow, line 55: This pretends that a property points to itself, causing the `RecursionError`. I can't tell if that's really what happened in the operational workflow, but that's my best guess. Somewhere we have a property that has a recursive call, nested somewhere.

VindeeR (approved):
This test looks good to go to me

kinow:
Thanks!

PR #3254, opened 2026-09-09
fix: clean statistics files from stats directory

### LINKED ISSUE #3199: autosubmit clean is not cleaning the stat files
Found by @ntorqulu in #3171 

> I was testing the clean command for stats and plots.
> 
>     For plots it correctly deletes old plots except the last two.
>     For stats it's currently not deleting any file, I think the problem is in _delete_stats_files_but_two_newest that searches always for the plot dir.
> 
> So in summary, clean is not cleaning the stat files; only the plots.

Ha1baraA11:
## Summary

Fixes #3199.

Statistics files are written under `<expid>/stats/`, but `autosubmit clean --stats` was scanning `<expid>/plot/`. Make the cleanup helper receive its target directory, so statistics cleanup retains the two newest files in `stats/` without touching plots.

## Testing

- `pytest -n 0 test/unit/monitor/test_monitor.py -q` — 64 passed
- Ruff, compile and diff checks passed
- The focused integration clean tests are blocked locally before assertions by an existing macOS multiprocessing database-path failure; the unmodified baseline fails at the same point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ several bug fixes and enhancements to improve the overall user experience.
 - Remove the unrelated `--filter_status` option from `autosubmit expid` #3241
 - Fix timeout guard is silently disabled for login/local jobs #3081
 - Fix CI ruff lint job failing on deleted files or single-commited branches #3166
+- Fix `clean` command to correctly delete files with `--stats` and `--plots` #3254 (thanks @Ha1baraA11)
 
 **New Features:**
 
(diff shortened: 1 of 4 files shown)


kinow (approved):
Tested locally, and the result were 2 PDFs in plot after `--plot`, and 2 files (1 PDF and 1 CSV) in the stats folder using `--stats`.

Then, I ran `monitor` and `stats` a few more times, and tested a call with both `--stats --plot`, and again 2 PDFs in plots, and 1 CSV and 1 PDF in the folders. Running again didn't delete anything.

I verified, and the files that remained were the most recently created.

Thus, everything is following the documentation of what the command is supposed to do.

Thanks for the tests too, @Ha1baraA11 . LGTM!

kinow:
Only added the changelog entry (and used the opportunity to add the missing one for @martiortis pull request yesterday, meaculpa). Merging. Thanks for the PR @Ha1baraA11

PR #3256, opened 2026-09-10
Fix split grouping mutating the job list

### LINKED ISSUE #3238: Create_splits_groups depletes self.jobs and the list must be restored afterwards
## Problem

In `JobGrouping._automatic_grouping` (autosubmit/job/job_grouping.py:~252) calling
`self._create_splits_groups()` apparently consumes/depletes `self.jobs`, so the code
must restore it right after:
    split_groups, split_groups_status = self._create_splits_groups()
    self.jobs = self.job_list.job_list   # restore, otherwise the group is left empty

## Where
- autosubmit/job/job_grouping.py:252 (inline TODO "depletes the self.jobs")

## Expected
`_create_splits_groups` should not consume `self.jobs` (copy internally), removing the
need for the restore line.

## Suggested
Investigate why the method depletes the list and fix it

1cbyc:
Closes #3238.

`_create_splits_groups()` passed the shared job list to `_create_groups()`, which removes grouped jobs in place. Automatic grouping then restored `self.jobs` by assigning `job_list.job_list`, making correctness depend on that separate alias.

This change performs split grouping on a temporary shallow copy and restores the original list object in a `finally` block. The caller-side repair is no longer needed. A focused regression verifies both list identity and contents after split grouping.

Validation:

- `pytest test/unit/test_job_grouping.py -q` (21 passed)
- `ruff check autosubmit/job/job_grouping.py test/unit/test_job_grouping.py`
- `python -m build`
- Full unit run: 2,394 passed, 41 skipped; three unrelated environment failures (two local-clone failures caused by the partial checkout's unavailable promisor object, and one CLI lookup failure because the virtualenv bin directory was not on `PATH`)

Co-authored-by: insisong <emmanuelisaacnsisong@gmail.com>

### DIFF
--- a/autosubmit/job/job_grouping.py
+++ b/autosubmit/job/job_grouping.py
@@ -28,10 +28,10 @@ if TYPE_CHECKING:
 
 class JobGrouping:
 
-    def __init__(self, group_by: str, jobs: list["Job"], job_list: "JobList", expand_list: list | None = None,
-                 expanded_status: list | None = None):
+    def __init__(self, group_by: str, jobs: list["Job"], job_list: "JobList", expand_list: str | None = None,
+                 expanded_status: list[int] | None = None):
         if not expand_list:
-            expand_list = []
+            expand_list = ""
         if not expanded_status:
             expanded_status = []
 
@@ -42,17 +42,17 @@ class JobGrouping:
         self.expand_list = expand_list
         self.expand_status = expanded_status
         self.automatic = False
-        self.group_status_dict = {}
-        self.ungrouped_jobs = []
+        self.group_status_dict: dict[str, Any] = {}
+        self.ungrouped_jobs: list[str] = []
 
     def group_jobs(self) -> dict[str, Any]:
         if self.expand_list:
             self._set_expanded_jobs()
 
-        jobs_group_dict = {}
-        blacklist = []
+        jobs_group_dict: dict[str, list[str]] = {}
+        blacklist: list[str] = []
 
-        groups_map = {}
+        groups_map: dict[str, str] = {}
         if self.group_by == 'automatic':
             self.automatic = True
             jobs_group_dict = self._automatic_grouping(groups_map)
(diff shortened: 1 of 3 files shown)


1cbyc:
Updated the branch to resolve the MyPy failures without changing the grouping behavior:

- corrected the `expand_list` input type to match the string values used by callers
- added concrete annotations for the grouping collections

Validation: the exact CI MyPy command passes, Ruff passes, and all 21 focused job-grouping tests pass.

1cbyc:
The final Python 3.12 integration shard completed successfully (`633 passed, 25 skipped, 5 xfailed`) and uploaded coverage, but GitHub marked the job `cancelled` after every step had completed successfully. All other checks are green. I cannot rerun upstream workflows without repository-admin access.

kinow:
That's  OK @1cbyc . Whenever the workflows fail one of the AS maintainers get a notification, and we review the changes and kick the jobs :-) I just did so. Thanks!

autosubmit/job/job_grouping.py
```diff
@@ -276,10 +276,14 @@ def _automatic_grouping(self, groups_map: dict[str, str]) -> dict[str, list[str]
...
-        self.group_by = 'split'
-        self._create_groups(jobs_group_dict, [])
+        jobs_group_dict: dict[str, list[str]] = {}
+        jobs = self.jobs
+        self.jobs = jobs.copy()
+        try:
+            self.group_by = 'split'
+            self._create_groups(jobs_group_dict, [])
+        finally:
+            self.jobs = jobs
```
kinow: Note to self: review later if there's any other way to fix the issue without copying `self.jobs` (@1cbyc  it's a bank holiday in Catalonia today, so the whole team is offline 👍 ).

kinow (commented):
Everything looks good, only added a note about one part of the code to review later. Thanks @1cbyc ! (it's holiday here! 🏖️ )

Issue #3209, opened 2026-08-31
Autosubmit runs a workflow, API creates perf stats reading AS data, then AS asks API the results...


kinow:
cc @pablogoitia , @ntorqulu , found while reviewing code in #3171 

Autosubmit runs the workflow, which produces some data, performing as well as it can.

The AS API then reads the data and creates the performance metrics. I am not sure if the code for that is entirely in the API or if it's in AS. Especially, because `autosubmit stats` probably creates the same metrics? (Also, another thing that is still pending to be confirmed.)

<img width="1026" height="567" alt="Image" src="https://github.com/user-attachments/assets/29c7113b-43a7-4c3e-98d6-bc816d185d51" />

Autosubmit 4 has always had a dependency on `requests`. There is some part of the code that tries to query the API to fetch the performance metrics for an experiment. That's used in the `autosubmit report` code.

So, AS will fetch the performance data from the API to report to the user. The performance data that comes from AS. But AS requests it from the API. I think only if the data is very expensive to be computed it'd make sense to store in the API.

But I think it'd make more sense if the code belonged to AS, and was imported by the API. Maybe this same code would be used by autosubmit stat, so users without the API (KIT, GANANA, Hanami external users, EDITO without the API kubernetes pod, etc.) would be able to access the same performance information.

Also, this means that any user without the API will simply never have those metrics reported in the `autosubmit report` code.

That way, we would be able to also remove the `requests` dependency, and have the code a bit simpler/easier to maintain, IMHO.

PR #3258, opened 2026-09-10
Pt2: Refactor experiment status traceability (Pt 1: https://github.com/BSC-ES/autosubmit/pull/2980)

### LINKED ISSUE #2049: Write keep alive signal in the DDBB

### LINKED ISSUE #2137: Extend experiment possible statuses to determine ARCHIVE, or DELETED state

ntorqulu:
Part 1 of this pr (was incorrectly closed): https://github.com/BSC-ES/autosubmit/pull/2980
Closes #2049, #2137
Related to https://github.com/BSC-ES/autosubmit/issues/2957: `RunningStatus` is now an enum

**Things done**
- Extended available experiment status in `experiment_status` table. Possible experiment status are: `NOT_RUNNING`, `RUNNING`, `ARCHIVED` and `DELETED`.
- Experiments created (`expid` and `create`) are marked as `NOT_RUNNING`. Implementation left open if we want to add more states in the future.
- Keep traceability of deleted experiments in both `autosubmit.db` and `as_times.db`. Once the experiment is deleted, it's kept in the database with status `DELETED` (tombstone).
- Added column `last_heartbeat` to table `experiment_status`. Each experiment creates a daemon thread (autosubmit-heartbeat-{expid}) that updates the `last_heartbeat` timestamp every 2 min while the experiment is running.
- Enforce unique indx on `name` in `experiment_status` table, kept most recent one.

**Check List**

- [x] Documentation updated.

### DIFF
--- a/autosubmit/database/db_common.py
+++ b/autosubmit/database/db_common.py
@@ -23,13 +23,14 @@ import sqlite3
 import subprocess
 from contextlib import suppress
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any
 
-from sqlalchemy import delete, func, insert, select, text, update
+from sqlalchemy import Connection, delete, func, insert, select, update
 from sqlalchemy.schema import CreateTable
 
 from autosubmit.config.basicconfig import BasicConfig
 from autosubmit.database import session, tables
+from autosubmit.history.experiment_status import ExperimentStatus
 from autosubmit.log.log import AutosubmitCritical, Log
 
 if TYPE_CHECKING:
@@ -318,14 +319,12 @@ def last_name_used(test=False, operational=False, evaluation=False):
     return result
 
 
-def delete_experiment(experiment_id):
+def delete_experiment(experiment_id: str) -> bool:
     """
-    Removes experiment from database. Anti-lock version.
+    Marks experiment as deleted in metadata tables. Anti-lock version.
 
     :param experiment_id: experiment identifier
-    :type experiment_id: str
-    :return: True if delete is successful
-    :rtype: bool
+    :return: True if operation is successful
     """
     fn = _delete_experiment
     if BasicConfig.DATABASE_BACKEND == 'postgres':
@@ -595,29 +594,24 @@ def _last_name_used(test=False, operational=False, evaluation=False):
     return row[0]
 
 
-def _delete_experiment(experiment_id):
+def _delete_experiment(experiment_id: str) -> bool:
     """
-    Removes experiment from database
+    Marks an experiment as deleted in database while preserving
+    the experiment row to avoid reusing assigned experiment IDs.
 
     :param experiment_id: experiment identifier
-    :type experiment_id: str
-    :return: True if delete is successful
-    :rtype: bool
+    :return: True if operation is successful
     """
-    check_db()
-
-    if not _check_experiment_exists(experiment_id, False):  # Reference the no anti-lock version.
+    if not _check_experiment_exists(
+        experiment_id, False
+    ):  # Reference the no anti-lock version.
         return True
     try:
-        (conn, cursor) = open_conn()
-    except DbException as e:
-        raise AutosubmitCritical("Could not establish a connection to database", 7001, str(e))
-    cursor.execute('DELETE FROM experiment '
-                   'WHERE name=:name', {'name': experiment_id})
-    row = cursor.fetchone()
-    if row is None:
-        Log.debug(f'The experiment {experiment_id} has been deleted!!!')
-    close_conn(conn, cursor)
+        ExperimentStatus(experiment_id).set_as_deleted()
+    except Exception as e:
+        raise AutosubmitCritical("Could not mark experiment as deleted", 7001, str(e))
+
+    Log.debug(f"The experiment {experiment_id} has been marked as deleted.")
     return True
 
 
@@ -860,30 +854,13 @@ def _delete_experiment_sqlalchemy(experiment_id: str) -> bool:
         experiment_id, False
     ):  # Reference the no anti-lock version.
         return True
+    try:
+        ExperimentStatus(experiment_id).set_as_deleted()
+    except Exception as e:
+        raise AutosubmitCritical("Could not mark experiment as deleted", 7001, str(e))
 
-    with _get_sqlalchemy_conn() as conn:
-        with conn.begin():
-            # Delete from experiment table
-            query = delete(tables.ExperimentTable).where(
-                tables.ExperimentTable.c.name == experiment_id  # type: ignore
-            )
-            result = conn.execute(query)
-
-            # Drop schema
-            conn.execute(text(f'DROP SCHEMA IF EXISTS "{experiment_id}" CASCADE'))
-
-            # Delete from experiment_status table
-            try:
-                query = delete(tables.ExperimentStatusTable).where(
-                    tables.ExperimentStatusTable.c.name == experiment_id  # type: ignore
-                )
-                conn.execute(query)
-            except Exception as e:
-                Log.debug(f"The experiment {experiment_id} has no status: {str(e)}")
-
-        if cast(int, result.rowcount) > 0:
-            Log.debug(f"The experiment {experiment_id} has been deleted!!!")
-        return True
+    Log.debug(f"The experiment {experiment_id} has been marked as deleted.")
+    return True
 
 
 def _get_experiment_id_sqlalchemy(name: str) -> int:
--- a/autosubmit/database/session.py
+++ b/autosubmit/database/session.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2025 Earth Sciences Department, BSC-CNS
+# Copyright 2015-2026 Earth Sciences Department, BSC-CNS
 #
 # This file is part of Autosubmit.
 #
@@ -42,7 +42,7 @@ def _resolve_engine(connection_url: str) -> Engine:
 class PostgreSQLEngineSingleton:
     """Singleton class to manage a single instance of the PostgreSQL engine."""
 
-    _instance: Engine = None
+    _instance: Engine | None = None
     _lock: threading.Lock = threading.Lock()
 
     @classmethod
@@ -53,6 +53,14 @@ class PostgreSQLEngineSingleton:
                 connection_url = BasicConfig.DATABASE_CONN_URL
                 cls._instance = _resolve_engine(connection_url)
         return cls._instance
+
+    @classmethod
+    def reset(cls) -> None:
+        """Dispose the cached engine if it exists, so a new one is built on next use."""
+        with cls._lock:
+            if cls._instance is not None:
+                cls._instance.dispose()
+                cls._instance = None
 
 
 def get_engine(db_path: str | Path) -> Engine:
--- a/autosubmit/database/tables.py
+++ b/autosubmit/database/tables.py
@@ -61,6 +61,7 @@ ExperimentStatusTable = Table(
     Column("status", Text, nullable=False),
     Column("seconds_diff", Integer, nullable=False),
     Column("modified", Text, nullable=False),
+    Column("last_heartbeat", Text, nullable=True),
 )
 """Stores the status of the experiments."""
 
@@ -153,7 +154,7 @@ JobDataTable = Table(
 
 """All these tables will go inside the $expid/db/job_list.db."""
 # Jobs table
-"""Table that holds the minium neccesary info about the experiment jobs."""
+"""Table that holds the minium necessary info about the experiment jobs."""
 JobsTable = Table(
     "jobs",
     metadata_obj,
--- a/autosubmit/experiment/manage.py
+++ b/autosubmit/experiment/manage.py
@@ -64,6 +64,7 @@ from autosubmit.helpers.version import get_version
 from autosubmit.history.experiment_history import (
     ExperimentHistory,
 )
+from autosubmit.history.experiment_status import ExperimentStatus
 from autosubmit.job.job_common import get_job_status
 from autosubmit.job.job_grouping import JobGrouping
 from autosubmit.job.job_list import JobList, load_job_list
@@ -335,6 +336,14 @@ def expid_fn(
         )
         Log.debug(f"Error calling save_update_details: {str(e)}")
 
+    try:
+        ExperimentStatus(exp_id).set_as_not_running()
+    except Exception as e:
+        Log.warning(
+            f"Could not update the status of experiment {exp_id}. "
+            f"The status will be fixed by the API worker. Error: {str(e)}"
+        )
+
     Log.result(f"Experiment {exp_id} created")
     return exp_id
 
@@ -441,13 +450,6 @@ def _delete_experiment(expid: str, force: bool) -> None:
 
     Log.info(f"Deleting experiment {expid}")
 
-    # Try to delete the experiment details
-    try:
-        ExperimentDetails(expid).delete_details()
-    except Exception as e:
-        Log.warning(f"Failed to delete DB details for experiment {expid}: {str(e)}")
-        raise
-
     try:
         _delete_expid(expid, force)
         Log.info(f"Experiment {expid} has been deleted")
@@ -458,7 +460,7 @@ def _delete_experiment(expid: str, force: bool) -> None:
 
 
 def _delete_expid(expid_delete: str, force: bool = False) -> None:
-    """Removes an experiment from the path and database.
+    """Removes an experiment from the path and marks it as deleted in the database.
 
     If the current user is eadmin and the -f flag has been sent, it deletes regardless of experiment owner.
 
@@ -517,7 +519,7 @@ def _delete_expid(expid_delete: str, force: bool = False) -> None:
             )
 
     message_parts = [
-        f"The {expid_delete} experiment was removed from the local disk and from the database.",
+        f"The {expid_delete} experiment was removed from local disk and marked as deleted in the database.",
         "Note that this action does not delete any data written by the experiment.",
         "Complete list of files/directories deleted:",
         "",
@@ -565,13 +567,6 @@ def _perform_deletion(
 
     is_sqlite = BasicConfig.DATABASE_BACKEND == "sqlite"
 
-    Log.info(f"Deleting experiment from {BasicConfig.DATABASE_BACKEND} database...")
-    try:
-        db_common.delete_experiment(expid_delete)
-        Log.result(f"Experiment {expid_delete} deleted from database")
-    except Exception as e:
-        error_message.append(f"Cannot delete experiment entry: {e}")
-
     Log.info("Removing experiment directory...")
     try:
         rmtree(experiment_path)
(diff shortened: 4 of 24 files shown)


ntorqulu:
### Important things to remember

1. Possible experiment statuses implemented in this pr
image

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -33,24 +36,72 @@
...
+        autosubmit_db_path = db_dir / "test.db"
+        with sqlite3.connect(autosubmit_db_path) as conn:
+            cursor = conn.cursor()
+            cursor.execute(
+                """
+                CREATE TABLE experiment (
+                    id INTEGER PRIMARY KEY,
+                    name TEXT NOT NULL,
+                    description TEXT NOT NULL,
+                    autosubmit_version TEXT
```
ntorqulu, line 68: **From Bruno's review:**

Maybe the test should do the same as the code, instead of writing the DDL here.

The install command calls db_common.py, which reads the autosubmit.sql to create the DB. You can call the db_common#create_db function passing the same sql file. Having the extra stuff in that file shouldn't matter. And that way, if one changes create_db, they can find other places where this is used and update...

For the DB + path, there should be something in conftext that may be reused here to create DB + initial DDL, maybe. Or maybe we could have a fixture just to create an empty DB 🤔 for later I guess.

test/integration/test_expid.py
```diff
@@ -869,6 +870,50 @@ def test_delete_experiment(mocker, tmp_path, autosubmit_exp):
...
+    assert 'Failed to delete experiment' in mocked_log.error.call_args_list[0][0][0]
+
+    # test eadmin force
+    delete_experiment(expids=f'{as_exp.expid}', force=True)
+    assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}").iterdir())
+    assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/data").iterdir())
+    assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/logs").iterdir())
+    assert all(as_exp.expid not in Path(f).name for f in Path(f"{run_dir}/metadata/structures").iterdir())
+    # Consult if the expid is still in the database (tombstone)
+    db_path = Path(BasicConfig.DB_PATH)
```
ntorqulu, line 910: **From Bruno's review**

Why use the f-string with run_dir, and `tests.db` here, instead of the `DB_PATH` constant that was used before? I think both point to the same location?

Image

test/integration/test_expid.py
```diff
@@ -869,6 +870,50 @@ def test_delete_experiment(mocker, tmp_path, autosubmit_exp):
...
+    with sqlite3.connect(db_path) as conn:
+        cursor = conn.cursor()
+        cursor.execute(f"SELECT name FROM experiment WHERE name='{as_exp.expid}'")
+        assert cursor.fetchone() is not None
```
ntorqulu, line 914: **From Bruno's review**

Maybe we can drop the cursor, and close the conn (not only the transaction).

```
# from contextlib import closing
    db_path = Path(BasicConfig.DB_PATH)
    with closing(sqlite3.connect(db_path)) as conn:
        row = conn.execute(
            f"SELECT name FROM experiment WHERE name='{as_exp.expid}'"
        ).fetchone()
    assert row is not None
```

But not important.

test/integration/conftest.py
```diff
@@ -539,12 +540,23 @@ def as_db(request: "FixtureRequest", tmp_path: "LocalPath", postgres_server: "Do
         raise ValueError(f'Unsupported database backend: {backend}')
 
     BasicConfig.read()
+    if backend == "postgres":
+        PostgreSQLEngineSingleton.reset()
```
ntorqulu, line 544: **From Bruno's review**

Argh. I think we can get rid of this singleton completely. For another issue. 👍

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -89,53 +140,342 @@ def test_experiment_status_db_manager(tmp_path: 'LocalPath', as_db: str, use_sql
...
+        as_times_db_path = db_dir / "as_times.db"
+        with sqlite3.connect(as_times_db_path) as conn:
+            cursor = conn.cursor()
+            cursor.execute("""
+                CREATE TABLE experiment_status (
+                    exp_id INTEGER PRIMARY KEY,
+                    name TEXT NOT NULL,
+                    status TEXT NOT NULL,
+                    seconds_diff INTEGER NOT NULL,
+                    modified TEXT NOT NULL
```
ntorqulu, line 219: **From Bruno's review**

Probably better to use the same mechanism AS/conftest use to create the DB. To avoid issues in the tests if/when we change that.

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -89,53 +140,342 @@ def test_experiment_status_db_manager(tmp_path: 'LocalPath', as_db: str, use_sql
...
+                )
+                """)
+            conn.commit()
+
+        # Initialize the manager
```
ntorqulu, line 224: **From Bruno's review**

s/z/s 😬 so we keep our comments/codebase closer to British or Euro English.

test/integration/history/database_managers/test_experiment_status_db_manager.py
```diff
@@ -33,24 +36,72 @@
...
+                CREATE TABLE experiment (
+                    id INTEGER PRIMARY KEY,
+                    name TEXT NOT NULL,
+                    description TEXT NOT NULL,
+                    autosubmit_version TEXT
+                )
+                """
+            )
+            cursor.executemany(
+                "INSERT INTO experiment (id, name, description, autosubmit_version) VALUES (?, ?, ?, ?)",
```
ntorqulu, line 73: **From Bruno's review**

And this one maybe we can use db_common's _save_experiment? Although you control the ID here, so maybe we can't use that function directly? Or maybe we could rely on autosubmit_exp for both SQLite and PG?

autosubmit/database/db_common.py
```diff
@@ -595,29 +594,24 @@ def _last_name_used(test=False, operational=False, evaluation=False):
...
-        (conn, cursor) = open_conn()
-    except DbException as e:
-        raise AutosubmitCritical("Could not establish a connection to database", 7001, str(e))
-    cursor.execute('DELETE FROM experiment '
-                   'WHERE name=:name', {'name': experiment_id})
-    row = cursor.fetchone()
-    if row is None:
-        Log.debug(f'The experiment {experiment_id} has been deleted!!!')
-    close_conn(conn, cursor)
+        ExperimentStatus(experiment_id).set_as_deleted()
```
ntorqulu, line 610: **From Bruno's review**

We are recreating an ExperimentStatus for every operation. The object seems to reload the BasicConfig, and create the database manager. Maybe we can improve this, somehow, to avoid recreating it in case we have a large simulation and this operation occurs a lot?

autosubmit/database/session.py
```diff
@@ -54,6 +54,14 @@ def get_instance(cls) -> Engine:
                 cls._instance = _resolve_engine(connection_url)
         return cls._instance
 
+    @classmethod
+    def reset(cls) -> None:
+        """Dispose the cached engine if it exists, so a new one is built on next use."""
+        with cls._lock:
+            if cls._instance is not None:
+                cls._instance.dispose()
+                cls._instance = None
```
ntorqulu, line 63: **From Bruno's review**

For another issue, but I think we could have something like this,

```py
_postgres_engine: Engine | None = None
_postgres_connection_url: str | None = None
_postgres_lock = threading.Lock()


def _get_postgres_engine(connection_url: str) -> Engine:
    global _postgres_engine, _postgres_connection_url

    with _postgres_lock:
            if _postgres_engine is not None:
                _postgres_engine.dispose()

            _postgres_engine = _resolve_engine(connection_url)
            _postgres_connection_url = connection_url

        return _postgres_engine
```

In Python, it's common devs with OOP from Java, JavaScript to create classes and instantiate things. But in reality, modules are also objects, created during module import/first use.

We can use that if we don't need class-hierarchy, OOP, etc...

So here the "singleton" is replaced by a global variable. We lock/create a barrier, then check if it exists and the URL changed. If so, dispose. If it does not exist, create a new. Or... well, just return the engine (le singleton 🇫🇷 😆 ).

I think that may work, and simplify the code (but it's Friday, and I cannot really trust my 🧠 so close to the weekend). I will create a separate issue later.

autosubmit/database/tables.py
```diff
@@ -61,6 +61,7 @@
     Column("status", Text, nullable=False),
     Column("seconds_diff", Integer, nullable=False),
     Column("modified", Text, nullable=False),
+    Column("last_heartbeat", Text, nullable=True),
```
ntorqulu, line 64: **From Bruno's review**

My sqlalchemy-fu is quite rusty 🙇

Can't we use DATETIME? I think SqlAlchemy supports it for sqlite too, https://docs.sqlalchemy.org/en/21/dialects/sqlite.html#sqlalchemy.dialects.sqlite.DATETIME

Even though in SQLite it will be probably stored as text anyway. Maybe it could be the last_heartbeat in UTC time? 😬 Not sure if this needs to be in UTC, or if for the heartbeat it'd make sense to have the system TZ.

autosubmit/experiment/manage.py
```diff
@@ -592,6 +587,17 @@ def _perform_deletion(
...
+    # Mark experiment status as deleted only after cleanup successful
+    if not error_message:
+        Log.info(
+            f"Updating experiment status in {BasicConfig.DATABASE_BACKEND} database..."
+        )
+        try:
+            db_common.delete_experiment(expid_delete)
+            Log.result(f"Experiment {expid_delete} marked as deleted in database")
+        except Exception as e:
+            error_message.append(f"Cannot update experiment metadata: {e}")
```
ntorqulu, line 599: **From Bruno's review**

Not related to your change directly, but I think this is a sort of XA/2PC (two-phase commit) problem, @ntorqulu -- https://en.wikipedia.org/wiki/Two-phase_commit_protocol

What AS does at the moment, I think, is remove the folders, and if that succeeds, then open a transaction and delete or mark as deleted (now) in the DB.

But if the DB step fails, then now the files were already removed. Not important as this is already possible to happen. I will test to confirm and open an issue for that later, some day 😬

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -55,31 +54,51 @@ def __init__(
...
+                modified text NOT NULL,
+                last_heartbeat text
+            );""")
         self.execute_statement_on_dbfile(self._as_times_file_path, create_table_query)
 
-    def set_existing_experiment_status_as_running(self, expid: str) -> None:
-        """ Set the experiment_status row as running. """
-        self.update_exp_status(expid, Models.RunningStatus.RUNNING)
+        # backward compatibility
+        self._add_column_if_missing("last_heartbeat", "text")
```
ntorqulu, line 70: **From Bruno's review**

It'll be great if someday we can fix https://github.com/BSC-ES/autosubmit/issues/1286 so that these migrations are more clearly separated.

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -55,31 +54,51 @@ def __init__(
...
+
+        # keep only latest row by name
+        self.execute_statement_on_dbfile(
+            self._as_times_file_path,
+            """DELETE FROM experiment_status
+               WHERE exp_id NOT IN (
+                   SELECT MAX(exp_id)
+                   FROM experiment_status
+                   GROUP BY name
```
ntorqulu, line 79: **From Bruno's review**

Is this needed for the API? I think not doing this doesn't break anything, but it's better to keep only the last one as the other ones are not used?

Looking at the as_times.db that I have, @ntorqulu I seem to recall that it was going to be deprecated? At least that's what my memory tells me today 😅 but I might be wrong.

It looks like the original design was to log every experiment status change. Here we will change and keep the last status only. If so, wouldn't it make sense to be a column of experiment? It's a ONE-to-ONE relationship between experiment and experiment_status, and in DB normalisation I think that's normally an indicator that something belongs/is part of another table.

I had to google that to remember the name 😆 Boyce-Codd normal form, or normal forms -- https://en.wikipedia.org/wiki/Database_normalization. I remember having to normally go through 1->3NF at least. But I didn't recall there were that many forms.

But trivia apart, I think if this table now will have one row per experiment, then it could be merged?

autosubmit/history/database_managers/experiment_status_db_manager.py
```diff
@@ -55,31 +54,51 @@ def __init__(
...
+            alter_query = (
+                f"ALTER TABLE experiment_status ADD COLUMN {column_name} {column_type};"
+            )
+            self.execute_statement_on_dbfile(self._as_times_file_path, alter_query)
+
+    def _column_exists(self, path: str, column_name: str) -> bool:
+        """Check whether a column exists in the experiment_status table for SQLite."""
+        query = "PRAGMA table_info(experiment_status);"
+        current_columns = [row[1] for row in self.get_from_statement(path, query)]
+        return column_name in current_columns
```
ntorqulu, line 101: **From Bruno's review**

As this is Python, maybe we can simplify and [EAFP](https://docs.python.org/2/glossary.html#term-eafp)
```
def _add_column_if_missing(self, column_name: str, column_type: str) -> None:
    try:
        self.execute_statement_on_dbfile(
            self._as_times_file_path,
            f"ALTER TABLE experiment_status ADD COLUMN {column_name} {column_type}",
        )
    except sqlite3.OperationalError as e:
        if "duplicate column name" not in str(e):
            raise
```

ntorqulu (commented):
Copied comments from last week Bruno's review in pr https://github.com/BSC-ES/autosubmit/pull/2980

autosubmit/history/database_managers/experiment_status_db_manager.py
ntorqulu: Implemented it on the SQLite manager and the SQL Alchemy manager.  
Previously two initializations against the db could crash. 
Now it swallows the duplicate column error instead of crashing.
I'm still thinking where should this be called...

An `ExperimentStatus` object is constructed in `expid`, `create`, `archive`, `unarchive`, `_prepare_run`, `run` and `delete`. It's not per heartbeat as the monitor reuses the same ExperimentStatus object.
The migration is executed every time the manager is constructed.
So, the `last_heartbeat` migration (add column if not present) should be a one-time migration, running once.
But I don't know exactly where to do it @kinow.

kinow:
@ntorqulu over the weekend [1] I thought about a corner case: users without the API and GUI. Is there any chance that those users would be missing anything if the API is never executed? Because I remember the API was doing some sort of step to update AS database?

Asking just to check if we have to think about a workaround for `monitor` or some other subcommand for users without the API.

[^1]: no idea why my brain decided to bring this up over this weekend :beach_umbrella: :sweat_smile:

ntorqulu:
> Is there any chance that those users would be missing anything if the API is never executed?

For sure the experiment statuses updates would be missing.
Currently in `master` AS backend only takes care of putting experiment statuses to `RUNNING`. Is the API with a background task that updates the statuses to `NOT RUNNING` in batches every 5 minutes.
So users that do not have the API running will never have the experiment statuses updated correctly in the db.
This is mainly what this pr is trying to solve, let the AS backend be the owner of the db and the API should only be relegated to read it.

I think the same happened with `stats` (https://github.com/BSC-ES/autosubmit/issues/3209).
I'll check if there're more commands that are partially owned by the API; the goal would be that the backend is the one in charge to avoid issues in this case, where AS is only ran with the backend.

kinow:
>  I'll check if there're more commands that are partially owned by the API; the goal would be that the backend is the one in charge to avoid issues in this case, where AS is only ran with the backend.

Exactly. Maybe we will have to modify the database in those commands. The only thing that bothers me is that I think those commands (stats, monitor) read statuses of workflow/jobs, but I think they do not modify the experiments table/database (I think). Maybe we will need a configuration flag (in the future), or something. Or maybe I'm just overthinking it haha.

Thanks Núria!

autosubmit/history/database_managers/experiment_status_db_manager.py
ntorqulu: > It looks like the original design was to log every experiment status change. Here we will change and keep the last status only

The schema design on master already had the `exp_id` as primary key in the `experiment_status` table. I don't know when this constrain was introduced, but it was not enforced on the already-used dbs (like the one in the hubs). So at least the code in master already enforces one row per experiment.

> Is this needed for the API? I think not doing this doesn't break anything, but it's better to keep only the last one as the other ones are not used?

It's partially needed...I would like to avoid as much as possible that the API does write operations on the db. Currently is the one in charge of updating the experiment statuses from RUNNING -> NOT RUNNING via a `DELETE+INSERT`. So the API first deletes all rows with the same exp_id and then inserts the new row with the status change, avoiding this way duplicates.
What could be done in the backend is replicate the api behaviour to avoid having to delete experiment status rows on database validation.
So:
- for the schema preparation, in `_validate_status_database` we do a `CREATE TABLE IF NOT EXISTS` (as before) and `ALTER ... ADD COLUMN` to add the heartbeat column if not present.
- and for the experiment status updates, I think I will replicate the code in the API: only when the experiment status needs to be inserted or updated clean the duplicated rows of this experiment.

> wouldn't it make sense to be a column of experiment?

Mmm I would go against it.
- `as_times.db` is a database that currently *could* be overwritten by the API. We're moving to avoiding this behavior, but this is the reality. I would not mix a db that the API does not touch (`autosubmit.db`) with one in which it has freedom (`as_times.db`) . This way we separate concerns and possible data corruption.
- the backend will write into the database the `last_heartbeat` value for experiments that are RUNNING every 2 min + every status transition. SQLite blocks the whole file for writes. This way we avoid blocking `autosubmit.db`
- The PK is not guaranteed in `as_times.db`. It would be a difficult data migration...

autosubmit/history/database_managers/experiment_status_db_manager.py
ntorqulu: I'll add a todo in `_add_column_if missing` for when issue #1286 is implemented.

autosubmit/experiment/manage.py
ntorqulu: @kinow do you think is better:
1. delete  files first, mark db later -> first delete the experiment files, then update the experiment status
2. mark db first, delete files later -> first mark the experiment status as DELETED, then delete the experiment files.

I would go with option 2, as this way the irreversible step (deleting files) is kept as the last step, which keeps the operation retriable.

PR #3261, opened 2026-09-10
Fix platform exclusivity during job refresh

### LINKED ISSUE #3234: Confirm whether CURRENT_EXCLUSIVE should be CURRENT_EXCLUSIVITY (FIXME in job.py)
## Problem
`autosubmit/job/job.py:2127` has a FIXME:

"Should be CURRENT_EXCLUSIVITY instead of CURRENT_EXCLUSIVE to match the platform
parameter?". Autosubmit may be reading the wrong key when updating job resource
parameters, producing jobs without the requested exclusivity.

## Where
- autosubmit/job/job.py (update_platform_associated_parameters, ~line 2127)

## Expected
Determine the correct parameter name used by platforms (slurm/PJM/ec…) and read
that key; add/adjust a test covering the exclusivity parameter.

## Suggested
Check platform headers/parameters (CURRENT_EXCLUSIVITY) and the docs

1cbyc:
## Summary

- use the platform `CURRENT_EXCLUSIVITY` value when a job-specific `CURRENT_EXCLUSIVE` override is absent
- preserve the existing job-level override behavior
- cover both parameter sources in the job refresh regression

## Validation

- `.venv/bin/pytest test/unit/test_job.py -q` (229 passed)
- `.venv/bin/ruff check autosubmit/job/job.py test/unit/test_job.py`
- `.venv/bin/python -m build`

Closes #3234

Co-authored-by: insisong <emmanuelisaacnsisong@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ several bug fixes and enhancements to improve the overall user experience.
 
 **Bug fixes:**
 
+- Preserve platform exclusivity when refreshing job parameters #3234
 - Remove the unrelated `--filter_status` option from `autosubmit expid` #3241
 - Fix timeout guard is silently disabled for login/local jobs #3081
 - Fix CI ruff lint job failing on deleted files or single-commited branches #3166
--- a/autosubmit/job/job.py
+++ b/autosubmit/job/job.py
@@ -2071,8 +2071,9 @@ class Job:
             self.shape = parameters.get("CURRENT_SHAPE", "")
             self.processors_per_node = parameters.get("CURRENT_PROCESSORS_PER_NODE", "1")
             self.nodes = parameters.get("CURRENT_NODES", "")
-            # FIXME: Should be ``CURRENT_EXCLUSIVITY`` instead of ``CURRENT_EXCLUSIVE`` to match the platform parameter?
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ several bug fixes and enhancements to improve the overall user experience.
 
 **Bug fixes:**
 
+- Preserve platform exclusivity when refreshing job parameters #3234
```
kinow, line 11: I'd change this to something more clear to users, indicating which option(s) are available now, what changed. We can take care updating the changelog.

kinow (commented):
Thanks for the PR @1cbyc . While this would fix one part of the problem, users would still have to know in platforms they use EXCLUSIVITY and in jobs they use EXCLUSIVE but Autosubmit falls back to EXCLUSIVITY.

From my comment in the linked issue:

> Maybe another option could be we define the final parameter name (EXCLUSIVE or EXCLUSIVITY), use the approach from #3261, update our docs to explain which option users must use, and also print a warning log if the users use the deprecated form.

I'll bring this up with others in some discussion to see if there are other options too. Will hold back on merging this, but thanks again! 👍

PR #3267, opened 2026-09-14
wrappers-split-3106

### LINKED ISSUE #3106: Capability of having one vertical wrapper per split
Hi,

We work in a workflow that already has:
 - Dates
 - Splits

We would like to have a single vertical wrapper per split, the same way that is done when using multiple dates. 

For example:

 - Workflow with [2 dates and 1 split](https://climatedt-wf.csc.fi/experiment/a3pg/graph):

Image

- Workflow with [1 date and 3 splits](https://climatedt-wf.csc.fi/experiment/a3pf/graph):

Image

Is there a way of having the same isolation of vertical wrappers for the splits? I tried to set MAX_WRAPPED but it didn't wrap vertically in the same split.

We may need to use chunks and members in the future to be used inside climateDT workflow. So it may not be an option of using chunks instead of splits.

Thanks!

dbeltrankyl:
fixes #3106

* Unified vertical code
* Fixed issues with jobs inside splits not being wrap up 
* Fixed issues with multiple ready jobs merged together 

**Check List**

- [x] Documentation updated.


dbeltrankyl:
Hello @VindeeR

Do you mind looking at this PR?

This fixes the issue with the splits, and a bug discovered in the deadlock mechanism that you reviewed not so long ago.

Aside of what I'm fixing in the PR, I'm also separating the run and inspect-cw deadlock code to improve it ( memory vs db)

VindeeR (approved):
Changes looking good to me 🙌 
Thanks for simplifying the wrappers

PR #3177, opened 2026-08-12
1405 general comments on autosubmit documentation

rajeshsuraj:
Style Improvements and general improvements from the issue.
moving pages to improve flow within the doc navigation

Closes #1405 

**Check List**

- [x] I have read `CONTRIBUTING.md`.
- [ ] Contains logically grouped changes (else tidy your branch by rebase).
- [ ] Does not contain off-topic changes (use other PRs for other changes).
- [ ] Applied any dependency changes to `pyproject.toml`.
- [ ] Tests are included (or explain why tests are not needed).
- [ ] Changelog entry included in `CHANGELOG.md` if this is a change that can affect users.
- [x] Documentation updated.
- [ ] If this is a bug fix, PR should include a link to the issue (e.g. `Closes #1234`).

PR #3273, opened 2026-09-16
[documentation] Update installation page, list requirements dynamically

### LINKED ISSUE #2599: [documentation] Make it clearer to users when they have to run `autosubmit configure` and `autosubmit install`
See this issue for an example of how our documentation can currently confuse users: https://github.com/BSC-ES/autosubmit/issues/2598#issuecomment-3292505357 (sorry, @pgierz).

We should aim to make the docs clearer.

But maybe we could also improve the `autosubmit configure` and `autosubmit install`, telling the user something like "This will configure the environment for a new installation, but there is already an existing installation configured at /etc/autosubmitrc. Do you want to proceed?".

### LINKED ISSUE #3195: Improve Installation page for Autosubmit docs
Image

Image

Update the dependency to be updated dynamic and fix/update conda docs which is out dated

rajeshsuraj:
Improving the Installation page.

Closes #3195 
Closes #2599

- [x] Documentation updated.


rajeshsuraj:
Current commit cherry picked from #3177 .

rajeshsuraj:
- [x] read min/max Python versions from pyproject.toml
- [x] show the license in the page (users installing it probably can benefit from knowing the license of the software they are about to install; something like Autosubmit is licensed under GNU...)
- [x] it'd be good to list all the python dependencies needed directly in a table, with a table. It could be hidden by default, collapsed, but visible if expanded [^1]
- [x] in pyproject.toml, I believe we are free to add anything like [mycustomproperty...]. But there's an ongoing PEP, [PEP 725](https://peps.python.org/pep-0725/). Let's try to use that [^2].
- [x] read PEP 725 information and display system dependencies automatically from pyproject.toml in the page

pyproject.toml
```diff
@@ -260,3 +260,21 @@ ignore = [
...
+    "dep:generic/rsync",
+    "dep:generic/sqlite3",
+    "dep:generic/subversion",
+    "dep:generic/tk",
+]
+
+[tool.autosubmit.docs.system-dependencies]
+git = { check = "git --version" }
+graphviz = { note = "2.40 is known not to work", check = "dot -v" }
+tk = { note = "python-tk on Debian and Ubuntu, tkinter on CentOS" }
```
kinow: Missing EOL...

docs/source/installation/index.rst
```diff
@@ -280,41 +191,41 @@ Mandatory parameters of /etc/autosubmit
...
-   path = /home/dbeltran/autosubmit/metadata/structures
+   path = /home/<user>/autosubmit/metadata/structures
 
    [historicdb]
-   path = /home/dbeltran/autosubmit/metadata/data
+   path = /home/<user>/autosubmit/metadata/data
 
    [historiclog]
-   path = /home/dbeltran/autosubmit/metadata/logs
+   path = /home/<user>/autosubmit/metadata/logs
```
kinow, line 209: 👍 thanks!

docs/source/installation/index.rst
```diff
@@ -53,218 +38,144 @@ The sequence of instructions to install Autosubmit and its dependencies with pip
...
-----------------------------------------------------------------
+.. _full-setup:
 
 .. _Shared-Filesystem:
 
-.. warning:: Keep in mind the .autosubmitrc precedence. If you, as a user, have a .autosubmitrc generated in the quick-installation, you have to delete or rename it before using the production environment installation.
+Full setup
+==========
+
+For a production environment with a shared database in ``/etc/autosubmitrc``, letting multiple users share and view others' experiments.
```
kinow: This sentence is a bit disconnected no? Like, letting the users share and view... why/then/etc? Or is it to let users? It's a bit odd....

docs/source/installation/index.rst
```diff
@@ -53,218 +38,144 @@ The sequence of instructions to install Autosubmit and its dependencies with pip
...
+
+Precedence between configuration files:
```
kinow: "Priority order of configuration files" ? I think precedence is normally not so common in technical docs when I read about "etc/" files, etc?

docs/source/installation/index.rst
```diff
@@ -53,218 +38,144 @@ The sequence of instructions to install Autosubmit and its dependencies with pip
...
+.. code-block:: bash
 
     # Generate a key pair for password-less ssh. PEM format is recommended as others can cause problems
     ssh-keygen -t rsa -b 4096 -C "email@email.com" -m PEM
 
     # Copy the public key to the remote machine
     ssh-copy-id -i ~/.ssh/id_rsa.pub user@remotehost
 
-    # Add your key to ssh agent ( if encrypted )
+    # Add your key to the ssh-agent (if encrypted)
```
kinow, line 116: Thanks for tidying up this page (which is quite important for new users!) 🙇

docs/source/installation/index.rst
```diff
@@ -53,218 +38,144 @@ The sequence of instructions to install Autosubmit and its dependencies with pip
...
+    # Run the experiment
+    autosubmit run a000
 
-For a very quick test, you can follow the next instructions to configure and run Autosubmit at the user level. Otherwise, please go directly to `How to configure Autosubmit <https://autosubmit.readthedocs.io/en/master/installation/index.html#how-to-configure-autosubmit>`_
+.. important::
+    In Autosubmit ``<= 4.1.16``, ``autosubmit configure`` created the directories itself. From the next release onwards, that is the responsibility of ``autosubmit install``. If you are upgrading, run ``autosubmit install`` after ``autosubmit configure``.
 
-.. code-block:: bash
+.. hint::
+    The ``dialog`` (GUI) library is optional. Without it, ``autosubmit configure`` prompts on the CLI. Use ``autosubmit configure -h`` to see all options.
```
kinow, line 96: (I think I never tested that option... I hope it's still working 😶‍🌫️ )

docs/source/installation/index.rst
```diff
@@ -53,218 +38,144 @@ The sequence of instructions to install Autosubmit and its dependencies with pip
...
+    autosubmit create a000
 
-    # Test autosubmit
-    autosubmit -v
+    # Run the experiment
+    autosubmit run a000
 
-For a very quick test, you can follow the next instructions to configure and run Autosubmit at the user level. Otherwise, please go directly to `How to configure Autosubmit <https://autosubmit.readthedocs.io/en/master/installation/index.html#how-to-configure-autosubmit>`_
+.. important::
+    In Autosubmit ``<= 4.1.16``, ``autosubmit configure`` created the directories itself. From the next release onwards, that is the responsibility of ``autosubmit install``. If you are upgrading, run ``autosubmit install`` after ``autosubmit configure``.
```
kinow: Maybe we don't need this note? After some time it'd be outdated. Maybe we leave users to find this in the changelog, and just explain how the installation works for the current version.

docs/source/installation/index.rst
```diff
@@ -53,218 +38,144 @@ The sequence of instructions to install Autosubmit and its dependencies with pip
...
-    # Where ~/.ssh/id_rsa is the path to your private key
+Then proceed to :ref:`quick-setup` or :ref:`full-setup`.
 
-    # run
-    autosubmit run a000
+Verify the install
+------------------
 
+.. hint::
+    ``autosubmit -v`` prints the installed version, ``autosubmit readme`` prints the README, and ``autosubmit changelog`` prints the changelog.
```
kinow: I'd leave this out. The -v will be shown later. But readme & changelog are in the kill-list :-)

docs/source/ext/dependencies_licenses.py
```diff
@@ -39,64 +43,201 @@
 TableHeader = Literal["Package", "Version Spec", "License"]
 ColumnWidth = Literal[30, 20, 50]
 
+# A ``License`` metadata field longer than this, or spanning several lines, is
+# the full licence text rather than a name, and is not usable as a table cell.
+_MAX_LICENSE_NAME_LENGTH = 64
```
kinow: Is 64 arbitrary, or is there a logic behind this number?

`Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International Public License` is normally used for datasets, but if that's used, then I guess we'd show it as `UNKNOWN`?

Maybe increase it to 128 or so, and print it. Then, one of us will see the long text in the generated documentation and update the code accordingly to handle any specific license.

This is probably better than reporting unknown while the license is known (i.e., we wouldn't know that the unknown is known but showing as unknown 😆 )

kinow (commented):
I haven't finished reviewing the directive yet. Will do later with more calm. But it's looking really good, @rajeshsuraj ! I'll take a look at the preview from ReadTheDocs too. Thanks! 🙇

docs/source/installation/index.rst
```diff
@@ -2,48 +2,33 @@
...
+   These must be available on the host machine:
 
-- Python dependencies: configobj>=5.0.6, argparse>=1.4.0 , python-dateutil>=2.8.2, matplotlib==3.4.3, numpy==1.21.6, py3dotplus>=1.1.0, pyparsing>=3.0.7, paramiko>=2.9.2, mock>=4.0.3, six>=1.10, portalocker>=2.3.2, networkx==2.6.3, requests>=2.27.1, bscearth.utils>=0.5.2, cryptography>=36.0.1, setuptools>=60.8.2, xlib>=0.21, pip>=22.0.3, ruamel.yaml, pythondialog, pytest, nose, coverage, PyNaCl==1.4.0, six>=1.10.0, requests, xlib, Pygments, packaging==19, typing>=3.7
+   .. system_dependencies::
 
-.. important:: ``dot -v`` command should contain "dot", pdf, png, SVG, Xlib in the device section.
+.. note:: The ``dot -v`` command should list ``dot``, ``pdf``, ``png``, ``SVG`` and ``Xlib`` in the device section.
 
-.. important:: The host machine has to be able to access HPCs/Clusters via password-less ssh. Ensure that the ssh key is in PEM format ``ssh-keygen -t rsa -b 4096 -C "email@email.com" -m PEM``.
+.. important:: The host machine must be able to reach HPCs and clusters via password-less SSH. Generate a PEM-format key with ``ssh-keygen -t rsa -b 4096 -C "email@email.com" -m PEM``.
```
kinow, line 18: Turn these admonitions/notes into text? The user would probably be fine reading notes about installation prerequisites in this section, and it's not very long.

docs/source/installation/index.rst
```diff
@@ -2,48 +2,33 @@
...
 
-To install autosubmit, execute the following:
-::
+Install Autosubmit
+==================
 
-    pip install autosubmit
+Autosubmit is distributed via PyPI and as source on GitHub. Choose one of the methods below.
```
kinow: Add links in PyPI to our package, and GitHub to source? You can save the link at the footer (I forget how that's done in rst... something like `.. linkA\n.. linkB`, then you can refer it multiple times in the text without re-writing the URL...

docs/source/installation/index.rst
rajeshsuraj: yeah it will quicky become stale , will remove it.

docs/source/installation/index.rst
rajeshsuraj: removing this as well

docs/source/installation/index.rst
rajeshsuraj: it does not look nice if i convert it to text so i kept as admonitions/notes.

docs/source/installation/index.rst
rajeshsuraj: image
its been written 10 yrs ago :skull_and_crossbones: , do we still support this feature or command ,
I will test it out and let you know

kinow: Whew, that might be some of the oldest code we still have around :laughing: 

Let me try that too now.

docs/source/ext/dependencies_licenses.py
```diff
@@ -39,64 +43,201 @@
 TableHeader = Literal["Package", "Version Spec", "License"]
 ColumnWidth = Literal[30, 20, 50]
 
+# A ``License`` metadata field longer than this, or spanning several lines, is
+# the full licence text rather than a name, and is not usable as a table cell.
+_MAX_LICENSE_NAME_LENGTH = 128
```
kinow, line 48: Make the comment a docstring comment? That way, the IDE can also show that when you hover the variable here or if it's imported elsewhere (at least PyCharm does that).

In Python that's a bit odd... but it goes "under" the variable.

```_MAX_LICENSE_...
"""A ``License`` metadata....
```

docs/source/installation/index.rst
```diff
@@ -53,218 +38,138 @@ The sequence of instructions to install Autosubmit and its dependencies with pip
...
 
-Create an ``/etc/autosubmitrc`` file or move it from ``$HOME/.autosubmitrc`` to ``/etc/autosubmitrc`` with the information as follows:
+.. _configure-autosubmit:
 
-Mandatory parameters of /etc/autosubmit
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``autosubmit configure``
+------------------------
+
+Create ``/etc/autosubmitrc`` (or move ``$HOME/.autosubmitrc`` to ``/etc/autosubmitrc``) with the sections below.
```
kinow, line 169: Users can also use an arbitrary location defining an environment variable, so these are not the only places to create `utosubmtirc`... You mention that above, maybe simplify here or remind users?

docs/source/installation/index.rst
kinow, line 1: Would be good if the lines were not so long. It's harder to read in an IDE, even on a wide monitor, and if someone changes one character, it may be harder and more error-prone to review -- https://sembr.org/

Maybe break the lines at 80, 120 more or less? e.g. from the Python docs source code: https://raw.githubusercontent.com/python/cpython/refs/heads/main/Doc/extending/embedding.rst

(We don't need to be so precise about line width IMO, but just something so it's easier to read/maintain.)

docs/source/installation/index.rst
```diff
@@ -53,218 +38,138 @@ The sequence of instructions to install Autosubmit and its dependencies with pip
     # Avoid interactive stuff
     export DEBIAN_FRONTEND=noninteractive
 
-    # Dependencies
-    apt install wget curl python3.9 python3.9-tk python3.9-dev graphviz -y -q
+    # System dependencies
+    apt install curl git graphviz python3 python3-pip python3-tk rsync sqlite3 subversion dialog -y -q
```
kinow, line 42: Maybe we can comment here that this example installs system dependencies for a Debian-based distribution, and tell users to look at our `pyproject.toml` file, or even link the part with `.. system_dependencies::`, and tell users from Suse/RHEL/Alpine/Rocky/etc., that they need to install those dependencies.

Also, OCD-nit-picking, but maybe we can move the flags before the utilities, maybe:

```rst
    # System dependencies
    apt install --force-yes --quiet \
      curl \
      git \
      graphviz \
      dialog \
      python3 \
      python3-pip \
      python3-tk \
      rsync \
      sqlite3 \
      subversion
```

pyproject.toml
```diff
@@ -260,3 +260,21 @@ ignore = [
...
+dependencies = [
+    "dep:generic/bash",
+    "dep:generic/curl",
+    "dep:generic/dialog",
+    "dep:generic/git@>=2.32",
+    "dep:generic/graphviz@>=2.38",
+    "dep:generic/rsync",
+    "dep:generic/sqlite3",
+    "dep:generic/subversion",
+    "dep:generic/tk",
```
kinow, line 274: Did some digging, and I think we can remove: curl, dialog, tk. I tried finding tk or dialog, but only found in the git history, e.g.,

```
kinow@ranma:~/Development/python/workspace/autosubmit/autosubmit$ git show 00ddc1eec
commit 00ddc1eec721a03348f6d0bfc4554855b99ddef9
Author: Joan Lopez <joan.lopez@bsc.es>
Date:   Thu Jul 28 11:32:55 2016 +0200

    Improving UX on configure steps

index 012850f10..d78ec3cb2 100644
@@ -30,7 +30,7 @@ except ImportError:
     # noinspection PyCompatibility
     from ConfigParser import SafeConfigParser
 
-# It is Python dialog available?
+# It is Python dialog available? (optional dependency)
 try:
     import dialog
 except Exception:
@@ -1139,7 +1139,7 @@ class Autosubmit:
             return False
 
         while local_root_path is None:
-            local_root_path = raw_input("Introduce Local Root path: ")
+            local_root_path = raw_input("Introduce path to experiments: ")
         local_root_path = local_root_path.replace('~', home_path)
         if not os.path.exists(local_root_path):
             Log.error("Local Root path does not exist.")
```

Nothing in our current code base :-) 

And add `"dep:generic/xdg-utils", as we use `xdg-open` when the user specifies `autosubmit create --plot`.

I think this is valid per PEP 725, https://peps.python.org/pep-0725/#dependencies-optional-dependencies?

```
[external.optional-dependencies]
plot = [
    "dep:generic/xdg-utils",
]
```

kinow:
Once this is merged, we may have to update our Docker images. I think they may be installing more/less system dependencies than we have here. Some may be needed for containers (e.g., tini).

docs/source/installation/index.rst
rajeshsuraj: I would maintain lines at 80 or 120 now , I use alt  + z to read in the IDE

kinow (approved):
@rajeshsuraj I just read your latest commit, and reviewed the code again before the build completed. I found nothing wrong with it, so leaving it approved :-)

Can you check if CICD build passes, do a final check on the generated RTD page (linked in the checks below), and if everything is OK, rebase/squash/merge etc (you know the drill :) ). 

Thank you a lot for working on this!!! We finally have good installation docs 🙏 🙇

PR #3275, opened 2026-09-16
Make STATUS start conditions exact, rename FAIL_OK → WEAK, fix docs

### LINKED ISSUE #3262: Understanding 'start conditions' dependency behaviour
Hi! In this [ClimateDT workflow issue](https://gitlab.earth.bsc.es/digital-twins/de_340-3/workflow/-/work_items/1770), the MONITOR_RESOURCES job was failing, and a possible solution was to have MONITOR_RESOURCES start after SIM had already started running. We checked, and this dependency start condition was already set up:

```
JOBS:
  MONITOR_RESOURCES:
    FILE: templates/performance/monitor_resources.sh
    PLATFORM: "%DEFAULT.HPCARCH%-login"
    DEPENDENCIES:
      SIM:
        STATUS: "RUNNING"
``` 

I did some tests, both with AS version 4.16.1 and 4.17.1, and it seems that the dependency works correctly; MONITOR_RESOURCES only started when SIM was RUNNING:

Image

Image

Image

But looking at the documentation, we saw this:

```
The statuses are ordered, so if you select RUNNING status, the task will be run if the parent is in any of the following statuses: RUNNING, QUEUING, HELD, SUBMITTED, READY, PREPARED, DELAYED, WAITING.
``` 
Could MONITOR_RESOURCES start when SIM is in any of those states? For instance, QUEUING? Are we misunderstanding the behaviour? 

It could be that the job was failing because SIM started running but didn't reach a certain point of the simulation. However, we wanted to make sure that we understand how Autosubmit works.

dbeltrankyl:
Fixes #3262.

Behavior changes:

- Made STATUS exact, with? / WEAK: true to relax it for all the statuses.
- The default is "COMPLETED" and COMPLETED/SKIPPED satisfy each other, so '?' is a no-op there.
- UNKNOWN/SUSPENDED are invalid start condition status ( raises AS critical)
- Renamed FAIL_OKAY → WEAK everywhere, including **the db columns.** cc @pablogoitia , @ntorqulu 

Refactor:

- Moved everything into job_common.Is_edge_satisfied, used by both the scheduler and the db. Removed the duplicated _edge_satisfied and the unused Status.LOGICAL_ORDER_SUCCESS_WORKFLOW. Configuration still accepts FAIL_OKAY/OPTIONAL/STATUS as case-insensitive aliases, and I strip them from the normalized configuration so there's no alias noise later on.
- Rewrote the docs and added CHANGELOG behavior/migration notes.

Off-topic:

Fixed get_engine(":memory:"), which was creating a stray :memory: file.

Example:

```yaml
EXPERIMENT:
  DATELIST: 20200101
  MEMBERS: fc0
  CHUNKSIZEUNIT: day
  CHUNKSIZE: 1
  NUMCHUNKS: 1
  CALENDAR: standard

JOBS:

  SIM:
    SCRIPT: | 
      sleep 1
    RUNNING: once
    PLATFORM: local
    WALLCLOCK: 00:05

  OTHER:
    SCRIPT: | 
      sleep 30
    RUNNING: once
    PLATFORM: local
    WALLCLOCK: 00:05


  MONITOR:
    FILE: monitor.sh
    RUNNING: once
    PLATFORM: local
    WALLCLOCK: 00:05
    DEPENDENCIES:
      SIM:
      OTHER:
        STATUS: 'RUNNING?'
```

image

**Check List**

- [x] Documentation updated.
- [x] If this is a bugfix, PR should include a link to the issue (e.g. `Closes #1234`).


dbeltrankyl:
ready to review ( if CI/CD passes)

autosubmit/config/configcommon.py
```diff
@@ -555,38 +568,56 @@ def _normalize_wallclock(data_fixed: {}) -> None:
...
         elif isinstance(dependencies, dict):
             for dependency, dependency_data in dependencies.items():
+                if isinstance(dependency_data, dict):
+                    # Keyword names are case-insensitive.
+                    dependency_data = {keyword.upper(): value for keyword, value in dependency_data.items()}
+                    # Backwards compatibility: map deprecated keyword names to their canonical form and
+                    # drop the aliases so the normalized dependencies contain no alias noise.
+                    if "STATUS" in dependency_data:
+                        dependency_data.setdefault("MIN_TRIGGER_STATUS", dependency_data.pop("STATUS"))
+                    for alias in ("FAIL_OK", "OPTIONAL"):
```
ntorqulu: I've tried to execute an experiment that had defined `FAIL_OK : True` to check the backward compatibility.
```
JOBS:

  SIM:
    SCRIPT: | 
      sleep 1
    RUNNING: once
    PLATFORM: local
    WALLCLOCK: 00:05

  OTHER:
    SCRIPT: |
      exit 1
    RUNNING: once
    PLATFORM: local
    WALLCLOCK: 00:05


  MONITOR:
    FILE: monitor.sh
    RUNNING: once
    PLATFORM: local
    WALLCLOCK: 00:05
    DEPENDENCIES:
      SIM:
      OTHER:
        FAIL_OK: True
``` 
Result:

Image

So `_normalize_dependencies` maps `FAIL_OK/OPTIONAL` to `WEAK` but `WEAK` is not accepting a failed parent. Is this intended? If so, should a message be displayed pointing that the parameter is no longer maintained?

ntorqulu: Well, when where the `FAIL_OK` and `OPTIONAL` parameters introduced? Where introduced in v4.2.0 or before?
If they were introduced in v4.2.0 should we maintain them? Or start with the base that it is not deployed so we don't need to convert the alias into `WEAK`?

autosubmit/database/db_manager_historical.py
```diff
@@ -81,7 +85,7 @@ def _save_historical_edges(self, graph: list[dict[str, Any]], run_id: int):
                 "min_trigger_status": edge.get("min_trigger_status"),
                 "completion_status": edge.get("completion_status"),
                 "from_step": edge.get("from_step"),
-                "fail_ok": edge.get("fail_ok"),
+                "weak": edge.get("weak"),
```
ntorqulu, line 88: Just to note, there's no migration, only addition of new columns.
So experiments that were running with 4.2.0 (leo's experiments I think) the job_list will have extra columns `fail_ok` and `weak`.
I don't see a problem in that if we don't ensure maintainability of versions that have not been deployed yet.

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -299,14 +299,31 @@ See :ref:`job_status_reference` for a complete reference and meanings.
...
+
+``COMPLETED`` (the default) and ``SKIPPED`` are both successful endings, so a parent in either one
+satisfies the other. Because of this, ``?`` has no effect on ``COMPLETED`` or ``SKIPPED``: the
+default dependency already runs once the parent completed **or** was skipped, and the workflow graph
+does not mark those edges as weak.
+
+.. note::
+   Since 4.2.0 ``STATUS`` is exact; previously a parent that had already moved past the selected
+   status also satisfied the dependency, and weak dependencies accepted a failed parent. The
+   keyword is now ``WEAK`` (case-insensitive); the deprecated names ``FAIL_OK`` and ``OPTIONAL``
```
ntorqulu: Should `FAIL_OK` and `WEAK` be deprecated or directly not mentioned?

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -320,15 +337,25 @@ The status are ordered, so if you select ``RUNNING`` status, the task will be ru
...
       POSTPROCESS:
         FILE: postprocess.sh
         DEPENDENCIES:
           SIM:
-            STATUS: 'RUNNING'
+            STATUS: 'RUNNING?'
         RUNNING: chunk
 
+In the example above, ``MONITOR`` starts only while ``SIM`` is running, while ``POSTPROCESS`` also
+starts if ``SIM`` already completed or failed.
```
ntorqulu: I think this contradict the paragraph above. *A failed parent is not accepted by ?*.
Should be completed or skipped?

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -1441,7 +1473,7 @@ Weak dependencies, work like this way:
     :figname: fig-dashed
     :width: 100%
     :align: center
-    :caption: Example showing the asim job starting only from chunk 3.
+    :caption: Example showing a weak dependency.
```
ntorqulu, line 1475: The example `jobs_dashed.yml` has `DEPENDENCIES: IT GET_FILES ?`.
It's not displaying correctly the weak dependencies in the generated image from the docs.

p.d. Changing topic to the API/GUI. The intend is to show also the weak dependencies in the graph view right? Currently they are seen as:

Image

But should be seen as how AS plot them:

Image

autosubmit/config/configcommon.py
```diff
@@ -555,38 +568,56 @@ def _normalize_wallclock(data_fixed: {}) -> None:
...
+                        if alias in dependency_data:
+                            dependency_data.setdefault("WEAK", dependency_data.pop(alias))
+
+                    status = dependency_data.pop("MIN_TRIGGER_STATUS", None)
+                    if status:
+                        status = status.upper()
+                        weak = bool(dependency_data.pop("WEAK", False))
```
ntorqulu: A little bit dangerous. If the user puts `WEAK: `false'`, it will be read as `True`.
Same for `FAIL_OK` and `OPTIONAL`.

autosubmit/config/configcommon.py
```diff
@@ -555,38 +568,56 @@ def _normalize_wallclock(data_fixed: {}) -> None:
...
 
         This function takes a string or dictionary of dependencies and normalizes them to a dictionary format.
         If the input is a string, it splits the string by spaces and converts each dependency to uppercase.
-        If the input is a dictionary, it converts each dependency key to uppercase and processes the status.
+        If the input is a dictionary, it uppercases the dependency keys and the keyword names and processes
+        the status.
 
-        Additionally, it checks for a ``?`` suffix in ``MIN_TRIGGER_STATUS``/``STATUS`` to set ``FAIL_OK``.
+        Additionally, it checks for a ``?`` suffix in ``MIN_TRIGGER_STATUS``/``STATUS`` and for the
+        ``WEAK`` keyword to mark the dependency as weak. The deprecated aliases ``STATUS``,
```
ntorqulu: Status is not deprecated right? Should be failed_ok and optional only, but again not sure if we should treat them as deprecated parameters if they have not been released yet.

autosubmit/job/job_list.py
```diff
@@ -384,14 +383,14 @@ def _add_edge_and_parent(
...
-                fail_ok=edge["fail_ok"]
+                weak=edge["weak"]
             )
         else:
             self.graph.edges[edge["e_from"], edge["e_to"]].update({
                 "min_trigger_status": edge["min_trigger_status"],
                 "completion_status": edge["completion_status"],
                 "from_step": edge["from_step"],
-                "fail_ok": edge["fail_ok"],
+                "weak": edge["weak"],
```
ntorqulu, line 393: From this file `job_list.py`, method `_check_relationships()`.
`WEAK` is not read, so its never merged in the dict. Only keys `MIN_TRIGGER_STATUS` and `FROM_STEP` are added to the dict.

So, an experiment where a job is defined with a weak dependency and a `*_FROM` key has a normalized dict with just `MIN_TRIGGER_STATUS` and `WEAK`.
I think this is a bug already in the code, not introduced in this pr. I can open an issue for this.

Experiment used:
```
EXPERIMENT:
  DATELIST: 20200101
  MEMBERS: fc0
  CHUNKSIZEUNIT: day
  CHUNKSIZE: 1
  NUMCHUNKS: 2
  CALENDAR: standard

JOBS:
  SIM:    {SCRIPT: 'sleep 0', RUNNING: chunk, PLATFORM: local}
  POST:
    SCRIPT: 'sleep 0'
    RUNNING: chunk
    PLATFORM: local
    DEPENDENCIES:
      SIM-1:
        STATUS: 'RUNNING?'
  POST_F:
    SCRIPT: 'sleep 0'
    RUNNING: chunk
    PLATFORM: local
    DEPENDENCIES:
      SIM-1:
        STATUS: 'RUNNING?'
        CHUNKS_FROM:
          all:
            CHUNKS_TO: 1
```

Image

ntorqulu (commented):
HI @dbeltrankyl, reviewed the pr.
My main question is whether parameters `FAIL_OKAY` and `OPTIONAL` should be kept as legacy for backward compatibility or not, taking into account that they were introduced in v4.2.0 and is not yet deployed (right?).

I'm missing review of this last part, `Fixed get_engine(":memory:"), which was creating a stray :memory: file.`, will check it tomorrow.

autosubmit/config/configcommon.py
dbeltrankyl: Thanks! I'll check the all the feedback in a few minutes

Just quick answer for this one 

> Well, when where the FAIL_OK and OPTIONAL parameters introduced? Where introduced in v4.2.0 or before?

I thought we had the OPTIONAL in 4.1.17.1, but apparently not. We can drop the fallback then thanks for noticing it

autosubmit/database/db_manager_historical.py
kinow: I discussed adding migrations with Luiggi during the Postgres rework. I think we have an issue for that...

kinow: Found it: https://github.com/BSC-ES/autosubmit/issues/1286

autosubmit/config/configcommon.py
dbeltrankyl: Ah, right. And yes removed all optional + fail_ok traces

autosubmit/config/configcommon.py
dbeltrankyl: True

I expected the user to introduce `Status: running?` instead of `WEAK: True`. However, it is better to be safe than sorry. (Weak is used more for normalization in our code.)

Changed to 
                        weak = str(dependency_data.pop("WEAK", False)).strip().lower() == "true"

autosubmit/database/db_manager_historical.py
dbeltrankyl: Thanks to both

@kinow, @ntorqulu  I'm working on the single-layer db  3114-single-db-layer branch ( not yet pushed and is in the early stages) but there I'm already doing a portable schema versioning to don't duplicate the work I'll do the fix in that branch

docs/source/userguide/defining_workflows/index.rst
dbeltrankyl: Not mentioned! thanks

docs/source/userguide/defining_workflows/index.rst
dbeltrankyl: Brain short circuit, I guess 😅, you're right

dbeltrankyl (commented):
I'll commit the feedback after this message, thanks!

autosubmit/database/session.py
```diff
@@ -66,13 +66,18 @@ def get_engine(db_path: str | Path) -> Engine:
...
     :param db_path: Path to the database file, only used for SQLite.
+    :return: An SQLAlchemy engine for the configured backend.
     """
     db_backend = BasicConfig.DATABASE_BACKEND
 
     if db_backend == "sqlite":
+        if str(db_path) == ":memory:":
+            # SQLite in-memory sentinel: do not materialize it as a file on disk.
+            return _resolve_engine("sqlite:///:memory:")
+
```
ntorqulu: About this one, I'll drop it from this pr to make it more scoped to the issue. I think @kinow will look into it in another pr.

dbeltrankyl: Ok!, it seems fair. I was just annoyed by the :memory: file to be honest

dbeltrankyl: dropped

kinow: Ha! just fixed it and wrote to Slack channel. I'm also utterly annoyed by that! Hahaha. https://github.com/BSC-ES/autosubmit/pull/3280

kinow: Oh, and it looks like I applied the same fix :) so if you or anyone else would like to review it.

autosubmit/config/configcommon.py
```diff
@@ -48,6 +49,11 @@
     from autosubmit.platforms.platform import Platform
 
 
+# Statuses accepted as a start condition (``STATUS``). ``UNKNOWN`` and ``SUSPENDED`` can never be
+# observed as a usable parent status, so they are not selectable.
```
kinow: Not sure what's selectable here. I understood the other part of the comment though.

autosubmit/config/configcommon.py
```diff
@@ -555,38 +568,56 @@ def _normalize_wallclock(data_fixed: {}) -> None:
...
         :param dependencies: The dependencies to normalize, either as a string or a dictionary.
-        :type dependencies: Union[str, dict]
+        :param job_name: Name of the job the dependencies belong to, used in error messages.
         :return: A dictionary with normalized dependencies.
-        :rtype: dict
+        :raises AutosubmitCritical: If a ``STATUS`` is not a valid start condition.
         """
         aux_dependencies = {}
         if isinstance(dependencies, str):
             for dependency in dependencies.upper().split(" "):
```
kinow: Here don't we have a multiple-spaces problem? Not in this change, @dbeltrankyl , but I opened the whole function in PyCharm and noticed this one...

Image

It doesn't seem to break or cause anything too bad doing a quick test, but debugger shows it's actually creating multiple empty dependencies.

Image

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -299,14 +299,30 @@ See :ref:`job_status_reference` for a complete reference and meanings.
...
-      - The task is unknown.
     * - ``COMPLETED``
       - The task is completed. # Default
-    * - ``SUSPENDED``
-      - The task is suspended.
 
-The status are ordered, so if you select ``RUNNING`` status, the task will be run if the parent is in any of the following statuses: ``RUNNING``, ``QUEUING``, ``HELD``, ``SUBMITTED``, ``READY``, ``PREPARED``, ``DELAYED``, ``WAITING``.
+Each ``STATUS`` is exact: the job runs only while the parent is in that status. Because statuses
+such as ``READY``, ``SUBMITTED``, ``QUEUING`` or ``RUNNING`` are transient, the job can miss the
+window and stay ``WAITING``.
```
kinow: I think I understand this, but I am not sure users will understand it. I think what it means is that the `STATUS` from the job in Autosubmit is checked against the latest status of the dependencies in the platforms. If the platform changed the status of the jobs faster than the Autosubmit scheduler (i.e., it is `WAITING` for the Autosubmit scheduler, but it is `COMPLETED` in a Slurm platform), then Autosubmit will not trigger dependencies to an intermediary status (e.g., `RUNNING`).

Maybe we can make it easier for workflow developers and new users?

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -299,14 +299,30 @@ See :ref:`job_status_reference` for a complete reference and meanings.
...
+
+``COMPLETED`` (the default) and ``SKIPPED`` are both successful endings, so a parent in either one
+satisfies the other. Because of this, ``?`` has no effect on ``COMPLETED`` or ``SKIPPED``: the
+default dependency already runs once the parent completed **or** was skipped, and the workflow graph
+does not mark those edges as weak.
+
+.. note::
+   Since 4.2.0 ``STATUS`` is exact; previously a parent that had already moved past the selected
+   status also satisfied the dependency, and weak dependencies accepted a failed parent. The
+   keyword is now ``WEAK`` (case-insensitive); ``UNKNOWN`` and ``SUSPENDED`` are not valid ``STATUS`` values and raise an
```
kinow, line 331: Maybe we should warn users in the command line if they use `OPTIONAL` or `FAIL_OK`?

I think we don't have many/any users of `OPTIONAL` or `FAIL_OK`, but we will break behaviour & backward compatibility between AS minor/patch/hotfix versions (not major versions), which is not ideal, so we probably have to at least announce that well to users (cc @f-macchia).

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -320,15 +336,25 @@ The status are ordered, so if you select ``RUNNING`` status, the task will be ru
...
       POSTPROCESS:
         FILE: postprocess.sh
         DEPENDENCIES:
           SIM:
-            STATUS: 'RUNNING'
+            STATUS: 'RUNNING?'
         RUNNING: chunk
 
+In the example above, ``MONITOR`` starts only while ``SIM`` is running, while ``POSTPROCESS`` also
+starts if ``SIM`` already completed or was skipped.
```
kinow: But not failed? Failed comes after RUNNING -- no ideally, but it still happens after in the sequence of possible states...

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -320,15 +336,25 @@ The status are ordered, so if you select ``RUNNING`` status, the task will be ru
...
 
-The ``FROM_STEP`` keyword can be used to select the **internal** step of the dependency that you want to check. The possible value is an integer. Additionally, the target dependency, must call to `%AS_CHECKPOINT%` inside their scripts. This will create a checkpoint that will be used to check the amount of steps processed.
+
+The ``FROM_STEP`` keyword can be used to select the **internal** step of the dependency that you want to check. It is used together with ``STATUS: 'RUNNING'``. The possible value is an integer. Additionally, the target dependency, must call to `%AS_CHECKPOINT%` inside their scripts. This will create a checkpoint that will be used to check the amount of steps processed.
```
kinow, line 365: I am not sure if an internal step of dependencies is defined elsewhere? If it is, maybe we can link here. Not important as it was already here before the change.

docs/source/userguide/defining_workflows/code/jobs_dashed.yml
```diff
@@ -10,6 +10,9 @@ JOBS:
 
   CALC_STATS:
     SCRIPT: 'sleep 0'
-    DEPENDENCIES: IT GET_FILES ?
+    DEPENDENCIES:
+      IT:
+      GET_FILES:
+        STATUS: 'FAILED?'
```
kinow: Here, `FAILED?` means there's a dependency to `GET_FILES` if it failed to run, but also if it's completed or skipped?

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -299,14 +299,30 @@ See :ref:`job_status_reference` for a complete reference and meanings.
...
-      - The task is suspended.
 
-The status are ordered, so if you select ``RUNNING`` status, the task will be run if the parent is in any of the following statuses: ``RUNNING``, ``QUEUING``, ``HELD``, ``SUBMITTED``, ``READY``, ``PREPARED``, ``DELAYED``, ``WAITING``.
+Each ``STATUS`` is exact: the job runs only while the parent is in that status. Because statuses
+such as ``READY``, ``SUBMITTED``, ``QUEUING`` or ``RUNNING`` are transient, the job can miss the
+window and stay ``WAITING``.
+
+Adding ``?`` to the status (or ``WEAK: true`` to the dependency) makes the dependency weak, so a
+parent that already finished **successfully** (``COMPLETED`` or ``SKIPPED``) also satisfies it. For
+example, ``RUNNING?`` runs the job while the parent is running, or once it completed or was skipped.
```
kinow, line 313: I think we don't talk about what happens when users specify both?

```yaml
JOBS:
  A: ...
  B: ...
  C:
    FILE: ok.sh
    DEPENDENCIES:
      A:
        WEAK: true
        STATUS: RUNNING
      B:
        WEAK: false
        STATUS: FAILED?
```

Here, I think `C`'s dependency to `B` is weak (true), even though we have `WEAK: false`. So the `STATUS?` has precedence over `WEAK`... that will probably not be very fun to validate later in #871 , but I haven't tried to think of alternatives yet (only the mode review is on right now 🤓 )

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -299,14 +299,30 @@ See :ref:`job_status_reference` for a complete reference and meanings.
...
+
+A failed parent is **not** accepted by ``?``. To run the job when the parent failed, use ``FAILED``
+(only on failure) or ``FAILED?`` (on failure, completion or skip).
+
+``COMPLETED`` (the default) and ``SKIPPED`` are both successful endings, so a parent in either one
+satisfies the other. Because of this, ``?`` has no effect on ``COMPLETED`` or ``SKIPPED``: the
+default dependency already runs once the parent completed **or** was skipped, and the workflow graph
+does not mark those edges as weak.
```
kinow, line 326: I think if users want COMPLETED, or SKIPPED, or FAILED (a general finished?), then users must use `FAILED?`?

docs/source/userguide/defining_workflows/index.rst
```diff
@@ -1394,12 +1420,14 @@ In this workflow you can see an illustrated example of ``SKIPPABLE`` parameter u
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In this workflow you can see an illustrated example of weak dependencies.
-
-Weak dependencies, work like this way:
+A weak dependency (``?`` suffix or ``WEAK: true``) is satisfied by a parent that already finished
+**successfully** (``COMPLETED`` or ``SKIPPED``). This is useful with transient start conditions such
+as ``RUNNING``, where the parent may have moved on before the job is evaluated. A failed parent is
+**not** accepted by a weak dependency; use ``STATUS: 'FAILED'`` (only on failure) or
+``STATUS: 'FAILED?'`` (on failure, completion or skip) for that.
```
kinow, line 1435: My Cylc-Fu is a bit rusty, but from I recall, in Cylc you could use the exact statuses, like running, failed, submitted... or three special aliases, `$job_name:finished`, `$job_name:failed`, `$job_name:succeeded`.

Finished was any final state, failed anything considered failed (I think there was only failed and submit-failed in Cylc? Can't recall if there was a skipped or timedout -- submit-failed means it failed to submit to the platform...), and succeeded I think had only one state [^1].

I think our way of reasoning is more complicated for users. You have the statuses, or you use `?` that means what we considered a finished successfully, which are statuses skipped or completed. But if a user wants only completed as successful, and they want the dependency only on completed or failed, but not skipped... then they must use which status?

```
STATUS: FAILED,COMPLETED
```

is not valid right?

Maybe @dmcclainbsc could review this PR too, at least the documentation, as he'll be one of the users (and he mentioned it in an internal meeting 😬 ). This way, I think we also don't run the risk of reviewing, merging, preparing a dev module, only then for A7 devs to find inconsistencies or issues, and then we have to do the process all over again...

Damien, you don't have to read reStructeredText. The rendered version of the latest commit is available at https://autosubmit--3275.org.readthedocs.build/en/3275/userguide/defining_workflows/index.html#start-conditions

[1]: In Cylc you can also create kind of custom statuses, by defining workflow outputs, and then starting tasks based on those outputs, so a job could have 30, 40 different statuses, but that's a different feature/discussion.

kinow: And as another example, ecFlow has "triggers".

```
task A
task B
  trigger A eq complete or A == aborted
```

Users can write this DSL for suites, or use PyFlow python with boolean logic to define how tasks/dependencies are triggered.

autosubmit/database/tables.py
```diff
@@ -221,7 +221,7 @@
     Column("min_trigger_status", String),
     Column("completion_status", String),
     Column("from_step", Integer),
-    Column("fail_ok", Boolean),
+    Column("weak", Boolean),
```
kinow, line 224: Couldn't we just use the `fail_ok` from the DB and eventually migrate it later?  I think we could just store the data in this boolean, add some comments in the code base, handle loading from fail_ok into weak in the object creation/read... and that's that?

Not super important, just asking in case it hasn't been considered and could solve some issues with API compatibility....

kinow (commented):
First pass through the code with the IDE, Dani. Skimmed through the Python code, and focused more on the docs and general idea to fix the issue linked.

docs/source/userguide/defining_workflows/index.rst
dbeltrankyl: In 4.1.17, these variables didn't exist as pointed by @ntorqulu

autosubmit/database/tables.py
dbeltrankyl: Same as the other comment, the fail_ok wasn't existing in 4.1.17 so, I think that now is the proper time to do the change, no?

dbeltrankyl:
Thanks @kinow 

Quick answered, two points. I'll revise the rest later

kinow:
A few other questions, from the tests, @dbeltrankyl :grimacing: 

What does this mean?

```yaml
JOBS:
  C:
      DEPENDENCIES:
          B:
              WEAK: true
```

C has a weak dependency on B? I think here WEAK doesn't mean anything, because there is no status, so that's ignored? (Something that would have to be validated in #871). Maybe we should drop `WEAK: bool` and use only `?` if that makes things simpler for users?

---

If I have a graph with three jobs, `A`, `B`, `C`, with the following dependencies:

```yaml
JOBS:
  A:
    FILE: ok.sh
    PLATFORM: LOCAL
    RUNNING: once
  B:
    FILE: ok.sh
    DEPENDENCIES:
      A:
        STATUS: COMPLETED
    WALLCLOCK: 00:05
    RUNNING: once
  C:
    FILE: ok.sh
    DEPENDENCIES:
      A:
        STATUS: FAILED
      B:
        STATUS: COMPLETED
    RUNNING: member
    WALLCLOCK: 00:05
```

The graph is this:

image

I believe C must run only if A has failed, and B has completed.

And B must run only if A has completed, not if A has failed.

I guess a user could have this with A failed, and the user manually forced B as completed via `setstatus`. And then in that case `C` would be triggered? 

But in a normal execution with A and B completed, I guess C would never be triggered? Cylc detects when the graph stalls. Are we doing anything with this case?

I did a quick test run with this branch, and I looks like Autosubmit runs A to completion, then triggers B, which runs fine too, and that's it. C is never triggered, and the user is never warned about anything?

```bash
Loading YAML: /home/kinow/autosubmit/a00i/conf/as_misc.yml
YAML configuration loaded:
  /home/kinow/autosubmit/a00i/conf/as_misc.yml
  /home/kinow/autosubmit/a00i/conf/autosubmit_a00i.yml
  /home/kinow/autosubmit/a00i/conf/expdef_a00i.yml
  /home/kinow/autosubmit/a00i/conf/jobs_a00i.yml
  /home/kinow/autosubmit/a00i/conf/platforms_a00i.yml
  /home/kinow/autosubmit/a00i/conf/version.yml

Expdef config file is correct
Platforms sections: OK
Jobs sections OK
Configuration files OK

local(log_recovery): Starting...
local(log_recovery): successfully connected.
Looking for new jobs...
Loading jobs and edges from database...
Checking 1 jobs for platform local
Job a00i_A changed from RUNNING to COMPLETED
Saving jobs to the database...
Jobs saved.
Saving jobs to the database...
Jobs saved.
Looking for new jobs...
Loading jobs and edges from database...
Saving jobs to the database...
Jobs saved.
File /home/kinow/autosubmit/a00i/proj/local_project/ok.sh exists
Section:B can submit 20 jobs at this time
Job: a00i_B submitted with job_id: 122416 and workflow commit: 
Saving jobs to the database...
Jobs saved.
Checking 1 jobs for platform local
Job a00i_B changed from SUBMITTED to RUNNING
Saving jobs to the database...
Jobs saved.
Checking 1 jobs for platform local
a00i_A_STAT_0 file have been transferred
local(log_recovery): Job a00i_A recovered 1 attempt(s).
Looking for new jobs...
Loading jobs and edges from database...
Checking 1 jobs for platform local
Job a00i_B changed from RUNNING to COMPLETED
Saving jobs to the database...
Jobs saved.
Saving jobs to the database...
Jobs saved.
Looking for new jobs...
Loading jobs and edges from database...
No more jobs to run.
Waiting for all logs to be updated
a00i_B_STAT_0 file have been transferred
local(log_recovery): Job a00i_B recovered 1 attempt(s).
local(log_recovery): Exiting.
Run successful
ROCRATE not present in experiment YAML configuration. No RO-Crate archive created.
```

autosubmit/config/configcommon.py
dbeltrankyl: Changed for 

```
# Statuses accepted as a start condition (``STATUS``). ``UNKNOWN`` and ``SUSPENDED`` can never be
# observed as a usable parent status, so they are not valid ``STATUS`` values.
```

autosubmit/config/configcommon.py
dbeltrankyl: Hii, I didn't touch this part ( bug per-existent) but doesn't hurt to fix it there, added the test and the fix

docs/source/userguide/defining_workflows/code/jobs_dashed.yml
dbeltrankyl: Yes

docs/source/userguide/defining_workflows/index.rst
dbeltrankyl: Yes, If both ``?`` and ``WEAK`` are set, ``?`` takes precedence.

I've doc it 

> that will probably not be very fun to validate later in #871 , 

I think we only need to check if at the end of the reload, WEAK is set to True or False because the Status will have the "?" stripped.

docs/source/userguide/defining_workflows/index.rst
dbeltrankyl: Yes, I reworded it

```
Use ``STATUS: 'FAILED'`` to run only when the parent failed, or ``STATUS: 'FAILED?'``
 to run once the parent has finished **regardless of the outcome**  (``COMPLETED``, ``SKIPPED`` or ``FAILED``).
```

docs/source/userguide/defining_workflows/index.rst
dbeltrankyl: This is intentional, it is what @dmcclainbsc asked to have in #3262 but I've reworded it as well

Basically, '?' only adds the successful endings (COMPLETED/SKIPPED)`

docs/source/userguide/defining_workflows/index.rst
dbeltrankyl: Nop it isn't we can open an issue

docs/source/userguide/defining_workflows/index.rst
dbeltrankyl: Hm, right now there is a gap

RUNNING + only FAILED is not possible 

Maybe it is better to drop "?" afterall and just do a list YAML?

I can add aliases as you suggest based on cylc

```
SUCCEEDED = {COMPLETED, SKIPPED},
FINISHED = {COMPLETED, SKIPPED, FAILED}.
```

And the users would have to configure: ( or a yaml list ) 

```
STATUS: 'RUNNING'
STATUS: 'RUNNING,FAILED'
STATUS: 'RUNNING,FINISHED' # The problem with FINISHED is that users won't be aware of this 
STATUS: 'RUNNING,COMPLETED,SKIPPED,FAILED' # but this is too long?
STATUS: 'FAILED'
STATUS: 'COMPLETED,SKIPPED'   # same as if the "STATUS" is not present
STATUS: 'COMPLETED'   # or make COMPLETED===SKIPPED? 
STATUS: 'FINISHED'
```


or if we keep the "?" ( I would drop this then, too complicate) 

```
STATUS: 'RUNNING'                # {RUNNING}
STATUS: 'RUNNING,FAILED'         # {RUNNING, FAILED}
STATUS: 'RUNNING,FAILED?'        # {RUNNING, FAILED, COMPLETED, SKIPPED}
STATUS: 'FAILED'                 # {FAILED}
STATUS: 'COMPLETED?'             # {COMPLETED, SKIPPED}   (default)
STATUS: 'FAILED?'                # {FAILED, COMPLETED, SKIPPED}
```

dbeltrankyl (commented):
Thanks @kinow ! can you do a quick look at the comment regarding the drop of "?" ? I can solve it if we agree on the change https://github.com/BSC-ES/autosubmit/pull/3275#discussion_r4061241558

dbeltrankyl:
> https://github.com/BSC-ES/autosubmit/pull/3275#discussion_r4061241558

And about this, the field in the db instead of FAIL_OK or WEAK it will be replaced by start_conditions or something like that... I need to think about it

PR #3281, opened 2026-09-21
[Bug] - PJM addresses a path for the channel instead of byte or str

### LINKED ISSUE #3278: [ Bug ] Error during log recovery: Expected unicode or bytes, got <class 'pathlib.PosixPath'>
This is not an issue that is blocking the runs, but we think there's probably an issue in the log recovery process. Tagging @kellekai, as he has also been running into the same issues being reported.

### Description

This is not a blocking issue, but I think the logs of the run are not being recovered properly or at least theres an issue ther. In the ASLOGS when we run the experiments, we get the following issue:

```bash
Using /home/Earth/kkeller/.ssh/config as SSH configuration file
Failed to authenticate with ssh-agent due to Authentication failed.
Trying to authenticate with other methods
fugaku(log_recovery): Error during log recovery: Expected unicode or bytes, got <class 'pathlib.PosixPath'>
  File "/home/Earth/kkeller/hanami-env/lib/python3.12/site-packages/autosubmit/platforms/platform.py", line 1087, in recover_platform_job_logs
    self.recover_job_log(jobs_db_manager, as_conf)
  File "/home/Earth/kkeller/hanami-env/lib/python3.12/site-packages/autosubmit/platforms/platform.py", line 1046, in recover_job_log
    report = job.retrieve_logfiles()
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/Earth/kkeller/hanami-env/lib/python3.12/site-packages/autosubmit/job/job.py", line 1532, in retrieve_logfiles
    if not self.platform.get_stat_file(self, attempt) or not self.stat_file_is_completed(attempt) or self.stat_registered(attempt):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/Earth/kkeller/hanami-env/lib/python3.12/site-packages/autosubmit/platforms/platform.py", line 737, in get_stat_file
    if self.check_file_exists(filename):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/Earth/kkeller/hanami-env/lib/python3.12/site-packages/autosubmit/platforms/pjmplatform.py", line 303, in check_file_exists
    self._ftpChannel.stat(Path(self.get_files_path(), src))
  File "/home/Earth/kkeller/hanami-env/lib/python3.12/site-packages/paramiko/sftp_client.py", line 491, in stat
    path = self._adjust_cwd(path)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/Earth/kkeller/hanami-env/lib/python3.12/site-packages/paramiko/sftp_client.py", line 949, in _adjust_cwd
    path = b(path)
           ^^^^^^^
  File "/home/Earth/kkeller/hanami-env/lib/python3.12/site-packages/paramiko/util.py", line 325, in b
    raise TypeError(f"Expected unicode or bytes, got {type(s)}")
TypeError: Expected unicode or bytes, got <class 'pathlib.PosixPath'>

Restoring SSH connection...
```

* Autosubmit version: HEAD in master
* Machine/VM/environment name (if applicable): bscautosubmit04
* Experiment ID (if applicable): abok
  * e.g. `<EXPID>_19900101_fc0_1_<TASK_NAME>` 
  * e.g. `/appl/AS/AUTOSUBMIT_DATA/<EXPID>/tmp/ASLOGS`
  * (And any other information you may have, the more, the merrier.)

### Reproducible Example

Create python env with HEAD of Autosubmit master
Copy experiment abok
Create and run 

### Expected Behaviour

No issues reported.

VindeeR:
Closes #3278 

It was being converted to path to sent the path instead of using string

**Check List**

- [ ] Documentation updated.


autosubmit/platforms/pjmplatform.py
```diff
@@ -300,7 +300,7 @@ def check_file_exists(
         while not file_exist and retries < max_retries:
             try:
                 # This return IOError if path does not exist
-                self._ftpChannel.stat(Path(self.get_files_path(), src))
+                self._ftpChannel.stat(str(Path(self.get_files_path(), src)))
```
kinow: This doesn't seem to be tested in our CICD, probably because we don't have the PJM containers, @VindeeR .

However, looking at the code, it seems to be the exact same code as in the Slurm platform?

If so, maybe we could move it to the parent ParamikoPlatform, and remove the function from both Slurm and PJM. That way, the code would now be tested in the ParamikoPlatform, via the Slurm container tests.

The PJM is also odd as it's doing a retries = 9999... if the intention was to stop the while loop, break would probably work better, or =max_retries+1... But I think it'd be possible to have a common function for both, that checks if a file exists via sftp?

And then this way we wouldn't have this (or similar) bugs appearing because we didn't test in our CICD...

kinow:
CICD needs a kick?

kinow (approved):
Thank you @VindeeR ! The part that's missing coverage was like that already, and since we are keeping most of the slurm code, just fixing the part about SFTPError, that should be safe. I think we won´t need a changelog as we "broke" it on master, without releasing it. Ready to be merged IMO. Thanks!

PR #3287, opened 2026-09-23
[Enhancement] - Organizing the mypy and type hint in the Job module

VindeeR:
Related #1117

Restructuring the organization and type hints in the Job module

**Check List**

- [ ] Documentation updated.


autosubmit/job/job_packages.py
```diff
@@ -194,7 +191,7 @@ def build_scripts(self, configuration: 'AutosubmitConfig') -> None:
         Log.debug("Building scripts")
         self._create_scripts(configuration)
 
-    def generate_scripts(self, configuration: 'AutosubmitConfig', only_generate: bool = False) -> None:
+    def generate_scripts(self, configuration:'AutosubmitConfig', only_generate: bool = False) -> None:
```
kinow, line 194: Is the correct no space, or one space? I think somewhere we started using double quotes too?

autosubmit/platforms/slurmplatform.py
```diff
@@ -305,20 +286,55 @@ def allocated_nodes() -> str:
...
+            except Exception as e:
+                if "garbage" in str(e).lower():
+                    sleep(2)
+                    retries = retries + 1
+                else:
+                    raise
+        if not file_exist and show_logs:
+            Log.warning(f"File {src} couldn't be found")
+        return file_exist
+
```
kinow: Rebase? I think you did that somewhere else, not on this PR?

autosubmit/platforms/slurmplatform.py
```diff
@@ -237,17 +227,15 @@ def get_check_all_jobs_cmd(self, jobs_id: str):
 
         :param jobs_id: ID of one or more jobs.
         :return: sacct command to all jobs.
-        :rtype: str
         """
         return f"sacct -n -X --jobs {jobs_id} -o jobid,State"
 
-    def get_estimated_queue_time_cmd(self, job_id: str):
+    @staticmethod
```
kinow, line 232: It doesn't hurt to make them static, but we may not need to make these functions static. They can be moved to the module or another file to make the objects lighter/smaller too. Depends whether they will be reused or not.

This is one warning I normally ignore in PyCharm code inspection.

autosubmit/job/job.py
```diff
@@ -3446,10 +3405,11 @@ def _finalize_wrapper_completion(self) -> bool:
 
         return True
 
-    def check_and_update_status(self, as_conf: AutosubmitConfig) -> bool:
-        """Check the status of the wrapper job and its inner jobs.
+    def check_and_update_status(self, as_conf: 'AutosubmitConfig') -> bool:
+        """
+        Check the status of the wrapper job and its inner jobs.
```
kinow: The other way around? """Check the status..."

autosubmit/job/job.py
```diff
@@ -3446,10 +3405,11 @@ def _finalize_wrapper_completion(self) -> bool:
 
         return True
 
-    def check_and_update_status(self, as_conf: AutosubmitConfig) -> bool:
-        """Check the status of the wrapper job and its inner jobs.
+    def check_and_update_status(self, as_conf: 'AutosubmitConfig') -> bool:
```
kinow, line 3389: double quotes here? "AutosubmitConfig"? I prefer single because I don't have to press Shift... but I think we were going the double-quotes way? 😆 So I'm adjusting my coding style to match that, maybe we can update just these ones we added in this PR.

autosubmit/job/job.py
```diff
@@ -3426,10 +3384,11 @@ def _check_wrapper_wallclock_and_handle(self) -> bool:
                 inner_job.new_status = Status.WAITING
         return True
 
-    def _sync_inner_job_statuses(self, as_conf: AutosubmitConfig) -> None:
-        """Persist status changes for inner jobs that have transitioned.
+    def _sync_inner_job_statuses(self, as_conf: 'AutosubmitConfig') -> None:
+        """
+        Persist status changes for inner jobs that have transitioned.
```
kinow: docstring, """Persist ?

autosubmit/job/job.py
```diff
@@ -3272,7 +3234,7 @@ def __init__(
             total_wallclock: str,
             num_processors: int,
             platform: 'ParamikoPlatform',
-            as_config: AutosubmitConfig,
+            as_config: 'AutosubmitConfig',
```
kinow, line 3221: Double quotes? Although now I see the single quotes in 'ParamikoPlatform'. @VindeeR , maybe we leave this change of single to double quotes to another PR...

autosubmit/job/job.py
```diff
@@ -3134,9 +3095,10 @@ def check_started_after(self, date_limit) -> bool:
             return False
 
     def check_running_after(self, date_limit) -> bool:
-        """Checks if the job was running after the given date
+        """
+        Checks if the job was running after the given date
```
kinow: Same line? """"Checks if...

autosubmit/job/job.py
```diff
@@ -3123,9 +3083,10 @@ def stat_registered(self, attempt: int) -> bool:
         return job_data is not None
 
     def check_started_after(self, date_limit) -> bool:
-        """Checks if the job started after the given date
+        """
+        Checks if the job started after the given date
```
kinow: docstring """Checks if... in other functions of this file too...

autosubmit/job/job.py
```diff
@@ -2585,7 +2555,8 @@ def init_platform(self, as_conf: AutosubmitConfig) -> None:
                 self.platform_name = as_conf.experiment_data.get("DEFAULT", {}).get("HPCARCH", "LOCAL")
             self.platform = self.submitter.platforms.get(self.platform_name)
 
-    def update_content_extra(self, as_conf: AutosubmitConfig, files: list[str]) -> list[str]:
+    @staticmethod
```
kinow, line 2555: See other commnet about staticmethod... this may actually hide where functions could be moved to module-level functions, for instance.

autosubmit/job/job_packages.py
VindeeR: As per [PEP](https://peps.python.org/pep-0008/#pet-peeves) it says to use a space after the colon

And at this [PEP](https://peps.python.org/pep-0484/#forward-references) reference it uses the single quote for forward import, but I'm not sure if it matters much I can change it to double quotes

autosubmit/platforms/slurmplatform.py
VindeeR: So you prefer to move it to a `util file` or just leave it as it was?

autosubmit/job/job.py
VindeeR: I thought we were using double for the Strings only and Forward import we would keep the single quotes

autosubmit/job/job_packages.py
kinow: Nah, single quote here is fine for me :man_shrugging: I thought we were going full double quotes. But let's not worry about this for now.

The GH UI is showing no space here, so I think we need to add the space back?

autosubmit/platforms/slurmplatform.py
kinow: No preference here, it depends :sweat_smile: 

We can make it static for now, but eventually when we review the code it might be moved to a util file, to the same file as _private (if only used here), or really made static if needed. It's just that I think the change here was more because the IDE complained that the function could be made static (not necessarily it needs to be made static).

autosubmit/job/job.py
kinow: Ah, if so that's OK too. I will adapt to what we choose. I was just remembering we were changing quotes, but not sure which one, nor where :laughing:

kinow:
CI needs a kick or two?

autosubmit/job/job.py
VindeeR: tbh it is a bit confusing and it is hard to keep track of which is which so same here, I'll try to adapt to whichever is best to everyone.

autosubmit/job/job_packages.py
VindeeR: Oh true I got confused with the spaces :laughing: sorry will fix it

autosubmit/job/job.py
kinow: Ditto! haha

kinow (approved):
Nothing else to add. Approving once the build passes.

kinow:
Missing coverage, but the changes are mainly adding types. +1 from me.

PR #3288, opened 2026-09-23
[Enhacement] - Removing the autosubmit_parameters

### LINKED ISSUE #3272: Replace autosubmit_parameters by docstrings to reduce memory
The `autosubmit_parameters` decorator had some patched-up code that I used in the past, in order to automate the documentation of Job parameters.

While that works, I realised that the objects created are now a bit "fat". Each `Job` now has a lot of strings attached. Besides the amount of memory used, Python may (or may not) intern the strings.

If the strings are interned, and given the nature of Autosubmit jobs, I think it could lead to either/both memory pressure or leaks/GC issues.

I have a notebook from April 2025 with some tests using docstrings: https://github.com/kinow/notebooks/tree/master/python/python-obj-docs

And the PR that deleted `autosubmit.py` introduced some code that reads docstrings to create the Autosubmit command line interface: https://github.com/BSC-ES/autosubmit/blob/9b1d85f5e243dd7f0c08d68a192203b8411570ae/autosubmit/scripts/_args.py#L256-L292

I believe combining the information from these two links, we should be able to keep the current documentation automation, but instead of using a decorator with Python strings, we could have comments (which can be discarded in production -- another interesting experiment/improvement for later!) that would help us creating leaner objects.

VindeeR:
Closes #3272

The autosubmit_parameters is the older and resource consuming way of adding the parameters to the class

**Check List**

- [ ] Documentation updated.


autosubmit/job/job.py
```diff
@@ -131,63 +131,6 @@ class RecoveryReport:
...
-        'job': {
-            'rootdir': 'Experiment folder path.',
-            'projdir': 'Project folder path.',
-            'nummembers': 'Number of members of the experiment.'
-        },
-        'project': {
-            'project.project_type': 'Type of the project.',
-            'project.project_destination': 'Folder to hold the project sources.'
-        }
-    }
```
kinow, line 189: This information needs to be retrieved via docstring instead, and used to produce the https://autosubmit.readthedocs.io/en/master/userguide/variables.html page (which is currently done here https://github.com/BSC-ES/autosubmit/blob/03897b378beb66e0b4eea3b96cdfd260dbffbb6b/docs/source/ext/autosubmit_variables.py 😬

autosubmit/job/job.py
```diff
@@ -131,63 +131,6 @@ class RecoveryReport:
...
-            'chunk_second_to_last_date': 'Chunk second to last date.',
-            'chunk_second_to_last_year': 'Chunk second to last year.',
-            'chunk_second_to_last_month': 'Chunk second to last month.',
-            'chunk_second_to_last_day': 'Chunk second to last day.',
-            'chunk_second_to_last_hour': 'Chunk second to last hour.',
-            'prev': 'Days since start date at the chunk\'s start.',
-            'chunk_first': 'True if the current chunk is the first, false otherwise.',
-            'chunk_last': 'True if the current chunk is the last, false otherwise.',
-            'chunk_end_date_last': 'End date of the last chunk, i.e. the experiment end boundary. Available to any date-aware job (notably RUNNING: date jobs).',
-            'ldate': 'Last date of the experiment (the run\'s final day, parallel to SDATE).',
```
VindeeR: apparently most if not all of those are calculated in runtime and not set in a specific variable with getter/setter but added to an dictionary so we cant find them as parameters.

PR #3298, opened 2026-09-24
fix: expand user home in local project paths

### LINKED ISSUE #3296: Expand project location using user path (~)
An easy new feature :-) 

### Problem

When I'm developing AS projects, it's quite common that my project is a local project, somewhere like `~/autosubmit-src/project`. The problem is that I always have to specify `/home/kinow`, as using `~` is not supported.

### Proposed Solution

Something like `pathlib.Path(bla).expanduser()`. Shouldn't be too difficult.

Unit tests have the ~ reset to the `tmp_path` of pytest (if not unit, then integraton tests). So we can also create a simple test project there, and make sure it works in an unit test (or an integration if needed)

fatihcvs:
`LOCAL.PROJECT_PATH` now expands a user-home prefix with `Path.expanduser()`. A project configured as `~/autosubmit-src/project` can be resolved without spelling out the absolute home directory. Empty paths still raise the existing error; absolute and ordinary relative paths retain their behavior.

Closes #3296.

Validation on Linux, Python 3.12.14:

- Before the fix, both new home-prefix regressions failed; the four preservation/error cases passed.
- `python -m pytest -n 0 --capture=sys test/unit/config test/unit/test_create.py test/unit/experiment`: **238 passed, 2 skipped**. Ran as an unprivileged user with POSIX temporary storage. The initial root run failed two ownership assertions because UID 0 is falsey; the unprivileged run passes those unchanged tests.
- Ruff on the two changed Python files and `git diff --check` passed.
- Full-suite and external Docker/HPC integration tests were not run.

**Check List**

- [x] Contains logically grouped changes.
- [x] Does not contain off-topic changes.
- [x] No dependency changes were needed.
- [x] Tests are included, using a temporary home directory and a real local project file.
- [x] Changelog entry included in `CHANGELOG.md`.
- [x] Documentation updated.
- [x] Related issue linked above.


kinow (approved):
Thank you @fatihcvs !

fatihcvs:
You're welcome! Thanks for reviewing and merging the fix.

